PR #6, opened 2024-03-30
chore: rename dracutdevs/dracut to dracut-ng/dracut-ng

LaszloGombos:
This change is required to initialize CI for dracut-ng.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -19,7 +19,7 @@ permissions:
 
 jobs:
     push_to_registry:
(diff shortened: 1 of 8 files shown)


Conan-Kudo:
We don't intend to change the CLI and configuration interfaces, right? Just the project? I think the potential for breakage would be ridiculous if we did more than that.

Conan-Kudo (approved):

Conan-Kudo:
Let's go...

PR #21, opened 2024-03-30
ci: improve the redability of job names

LaszloGombos:
Changes the job name from
"Integration Test / test (arch:latest, 10)" to
"Integration Test / 10 on arch".

This change makes the job names shorter and more readable.

(Cherry-picked commit from dracutdevs/dracut#2385)

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -9,6 +9,7 @@ env:
 
 jobs:
     test:
+        name: ${{ matrix.test }} on ${{ matrix.container }}
         runs-on: ubuntu-latest
         timeout-minutes: 30
         concurrency:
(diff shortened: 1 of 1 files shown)


Conan-Kudo (changes requested):
Can you change the commit message to add `Co-authored-by: Laszlo Gombos <laszlo.gombos@gmail.com>`. Also there's a few spelling errors in the commit message that would be great to clean up here.

.github/workflows/integration.yml
```diff
@@ -1,4 +1,4 @@
-name: Integration Test
+name: Test
```
Conan-Kudo, line 1: I'm not sure we should rename this.

Conan-Kudo (changes requested):

Conan-Kudo (approved):

PR #23, opened 2024-03-30
revert(ci): install multipath-tools into openSUSE container

LaszloGombos:
The way multipath works in SUSE distros differs from upstream, causing some tests to fail.

This reverts commit c08ae406bc7c7d815b0911d9429f400719afc2ae.

Fixes https://github.com/dracutdevs/dracut/commit/c08ae406

(Cherry-picked commit from dracutdevs/dracut#2499)

This should make tests 3,13,14 green for openSUSE, but since this is a CI container change it will only take effect in subsequent PRs (this is just how CI is setup for now).


Conan-Kudo (changes requested):
This needs rebasing.

Conan-Kudo (approved):

PR #42, opened 2024-03-31
test(FULL SYSTEMD): no need to include dbus to the target rootfs

LaszloGombos:
systemd does not depend on dbus.

## Checklist

(Cherry-picked commit from dracutdevs/dracut#2549)

Ubuntu has been carrying this PR - https://changelogs.ubuntu.com/changelogs/pool/main/d/dracut/dracut_060+5-1ubuntu2/changelog

Debian just landed it - https://salsa.debian.org/debian/dracut/-/commit/700f2c9b75fd3486425fe941192790f4be505076

### DIFF
--- a/test/TEST-04-FULL-SYSTEMD/test.sh
+++ b/test/TEST-04-FULL-SYSTEMD/test.sh
@@ -52,7 +52,7 @@ test_setup() {
 
     # Create what will eventually be our root filesystem onto an overlay
     "$DRACUT" -l --keep --tmpdir "$TESTDIR" \
-        -m "test-root dbus" \
+        -m "test-root systemd" \
         -I "ldconfig" \
         -i ./test-init.sh /sbin/test-init \
         -i ./fstab /etc/fstab \


test/TEST-04-FULL-SYSTEMD/test-init.sh
```diff
@@ -37,5 +37,4 @@ if getargbool 0 rd.shell; then
     setsid $CTTY sh -i
 fi
 echo "Powering down."
-systemctl --no-block poweroff
-exit 0
+poweroff -f
```
Conan-Kudo, line 40: Why are we changing the commands here?

Conan-Kudo (changes requested):

test/TEST-04-FULL-SYSTEMD/test-init.sh
dracutng: To align with all the other tests - e.g. https://github.com/dracut-ng/dracut-ng/blob/main/test/TEST-01-BASIC/test-init.sh . I can take this hunk out of this PR to have a separate discussion.

Conan-Kudo (approved):

PR #47, opened 2024-03-31
ci: run legacy network tests on openSUSE instead of Fedora

LaszloGombos:
## Changes

openSUSE has been actively maintaining network-legacy dracut module.

Fedora is no longer packaging network-legacy dracut module ([Remove network-legacy module](https://src.fedoraproject.org/rpms/dracut/c/a1ebaf27b616010bc672be9409ff42b8234b008d)), so it is not a good container to test the network-legacy dracut module.

## Checklist

CC @aafeijoo-suse

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -68,7 +68,6 @@ jobs:
                 ]
                 network: [
                         "network-manager",
-                        "network-legacy",
                         #"systemd-networkd",
                         #"connman",
                 ]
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
I don't think this module is compatible with openSUSE either, as the legacy network stuff between Fedora and openSUSE differed just enough to not be compatible.

We should just drop network-legacy instead.

Conan-Kudo:
Ehh, whatever, we can do that later.

Conan-Kudo (approved):

PR #53, opened 2024-03-31
feat(network): include 98-default-mac-none.link if it exists

dracutng:
## Changes

In Fedora Linux there was a new 98-default-mac-none.link file added to set the MACAddressPolicy=none for bond/bridge/team devices. We'd like for this policy to apply in the initramfs as well. See

- https://fedoraproject.org/wiki/Changes/MAC_Address_Policy_none
- https://src.fedoraproject.org/rpms/systemd/pull-request/100#

## Changes

(Cherry-picked commit from dracutdevs/dracut#2224)

CC @dustymabe


Conan-Kudo:
I think if someone wants to pull it out in network-legacy later, we'd take a pull request for it. But for now, let's land it, because I think it's a reasonable default anyway.

Conan-Kudo (approved):

PR #61, opened 2024-04-01
fix(dmsquash-live): use the overlay size with thin provisioning     

dracutng:
Using the overlay size in the device mapper table allows the filesystem in the base image to be extended at runtime if the overlay is larger than the base image (i.e. this makes resize2fs /dev/mapper/live-rw work when rd.live.overlay.thin is used).

## Changes

This patch changes the upper-bound of the live-rw device when the user requests to create a thin-provisioned overlay by specifying the rd.live.overlay.thin kernel parameter. The size of the overlay (typically 32GiB, but it can be specified by the user with a kernel parameter) is used rather than the size of the base image.

## Checklist

Users of Fedora Linux have been reporting this bug here: https://discussion.fedoraproject.org/t/fedora-liveos-root-system-and-available-ram/82531/1

CC @gregory-lee-bartholomew

(Cherry-picked commit from dracutdevs/dracut#2604)

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -298,7 +298,7 @@ do_live_overlay() {
         dmsetup message /dev/mapper/live-overlay-pool 0 "create_thin 0"
 
         # Create a snapshot of the base image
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
This patch needs the commit message reworked.

Conan-Kudo (approved):

PR #58, opened 2024-03-31
ci: only run C code scan if src/ directory changed

LaszloGombos:
## Changes

Only run scanner if src/ directory changed

(Cherry-picked commit from dracutdevs/dracut#2230)

Positive feedback from @aafeijoo-suse @mrc0mmand

PR #63, opened 2024-04-01
fix(Makefile): release is now just made out of a git sha

dracutng:
## Changes

Remove dist make target as it is no longer used or maintained.

(Cherry-picked commit from dracutdevs/dracut#2504)

Positive feedback from @pvalena

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,17 +233,6 @@ clean:
 	$(RM) dracut-cpio src/dracut-cpio/target/release/dracut-cpio*
 	$(MAKE) -C test clean
 
-dist: dracut-$(DRACUT_MAIN_VERSION).tar.xz
-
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

Conan-Kudo:
Looks like the CI broke with #58, so prepared a revert in #65.

pvalena:
> > I don't know if we want this... I think we want to continue to produce release tarballs.
> 
> You mean tar.gz just from source code ? Github can do that from a tag like https://github.com/dracutdevs/dracut/releases/tag/059
> 
> I think last time this make target was used is tag 056 - https://github.com/dracutdevs/dracut/releases/tag/056 . Perhaps to sign the tar ball..

Exactly, `source-git` (packit) and git releases fix this.

PR #65, opened 2024-04-01
revert(ci): only run C code scan if src/ directory changed

Conan-Kudo:
This breaks our required CI status conditions as a required check now does not always run.

This reverts commit cffcbf22717dba9fd296772673fee4b0d286f9ad.


Conan-Kudo:
Yeah, that's fine then.

PR #67, opened 2024-04-02
Skip creating the dracut initrd when initrd is provided or a different initrd generator is configured

Conan-Kudo:
This pull request changes...

## Changes

`kernel-install` respects the `KERNEL_INSTALL_INITRD_GENERATOR` to select initramfs generators, especially now to not invoke dracut at all if unwanted.

## Checklist

Rebased version of https://github.com/dracutdevs/dracut/pull/1825 by @keszybz pulled from Fedora dracut patchset.

### DIFF
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -6,38 +6,19 @@ BOOT_DIR_ABS="$3"
 KERNEL_IMAGE="$4"
 
 # If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
-# So, let's skip to create initrd.
+# In this case, do not create the initrd.
 if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
     exit 0
 fi
 
-# Do not attempt to create initramfs if the supplied image is already a UKI
-if [[ "$KERNEL_INSTALL_IMAGE_TYPE" = "uki" ]]; then
+# Skip this plugin if we're using a different generator. If nothing is specified,
+# assume we're wanted since we're installed.
+if [ "${KERNEL_INSTALL_INITRD_GENERATOR:-dracut}" != "dracut" ]; then
     exit 0
 fi
 
-# Mismatching the install layout and the --uefi/--no-uefi opts just creates a mess.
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
I don't know why this is broken for test-04 on opensuse, but for now, convert to draft...

AndrewAmmerlaan:
This would remove auto-detection of the layout (i.e. build an uki if layout=uki and uki_generator=dracut) and also removes installing what we generate to the staging area which will break installing an initrd for systemd-boot.

I'm not opposed to the idea of this change, but the way its been done here is wrong and will break things (at least on our end). If we do this, then I think the correct way is to simply remove the fallback code path that I kept for backwards compatibility reasons before. I.e replace what's in the else with ˋexit 0ˋ

Conan-Kudo:
@keszybz do you have any feedback or interest in revising this to land? We're trying to get dracut-ng in shape for Fedora...

AndrewAmmerlaan:
> @AndrewAmmerlaan Would you be interested in submitting an alternative PR?

I can do that but I think we should first determine what problem we are solving because this is not yet completely clear to me. As I mentioned above the dracut plugin is already skipped if dracut is not requested by initrd_generator and/or uki_generator. Is the problem that this behaviour is currently conditional to the layout setting (i.e. it falls back to the old behaviour if the layout is not set to bls or uki)? Because this is an easy fix if we are willing to sacrifice the backwards compatibility with the "old behaviour".

Conan-Kudo:
That's a good question. It looks like @keszybz was trying to adapt `kernel-install` to match updated semantics from systemd: https://github.com/systemd/systemd/pull/22175. That was originally done in https://github.com/dracutdevs/dracut/pull/1691 and later in https://github.com/dracutdevs/dracut/pull/1825.

keszybz:
Those patches are quite old. Similar logic was merged in a patch from @AndrewAmmerlaan, so it's possible that it's all good now. At least it seems the first patch is not needed.

But I think the second patch is still useful.

Conan-Kudo:
@AndrewAmmerlaan It seems @keszybz would like you to submit a new pull request for this. Would you be willing to do so?

AndrewAmmerlaan:
> @AndrewAmmerlaan It seems @keszybz would like you to submit a new pull request for this. Would you be willing to do so?

I opened a new one here: https://github.com/dracut-ng/dracut-ng/pull/111

Conan-Kudo:
Closing for #111

PR #73, opened 2024-04-02
consolidated s390 device configuration

dracutng:
Consolidate the persistent and dynamic configuration of s390-specific devices in Linux distributions by delegating the configuration to the existing framework zdev from s390-tools.

This pull request completes distribution independent s390 device configuration in initrds built by dracut.
This also fixes a few known and newly discovered bugs in addition to the changes described below. Please see the commit descriptions for more bug fix information.
Some of the commits [see their description] depend on certain commits from ibm-s390-linux/s390-tools#158 and thus https://github.com/ibm-s390-linux/s390-tools/releases/tag/v2.31.0.

Zdev's job is to perform low-level configuration after which the user gets architecture-independent objects such as block devices, SCSI devices, or network interfaces. Those can and should in turn be configured with existing common code mechanisms. So there's a clear separated layering for configuration duties.

In particular, the s390-specific devices currently are: DASD (traditional disk), ZFCP (scsi), and ZNET representing channel-attached network (QETH incl. OSA and HiperSockets, LCS, CTC). Zdev has a stable command line user interface and abstracts from sysfs and from a persistent configuration representation. Zdev encapsulates configuration details. Systems management code can simply delegate configuration to zdev and thus reduce architecture-specific code.

This improves user experience, serviceability, maintainability, and reduces test effort.

## Changes
Dracut had 3 different plugin module sets to configure s390 devices. Those module sets partly competed with each other and partly contradicted each other such as activating too many devices for a memory-limited kdump environment.

original modules from Red Hat: dasd, dasd_mod, znet, zfcp.
additional modules from SUSE because the Red Hat modules were tightly coupled with the Red Hat specific persistent configuration mechanism and SUSE needed their own persistent mechanism: dasd_rules, qeth_rules, zfcp_rules.
Along with zdev as configuration backend, IBM introduced https://github.com/ibm-s390-linux/s390-tools/tree/master/zdev/dracut

making use of zdev as consolidated configuration mechanism: zdev, zdev-kdump.
This change consolidates the 3 dracut module sets into one single module set. The user syntax with rd.dasd, rd.znet, and rd.zfcp remains the same; only the backend implementation changes to zdev.
Move dasd and zfcp into zdev of s390-tools (modules could alternatively exist in dracut but they are so tightly coupled with zdev that s390-tools seemed natural).
It supports dracut features: hostonly, hostonly-cmdline, print-cmdline.
Update znet and cms (cmsdasd & cmsconffile feature used by anaconda installer) and keep them in dracut because other code depends on these modules.

## Checklist

(Cherry-picked commit from dracutdevs/dracut#2534)

CC @steffen-maier @dtardon @tblume @lnykryn


tblume:
These changes are exclusively for the s390x architecture and mostly bound to s390x hardware.
Seems we don't have any test coverage therefore.
Hence we will do some testing downstream and I'd recommend the other distros to do so as well.

tblume:
> Lot of good work here and just wanted to express that I support the goal of this PR and how it is trying to achieve this goal (to avoid some unnecessary surprises).

Yes, agreed.

> > Hence we will do some testing downstream and I'd recommend the other distros to do so as well.
> 
> This is indeed the case. If possible please share the timeline of your testing so that we can have an idea of the timeframe and which release we should target.

I will try to do it in the next 2 weeks (depending on my incoming tasks).
 
> > I suppose we could theoretically move znet into s390-tools upstream as well...
> 
> I hope s390-tools community has a better handle on s390x testing so that is one another reason why it make sense to move some dracut modules there.

Indeed.

Conan-Kudo (approved):
I don't have a ton of knowledge of IBM Z, but the changes seem reasonable to me.

aafeijoo-suse:
@steffen-maier what did you mean with https://github.com/dracutdevs/dracut/pull/2632#issuecomment-2069829971, was that something important to consider? This PR was merged without taking that comment into account, or the "step 2" PR https://github.com/dracutdevs/dracut/pull/2632

steffen-maier:
@aafeijoo-suse, https://github.com/dracutdevs/dracut/pull/2632 is intentionally a separate PR and in draft state, so it is actually good that it was not merged here with dracut-ng.

> This is the final cleanup as very last packaging step after https://github.com/dracutdevs/dracut/pull/2534. The PR split was triggered by https://github.com/dracutdevs/dracut/pull/2534#issuecomment-1933744932. Intentionally only a draft PR for the time being, until Fedora packagers give a go after package updates for step 1 in 2534 could migrate existing persistent s390 device config and thus drop the depency on the removed dracut modules in this PR here.

The comment link in the above quote from the 2632 PR description refers to @sharkcz's comment:

> There is no single "upgrade point" in Fedora (or ELN), so the migration needs to be done via a regular package update. And this will need some coordination and setting versioned dependencies between the involved packages (dracut, s390utils, ...)

Maybe my update in https://github.com/dracutdevs/dracut/pull/2632#issuecomment-2069829971 was pointless. I did not know about dracut-ng and wanted the PR 2632 not to be forgotten. For the time being, 2632 should stay in draft mode until Fedora packaging could come to a state where the part 1 (https://github.com/dracutdevs/dracut/pull/2534 or this #73 here) has been fully integrated.

The changes for 2534 I mentioned my comment you referenced did not (and will not) happen but were also not necessary (the changes were about moving the znet dracut module also to s390-tools as discussed in https://github.com/dracutdevs/dracut/pull/2534#issuecomment-2015081652 and subsequent comments). So what was merged here with #73 is fine for now.

PR #96, opened 2024-04-04
fix: wait 12 hours before halt on media check fail

dracutng:
## Changes

If a modesetting driver has been loaded by the time the media check happens, halting the system (as we currently do immediately if the check fails) blanks the screen, which is confusing for the user. This adds a warning message and a 12-hour wait before the system is eventually halted, so the user can see the media check failure and (presumably) reboot and fix the medium.

It also tweaks the text of the failure message not to call it a "CD check", since it's not 1998 any more.

https://bugzilla.redhat.com/show_bug.cgi?id=2246410

This pull request changes...

## Checklist


pvalena:
LGTM.

pvalena (approved):

PR #97, opened 2024-04-04
fix(test): do not omit dracut modules for initramfs.testing

LaszloGombos:
## Changes

We want to know if including dracut modules makes the test fail instead of masking issues by omitting dracut modules.

This PR increase test coverage and makes the tests simpler to maintain.

This PR exposed two issues (either with dracut itself or with the test suite)
- test 04 an issue with multipath dracut module for full systemd test. 
 - test 12 on Ubuntu (only), that required explicitly omitting dbus to keep the test passing. 
 
 These exposed new issues confirms the reasoning behind why this PR increases test coverage.

(Cherry-picked commit from dracutdevs/dracut#2258)

### DIFF
--- a/test/TEST-02-SYSTEMD/test.sh
+++ b/test/TEST-02-SYSTEMD/test.sh
@@ -64,7 +64,6 @@ test_setup() {
     # make the man command succeed always
     "$DRACUT" -l -i "$TESTDIR"/overlay / \
         -a "test systemd" \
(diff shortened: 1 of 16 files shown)


bdrung:
Did you open a bug for test 12 on Ubuntu needing to omit networking?

Conan-Kudo (approved):

Issue #103, opened 2024-04-04
Release tarballs do not include dracut.html


LaszloGombos:
This is a regression from earlier releases. Release https://github.com/dracutdevs/dracut/releases/tag/056 included dracut.html in the release.

PR #104, opened 2024-04-04
fix(release): maintain dracut.html in the source tree

LaszloGombos:
## Changes

Plan to update dracut.html as part of release generation together with [dracut-version](https://github.com/dracut-ng/dracut-ng/blob/main/dracut-version.sh ).

Fixes https://github.com/dracut-ng/dracut-ng/issues/103

CC @pvalena

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,6 +22,11 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Install dependencies
+        run: |
+          sudo apt -y update
+          sudo apt -y install asciidoc
+
       - name: Build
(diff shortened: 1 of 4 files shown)


pvalena:
Right, we need this, as we don't have tarballs. 

The regeneration should be part of [tools/release.sh](https://github.com/dracut-ng/dracut-ng/blob/main/tools/release.sh), right?

Briefly checking the html contents - LGTM (but I've not verified by regenerating it). It would probably make sense to create some github action for regenerating it when corresponding files are changed. But that's a stretch-goal :).

pvalena (approved):

Conan-Kudo (approved):
I'm not sure I love this, but I don't really care enough to say anything against it either.

Issue #106, opened 2024-04-05
Enable test for dracut-cpio


LaszloGombos:
[enable_dracut_cpio](https://github.com/dracut-ng/dracut-ng/blob/main/Makefile#L209) does not seem to be tested on any of the CI test containers (TEST-62 is skipped). 

References
 - https://github.com/dracutdevs/dracut/pull/1531
 - https://github.com/dracutdevs/dracut/issues/1662
 
 CC @ddiss (author) @bdrung (performance interest)

ddiss:
Thanks for the ping, we enable `dracut-cpio` at build time for openSUSE and some versions of SLES. However, it requires explicit `enhanced-cpio` configuration otherwise regular GNU `cpio` is used by default.

ddiss:
While we're here, is there any documentation regarding the intentions / backing of this fork?

PR #107, opened 2024-04-05
test(DRACUT-CPIO): enable building dracut-cpio when cargo is installed

LaszloGombos:
## Changes

Since the test is well written, it will be just skipped if cargo is not in the test container.
    
Add cargo to the openSUSE test container.

Fixes https://github.com/dracut-ng/dracut-ng/issues/106

CC @ddiss

### DIFF
--- a/test/TEST-63-DRACUT-CPIO/test.sh
+++ b/test/TEST-63-DRACUT-CPIO/test.sh
@@ -21,19 +21,16 @@ test_dracut_cpio() {
 
     mkdir -p "$tdir"
 
-    # VM script to print sentinel on boot
-    # write to kmsg so that sysrq messages don't race with console output
     cat > "$tdir/init.sh" << EOF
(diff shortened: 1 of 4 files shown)


ddiss (approved):
I'm not to familiar with github workflows, but the changes look reasonable to me.

Conan-Kudo:
Is there a reason this is only openSUSE? Also, what does `dracut-cpio` even do?

test/TEST-63-DRACUT-CPIO/test.sh
```diff
@@ -24,16 +24,15 @@ test_dracut_cpio() {
     # VM script to print sentinel on boot
     # write to kmsg so that sysrq messages don't race with console output
```
ddiss, line 25: you can probably get rid of the above line, although won't the race still be a problem with poweroff triggering the shutdown instead of sysrq?

Conan-Kudo (approved):

PR #111, opened 2024-04-05
fix(install.d): simplify and use what kernel-install gives us

AndrewAmmerlaan:
drops legacy fallback path, exit always if we are not chosen as initrd_ generator. Exit if we are already installing an UKI. Exit if an initrd image is already specified by kernel-install.

Follow-up of https://github.com/dracut-ng/dracut-ng/pull/67


install.d/50-dracut.install
```diff
@@ -16,102 +16,64 @@ if [[ "$KERNEL_INSTALL_IMAGE_TYPE" = "uki" ]]; then
...
+    UEFI_OPTS="--uefi"
+fi
 
-            read -r -d '' -a line < /proc/cmdline
-            for i in "${line[@]}"; do
-                [[ ${i#initrd=*} != "$i" ]] && continue
-                BOOT_OPTIONS+=("$i")
-            done
-        fi
+if [[ -f ${KERNEL_IMAGE%/*}/$IMAGE ]]; then
```
pvalena, line 33: AFAIU, this would make it noop when reinstalling `kernel-core`, which is not a good idea I think. I'd rather keep the previous behaviour.

install.d/50-dracut.install
```diff
@@ -2,12 +2,12 @@
...
 KERNEL_VERSION="$2"
+#shellcheck disable=SC2034
 BOOT_DIR_ABS="$3"
 KERNEL_IMAGE="$4"
 
-# If KERNEL_INSTALL_MACHINE_ID is defined but empty, BOOT_DIR_ABS is a fake directory.
-# So, let's skip to create initrd.
-if ! [[ ${KERNEL_INSTALL_MACHINE_ID-x} ]]; then
+# If the initrd was provided on the kernel command line, we shouldn't generate our own.
+if [[ "$COMMAND" != "add" || "$#" -gt 4 ]]; then
```
pvalena, line 10: With this, the support for `remove` is dropped. I can't support this.

pvalena (changes requested):

install.d/50-dracut.install
AndrewAmmerlaan: Removal is handled by other kernel-install plugins. 

This plugin should only generate an initrd or uki.efi and put it on the temporary staging area (`KERNEL_INSTALL_STAGING_AREA`), it does not install the inird or uki.efi to /boot or ESP, because where we want it depends on the `KERNEL_INSTALL_LAYOUT`. Since we have ` BOOT_DIR_ABS="$KERNEL_INSTALL_STAGING_AREA"` and the staging area is unset when removing the kernel the `remove` case never actually did anything.

install.d/50-dracut.install
pvalena: Yeah, you're right; it preserves the same behaviour. Sorry for the churn.

install.d/50-dracut.install
pvalena: Oh, ok. I didn't know it's obsoleted (if it was ever used).

AndrewAmmerlaan: I suppose it was used before we changed the plugin to respect and use the `KERNEL_INSTALL_STAGING_AREA`.

lnykryn: > Removal is handled by other kernel-install plugins.
> 

Is it still the grub plugin? If so, that's a thing that should be fixed.

AndrewAmmerlaan: AFAIK there is no upstream grub plugin, neither in grub nor in systemd. We have one [downstream in Gentoo](https://github.com/projg2/installkernel-gentoo/blob/master/hooks/systemd/90-compat.install) that handles installing (from staging area) and removing the initramfs when the kernel is installed or removed. This mirrors what the upstream systemd-boot plugin does, it also handles installing (from staging area) and removing initramfs along with the kernel.

pvalena (commented):
With all the comments, let's check how this is handled in Fedora.

pvalena:
> Thanks for helping with the review @pvalena .

np

> Just to clarify are you saying that you're planning to test this on Fedrora and we should wait until you have a chance to tests and report back ?

Yes, I do plan to test this, but currently I'm working on something else... so I'll get to it later on (either when it's merged or before). But if someone wants, they can do the review (& ideally test on Fedora and/or elsewhere) as well.

_ _ _ _
Oh, I thought you need 2 person approval... It's just 1, sadly.

pvalena:
> Even the 2 person approval can happen in a minute, so that does not really buy you time.
> 
> Landing it sooner might actually helps with testing (for examples it goes to dracut-git AUR package in Arch so that users can report back with issues).

I don't mind it landing sooner - I was just hoping more ppl are reviewing the PRs that get merged (not just one per PR + author), including this one.

nabijaczleweli:
Why did you CC me on this? Like all kernel-install drop-ins, this is effectively fedora-specific code (yes, fedora distributes its package manager's kernel and bootloader hooks as part of systemd upstream. no, I don't know why they chose to do this).

AndrewAmmerlaan:
> Like all kernel-install drop-ins, this is effectively fedora-specific code 

Well no, it's not Fedora specific, we use `kernel-install` in Gentoo too.

pvalena:
Feel free to go ahead - I will test it later, together with the whole release, due to current lack of time.

AndrewAmmerlaan:
> https://bugzilla.redhat.com/show_bug.cgi?id=2274739

I think this PR should make the downstream patch mentioned here unnecessary. I have asked the reporter to test this.

AndrewAmmerlaan:
> > https://bugzilla.redhat.com/show_bug.cgi?id=2274739

Comment from this bug report:

> That PR works fine for my use case (upstream dracut/dracut-ng without that patch does too). I like the much simpler 
> script. I believe it would allow Fedora's downstream patch to be reduced to just:
> 
>     -if [[ "$KERNEL_INSTALL_INITRD_GENERATOR" = "dracut" ]]; then
>     +if [[ "${KERNEL_INSTALL_INITRD_GENERATOR:-dracut}" = "dracut" ]]; then
> 
> (That would fix this issue because the downstream patch would no longer conflict with my setup.)

I applied this change here, I think it makes sense to assume we are wanted if no preference is set, and this would eliminate the need for a patch downstream in Fedora.

PR #113, opened 2024-04-06
ci: add more packages to the Arch container to enable more testing

LaszloGombos:
## Changes

Install tgt with pacman, instead of managing tgt dependencies (such us perl-config-general) ourselves.

Install new packages that can help with testing - such as tpm2-tools, busybox, nvme-cli

### DIFF
--- a/test/container/Dockerfile-Arch
+++ b/test/container/Dockerfile-Arch
@@ -4,17 +4,25 @@ RUN pacman --noconfirm -Syu \
     asciidoc \
     astyle \
     base-devel \
+    bluez \
     btrfs-progs \
+    busybox \
+    cargo \
+    cifs-utils \
     connman \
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
I'm really not sure I want to do this. I would prefer that for the Arch platform that we stay within the realm of baseline tooling here. Arch is already difficult enough as it is...

freswa:
If I'm not mistaken, `perl-config-general` lives in `[extra]` these days, so no need to build it from AUR which is probably outdated anyways. Can you confirm `tgt` is the only missing package? It has enough votes, so that I could move it to `[extra]` as well.

freswa:
`tgt` is now in `[extra]` as well.

Conan-Kudo (approved):

PR #114, opened 2024-04-06
fix(dracut): move hooks directory from /usr/lib to /var/lib

dracutng:
## Changes

Since https://github.com/systemd/systemd/commit/ffc1ec73, /usr is mounted as read-only in the initramfs by default.

Fixes https://github.com/dracutdevs/dracut/issues/2588

(cherry picked from dracutdevs/dracut#2591)

CC @aafeijoo-suse

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -183,42 +183,42 @@ You are encouraged to provide a README that describes what the module is for.
 
 init has the following hook points to inject scripts:
 
-`/lib/dracut/hooks/cmdline/*.sh`
+`/var/lib/dracut/hooks/cmdline/*.sh`
    scripts for command line parsing
 
-`/lib/dracut/hooks/pre-udev/*.sh`
+`/var/lib/dracut/hooks/pre-udev/*.sh`
    scripts to run before udev is started
 
-`/lib/dracut/hooks/pre-trigger/*.sh`
+`/var/lib/dracut/hooks/pre-trigger/*.sh`
    scripts to run before the main udev trigger is pulled
 
-`/lib/dracut/hooks/initqueue/*.sh`
+`/var/lib/dracut/hooks/initqueue/*.sh`
    runs in parallel to the udev trigger
    Udev events can add scripts here with /sbin/initqueue.
    If /sbin/initqueue is called with the "--onetime" option, the script
    will be removed after it was run.
-   If /lib/dracut/hooks/initqueue/work is created and udev >= 143 then
+   If /var/lib/dracut/hooks/initqueue/work is created and udev >= 143 then
    this loop can process the jobs in parallel to the udevtrigger.
    If the udev queue is empty and no root device is found or no root
    filesystem was mounted, the user will be dropped to a shell after
    a timeout.
    Scripts can remove themselves from the initqueue by "rm $job".
 
-`/lib/dracut/hooks/pre-mount/*.sh`
+`/var/lib/dracut/hooks/pre-mount/*.sh`
    scripts to run before the root filesystem is mounted
    Network filesystems like NFS that do not use device files are an
    exception. Root can be mounted already at this point.
 
-`/lib/dracut/hooks/mount/*.sh`
+`/var/lib/dracut/hooks/mount/*.sh`
    scripts to mount the root filesystem
    If the udev queue is empty and no root device is found or no root
    filesystem was mounted, the user will be dropped to a shell after
    a timeout.
 
-`/lib/dracut/hooks/pre-pivot/*.sh`
+`/var/lib/dracut/hooks/pre-pivot/*.sh`
    scripts to run before latter initramfs cleanups
 
-`/lib/dracut/hooks/cleanup/*.sh`
+`/var/lib/dracut/hooks/cleanup/*.sh`
    scripts to run before the real init is executed and the initramfs
    disappears
    All processes started before should be killed here.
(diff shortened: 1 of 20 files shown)


freswa:
It's just a feeling, but `/var/lib` is a directory which is used to store data and managed by the software itself.
Hook dirs imho belong to either `/usr` when another package provides them, or `/etc` if the administrator/user would like to create them.

Just my 2 cents :)

bdrung:
I agree with @fre

bdrung (changes requested):
I agree with @freswa, the hooks should live in `/usr` and or `/etc`. Generating dynamic configuration should be better placed in `/run` instead.

aafeijoo-suse:
You can read the old discussion which led us to choose `/var` as the "least bad option" here: https://bugzilla.opensuse.org/show_bug.cgi?id=1218068

Conan-Kudo (changes requested):
I am not sure this is a good change on its own. This will break backwards compatibility for anyone already using the existing hooks directory for packaged hooks, and the new path doesn't work on systems where `/var` is not available during transactions (e.g. RPM-OSTree).

aafeijoo-suse:
> I am not sure this is a good change on its own. This will break backwards compatibility for anyone already using the existing hooks directory for packaged hooks,

There is a compat symlink to the old location.

https://github.com/dracut-ng/dracut-ng/pull/114/files#diff-1a92fcdd8eaf1d31c9913c0e2f96022f82331a5ba7a2d89802927bae30484d1eR51

`/usr` is read-only in systemd git main, so a change is required.

> and the new path doesn't work on systems where `/var` is not available during transactions (e.g. RPM-OSTree).

I you follow the discussion in https://bugzilla.opensuse.org/show_bug.cgi?id=1218068 and https://github.com/dracutdevs/dracut/pull/2591, I initially suggested a directory like `/.dracut`, but we were trying to guess if `/var` could be used without breaking anything.

pvalena:
> > There is a compat symlink to the old location.
> 
> One option - that I am more and more considering - is to only add the compat symlink and do not change the documentation. This would make the compat symlink an implementation detail and we would still recommend out of tree dracut modules to use e.g. /usr/lib instead of /var/lib .
> 
> Any thoughts on this ?

Just to be clear, that would mean 2 locations:
 -  (1) for user-created (packaged) hooks; and
 -  (2) for runtime-created (generated) hooks

Those need to be then handled both - e.g. by copying all of the (1) to (2) prior to generating new ones. I'm fine with `/var` but IMHO if done correctly, we could use:
 - `/usr/lib` for (1)
 - `/run/dracut` for (2)

Is there some reason for not doing this? Is there some reason for not using `/run` that I've missed?

pvalena:
> ```
>     # symlink to old hooks location for compatibility
>     ln_r /var/lib/dracut/hooks /lib/dracut/hooks
> ```
> 
> takes care about both cases already (assuming /var is writable).

Hmm I didn't realize this before, but now seeing it, I think it doesn't work in case there're some pre-existing files in `/lib/dracut/hooks`.

The original intent is to create a symlink `/lib/dracut/hooks` (to pointing to directory `/var/lib/dracut/hooks`), which is fine, but if there already is a directory `/lib/dracut/hooks`, then it will create a symlink **in** that directory instead (creating symlink `/lib/dracut/hooks/hooks`; [test log](https://gist.github.com/pvalena/f1e87e58913dc2c6d8948b442be6b023)).

aafeijoo-suse:
> > ```
> >     # symlink to old hooks location for compatibility
> >     ln_r /var/lib/dracut/hooks /lib/dracut/hooks
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > takes care about both cases already (assuming /var is writable).
> 
> Hmm I didn't realize this before, but now seeing it, I think it doesn't work in case there're some pre-existing files in `/lib/dracut/hooks`.
> 
> The original intent is to create a symlink `/lib/dracut/hooks` (to pointing to directory `/var/lib/dracut/hooks`), which is fine, but if there already is a directory `/lib/dracut/hooks` (empyty or not), then it will create a symlink **in** that directory instead (creating symlink `/lib/dracut/hooks/hooks`; [test log](https://gist.github.com/pvalena/f1e87e58913dc2c6d8948b442be6b023)).

FTR, we have been using the current proposal since systemd introduced the change, otherwise Tumbleweed with systemd from git main would not work. So, it works. That said, I think your idea of copying the hooks to /run at boot may also be valid, I could give it a try (maybe on a separate PR) if that's the desired path to follow.

lnykryn:
This is probably a stupid idea, but shouldn't we just set ProtectSystem=no for the hooks unit files? I understand the reasoning behind the systemd change. If you build initrd with let's say mkosi-initrd you mostly don't need to touch /usr. But Dracut is unfortunately built in a way where it does a lot of "self-modifications" (I mean scripts that generate scripts....)

Conan-Kudo:
I'm closing this as #186 was merged.

aafeijoo-suse:
> I'm closing this as #186 was merged.

That's a "good" way to end an ongoing discussion.

AdamWill:
I think it'd be reasonable to re-open this, as #186 / #194 is really more of a band-aid than a long term "fix". It does seem like it would be much cleaner in the long term to really change dracut's own idea of where the hooks go, not just kinda trick it with a symlink. But it's also clearly more disruptive.

PR #118, opened 2024-04-06
fix(systemd): explicitly install some libs that will not be statically included

dracutng:
## Changes

Some required libraries that used to be statically included are in the process to be opened via `dlopen()`.

References:
- https://github.com/systemd/systemd/pull/31131
- https://github.com/systemd/systemd/pull/31550
- https://github.com/systemd/systemd/pull/32019

Closes https://github.com/dracutdevs/dracut/issues/2642

(cherry picked from dracutdevs/dracut#2643)


Conan-Kudo (approved):
😩

PR #120, opened 2024-04-06
fix(dracut.sh): omit compressed kernel modules from find searching exec files

dracutng:
## Changes

Although the kernel modules are not included yet because they are not executable, this speeds up the `find`.

## Checklist


thesamesam (approved):
We could apply https://github.com/dracutdevs/dracut/pull/2321#discussion_r1180540666 safely, I think, but the change is fine as it is.

PR #123, opened 2024-04-06
fix(dracut.sh): abort if Bash is in POSIX mode

dracutng:
## Changes

Fixes issue #2522


thesamesam (commented):
This is quirky enough that it might be nice to have a comment above it linking to the bug. I raised my eyebrow when I read the change first :)

bdrung (commented):
Why not unset `POSIXLY_CORRECT` instead?

PR #131, opened 2024-04-06
ci: improve Gentoo container

LaszloGombos:
## Changes

Switch from musl and OpenRC to glibc and systemd for Gentoo.
Use Alpine instead for musl and OpenRC testing.

Gentoo musl container is not compatible with actions/checkout version >=2 due to musl. Alpine container (with musl) does not have this issue, so this PR would allow us to remove the actions/checkout@1 workarounds as seen on https://github.com/dracut-ng/dracut-ng/pull/129 .

This PR does not enable yet any testing on the new Alpine container, so for now the Manualtest Action needs to be used for that.


Conan-Kudo:
This PR has nothing to do with Gentoo?

Conan-Kudo (approved):

Conan-Kudo:
Anyway, I don't have any issues with this, even if I'm a bit confused by it.

Issue #130, opened 2024-04-06
With dracut 100, nvme kernel drivers are not included in hostonly mode


dalto8:
**Describe the bug**
When building the initrd with `dracut --hostonly --no-hostonly-cmdline` nvme kernel drivers are not included even though there are nvme devices installed and used in the system.  Downgrading to 059 includes these drivers. 

**Distribution used**
Arch Linux

**Dracut version**
100

**Init system**
systemd

Please let me know if I can provide any further information.

dalto8:
Here is the log with dracut -v https://pastebin.mozilla.org/vbT0oBGO

I suspect it will work if I add the driver manually since it is added when it isn't in hostonly mode.

If needed, I can build and test v60.

dalto8:
This is in the logs:
```
dracut[I]: *** Including module: kernel-modules ***
```

Is there something else I should check?

dalto8:
> I see no change around the nvme driver (file changed but not the line) - https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/90kernel-modules/module-setup.sh#L30

It isn't even reaching that line.  It isn't entering this `if` block: https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/90kernel-modules/module-setup.sh#L107

dalto8:
I built the Arch packages for 60-1 and 60-2.(The latter removes some patches).

Both those versions include the nvme driver.  Only when upgrading to 100 does it stop being included.

freswa:
@dalto8 Could you post your dracut config and the command you run dracut with please?

dalto8:
> @dalto8 Could you post your dracut config and the command you run dracut with please?

The command is `dracut --hostonly --no-hostonly-cmdline initrd 6.8.2-arch2-1`

The config file is unchanged but I do have a drop in that has this in it.

```
omit_dracutmodules+=" network cifs nfs nbd brltty i18n "
#hostonly_mode=strict
compress="zstd"
```
I commented out `hostonly_mode=strict` as my first troubleshooting step.

> Perhaps caused by this #36 ?

Let me revert that and test.

dalto8:
> Perhaps caused by this #36 ?

Yes.  That seems be it.  Specifically, the change to [line 1670](https://github.com/dracutng/dracut-ng/blob/1cb7754d1278814786e6d98e4915cf59fb667cc5/dracut.sh#L1670).

If I revert that change, it works again.

[...]

PR #133, opened 2024-04-07
revert: "fix(dracut.sh): do not add device if find_block_device returns an error"

LaszloGombos:
Partial reverts dracut-ng/dracut-ng#36

Fixes https://github.com/dracut-ng/dracut-ng/issues/130


Conan-Kudo (approved):

dalto8:
This reverts the wrong line I think.  It is the change to https://github.com/dracutng/dracut-ng/blob/1cb7754d1278814786e6d98e4915cf59fb667cc5/dracut.sh#L1670 that causes the issue in my testing.

Issue #135, opened 2024-04-07
systemd enabled Gentoo container fails tests


LaszloGombos:
After systemd is enabled for the Gentoo CI container tests are failing.

Hope to fix the container instead of reverting the change as we want to move forward and have systemd in Gentoo container.

See https://github.com/dracut-ng/dracut-ng/commit/76963537edcfbad21c5fa096834e982564653808

Error message

```
Could not find a Linux kernel version 6.6.21-gentoo-dist to test with!
Please install linux.
```

Link to failed run - https://github.com/dracut-ng/dracut-ng/actions/runs/8589738665/job/23536300253

```
ls /boot/
kernel-6.6.21-gentoo-dist
```

AndrewAmmerlaan:
Those test have in common that they are testing encryption and RAID. I suspect some systemd components are missing from the initramfs because they are not installed by the systemd package.

We probably need to enable some USE flag for `sys-apps/systemd`, the question though is which one:
```
[I] sys-apps/systemd
Available versions:  254.8-r1(0/2)^t (~)254.9-r1(0/2)^t (~)254.10(0/2)^t 255.3-r1(0/2)^t (~)255.4(0/2)^t{gpkg} **9999(0/2)*l^t {acl apparmor audit boot cgroup-hybrid cryptsetup curl +dns-over-tls elfutils fido2 +gcrypt gnutls homed http idn importd iptables +kernel-install +kmod +lz4 lzma +openssl pam pcre pkcs11 policykit pwquality qrcode +resolvconf +seccomp secureboot selinux split-usr +sysv-utils test tpm ukify vanilla xkb +zstd ABI_MIPS="n32 n64 o32" ABI_S390="32 64" ABI_X86="32 64 x32" PYTHON_SINGLE_TARGET="python3_10 python3_11 python3_12"}
[...]

PR #136, opened 2024-04-07
ci: adopt Gentoo container to run the test suite

LaszloGombos:
## Changes

Tell the installkernel package to use Debians installkernel rather then systemd's kernel-instal

Fixes: #135

Remaining test failures (04, 10, 12, 13, 18) should recover after the PR lands and test runs will pick up the new container version

### DIFF
--- a/test/container/Dockerfile-Gentoo
+++ b/test/container/Dockerfile-Gentoo
@@ -17,6 +17,7 @@ FROM docker.io/gentoo/stage3:$TAG as kernel
 COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo
 # disable initramfs generation, only need the kernel image itself
 RUN echo 'sys-kernel/gentoo-kernel-bin -initramfs' > /etc/portage/package.use/kernel
+RUN echo 'sys-kernel/installkernel -systemd' >> /etc/portage/package.use/kernel
 RUN emerge -qv sys-kernel/gentoo-kernel-bin
 
 FROM docker.io/gentoo/stage3:$TAG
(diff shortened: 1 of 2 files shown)


AndrewAmmerlaan (commented):
It's called `kernel-` on systemd based systems (using systemd's `kernel-install`). On OpenRC systems it's named `vmlinuz-` or `vmlinux-` (using Debian's `installkernel`).

`kernel-` is also one of the valid glob's that `grub-mkconfig` considers.

AndrewAmmerlaan:
Hmm, UKI test is failing. 

The "correct" way of building an UKI on Gentoo is to enable the `uki` USE flag on `sys-kernel/installkernel`.

AndrewAmmerlaan:
As a quick fix I recommend changing:
```
RUN echo 'sys-kernel/gentoo-kernel-bin -initramfs' > /etc/portage/package.use/kernel
```
to 
```
RUN echo 'sys-kernel/gentoo-kernel-bin -initramfs' > /etc/portage/package.use/kernel
RUN echo 'sys-kernel/installkernel -systemd dracut' > /etc/portage/package.use/kernel
```
To tell the installkernel package to use Debians `installkernel` rather then systemd's `kernel-instal`

Conan-Kudo (approved):

AndrewAmmerlaan (approved):

AndrewAmmerlaan:
```
dracut[F]: Can't find a kernel image '/boot/vmlinuz-6.6.21-gentoo-dist' to create a UEFI executable
```
This should resolve itself once newer kernel versions are stable. I pushed a fix a while back to install a symlink to the kernel in `/lib/modules/vmlinuz` where it should be picked up by dracut automatically without having to explicitly set `--kernel-image ...`.

```

--kernel-image <file>
Specifies the kernel image, which to include in the UEFI executable. The default is /lib/modules/<KERNEL-VERSION>/vmlinuz or /boot/vmlinuz-<KERNEL-VERSION>.
```

PR #142, opened 2024-04-08
fix(nvmf): move /etc/nvme/host{nqn,id} requirement to hostonly

dracutng:
## Changes

When creating initramfs for universal boot image such as an installer, we can't include any machine-specific IDs. Let's move the check for /etc/nvme/hostnqn and /etc/nvme/hostid files presence to the hostonly section to avoid unsatisfied requirements.

## Checklist

### DIFF
--- a/modules.d/95nvmf/module-setup.sh
+++ b/modules.d/95nvmf/module-setup.sh
@@ -3,8 +3,6 @@
 # called by dracut
 check() {
     require_binaries nvme jq || return 1
-    [ -f /etc/nvme/hostnqn ] || return 255
-    [ -f /etc/nvme/hostid ] || return 255
 
     is_nvmf() {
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

mwilck (approved):
Actually, even for hostonly setups, static hostnqn is not strictly necessary, as it can be provided from the NBFT. 

But in practice I suppose it will always be present in NVMeoF boot scenarios. And that would be a separate change anyway.

tbzatek:
> Actually, even for hostonly setups, static hostnqn is not strictly necessary, as it can be provided from the NBFT.

The `95nvmf` module still supports the old (non-NBFT) way of making NVMeoF connections - the `rd.nvmf.discover` arguments.

PR #143, opened 2024-04-08
feat(hwdb): install hwdb on demand when module is needed

dracutng:
## Changes

Adding a module to install hwdb. Further extensions might make only selected part of hwdb installable, to save space. The module is not included by default.

Including the module adds 2MB of compressed data (on Fedora, the file has 12MB).

hwdb is needed in case of custom HW, like a keyboard/mouse or various interfaces.

Original PR: https://github.com/dracutdevs/dracut/pull/1681

## Checklist

Fixes #

### DIFF
--- a/modules.d/95hwdb/module-setup.sh
+++ b/modules.d/95hwdb/module-setup.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+check() {
+    return 255
+}
+
+# called by dracut
+install() {
+    local hwdb_bin
+
+    # systemd-hwdb ships the file in /etc, with /usr/lib as an alternative.
+    # The alternative location is preferred, as we can consider it being user
+    # configuration.
+    hwdb_bin="${udevdir}"/hwdb.bin
+
+    if [[ ! -r "${hwdb_bin}" ]]; then
+        hwdb_bin="${udevconfdir}"/hwdb.bin
+    fi
+
+    if [[ $hostonly ]]; then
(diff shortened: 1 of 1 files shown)


pvalena:
I'm not sure what's up with shfmt though.

pvalena:
I can fix this, but why is this required? It doesn't make sense to me...

modules.d/95hwdb/module-setup.sh
```diff
@@ -0,0 +1,27 @@
...
+    return 255
+}
+
+# called by dracut
+install() {
+    local hwdb_bin
+
+    # systemd-hwdb ships the file in /etc, with /usr/lib as an alternative.
+    # The alternative location is preferred, as we can consider it being user
+    # configuration.
```
Conan-Kudo, line 15: This comment is confusing. Do you mean we prefer the `/usr/lib` location or the `/etc` location?

Conan-Kudo (changes requested):

modules.d/95hwdb/module-setup.sh
pvalena: Yes, that's what I meant.

From man page:
```
       --usr
           Generate in /usr/lib/udev instead of /etc/udev.
```

pvalena: Usually, it should be only in one place. It's unclear which one, but copying both would be even more confusing. Although copying what `systemd-hwdb query ` uses might be better...

pvalena (changes requested):
I'll switch the order of priorities, to behave the same way as `systemd-hwdb query` does.

PR #144, opened 2024-04-08
fix(test): fixup Gentoo CI

Nowa-Ammerlaan:
we download systemd docker image, so we want systemd, not systemd-utils

equery f is used to list files installed by a package on Gentoo

Closes: https://github.com/dracut-ng/dracut-ng/issues/135


AndrewAmmerlaan:
This should also fix the "No /boot/vmlinuz-x.y.z` error I saw before in your other PR.

PR #160, opened 2024-04-12
feat(dracut.sh): make initramfs-${kernel}.img filename configurable

### LINKED ISSUE #99: detect and generate initrd.img-${kernel}
https://www.freedesktop.org/wiki/Specifications/BootLoaderSpec/ refers to initramfs as "initrd".

One of the most popular distro's (Debian) expects /boot/initrd.img-${kernel} instead of initramfs-${kernel}.img, see https://kernel-team.pages.debian.net/kernel-handbook/ch-packaging.html#s-arch-dep

Let's look for a way to detect in a reliable way what should be the default output file that is more inclusive.

Possibly related openSUSE patch downstream: https://github.com/dracutdevs/dracut/commit/f68649cf55ca4545ed8304435c9c7315846c45a4

LaszloGombos:
## Changes

The ambition is to propose a minimum viable patch and replace simple downstream patches like https://github.com/openSUSE/dracut/commit/1442c9d29

The `initrdname` configuration option is required to match a `initr*${kernel}*` file pattern.

Distributions are incentived but not required to follow the upcoming https://github.com/uapi-group/specifications/issues/100 standard (assuming positive resolution).

## Checklist

Fixes #99

Tested with the following config file entry

```
initrdname=initrd-${kernel} 
```

### DIFF
--- a/dracut-initramfs-restore.sh
+++ b/dracut-initramfs-restore.sh
@@ -46,8 +46,15 @@ elif mountpoint -q /efi; then
 elif mountpoint -q /boot/efi; then
     IMG="/boot/efi/$MACHINE_ID/$KERNEL_VERSION/initrd"
 else
-    echo "No initramfs image found to restore!"
-    exit 1
+    files=("/boot/initr*${KERNEL_VERSION}*")
+    if [ "${#files[@]}" -ge 1 ] && [ -e "${files[0]}" ]; then
+        IMG="${files[0]}"
+    elif [[ -f /boot/initramfs-linux.img ]]; then
+        IMG="/boot/initramfs-linux.img"
+    else
+        echo "No initramfs image found to restore!"
+        exit 1
+    fi
 fi
 
 cd /run/initramfs
--- a/dracut.sh
+++ b/dracut.sh
@@ -1140,6 +1140,9 @@ if ! [[ $outfile ]]; then
         mkdir -p "$dracutsysrootdir$efidir/Linux"
         outfile="$dracutsysrootdir$efidir/Linux/linux-$kernel${MACHINE_ID:+-${MACHINE_ID}}${BUILD_ID:+-${BUILD_ID}}.efi"
     else
+        if ! [[ $initrdname ]]; then
+            initrdname="initramfs-${kernel}.img"
+        fi
         if [[ -d "$dracutsysrootdir"/efi/loader/entries || -L "$dracutsysrootdir"/efi/loader/entries ]] \
             && [[ $MACHINE_ID ]] \
             && [[ -d "$dracutsysrootdir"/efi/${MACHINE_ID} || -L "$dracutsysrootdir"/efi/${MACHINE_ID} ]]; then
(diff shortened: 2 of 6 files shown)


freswa (approved):
lgtm

dracut-initramfs-restore.sh
```diff
@@ -46,8 +46,15 @@ elif mountpoint -q /efi; then
 elif mountpoint -q /boot/efi; then
     IMG="/boot/efi/$MACHINE_ID/$KERNEL_VERSION/initrd"
 else
-    echo "No initramfs image found to restore!"
-    exit 1
+    files=("/boot/initr*${KERNEL_VERSION}*")
```
Conan-Kudo, line 49: How is this guaranteed to work? Someone could set it to be `$KERNEL_VERSION-initramfs`...

man/dracut.conf.5.asc
```diff
@@ -320,6 +320,12 @@ Logging levels:
    If set to _yes_, try to execute tasks in parallel (currently only supported
    for _--regenerate-all_).
 
+*initrdname=*"_<filepattern>_"::
+    Specifies the file name for the generated initramfs if it is not set otherwise.
+    The initrdname configuration option is required to match the _initr*${kernel}*_
+    file pattern and only one file with this pattern should exists in the
+    directory where initramfs is loaded from.
```
Conan-Kudo, line 327: The default name needs to be documented.

man/dracut.8.asc
```diff
@@ -18,13 +18,8 @@ DESCRIPTION
...
 If <kernel version> is omitted, then the version of the actual running
-kernel is used. If <image> is omitted or empty, depending on bootloader
-specification, the default location can be
-_/efi/<machine-id>/<kernel-version>/initrd_,
-_/boot/<machine-id>/<kernel-version>/initrd_,
-_/boot/efi/<machine-id>/<kernel-version>/initrd_,
-_/lib/modules/<kernel-version>/initrd_ or
-_/boot/initramfs-<kernel-version>.img_.
+kernel is used. If <image> is omitted or empty, the default location will be
+determined by the local configuration or Linux distribution policy.
```
Conan-Kudo, line 22: This is not actually true. We are only adding customization for the file name, not the paths. So this is a bit excessive.

Conan-Kudo (changes requested):

dracut.sh
```diff
@@ -1093,6 +1093,13 @@ drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
 [[ $sbat_l ]] && sbat="$sbat_l"
 [[ $machine_id_l ]] && machine_id="$machine_id_l"
 
+if ! [[ $outfile ]] && [[ $initrdname ]]; then
+    outfilename="$initrdname"
```
aafeijoo-suse, line 1097: `outfilename` is assigned and never used.

dracut.sh
```diff
@@ -1093,6 +1093,13 @@ drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
...
+else
+    # shellcheck disable=SC2034
+    outfilename="initramfs-${kernel}.img"
```
aafeijoo-suse, line 1100: `outfilename` is assigned and never used.

dracut.sh
```diff
@@ -1093,6 +1093,13 @@ drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
 [[ $sbat_l ]] && sbat="$sbat_l"
 [[ $machine_id_l ]] && machine_id="$machine_id_l"
 
+if ! [[ $outfile ]] && [[ $initrdname ]]; then
```
aafeijoo-suse, line 1096: The condition `if [[ $initrdname ]]` should be moved within the next `if ! [[ $outfile ]]`, i.e., if `outfile` has any value, `initrdname` is useless.

aafeijoo-suse (changes requested):

man/dracut.8.asc
pvalena: I agree with the change- documentation is aimed at users; who either use the default, or can override it explicitly.

pvalena (approved):
LGTM.

Conan-Kudo (approved):

PR #161, opened 2024-04-13
ci: enable Void Linux for basic tests with runit

LaszloGombos:
## Changes

Void is an well maintained distribution that can enable Dracut to be tested for compatibility with runit . 

Only adding the test container is required, the test is already passing on Void Linux (due to how the CI is designed we will only see the pass after the PR lands).

@classabbyamp, @zdykstra and @ahesford has been a significant supporter of Dracut both via code contributions and PR reviews. 

## Checklist

### DIFF
--- a/test/container/Dockerfile-Void
+++ b/test/container/Dockerfile-Void
@@ -1,0 +1,44 @@
+FROM ghcr.io/void-linux/void-glibc-full
+
+RUN xbps-install -Syu xbps && xbps-install -yu \
+    asciidoc \
+    base-devel \
+    bash \
+    binutils \
+    btrfs-progs \
+    busybox \
+    cargo \
+    cpio \
+    cryptsetup \
+    curl \
+    dhclient \
+    dmraid \
+    dosfstools \
+    e2fsprogs \
+    edk2-ovmf \
+    eudev \
+    git \
+    gnupg \
+    gummiboot \
+    iputils \
+    kbd \
+    libkmod-devel \
+    linux \
+    lvm2 \
+    make \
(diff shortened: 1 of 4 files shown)


zdykstra:
This is excellent to see. Knowing that tests pass on Void (with it's lack of systemd) will be a significant win for Void users.

ahesford:
I don't know much about GitHub CI or the mech mechanics of Dracut tests, but it would be wonderful to see Void in the official test pipeline!

test/container/Dockerfile-Void
```diff
@@ -0,0 +1,43 @@
+FROM ghcr.io/void-linux/void-glibc-full
+
+RUN xbps-install -Syu \
```
classabbyamp, line 3:
```suggestion
RUN xbps-install -Syu xbps && xbps-install -yu \
```
ensure the container build works if xbps needs an update

test/container/Dockerfile-Void
```diff
@@ -0,0 +1,43 @@
...
+    open-iscsi \
+    openssh \
+    parted \
+    pigz \
+    procps-ng \
+    qemu \
+    squashfs-tools \
+    sudo \
+    util-linux \
+    xz \
```
classabbyamp, line 43:
```suggestion
    xz \
    && rm -rf /var/cache/xbps
```
remove any cached package files to save a little space

classabbyamp (commented):
thanks for adding this!

classabbyamp (approved):

Conan-Kudo (approved):

PR #167, opened 2024-04-13
fix(dracut.sh): include efi mountpoint for hostonly

dalto8:
When the ESP is mounted at /efi on it's own device, support for the underlying device is excluded from the initramfs.

Since /efi is an increasingly common location for the ESP to be mounted, it should probably be included here.

## Changes
Adds /efi to the list of mountpoints to check for block devices when using `hostonly`

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1664,7 +1664,8 @@ if [[ $hostonly ]] && [[ $hostonly_default_device != "no" ]]; then
         "/usr/lib64" \
         "/boot" \
         "/boot/efi" \
-        "/boot/zipl"; do
+        "/boot/zipl" \
+        "/efi"; do
         mp=$(readlink -f "$dracutsysrootdir$mp")
         mountpoint "$mp" > /dev/null 2>&1 || continue
         _dev=$(find_block_device "$mp")


aafeijoo-suse (commented):
IMHO this is wrong, please read https://github.com/dracutdevs/dracut/issues/2044

dalto8:
It seems like we should either have both `/efi` and `/boot/efi` or neither depending on which logic you prefer to follow.

Conan-Kudo (approved):
Not a fan of `/efi` but I am unfortunately aware of people using that mount point.

dalto8:
> Is this currently the case on some Arch installations ? Any pointers to code/installer that sets this up ? Thanks !

I think many of the distros using systemd-boot are using `/efi`.  I know that [EndeavourOS is](https://github.com/endeavouros-team/calamares/blob/calamares/src/modules/partition/PartitionViewStep.cpp#L467) which uses dracut by default.  It is also mentioned as one of the primary options in the [Arch wiki](https://wiki.archlinux.org/title/EFI_system_partition#Typical_mount_points) so many Arch users use it.  Gentoo makes the [same recommendation](https://wiki.gentoo.org/wiki/EFI_System_Partition).

PR #169, opened 2024-04-13
ci: enable testing on arm64 

LaszloGombos:
Enable testing on arm64 using the Debian container.

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -29,15 +29,19 @@ jobs:
             fail-fast: false
             matrix:
                 config:
-                    - { dockerfile: 'Dockerfile-Fedora-latest',     tag: 'fedora:latest' }
-                    - { dockerfile: 'Dockerfile-OpenSuse-latest',   tag: 'opensuse:latest' }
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
I would prefer to have Fedora and Debian tested in both architectures.

Conan-Kudo:
We could use native AArch64 machines couldn't we?

bdrung:
I have no clear preference whether the arm64 tests should run on Debian or Ubuntu, but please keep the amd64 tests. Then it will be easier to nail down failures (i.e. is the failure on the distro or the arch?).

bdrung:
AArch64 runners are in beta: https://github.blog/changelog/2023-10-30-accelerate-your-ci-cd-with-arm-based-hosted-runners-in-github-actions/

A note from my tests to run the tests on other architectures: The IDE device needs to be replaced by virtio. See https://github.com/dracutdevs/dracut/issues/2493

Conan-Kudo:
No, I think you're fine, but let's not make Debian not get tested on x86_64. Maybe a second matrix for ARM?

Conan-Kudo (approved):

PR #170, opened 2024-04-14
ci: gummiboot-stub no longer supports UEFI image with latest kernel

### LINKED ISSUE #139: test 18 fails on Alpine
https://github.com/dracut-ng/dracut-ng/actions/runs/8592248172

```
Initramfs unpacking failed: invalid magic at start of compressed archive
```

LaszloGombos:
## Changes

Remove ovmf from Alpine container to explicitly skip UEFI test.
Replace gummiboot with systemd-boot-efistub in the Void container.
Be consistent and remove sudo from all test containers.

## Checklist

Fixes #139

CC @classabbyamp


Conan-Kudo (approved):
Huh, I didn't think gummiboot was still around...

PR #171, opened 2024-04-14
fix(network-manager,dbus): improve dependency checking

LaszloGombos:
## Changes

Only network-manager requires busctl. Some distributions do not package busctl as part of dbus package.

This allows dbus dracut module to be installed even if busctl is not available to satisfy e.g. bluetooth module installation requirements.

## Checklist

### DIFF
--- a/modules.d/06dbus-broker/module-setup.sh
+++ b/modules.d/06dbus-broker/module-setup.sh
@@ -6,7 +6,6 @@
 check() {
 
     # If the binary(s) requirements are not fulfilled the module can't be installed
-    require_binaries busctl || return 1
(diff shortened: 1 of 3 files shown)


aafeijoo-suse:
> AFAIK `busctl` is just a tool to manually introspect the bus from the emergency shell added for completeness with the dracut dbus modules, so no one "requires" it.

Ok, I just saw the why:

https://github.com/dracut-ng/dracut-ng/blob/3107d80cfc538278b37926ed95ca23c49f5c7f02/modules.d/35network-manager/nm-initrd.service#L18

Anyway, I would not remove it from the dbus modules. The main action should be to get rid of dbus of fix it.

PR #172, opened 2024-04-14
fix(dracut-init.sh): force to perform the actual action

LaszloGombos:
## Changes

Be consistent and always perform the actual action you would expect from the name.


Conan-Kudo (approved):
I could go either way on this, but it doesn't seem to be _bad_, so sure.

PR #174, opened 2024-04-14
test(LVM-THIN): make sure dm_thin_pool Linux kernel module is loaded

### LINKED ISSUE #173: LVM-THIN fails on Void
Test [TEST-17-LVM-THIN](https://github.com/dracut-ng/dracut-ng/blob/main/test/TEST-17-LVM-THIN/test.sh) is the only failing test on Void. Could be an issue with the test case code itself, or we might need to skip this test on Void.

Error:

```
thin-pool: Required device-mapper target(s) not detected in your kernel.
```

Full log: https://github.com/dracut-ng/dracut-ng/actions/runs/8679836562

Perhaps `modprobe dm_thin_pool`

LaszloGombos:
It might be that the lvm dracut module should be improved instead of working around it in the test.

Regardless, I wanted to share what it takes to pass this test on Void.

## Checklist

Fixes #173

CC @classabbyamp

### DIFF
--- a/test/TEST-17-LVM-THIN/create-root.sh
+++ b/test/TEST-17-LVM-THIN/create-root.sh
@@ -11,6 +11,7 @@ udevadm control --reload
 udevadm settle
 
 set -ex
+modprobe dm_thin_pool
 for dev in /dev/disk/by-id/ata-disk_disk[123]; do
     lvm pvcreate -ff -y "$dev"
 done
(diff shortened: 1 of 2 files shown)


classabbyamp:
thanks for keeping us downstreams in the loop, but I think for most things it'd be more relevant to tag the maintainer of void's dracut package, @zdykstra

Conan-Kudo (approved):

PR #175, opened 2024-04-14
feat(systemd-bsod): dracut module for systemd-bsod

LaszloGombos:
## Changes

Include https://www.freedesktop.org/software/systemd/man/latest/systemd-bsod.service.html

bsod depends on qrencode. Add qrencode to Arch and Fedora container to trigger testing.

While at it add astyle, jq to the Fedora container as well as it was missing for testing some of the dracut modules.

## Checklist

### DIFF
--- a/modules.d/01systemd-bsod/module-setup.sh
+++ b/modules.d/01systemd-bsod/module-setup.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Prerequisite check(s) for module.
+check() {
+    # If the binary(s) requirements are not fulfilled the module can't be installed
+    require_binaries "$systemdutildir"/systemd-bsod || return 1
+
+    # Return 255 to only include the module, if another module requires it.
+    return 255
+}
+
+# Module dependency requirements.
+depends() {
+    # This module has external dependency on other module(s).
+    echo systemd-journald
+    # Return 0 to include the dependent module(s) in the initramfs.
+    return 0
+}
+
+# Install the required file(s) for the module in the initramfs.
+install() {
+    inst_multiple \
+        "$systemdsystemunitdir"/systemd-bsod.service \
+        "$systemdsystemunitdir"/initrd.target.wants/systemd-bsod.service \
+        "$systemdutildir"/systemd-bsod
+
+    inst_libdir_file "libqrencode.so*"
+}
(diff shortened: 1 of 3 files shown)


aafeijoo-suse (changes requested):
> ## Checklist
> 
>     * [x]  I have tested it locally

This cannot work without adding the binary executed by the new service and its required libraries.

I would also make this optional, creating a new module only for this or adding it to the systemd-journald module.

modules.d/01systemd-bsod/module-setup.sh
```diff
@@ -0,0 +1,29 @@
...
+    echo systemd
+    # Return 0 to include the dependent module(s) in the initramfs.
+    return 0
+}
+
+# Install the required file(s) for the module in the initramfs.
+install() {
+    inst_multiple -o \
+        "$systemdsystemunitdir"/systemd-bsod.service \
+        "$systemdsystemunitdir"/sysinit.target.wants/systemd-bsod.service \
```
aafeijoo-suse, line 26: `s/sysinit.target.wants/initrd.target.wants/`

modules.d/01systemd-bsod/module-setup.sh
```diff
@@ -0,0 +1,29 @@
...
+        "$systemdutildir"/systemd-bsod \
+        qrencode
```
aafeijoo-suse, line 28: `systemd-bsod` does not require or use the `qrencode` binary, it calls `libqrencode` via dlopen, so you should add it with `inst_libdir_file`.

modules.d/01systemd-bsod/module-setup.sh
```diff
@@ -0,0 +1,29 @@
...
+    require_binaries "$systemdutildir"/systemd-bsod || return 1
+
+    # Return 255 to only include the module, if another module requires it.
+    return 255
+}
+
+# Module dependency requirements.
+depends() {
+    # This module has external dependency on other module(s).
+    echo systemd
```
aafeijoo-suse, line 17: `systemd-bsod.service` is just continuously watching the journal, even its source code is [on the same level of hierarchy](https://github.com/systemd/systemd/tree/main/src/journal), so it should depend only on `systemd-journald`.

We should start actually using the available 01systemd modules in dracut and break out the 00systemd monster. Maybe I submit a PR with that.

aafeijoo-suse (changes requested):

aafeijoo-suse (approved):
Module working as expected.

Conan-Kudo (approved):

PR #177, opened 2024-04-14
fix(lvm): include 69-dm-lvm.rules

LaszloGombos:
See https://github.com/lvmteam/lvm2/blob/main/udev/69-dm-lvm.rules.in

CC @teigland @mwilck

### DIFF
--- a/modules.d/90lvm/module-setup.sh
+++ b/modules.d/90lvm/module-setup.sh
@@ -86,6 +86,10 @@ install() {
     # debian udev rules
     inst_rules 56-lvm.rules 60-persistent-storage-lvm.rules
 
+    if dracut_module_included "systemd"; then
+        inst_multiple -o "$udevrulesdir"/69-dm-lvm.rules
+    fi
+
     inst_script "$moddir/lvm_scan.sh" /sbin/lvm_scan
     inst_hook cmdline 30 "$moddir/parse-lvm.sh"
 


teigland:
This udev rule implements general LV autoactivation, which has not been wanted in the initrd, and conflicts with the existing direct activation of LVs named by rd.lvm.lv.

mwilck:
> This udev rule implements general LV autoactivation, which has not been wanted in the initrd, and conflicts with the existing direct activation of LVs named by rd.lvm.lv.

What exactly is the conflict? Can anything go badly wrong if an LV is activated more than once?

IMO autoactivation in the initramfs has its pros and cons. On systems with a huge number of LVs, you often want to limit the activation to what's necessary to boot the system. OTOH, relying on hard-coded or command-line passed `rd.lvm.lv` makes the setup unflexible. How does it work with non-hostonly initrds? I observe that there's no `rd.lvm.lv=*` or option, or anything like that that would override the hard-coded behavior.

teigland:
> I noticed that mkinitcpio installs this rule - https://github.com/archlinux/mkinitcpio/blob/master/install/lvm2#L39 . Do you consider that as a bug in mkinitcpio (apologies for the slightly out of scope question).

I'm not familiar with what's happening in that url, so I can't really say.  An initrd scheme could be designed to use general LV autoactivation, and maybe that's what's going on.   But, there are problems with doing that, e.g. advanced LV types can't be managed well in the initrd, and activating many non-root LVs can delay finding and activating the actual root LV (to the point of boot timing out.)

teigland:
> What exactly is the conflict? Can anything go badly wrong if an LV is activated more than once?

Someone would have to try it to find out, and it might even "work" for the most part, but running two parallel activation schemes would be a mess of a design and something we'd not support doing.

> IMO autoactivation in the initramfs has its pros and cons. On systems with a huge number of LVs, you often want to limit the activation to what's necessary to boot the system. OTOH, relying on hard-coded or command-line passed `rd.lvm.lv` makes the setup unflexible. How does it work with non-hostonly initrds? I observe that there's no `rd.lvm.lv=*` or option, or anything like that that would override the hard-coded behavior.

I largely agree, and I'd be interested in reviewing any new designs in this area.   I made some comments some years ago about possibly using partition types, or special reserved LV names to select root LVs without needing to name them on the command line.

PR #180, opened 2024-04-15
feat(tpm2-tss): add tpm2.target and systemd-tpm2-generator

dracutng:
## Changes

https://github.com/systemd/systemd/commit/4e1f0037 added a new `tpm2.target`, which is now used by `systemd-pcrphase-initrd.service`, and `systemd-tpm2-generator`, which adds a `Wants=` dependency from `sysinit.target` to `tpm2.target` when it detects that the firmware discovered a TPM2 device but the kernel didn't.

## Checklist

Fixes #


aafeijoo-suse:
image

I think there is no point in reviewing my own patch, even if a bot grabs it from another repository. Someone else should do it.

PR #197, opened 2024-04-18
perf(systemd-udevd): remove duplicate rules

LaszloGombos:
## Changes

During a typical systemd-enabled initrd generation both [udev-rules](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/95udev-rules/module-setup.sh) and systemd-udevd gets installed. This leads to duplicate file copying, that can be easily avoided just by removing the duplicate rules.

The performance benefit of this PR is negligent. More importantly this PR makes it easier to read, reason and maintain the code.

Added explicit dependency to make sure there is no regression for the `dracut -m systemd-udevd` invocation.

## Checklist

### DIFF
--- a/modules.d/01systemd-udevd/module-setup.sh
+++ b/modules.d/01systemd-udevd/module-setup.sh
@@ -20,7 +20,7 @@ check() {
 depends() {
 
     # This module has external dependency on other module(s).
-    echo systemd systemd-sysctl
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
This one needs rebasing to resolve conflicts.

Conan-Kudo (approved):

PR #198, opened 2024-04-18
fix(test): fixup Gentoo CI

### LINKED ISSUE #135: systemd enabled Gentoo container fails tests

LaszloGombos:
Improve systemd installation by enabling cryptsetup flag.

Remove non-systemd craft from Gentoo, since now the project has Alpine and Void container to test with.

Add some extra packages to enable more test coverage.

## Checklist

Fixes #135

### DIFF
--- a/test/container/Dockerfile-Gentoo
+++ b/test/container/Dockerfile-Gentoo
@@ -1,70 +1,53 @@
-# ARG TAG=musl # to enable musl instead of glibc
-
-ARG TAG=systemd
-
 FROM docker.io/gentoo/portage:latest as portage
 
-# kernel and its dependencies in a separate builder
-FROM docker.io/gentoo/stage3:$TAG as kernel
+FROM docker.io/gentoo/stage3:systemd
 COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo
 
 # Speed-up using binpkgs
 RUN echo "MAKEOPTS=\"-j$(nproc) -l$(nproc)\"" >> /etc/portage/make.conf
 RUN echo "EMERGE_DEFAULT_OPTS=\"-j$(nproc) -l$(nproc)\"" >> /etc/portage/make.conf
 RUN echo "FEATURES=\"getbinpkg binpkg-ignore-signature parallel-fetch parallel-install pkgdir-index-trusted\"" >> /etc/portage/make.conf
+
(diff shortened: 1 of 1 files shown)


test/container/Dockerfile-Gentoo
```diff
@@ -1,70 +1,53 @@
-# ARG TAG=musl # to enable musl instead of glibc
-
-ARG TAG=systemd
-
 FROM docker.io/gentoo/portage:latest as portage
 
-# kernel and its dependencies in a separate builder
-FROM docker.io/gentoo/stage3:$TAG as kernel
+FROM docker.io/gentoo/stage3:systemd
```
AndrewAmmerlaan, line 3: Ideally we would test both openrc and systemd, but not sure if this is feasible. Some tests will require functionality from systemd, but most should work in an openrc container as well. And openrc is still more or less "the default" for Gentoo.

AndrewAmmerlaan: > Our CI currently not able to handle the '2 containers from one Dockerfile'.

What about making two copies of the Dockerfile, `Dockerfile-Gentoo-systemd` and `Dockerfile-Gentoo-openrc`?

> Given the situation, would you want systemd or OpenRC for Gentoo ?

If we must choose, then I'd personally pick systemd because this should allow testing all features, whereas not everything is going to be possible with openrc.  Though I suspect someone will eventually yell at me for saying this.

AndrewAmmerlaan (approved):
Alright let's do it like this then for now. And then maybe later we can revisit this and make a second docker file for openrc based Gentoo.

Conan-Kudo (approved):
I think this is a good plan, maybe one of the @dracut-ng/gentoo-maint folks can contribute an openrc variant to add to the suite.

PR #205, opened 2024-04-20
ci: run GETARG test on arm64

LaszloGombos:
## Changes

This is an easy test to start with as it does not require a VM.

Only set CPU to IvyBridge-v2 for AMD64.

## Checklist

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -210,3 +210,28 @@ jobs:
 
             -   name: "${{ matrix.container }} TEST-${{ matrix.test }}"
                 run: ./tools/test-github.sh "TEST-${{ matrix.test }}" ${{ matrix.test }}
+    arm64:
+        name: ${{ matrix.test }} on ${{ matrix.container }} on arm64
+        runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/integration.yml
```diff
@@ -210,3 +210,28 @@ jobs:
...
+        name: ${{ matrix.test }} on ${{ matrix.container }} on arm64
+        runs-on: ubuntu-latest
+        timeout-minutes: 30
+        concurrency:
+            group: arm64-${{ github.workflow }}-${{ github.ref }}-${{ matrix.container }}-${{ matrix.test }}
+            cancel-in-progress: true
+        strategy:
+            matrix:
+                container: [
+                        "debian",
```
Conan-Kudo, line 223: Can we put Fedora in here too?

Conan-Kudo (changes requested):

.github/workflows/integration.yml
Conan-Kudo: Well, we can fix that in a separate PR then.

Conan-Kudo (approved):

PR #213, opened 2024-04-21
doc: remove incomplete and distribution specific information

LaszloGombos:
I find it confusing to try to list incomplete and distribution specific package names.

Incorporate README.generic into man/dracut.usage.asc.


Conan-Kudo:
I think it probably makes sense to cover this in some meaningful way in the man pages.

Conan-Kudo (approved):

Issue #60, opened 2024-04-01
dbus in initramfs concerns 


LaszloGombos:
There has been some concerns raised about the consequences of including dbus and stystemd together in the generated initramfs. 

This could potentially impact dracut modules that pulls in dbus dracut module as a dependency such as [network-manager](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/35network-manager/module-setup.sh#L13) dracut module.

https://github.com/dracutdevs/dracut/issues/2378

LaszloGombos:
CC @BtbN for visibility

PR #218, opened 2024-04-21
fix(systemd-networkd): dbus is not a mandatory dependency

LaszloGombos:
## Changes

systemd-networkd does not require dbus, see https://github.com/dracutdevs/dracut/issues/2378#issuecomment-1668136681 .

Now that the project has a systemd-networkd [test case](https://github.com/dracut-ng/dracut-ng/pull/147), it is observable that dropping dbus from the generated initrd does not break the test case.

## Checklist

Fixes (partially): #60 https://github.com/dracutdevs/dracut/issues/2378

### DIFF
--- a/modules.d/01systemd-networkd/module-setup.sh
+++ b/modules.d/01systemd-networkd/module-setup.sh
@@ -22,7 +22,7 @@ check() {
 depends() {
 
     # This module has external dependency on other module(s).
-    echo dbus kernel-network-modules systemd-sysusers
+    echo kernel-network-modules systemd-sysusers
     # Return 0 to include the dependent module(s) in the initramfs.
     return 0
 


Conan-Kudo (approved):

aafeijoo-suse (commented):
As has already been said here https://github.com/dracutdevs/dracut/pull/2476#pullrequestreview-1566686395, this module is requiring a binary which requires dbus.

https://github.com/dracut-ng/dracut-ng/blob/c6906fea4df28428aa889c7bb8197689ae8107ee/modules.d/01systemd-networkd/module-setup.sh#L10

Pasting the actual error, because apparently it's hard to believe.

```
sh-5.2# networkctl
Failed to connect to system bus: No such file or directory
```

That said, this doesn't mean that `networkctl` is needed for systemd-networkd to work.

PR #223, opened 2024-04-22
Remove False Positive Errors When Using 'livenet' And Squashfs

dracutng:
## Changes

Remove the following false positive error messages when using a squashfs image using the livenet module

`modprobe` can't `squashfs` driver when actually it is built-in in the kernel
`dmsquash-live-root` trying to verify the ISO checksum for a squashfs filesystem file image

## Checklist

Fixes 
- One fix was provided and integrated long ago, but it got broken again. See d0cd7cd

(cherry picked from dracutdevs/dracut#2595)


FedericoVaga:
Hi @LaszloGombos . It looks OK to me, it is the same patch I pushed to `dracut`.

PR #235, opened 2024-04-24
fix(numlock): use the same shebang as other dracut modules

LaszloGombos:
## Changes

Fix what was pointed out at https://github.com/dracut-ng/dracut-ng/pull/226#pullrequestreview-2019137318

## Checklist

### DIFF
--- a/modules.d/90numlock/module-setup.sh
+++ b/modules.d/90numlock/module-setup.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/bash
+#!/bin/bash
 # This file is part of dracut.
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

aafeijoo-suse (commented):
> Fix what was pointed out at https://github.com/dracut-ng/dracut-ng/pull/226#pullrequestreview-2019137318

Ah, then there is still another one in this module.

https://github.com/dracut-ng/dracut-ng/blob/efa0268881df36824c72020e6f8b8966cc189375/modules.d/90numlock/numlock.sh#L1

PR #238, opened 2024-04-24
ci: move the ISCSI test to run only on systemd-networkd

### LINKED ISSUE #240: test 30 is flaky with network-manager and occasionally with network-legacy
As seen - e.g. https://github.com/dracut-ng/dracut-ng/pull/234#pullrequestreview-2019740967

LaszloGombos:
## Changes

Disable two of the four test scenarios (the non-`netroot=iscsi` cases) for the ISCSI test and moved the ISCSI test to run only on systemd-networkd (and not on network-manager or network-legacy). This resolved the flakiness with the ISCSI test.

## Checklist

Fixes #240

### DIFF
--- a/test/TEST-30-ISCSI/test.sh
+++ b/test/TEST-30-ISCSI/test.sh
@@ -78,10 +78,11 @@ run_client() {
 do_test_run() {
     initiator=$(iscsi-iname)
 
-    run_client "root=dhcp" \
-        "root=/dev/root netroot=dhcp ip=enp0s1:dhcp" \
-        "rd.iscsi.initiator=$initiator" \
-        || return 1
+    # TODO: does not work well
+    # run_client "root=dhcp" \
+    #     "root=/dev/root netroot=dhcp ip=enp0s1:dhcp" \
+    #     "rd.iscsi.initiator=$initiator" \
+    #     || return 1
 
     run_client "netroot=iscsi target0" \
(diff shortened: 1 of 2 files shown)


lnykryn:
I don't know if this is still valid, but in the past I used NM and networked on one machine. NM for "main" networking and networked for systemd-nspawn instances.
From my point of view NM in initrd gets much more testing than networkd, so I am not sure I like this change, on the other hand, we are not building networkd at all in rhel, so I don't have strong opinion here.

test/TEST-30-ISCSI/test.sh
```diff
@@ -78,10 +78,11 @@ run_client() {
...
 
-    run_client "root=dhcp" \
-        "root=/dev/root netroot=dhcp ip=enp0s1:dhcp" \
-        "rd.iscsi.initiator=$initiator" \
-        || return 1
+    # TODO: does not work well
+    # run_client "root=dhcp" \
+    #     "root=/dev/root netroot=dhcp ip=enp0s1:dhcp" \
+    #     "rd.iscsi.initiator=$initiator" \
+    #     || return 1
```
Conan-Kudo, line 85: I'm not super-stoked about keeping around TODOs like this. Can we capture this some other way to get fixed properly?

Conan-Kudo (changes requested):

PR #254, opened 2024-04-27
feat(dracut.sh): make hostonly default

LaszloGombos:
## Changes

hostonly is the recommended default configuration. It should be the default even without a configuration file.

## Checklist

[Fedora](https://src.fedoraproject.org/rpms/dracut/blob/rawhide/f/dracut.spec#_253) and [openSUSE](https://github.com/openSUSE/dracut/blob/SUSE/059/suse/dracut.spec#L164) package a configuration file that makes hostonly the default for those distributions. Some other distributions (e.g. Gentoo or Arch) however do not do this (it is unknown if this is intentional or not). 

Some distributions have strong desire to NOT customize upstream - which in dracut's case could lead to software is not used in its recommended configuration by the users.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1051,6 +1051,7 @@ drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
 [[ $prefix_l ]] && prefix=$prefix_l
 [[ $prefix == "/" ]] && unset prefix
 [[ $hostonly_l ]] && hostonly=$hostonly_l
+[[ $hostonly ]] || hostonly="-h"
 [[ $hostonly_cmdline_l ]] && hostonly_cmdline=$hostonly_cmdline_l
 [[ $hostonly_mode_l ]] && hostonly_mode=$hostonly_mode_l
 [[ $hostonly == "yes" ]] && ! [[ $hostonly_cmdline ]] && hostonly_cmdline="yes"
--- a/man/dracut.conf.5.asc
+++ b/man/dracut.conf.5.asc
@@ -119,7 +119,7 @@ Configuration files must have the extension .conf; other extensions are ignored.
 *hostonly=*"__{yes|no}__"::
     Host-only mode: Install only what is needed for booting the local host
     instead of a generic host and generate host-specific configuration
-    (default=no).
+    (default=yes).
 
 *hostonly_mode=*"__{sloppy|strict}__"::
     Specify the host-only mode to use (default=sloppy).


zdykstra (commented):
I'm firmly against hostonly being the default for users.  The only tangible benefit, and it's a modest one at that, is that you potentially save a small amount of space for each initramfs that's built at the cost of flexibility when hardware changes.

From a Void perspective, we'll just need to account for this in our ISO build scripts to ensure that a full-fat initramfs is built.  It's possible that we'd just do the opposite of Fedora and openSUSE and patch things to have hostonly off by default in the package.

freswa:
From an Arch perspective I second what @zdykstra wrote. Also hostonly sometimes lead to broken initramfs when new kernel features are not aligned with a corresponding dracut commit and release. So with unexperienced users, we'll run into more support requests with a questionable gain in speed (?). Experienced users can benefit from hostonly quite easily, they just need to rtfm :)

So we'd prefer the current `hostonly=false` default

dalto8:
There are definitely pros and cons to using `hostonly` vs using `nohostonly`.

However, if this is true:
> hostonly is the recommended default configuration

Why would the recommended default ever not be shipped as the default?  Distros can change the default in their packaging if they prefer different behavior.

pvalena:
I agree here from the standpoint of not changing the defaults which are here for years, as it might affect user experience. From distro perspective, Fedora sets it as default in configuration, which is more transparent to users as well. Not sure what's the benefit here?

lnykryn:
I don't think this will impact Fedora or RHEL since the built-in default is always overridden by the config in /usr.
But as @zdykstra wrote, this can be a problem for iso-building tools.

bdrung (commented):
The Ubuntu standpoint is to have `hostonly=false` as default: We want that user can replace hardware and still be able to boot. Taking a disk from one machine, plug it into a different machine, and boot should work out of the box.

Switching to `hostonly=true` should be a deliberate decision by the user. Experienced users are expected to know how to rebuild their initrd when they fiddle around with their hardware setup.

Conan-Kudo:
From my perspective as an upstream person, I don't think it makes sense to change the default. The idea is that the defaults should always _do the right thing_ and hostonly mode is mainly useful in distributions that have the requisite integration that generates the initramfs correctly for the target environment. But dracut can be used to pregenerate initramfs images (as is done when using dracut to create UKIs, live USB, and rescue images), and those should be generic images instead.

There's also a long story of expectations from third party tools, and changing this isn't worth the breakage.

Conan-Kudo:
I'm closing this as it's pretty clear we don't want this.

jozzsi:
> The Ubuntu standpoint is to have `hostonly=false` as default: We want that user can replace hardware and still be able to boot. Taking a disk from one machine, plug it into a different machine, and boot should work out of the box.

Coming here from https://github.com/dracut-ng/dracut-ng/issues/674, which is another manifestation of the fact that even experienced dracut users find non-hostonly confusing.

Hostonly actually has two modes, mainly for the reason mentioned by @bdrung  - see [ --hostonly-mode](https://github.com/dracut-ng/dracut-ng/blob/main/man/dracut.8.asc?plain=1#L362) . Hostonly does not imply that user making simple storage changes would lead to unbootable system - only the `--hostonly-mode=strict` option does that (which is not the default even in hostonly mode).

Non-hostonly mode really is only for the live-iso boot use-case.

bdrung:
I have to revisit my previous comment, because I misunderstood the meaning of `hostonly=false`. Now I support making hostonly the default and to have a hostonly mode that supports major hardware changes: https://github.com/dracut-ng/dracut-ng/issues/748

PR #257, opened 2024-04-29
feat(90crypt): add support for AF_UNIX key files

kszczek:
systemd v248 introduced support for using AF_UNIX stream sockets as key files in `/etc/crypttab`. This PR enhances the 90crypt module to identify socket units with matching socket file paths. It then includes the first matching socket unit along with its corresponding service unit. This correspondence is determined by checking the `Service=` option in the socket unit or, if that's not available, by replacing the `.socket` suffix with either `@.service` or `.service`, depending on the `Accept=` option (see man page for systemd.socket).

Futhermore, this functionality handles sockets located under `/run/cryptsetup-keys.d/`, which are automatically discoverable by the systemd-cryptsetup utility when the key file field in `/etc/crypttab` is empty.

### `/usr/local/bin/get-secret`
```bash
#!/bin/bash
echo -n "secret"
```

### `/etc/dracut.conf.d/get-secret.conf`
```plaintext
install_items+=" /usr/local/bin/get-secret "
```

### `/etc/systemd/system/root-volume-key.socket`
```plaintext
[Unit]
Description=Root Volume Key Socket
DefaultDependencies=no
Conflicts=multi-user.target shutdown.target
Before=cryptsetup-pre.target multi-user.target shutdown.target
Wants=cryptsetup-pre.target

[Socket]
ListenStream=/run/cryptsetup-keys.d/luks-<UUID>.key
SocketUser=root
SocketGroup=root
SocketMode=0400
DirectoryMode=0750
Accept=yes
MaxConnections=1
RemoveOnStop=yes

[Install]
WantedBy=sockets.target
```

### `/etc/systemd/system/root-volume-key@.service`
```plaintext
[Unit]
Description=Root Volume Key Service
DefaultDependencies=no
After=root-volume-key.socket
Requisite=root-volume-key.socket
RefuseManualStart=yes
RefuseManualStop=yes

[Service]
Type=exec
ExecStart=/bin/bash /usr/local/bin/get-secret
StandardOutput=socket
StandardError=kmsg+console
```

### `/etc/crypttab`
```plaintext
luks-<UUID> UUID=<UUID> none discard
```

Also, make sure to enable the `cryptsetup-keyfile.socket` before regenerating the initramfs.

## Checklist


Conan-Kudo:
This seems reasonable, but I'd like a second opinion. @DaanDeMeyer, could you take a look?

modules.d/90crypt/module-setup.sh
```diff
@@ -92,6 +92,73 @@ cmdline() {
...
+        local volume_name=$1
+        local socket_path=$2
+
+        # ignore paths followed by a device specification
+        if [[ "$socket_path" == *":"* ]]; then
+            return
+        fi
+
+        # if no explicit path is provided, try to include units for auto-discoverable keys
+        if [[ "$socket_path" = "-" || "$socket_path" = "none" ]]; then
```
Conan-Kudo, line 106: Can you use `==` here too? Also please split this into `[[ ... ]] || [[ ... ]]`

Conan-Kudo (changes requested):

modules.d/90crypt/module-setup.sh
kszczek: Done. Also dropped the quotes around `"$socket_path"`, as linter also complained about this.

DaanDeMeyer:
I'm neither familiar enough with cryptsetup nor dracut to give any real feedback on this :(

kszczek:
Should be fixed. Might be worth considering to let the "low cost" workflows run without approval, like the commit message checker or linter. This would reduce the back and forth.

pothos:
I think that won't be used in Flatcar because Flatcar ships a prebuilt initrd which has no crypttab enries and relies on invoking the systemd-crypttab-generator directly for a synthesized entry only visible to it.

Conan-Kudo (approved):

PR #262, opened 2024-05-02
feat(systemd-cryptsetup): new module for systemd-cryptsetup

LaszloGombos:
Separate out systemd dependent part of crypt into it's own module.

This new `systemd-cryptsetup` module is dependent on the `crypt` module, so it inherits all the `check()` and `install()` etc functionality, that we do not need to duplicate.

## Checklist

### DIFF
--- a/modules.d/90crypt/module-setup.sh
+++ b/modules.d/90crypt/module-setup.sh
@@ -31,9 +31,6 @@ depends() {
             deps+=" tpm2-tss"
         fi
     fi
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
@DaanDeMeyer Could you please look at this too?

Conan-Kudo:
I think this split makes sense, but I'm not sure about the name of the module (it might make sense to call it `systemd-cryptsetup`)...

Conan-Kudo (approved):

PR #270, opened 2024-05-03
fix(install.d): Fix failure when kernel-install command is not `add`

chenxiaolong:
PR #195 added `:?` checks to ensure that required variables are set, but this is done too early when using kernel-install's other modes, like `kernel-install remove <version>`. This commit moves the `$COMMAND` check to the top before other variables are set.


install.d/51-dracut-rescue.install
```diff
@@ -3,15 +3,16 @@
...
-KERNEL_IMAGE="${4:?}"
 
 # If the initrd was provided on the kernel command line, we shouldn't generate our own.
 if [[ "$COMMAND" = "add" && "$#" -gt 4 ]]; then
     exit 0
 fi
 
+KERNEL_VERSION="${2:?}"
+BOOT_DIR_ABS="${3%/*}/0-rescue"
+KERNEL_IMAGE="${4:?}"
```
AndrewAmmerlaan, line 14: This won't help here because this script also runs on `kernel-install remove` (unlike `50-dracut.install`, note that the if statement above is slightly different)

I suggest to instead simply remove the `:?` on `KERNEL_IMAGE` which is the one that is not going to be set when we call remove. Systemd's `90-loadentry.install` has this, so doing the same here should be fine:
```
COMMAND="${1:?}"
KERNEL_VERSION="${2:?}"
ENTRY_DIR_ABS="${3:?}"
KERNEL_IMAGE="$4"
```

And then for symmetry reasons I would suggest to also fix the problem in this same way in `50-dracut.install`

AndrewAmmerlaan (commented):
Oops, this is my fault. I tested installing but forgot about also testing removing. Thanks for the fix.

install.d/51-dracut-rescue.install
chenxiaolong: Thanks! I overlooked that `51-dracut-rescue.install` had a slightly different `if` statement. I thought I tested it, but I forgot I had symlinked `/etc/kernel/install.d/51-dracut-rescue.install` to `/dev/null` (whoops).

This should be fixed now.

install.d/50-dracut.install
```diff
@@ -3,8 +3,8 @@
 COMMAND="${1:?}"
 KERNEL_VERSION="${2:?}"
 #shellcheck disable=SC2034
-BOOT_DIR_ABS="$3"
-KERNEL_IMAGE="${4:?}"
+BOOT_DIR_ABS="${3:?}"
```
AndrewAmmerlaan, line 6: We don't actually use `BOOT_DIR_ABS` in this script, since the images we generate are copied to the staging area. So we don't have to enforce that $3 is set.

This is also why we have `#shellcheck disable=SC2034` here, shellcheck has a minor complaint about the variable not being used otherwise. For symmetry reasons I kept it in anyway, we do use it in `51-dracut-rescue.install`.

chenxiaolong: That makes sense. Fixed!

AndrewAmmerlaan (approved):

chenxiaolong:
Pushed one more fix for the commit message due to:
```
   Error: [C015] Description should not start with a repetition of the tag
  fix(install.d): fix failure when kernel-install command is not `add`
                  ^~~
```

AndrewAmmerlaan (approved):

PR #275, opened 2024-05-03
fix(35-network-manager): let the kernel generate a UUID for /etc/machine-id

AndrewAmmerlaan:
Original PR: https://github.com/dracutdevs/dracut/pull/534 CC @floppym 
Downstream bug: https://bugs.gentoo.org/677554

This allows the module to work without systemd.


floppym:
The dependency on dbus was introduced in 112f03f9e225a790cbc6378c70773c6af5e7ee34, a couple years after https://github.com/dracutdevs/dracut/pull/534 was initially opened.

So yes, network-manager currently requires systemd, but that was not previously the case.

AndrewAmmerlaan:
In Gentoo `net-misc/networkmanager` always depends on `sys-apps/dbus`, however `sys-app/dbus` only depends on `sys-apps/systemd` if the systemd flag is enabled. So perhaps the best strategy here is to try and break the dependency on systemd in the dbus module. I have not yet looked into the details though.

PR #276, opened 2024-05-03
fix(dracut): add support for RISC-V EFI

LaszloGombos:
## Changes

From: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/patches/dracut/dracut-riscv64-efi.patch

## Checklist

CC @valentindavid


Conan-Kudo:
Can you change this commit to be authored by the original author and drop the unstable link from the commit message? I think retaining the original author should be sufficient attribution.

Conan-Kudo (approved):

Issue #282, opened 2024-05-06
Gentoo container no longer builds (starting May5)


LaszloGombos:
https://github.com/dracut-ng/dracut-ng/actions/workflows/container.yml

CC @AndrewAmmerlaan

AndrewAmmerlaan:
At first glance this seems to be caused by the 17 to 23 profile upgrade, we might have a mismatch in the profile of the docker image versus the profile of the binpkg host. Will investigate later.

PR #283, opened 2024-05-06
fix(Dockerfile-Gentoo): add --deep

AndrewAmmerlaan:
This should help resolving the depgraph when the binhost pkgs require newer glibc then the docker image provides.

Also remove conflicting --quiet and --verbose, keep only --quiet

Closes: https://github.com/dracut-ng/dracut-ng/issues/282


AndrewAmmerlaan:
@LaszloGombos This resolves the problem, but there is still something weird going on. The docker images are still on the 17.1 profiles, the binhost is set correctly to the 17.1 version, yet somehow the binpkg we fetch from this host requires profile 23.0 features.

It's a problem on our end though, I'll ask around to see if we can resolve it somehow.

PR #297, opened 2024-05-19
feat: only install network modules in non-hostonly mode if requested

### LINKED ISSUE #132: test 12 on Ubuntu (only) requires explicitly omitting dbus to keep the test passing.
Discovered by https://github.com/dracut-ng/dracut-ng/pull/97

LaszloGombos:
## Changes

In most setups, network and related modules are not needed and only introduce potential extra problems (example https://github.com/dracut-ng/dracut-ng/issues/132). If you need the network modules in non-hostonly mode, just include it explicitly (you probably already include it explicitly, eg for live iso use-case).

Some network dependent modules (such as [ssh-client](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/95ssh-client/module-setup.sh#L19)) always had this policy. This PR align this policy to all network modules.

In hostonly mode dracut will detect and pull in the required dracut modules to boot - including networking modules.

**This PR intentionally breaks backwards compatibility with the rational, that it improves performance for most users at the expense of a few setups needing extra configuration with a one-time migration.**

An example of "improves performance".
On Arch (or any other distro where the default is no-hostonly), I install cifs package to my rootfs as I need to access some data from a remote cifs server. cifs however is not needed to boot the system. Just by installing cifs, next time a regenerate initrd, now my intird includes all the networking infrastructure and cifs, even though this is not required, resulting in increased initrd size and slower boot.

## Checklist

Fixes #132

CC @berolinux

### DIFF
--- a/test/TEST-12-RAID-DEG/test.sh
+++ b/test/TEST-12-RAID-DEG/test.sh
@@ -97,7 +97,6 @@ test_setup() {
     chmod 0600 /tmp/key
 
     test_dracut \
-        -o "dbus" \
         -i "./cryptroot-ask.sh" "/sbin/cryptroot-ask" \
         -i "/tmp/mdadm.conf" "/etc/mdadm.conf" \
         -i "/tmp/crypttab" "/etc/crypttab" \
--- a/modules.d/95cifs/module-setup.sh
+++ b/modules.d/95cifs/module-setup.sh
@@ -12,7 +12,7 @@ check() {
         return 255
     }
 
-    return 0
+    return 255
 }
 
 # called by dracut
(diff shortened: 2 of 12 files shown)


test/TEST-12-RAID-DEG/test.sh
```diff
@@ -97,7 +97,6 @@ test_setup() {
     chmod 0600 /tmp/key
 
     test_dracut \
-        -o "dbus" \
```
freswa, line 100: What the relation with the network stack here?

freswa (commented):
Besides one question, lgtm

bdrung:
I am against this change. Network modules needed for booting should be included by default if that is installed on the host. Experts can disable those unwanted modules, but the default should work out of the box.

Debian/Ubuntu moved the network related modules (like cifs, nbd, nfs, etc) into the separate dracut-network package. Removing this package will result in what you want.

PR #298, opened 2024-05-19
fix: /etc/modprobe.d --> /run/modprobe.d

### LINKED ISSUE #253: with systemd 256 dracut cannot write hooks to /lib/dracut/hooks during boot because /lib is /usr/lib, which is read-only with new ProtectSystem feature
**Describe the bug**
systemd 256 has a new `ProtectSystem` feature which makes `/usr` read-only. This is set to default to on in initramfs environments. dracut's `hookdir`, where it writes hooks generated during the boot process, is `/lib/dracut/hooks`. On Fedora, `/lib` is a symlink to `/usr/lib`, so dracut winds up trying to write hooks to a read-only filesystem.

**Distribution used**
Fedora Rawhide.

**Dracut version**
060.

**Init system**
systemd.

**To Reproduce**
Update a Fedora Rawhide system to systemd 256 (from https://koji.fedoraproject.org/koji/buildinfo?buildID=2444479 ). Re-generate the initramfs with `dracut -f`. Reboot. You will see errors like `//lib/dracut/hooks/cmdline/00-parse-root.sh: line 28: /lib/dracut/hooks/initqueue/finished/devexists-\x2fdev\x2fdisk\x2fby-uuid\x2ff18e44fc-458a-483d-ae4f-fe5ee8ba011c.sh: Read-only file system`.

**Expected behavior**
There should be no such errors.

**Additional context**
@poettering pointed out the likely cause of these errors. There is an additional problem with systemd 256 which will cause the system to fail to boot at all unless dracut has [this change applied](https://github.com/dracut-ng/dracut-ng/pull/118); I will probably send a dracut build with that backported soon. With that patched, the system may boot (unless any boot-time hooks are actually vital to boot in your configuration) and the errors may not be so obvious, but I think they will still happen.

I've also filed this on systemd at https://github.com/systemd/systemd/issues/32511 ; it seems reasonable to have a bug report on each side, because we may want to make some adjustment in systemd in the short term, but in the long term we probably want dracut to change the hook directory to something else. It looks like this might be rather disruptive, though.

LaszloGombos:
## Changes

Change /etc/modprobe.d path to /run/modprobe.d in hooks.  

## Checklist

Fixes #253 (partially)

### DIFF
--- a/modules.d/01fips/fips.sh
+++ b/modules.d/01fips/fips.sh
@@ -97,7 +97,6 @@ fips_load_crypto() {
     read -d '' -r FIPSMODULES < /etc/fipsmodules
 
     fips_info "Loading and integrity checking all crypto modules"
-    mv /etc/modprobe.d/fips.conf /etc/modprobe.d/fips.conf.bak
     for _module in $FIPSMODULES; do
         if [ "$_module" != "tcrypt" ]; then
             if ! nonfatal_modprobe "${_module}" 2> /tmp/fips.modprobe_err; then
@@ -113,7 +112,10 @@ fips_load_crypto() {
             fi
         fi
     done
-    mv /etc/modprobe.d/fips.conf.bak /etc/modprobe.d/fips.conf
+    if [ -f /etc/fips.conf ]; then
+        mkdir -p /run/modprobe.d
+        cp /etc/fips.conf /run/modprobe.d/fips.conf
+    fi
 
     fips_info "Self testing crypto algorithms"
     modprobe tcrypt || return 1
--- a/modules.d/90kernel-modules/parse-kernel.sh
+++ b/modules.d/90kernel-modules/parse-kernel.sh
@@ -1,11 +1,7 @@
 #!/bin/sh
 
-_modprobe_d=/etc/modprobe.d
-if [ -d /usr/lib/modprobe.d ]; then
-    _modprobe_d=/usr/lib/modprobe.d
-elif [ -d /lib/modprobe.d ]; then
-    _modprobe_d=/lib/modprobe.d
-elif [ ! -d $_modprobe_d ]; then
+_modprobe_d=/run/modprobe.d
+if [ ! -d $_modprobe_d ]; then
     mkdir -p $_modprobe_d
 fi
 
@@ -17,8 +13,6 @@ for i in $(getargs rd.driver.pre -d rdloaddriver=); do
         done
     )
 done
-
-[ -d /etc/modprobe.d ] || mkdir -p /etc/modprobe.d
 
 for i in $(getargs rd.driver.blacklist -d rdblacklist=); do
     (
--- a/modules.d/95lunmask/parse-lunmask.sh
+++ b/modules.d/95lunmask/parse-lunmask.sh
@@ -32,8 +32,9 @@ for lunmask_arg in $(getargs rd.lunmask); do
         IFS="$OLDIFS"
         if [ -d /sys/module/scsi_mod ]; then
             printf "manual" > /sys/module/scsi_mod/parameters/scan
-        elif [ ! -f /etc/modprobe.d/95lunmask.conf ]; then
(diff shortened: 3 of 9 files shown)


jrelvas-ipc:
Applied this on top of `dracut-101-1.fc41`. 

There's no longer any "Read-only filesystem" errors logged. I also didn't spot any regressions during boot.

Patch seems to work as advertised.

modules.d/98dracut-systemd/dracut-pre-udev.sh
```diff
@@ -31,7 +27,7 @@ for i in $(getargs rd.driver.pre -d rdloaddriver=); do
     )
 done
 
-[ -d /etc/modprobe.d ] || mkdir -p /etc/modprobe.d
+[ -d /run/modprobe.d ] || mkdir -p /run/modprobe.d
```
aafeijoo-suse, line 30: This line can be removed.

modules.d/90kernel-modules/parse-kernel.sh
```diff
@@ -18,7 +14,7 @@ for i in $(getargs rd.driver.pre -d rdloaddriver=); do
     )
 done
 
-[ -d /etc/modprobe.d ] || mkdir -p /etc/modprobe.d
+[ -d /run/modprobe.d ] || mkdir -p /run/modprobe.d
```
aafeijoo-suse, line 17: This line can be removed.

modules.d/95lunmask/parse-lunmask.sh
```diff
@@ -33,7 +33,8 @@ for lunmask_arg in $(getargs rd.lunmask); do
         if [ -d /sys/module/scsi_mod ]; then
             printf "manual" > /sys/module/scsi_mod/parameters/scan
         elif [ ! -f /etc/modprobe.d/95lunmask.conf ]; then
```
aafeijoo-suse, line 35: This condition must be changed to `[ ! -f /run/modprobe.d/95lunmask.conf ]`

modules.d/01fips/fips.sh
```diff
@@ -97,7 +97,7 @@ fips_load_crypto() {
     read -d '' -r FIPSMODULES < /etc/fipsmodules
 
     fips_info "Loading and integrity checking all crypto modules"
-    mv /etc/modprobe.d/fips.conf /etc/modprobe.d/fips.conf.bak
+    mv /run/modprobe.d/fips.conf /run/modprobe.d/fips.conf.bak
```
aafeijoo-suse, line 100: You're trying to move a file that does not exist:

https://github.com/dracut-ng/dracut-ng/blob/25dabef1dab4612e31be84af3e262dc187ec4158/modules.d/01fips/module-setup.sh#L48

Now FIPS is broken, this workaround of renaming the fips.conf file before modprobing the fips modules is not valid anymore.

aafeijoo-suse (approved):
The following line is not necessary anymore:

https://github.com/dracut-ng/dracut-ng/blob/25dabef1dab4612e31be84af3e262dc187ec4158/modules.d/01fips/module-setup.sh#L43

Apart from that, I just tested 2 x86_64 setups (1 - / unencrypted, 2 - separate /boot and / encrypted) and it works. But someone else with more FIPS experience should take a look at this before merging it.

modules.d/01fips/fips.sh
```diff
@@ -113,7 +112,10 @@ fips_load_crypto() {
...
         fi
     done
-    mv /etc/modprobe.d/fips.conf.bak /etc/modprobe.d/fips.conf
+    if [ -f /etc/fips.conf ]; then
+        mkdir -p /run/modprobe.d
+        cp /etc/fips.conf /run/modprobe.d/fips.conf
+    fi
 
     fips_info "Self testing crypto algorithms"
     modprobe tcrypt || return 1
```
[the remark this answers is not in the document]
dtardon: Ah, that disables just the module's internal aliases, not the module itself.

dtardon (commented):
LGTM

Conan-Kudo (approved):

PR #299, opened 2024-05-19
fix(dracut): bsdcpio compatibility

LaszloGombos:
## Changes

From https://github.com/OpenMandrivaAssociation/dracut/blob/master/dracut-044-bsdcpio-compat.patch

Make cpio invocations more compatible with bsdcpio -- the mode indicator has to be the first argument.

## Checklist


Conan-Kudo (approved):
This is ugly, but okay I guess...

Issue #305, opened 2024-05-21
Support exit with failure in case a test environment is unsuitable


bdrung:
In the Debian/Ubuntu package I want to run all test cases and explicitly disable test that are not suitable. Tests that are skipped because `test_check` fails should fail the test instead of silently skipping it. Otherwise missing dependencies will not be caught by the test.

Please add an option to make the test fail if `test_check` is not satisfied.

bdrung:
It's enough to be for the entire testsuite, e.g. by setting an environment variable.

PR #306, opened 2024-05-22
test: when DRACUT_SKIP_TEST_CHECK is set skip running test_check

LaszloGombos:
## Changes

Add an option to force run the test without executing test_check.

## Checklist

Fixes #

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -36,9 +36,16 @@ test_dracut() {
         "$@" || return 1
 }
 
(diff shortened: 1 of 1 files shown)


bdrung:
Skipping the test check is not what I want for #305. In case the environment lacks the needed tools, `DRACUT_SKIP_TEST_CHECK` would just ignore the test check and fail later on. Finding the root cause will probably be harder than what the test check would provide.

PR #311, opened 2024-05-23
fix(fips): Remove /dev/{random,urandom} pre-creation

vittyvk:
Random device nodes were required to workaround libgcrypt initialization problem (see https://bugzilla.redhat.com/show_bug.cgi?id=1401444) which is gone since libgcrypt-1.10 (2022): modern Linux kernels (3.17+) support getrandom() syscall and libgcrypt has switch to using getentropy() (glibc-2.25+).

The requirement to run dracut as root (which is needed to mknod /dev/{random,urandom}) is a particular problem for building UKIs in distro build systems where packages are not built by a privileged user.

Note, dracut itself always pre-creates /dev/{null,kmsg,console,random,urandom} devices when running privileged so the patch has no effect on 'traditional' setup.

## Changes
'fips' module doesn't pre-create /dev/{random,urandom} anymore and thus can run unpriviliged.

## Checklist

### DIFF
--- a/modules.d/01fips/module-setup.sh
+++ b/modules.d/01fips/module-setup.sh
@@ -67,16 +67,4 @@ install() {
     inst_multiple sha512hmac rmmod insmod mount uname umount grep sed cut find sort
 
     inst_simple /etc/system-fips
-    [ -c "${initdir}"/dev/random ] || mknod "${initdir}"/dev/random c 1 8 \
-        || {
-            dfatal "Cannot create /dev/random"
-            dfatal "To create an initramfs with fips support, dracut has to run as root"
-            return 1
-        }
-    [ -c "${initdir}"/dev/urandom ] || mknod "${initdir}"/dev/urandom c 1 9 \
-        || {
-            dfatal "Cannot create /dev/urandom"
-            dfatal "To create an initramfs with fips support, dracut has to run as root"
-            return 1
-        }
 }


vittyvk:
> Earlier related discussion [dracutdevs/dracut#2331](https://github.com/dracutdevs/dracut/pull/2331)

Missed this one, thanks! Cc: @cgwalters :-)

There are probably other ways to create /dev/{random,urandom} in the image but I don't think we need them. I've tested RHEL9 and Fedora rawhide guests with no /dev/{random,urandom} in FIPS enabled mode and nothing blew up.

vittyvk:
> ```
>   Error: [C003] Description should not start with a capital case letter
>   fix(fips): Remove /dev/{random,urandom} pre-creation
>              ^
>              r
> ```

Fixed!

cgwalters:
I'm OK with this, it's always nice to delete code; but I do think relying on all the userspace underneath systemd to always use `getrandom()` is somewhat dicey.

vittyvk:
> I'm OK with this, it's always nice to delete code; but I do think relying on all the userspace underneath systemd to always use `getrandom()` is somewhat dicey.

We're not getting rid of /dev/{random,urandom} completely, of course :-)

This particular piece of code was needed when libgcrypt's **initialization** code in FIPS mode was requiring /dev/urandom. As systemd was linked with it, it had no chance to mount devtmpfs to /dev/. AFAIU systemd mounts /dev pretty early so all units which start after are OK.

cgwalters:
Yes, when I said "underneath systemd" I meant "before systemd".

Conan-Kudo (approved):

pvalena:
> I guess the only way to figure out if such code exists is to keep expanding usage of 'unprivileged-generated' initramfs (dracut not running under 'root') as privileged dracut always injects /dev/{null,kmsg,console,random,urandom}. FWIW, I'm not aware of any issues of such kind reported against Fedora or RHEL UKI which exist since Jan, 2023. (and to be honest I'm more worried about the missing /dev/null than /dev/urandom :-)

Wouldn't it be also good to create the devices in an uprivilleged way? I think that was discussed previously. Is that still planned?

vittyvk:
@pvalena It can certainly be done but is this really needed? IMO we should actually try to go the other way around: try dropping device creation for privileged user as well, i.e.

```
diff --git a/dracut.sh b/dracut.sh
index 7b9bc459be19..841a3cc85557 100755
--- a/dracut.sh
+++ b/dracut.sh
@@ -1900,13 +1900,6 @@ if [[ $kernel_only != yes ]]; then
         # shellcheck disable=SC2174
         mkdir -m 0755 -p "${initdir}/lib/dracut/hooks/$_d"
     done
-    if [[ $EUID == "0" ]] && ! [[ $DRACUT_NO_MKNOD ]]; then
-        [[ -c ${initdir}/dev/null ]] || mknod "${initdir}"/dev/null c 1 3
-        [[ -c ${initdir}/dev/kmsg ]] || mknod "${initdir}"/dev/kmsg c 1 11
-        [[ -c ${initdir}/dev/console ]] || mknod "${initdir}"/dev/console c 5 1
-        [[ -c ${initdir}/dev/random ]] || mknod "${initdir}"/dev/random c 1 8
-        [[ -c ${initdir}/dev/urandom ]] || mknod "${initdir}"/dev/urandom c 1 9
-    fi
 fi
 
 _isize=0 #initramfs size

```
and see if anyone complains. As we already have UKIs produced in non-privileged mode where these devices are not created and no complains, it may happen that this is just a not-really-needed legacy. Or, we will find a good justification why we need to do unprivileged creation.

Just my $0.02, it's very possible I'm missing something important here.

vittyvk:
FWIW, I looked at Fedora-40 and 'filesystem' package does not contain any device file in '/dev', this means that the 'real' system always counts on mounted devtmpfs. I'm not sure if it can be used as a clear indication that static device nodes are never needed: I seriously double there's a significant userbase which doesn't use initramfs at all (and for everything with an initramfs /dev will be populated from initramfs).

PR #317, opened 2024-05-24
fix(systemd-ask-password): no graphical output in aarch64

aafeijoo-suse:
A password cannot be entered if there is no graphical output during boot, as is the case in aarch64, where `efifb` does not work with `qemu-system-aarch64`:
- `virtio-gpu-pci` does not expose a linear framebuffer
- `virtio-vga` is not supported
- `ramfb` is not enough

Therefore, depend on the `drm` module if `virtio_gpu` is loaded on the system.

Issue detected on aarch64 encrypted images, where there is no password prompt on the default console.

## Checklist

### DIFF
--- a/modules.d/01systemd-ask-password/module-setup.sh
+++ b/modules.d/01systemd-ask-password/module-setup.sh
@@ -19,6 +19,19 @@ check() {
 # Module dependency requirements.
 depends() {
 
+    if [[ $hostonly ]]; then
+        # A password cannot be entered if there is no graphical output during boot,
+        # as is the case in aarch64, where efifb does not work with qemu-system-aarch64:
+        # - virtio-gpu-pci does not expose a linear framebuffer
+        # - virtio-vga is not supported
+        # - ramfb is not enough
+        # Therefore, depend on the drm module if virtio_gpu is loaded on the system.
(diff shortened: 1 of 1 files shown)


modules.d/01systemd-ask-password/module-setup.sh
```diff
@@ -19,6 +19,20 @@ check() {
 # Module dependency requirements.
 depends() {
 
+    if [[ $hostonly ]]; then
+        # A password cannot be entered if there is no graphical output during
+        # boot, as is the case in aarch64, where:
+        # - efifb does not work with qemu-system-aarch64
+        # - virtio-gpu-pci does not expose a linear framebuffer
+        # - virtio-vga is not supported
+        # - ramfb is not enough
```
Vogtinator, line 28:
```suggestion
        # efifb does not work with qemu-system-aarch64:
        # - virtio-gpu-pci does not expose a linear framebuffer
        # - virtio-vga is not supported
        # - ramfb is not enough
```

aafeijoo-suse: Yes, changed. Thanks.

Conan-Kudo (approved):

PR #326, opened 2024-05-27
refactor: use inst_sysusers to install sysusers files

LaszloGombos:
## Changes

Refactor the way systemd-sysusers config files gets installed to make it easier for distributions to customize.

## Checklist

CC @AndrewAmmerlaan

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -709,6 +709,15 @@ inst_libdir_file() {
     [[ ${#_files[@]} -gt 0 ]] && inst_multiple "${_files[@]}"
 }
 
+# install sysusers files
+inst_sysusers() {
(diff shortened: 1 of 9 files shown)


AndrewAmmerlaan (commented):
Something like this would probably help, but not sure yet how to deal with the `acct-group`/`acct-user` situation because sometimes both a group and a user file exist. And other times it is just the one or the other.

AndrewAmmerlaan (approved):
>  Yes this would install both acct-group/acct-user, but is there a situation where that is particularly harmful ?

That should not be a problem.

I'm a bit confused though about what it is `-H` does, I couldn't find it in `man dracut.modules`.

Conan-Kudo (approved):

PR #335, opened 2024-06-05
fix(dracut): do not add a hostonly dependency to the device containing the ESP

aafeijoo-suse:
Since a956a464, a new device unit is created for `/boot/efi` (also for `/efi` since 4a6a4ac6) configured with `Wants=initrd.target`.

```
localhost:~ # lsblk -o UUID,MOUNTPOINT /dev/vda1
UUID                                 MOUNTPOINT
96D6-4309                            /boot/efi
localhost:~ # lsinitrd | grep "4309\.device"
drwxr-xr-x   1 root     root            0 Jun  5 10:59 etc/systemd/system/dev-disk-by\x2duuid-96D6\x2d4309.device.d
-rw-r--r--   2 root     root           46 Jun  5 10:59 etc/systemd/system/dev-disk-by\x2duuid-96D6\x2d4309.device.d/timeout.conf
lrwxrwxrwx   1 root     root           42 Jun  5 10:59 etc/systemd/system/initrd.target.wants/dev-disk-by\x2duuid-96D6\x2d4309.device -> ../dev-disk-by\x2duuid-96D6\x2d4309.device
```

Include just the filesystem drivers of the ESP in hostonly mode, without adding a hard dependency to a device unit that it is not needed at this point in the boot process.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1599,6 +1599,23 @@ fi
 
 declare -A host_fs_types
 
+_get_fs_type() {
+    [[ $1 ]] || return
+    if [[ -b /dev/block/$1 ]]; then
+        ID_FS_TYPE=$(get_fs_env "/dev/block/$1") && host_fs_types["$(readlink -f "/dev/block/$1")"]="$ID_FS_TYPE"
+        return 1
+    fi
+    if [[ -b $1 ]]; then
+        ID_FS_TYPE=$(get_fs_env "$1") && host_fs_types["$(readlink -f "$1")"]="$ID_FS_TYPE"
+        return 1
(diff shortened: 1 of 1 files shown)


dalto8:
Have this been tested in a situation where the initramfs is hosted in the ESP and the ESP is the only thing that would bring it that device?

aafeijoo-suse:
> Have this been tested in a situation where the initramfs is hosted in the ESP and the ESP is the only thing that would bring it that device?

Yes. This question should not be necessary knowing how the boot process works, but can try this patch with sd-boot.

PR #337, opened 2024-06-06
WIP: feat(plymouth-ng): alternative module to use plymouth as boot splash

zboszor:
This is a rebase of https://github.com/dracutdevs/dracut/pull/1774  over dracut-ng. Only the last commit is missing that added testing the boot-animation module because there were conflicts. @LaszloGombos 

## Changes

Added new modules: plymouth-ng and boot-animation, the second being a meta-module.

plymouth-ng is an alternative to the plymouth module. The difference is that it doesn't rely on plymouth's "populate initrd" script.

## Checklist

Fixes #

### DIFF
--- a/modules.d/50boot-animation/module-setup.sh
+++ b/modules.d/50boot-animation/module-setup.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    return 255
+}
+
+# called by dracut
+depends() {
+    # shellcheck disable=SC2043
+    for module in plymouth-ng; do
+        if dracut_module_included "$module"; then
+            splash_backend="$module"
+            break
+        fi
+    done
+
+    if [ -z "$splash_backend" ]; then
+        if [[ -e $dracutsysrootdir$systemdsystemunitdir/plymouth-start.service ]]; then
+            splash_backend="plymouth-ng"
+        fi
+    fi
+    echo "$splash_backend"
+    return 0
+}
+
+# called by dracut
+installkernel() {
+    return 0
+}
+
+# called by dracut
+install() {
+    return 0
+}
--- a/modules.d/50plymouth-ng/module-setup.sh
+++ b/modules.d/50plymouth-ng/module-setup.sh
@@ -1,0 +1,150 @@
+#!/bin/bash
+
+pkglib_dir() {
+    local _dirs="/usr/lib/plymouth /usr/lib64/plymouth /usr/libexec/plymouth"
+    if find_binary dpkg-architecture &> /dev/null; then
+        local _arch
+        _arch=$(dpkg-architecture -qDEB_HOST_MULTIARCH 2> /dev/null)
+        [ -n "$_arch" ] && _dirs+=" /usr/lib/$_arch/plymouth"
+    fi
+    for _dir in $_dirs; do
+        if [[ -x "$dracutsysrootdir""$_dir"/plymouthd-fd-escrow ]]; then
+            echo "$_dir"
+            return
+        fi
+        if [[ -x "$dracutsysrootdir""$_dir"/plymouth-populate-initrd ]]; then
+            echo "$_dir"
+            return
+        fi
+    done
+}
+
+plugin_dir() {
+    local _dirs="/usr/lib/plymouth /usr/lib64/plymouth /usr/libexec/plymouth"
+    if find_binary dpkg-architecture &> /dev/null; then
+        local _arch
+        _arch=$(dpkg-architecture -qDEB_HOST_MULTIARCH 2> /dev/null)
+        [[ -n $_arch ]] && _dirs+=" /usr/lib/$_arch/plymouth"
+    fi
+    for _dir in $_dirs; do
+        if [[ -d "$dracutsysrootdir""$_dir"/renderers ]]; then
+            echo "$_dir"
+            return
+        fi
+    done
+}
+
+# $1: Configuration file
+# $2: Setting name
+function read_setting_from_file() {
+    grep -v '^#' "$1" 2> /dev/null \
+        | awk 'BEGIN { FS="[=[:space:]]+"; OFS=""; ORS="" } $1 ~/^'"$2"'$/ { print $2 }'
+}
+
+# $1: Configuration file
+function set_configured_theme_path_from_file() {
+    [[ -z $CONFIGURED_THEME_DIR ]] && CONFIGURED_THEME_DIR=$(read_setting_from_file "$1" "ThemeDir")
+    [[ -z $CONFIGURED_THEME_DIR ]] && CONFIGURED_THEME_DIR=/usr/share/plymouth/themes
+    CONFIGURED_THEME_NAME=$(read_setting_from_file "$1" "Theme")
+}
+
+function set_theme_dir() {
+    set_configured_theme_path_from_file "$dracutsysrootdir"/etc/plymouth/plymouthd.conf
+    if [[ -z $CONFIGURED_THEME_DIR ]] || [[ ! -d "$dracutsysrootdir$CONFIGURED_THEME_DIR/$CONFIGURED_THEME_NAME" ]]; then
+        set_configured_theme_path_from_file "$dracutsysrootdir"/usr/share/plymouth/plymouthd.defaults
+    fi
+
+    if [[ -n $CONFIGURED_THEME_DIR ]] && [[ -d $dracutsysrootdir$CONFIGURED_THEME_DIR ]]; then
+        PLYMOUTH_THEME_DIR="$CONFIGURED_THEME_DIR"
+    else
+        PLYMOUTH_THEME_DIR=/usr/share/plymouth/themes
+    fi
+    PLYMOUTH_THEME_NAME=${CONFIGURED_THEME_NAME:-spinner}
+}
+
+# called by dracut
+check() {
+    [[ "$mount_needs" ]] && return 1
+    [[ $(pkglib_dir) ]] || return 1
+
+    require_binaries plymouthd plymouth plymouth-set-default-theme || return 1
+
+    set_theme_dir
+    [[ -n $PLYMOUTH_THEME_NAME ]] || return 1
+
+    return 255
+}
+
+# called by dracut
+depends() {
+    echo drm
+}
+
+# called by dracut
+install() {
+    PKGLIBDIR=$(pkglib_dir)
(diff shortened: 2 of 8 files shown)


LaszloGombos:
I would be interested to collect feedback from Linux distributions to see if they would be interested (or perhaps even prefer) the new `plymouth-ng` module over the existing `plymouth` module.

@zboszor Would you be able to share if this new `plymouth-ng` is already in use somewhere else ?

zboszor:
> I would be interested to collect feedback from Linux distributions to see if they would be interested (or perhaps even prefer) the new `plymouth-ng` module over the existing `plymouth` module.

Sure, do that.

The goal of this new module was that plymouth's populate initrd script didn't behave completely correctly when using a cross-compiled sysroot.
 
> @zboszor Would you be able to share if this new `plymouth-ng` is already in use somewhere else ?

Yes, it's in use. I can't disclose the company name, but we use a Yocto based cross-compiled OS.

LaszloGombos:
It does not have to be in the next release, but ideally upstream should only have one plymouth module eventually, so I hope to get a consensus from distributions that `plymouth-ng` is a better alternative to `plymouth`.

freswa (changes requested):
It's neither in our main repos nor in the AUR. Personally I didn't even know it exist. So for now that's a NACK from my side.

LaszloGombos:
> plymouth-ng

To clarify, there is no Linux project/package named `plymouth-ng` . This PR proposes a new dracut module called `plymouth-ng` to take advantage of the good old `plymouth` Linux package.

freswa:
Sorry that's my fault. In that case we don't have objections.

freswa (approved):

modules.d/50boot-animation/module-setup.sh
```diff
@@ -0,0 +1,35 @@
...
+    # shellcheck disable=SC2043
+    for module in plymouth-ng; do
+        if dracut_module_included "$module"; then
+            splash_backend="$module"
+            break
+        fi
+    done
+
+    if [ -z "$splash_backend" ]; then
+        if [[ -e $dracutsysrootdir$systemdsystemunitdir/plymouth-start.service ]]; then
```
LaszloGombos, line 19: Do you know if this condition is true in at least one of the CI containers (Arch, Debian, Gentoo, Debian, Fedora, openSUSE) ? 

In other words is this line tested by the current CI ? 

Thanks !

bdrung (changes requested):
Thanks for submitting the alternative plymouth module. Before going into details of the code review, I have two remarks addressed:

1. I don't like having two competing implementations in dracut that are supposed to do the same task. This makes the code bigger and add conditional checks. We would need to maintain and debug both implementations. Instead we should just ship one implementation and make this implementation as good as possible (make all customers happy with it).
2. Can you provide reasoning why we should drop the current implementation and use the one that you propose? What problem does the current one has? What does your make better? Does it has any (potential) drawback?

We have some open bug reports against the plymouth module in Ubuntu and I plan to look closer at that code part soon. I have no general objections to drop the current implementation and introduce a new one.

zboszor:
> Thanks for submitting the alternative plymouth module. Before going into details of the code review, I have two remarks addressed:
> 
>     1. I don't like having two competing implementations in dracut that are supposed to do the same task. This makes the code bigger and add conditional checks. We would need to maintain and debug both implementations. Instead we should just ship one implementation and make this implementation as good as possible (make all customers happy with it).

The current implementation evolved at https://github.com/dracutdevs/dracut/pull/1774. The whole story is there.

Long story short, it was the original maintainers' request that since plymouth isn't the only boot splash provider application, a meta-package called boot-animation should be provided that can use any of the underlying (currently theoretical) backends.

>     2. Can you provide reasoning why we should drop the current implementation and use the one that you propose? What problem does the current one has? What does your make better? Does it has any (potential) drawback?

There are cases when populate initrd script in plymouth doesn't work properly. See https://github.com/dracut-ng/dracut-ng/pull/337#issuecomment-2156130132

dracut's own installer program is much smarter regarding library dependencies when used in a cross-compiled sysroot.

> We have some open bug reports against the plymouth module in Ubuntu and I plan to look closer at that code part soon. I have no general objections to drop the current implementation and introduce a new one.

modules.d/50boot-animation/module-setup.sh
zboszor: systemd based distros with plymouth installed should have it.

With `dracutsysrootdir=""` or `dracutsysrootdir="/"`:
```
[zozo@fedora ~]$ export systemdsystemunitdir=/usr/lib/systemd/system
[zozo@fedora ~]$ if [[ -e $dracutsysrootdir$systemdsystemunitdir/plymouth-start.service ]]; then echo file exists ; fi
file exists
```

LaszloGombos:
> The goal of this new module was that plymouth's populate initrd script didn't behave completely correctly when using a cross-compiled sysroot.

This seems a use-case most distributions and dracut users will not run into. We should consider accommodating this use-case, but not at the expense of an intrusive change like a meta module and the complexity it comes with.  I understand that an earlier advice asked for the meta module, but give the nature of the use-case, I hope to change change the existing plymouth module instead.

AndrewAmmerlaan:
> The goal of this new module was that plymouth's populate initrd script didn't behave completely correctly when using a cross-compiled sysroot.

I have not really dug into this yet, but this sounds like the fix should be in plymouth. Or am I missing something?

zboszor:
> > The goal of this new module was that plymouth's populate initrd script didn't behave completely correctly when using a cross-compiled sysroot.
> 
> This seems a use-case most distributions and dracut users will not run into. We should consider accommodating this use-case, but not at the expense of an intrusive change like a meta module and the complexity it comes with. I understand that an earlier advice asked for the meta module, but give the nature of the use-case, I hope to change change the existing plymouth module instead.

The meta module advice was an orthogonal, but related problem. There are other fish in the sea than plymouth. See https://en.wikipedia.org/wiki/Bootsplash#Linux_distributions

zboszor:
> > The goal of this new module was that plymouth's populate initrd script didn't behave completely correctly when using a cross-compiled sysroot.
> 
> I have not really dug into this yet, but this sounds like the fix should be in plymouth. Or am I missing something?

I am all for copying dracut-install into plymouth.

Frankly, I would just rename plymouth-ng back to plymouth and replace the old dracut module with the new one. That was my original intention.

LaszloGombos:
> Frankly, I would just rename plymouth-ng back to plymouth and replace the old dracut module with the new one. That was my original intention.

Can you please upload the version that just renames plymouth-ng back to plymouth and replace the old dracut module with the new one ?

zboszor:
> > Frankly, I would just rename plymouth-ng back to plymouth and replace the old dracut module with the new one. That was my original intention.
> 
> Can you please upload the version that just renames plymouth-ng back to plymouth and replace the old dracut module with the new one ?

I guess this also implies getting rid of the boot-animation meta module?

modules.d/50plymouth-ng/plymouth-pretrigger.sh
```diff
@@ -0,0 +1,27 @@
...
+            --subsystem-match=acpi \
+            > /dev/null 2>&1
+
+        udevadm settle --timeout=180 2>&1 | vinfo
+
+        info "Starting plymouth daemon"
+        mkdir -m 0755 /run/plymouth
+        read -r consoledev rest < /sys/class/tty/console/active
+        consoledev=${consoledev:-tty0}
+        [ -x /lib/udev/console_init -a -e "/dev/$consoledev" ] && /lib/udev/console_init "/dev/$consoledev"
```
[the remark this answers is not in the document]
pvalena: +1

modules.d/50plymouth-ng/plymouth-pretrigger.sh
```diff
@@ -0,0 +1,27 @@
...
+        plymouthd --attach-to-session --pid-file /run/plymouth/pid
+        plymouth --show-splash 2>&1 | vinfo
+        # reset tty after plymouth messed with it
+        [ -x /lib/udev/console_init -a -e "/dev/$consoledev" ] && /lib/udev/console_init "/dev/$consoledev"
```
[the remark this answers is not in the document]
pvalena: +1

modules.d/50plymouth-ng/module-setup.sh
```diff
@@ -0,0 +1,150 @@
...
+
+    inst_multiple \
+        /usr/share/plymouth/plymouthd.defaults \
+        /etc/plymouth/plymouthd.conf
+
+    inst_multiple -o /usr/share/plymouth/themes/default.plymouth
+
+    inst_multiple -o "/usr/share/plymouth/*.png"
+
+    inst_multiple -o /etc/system-release
```
pvalena, line 121: This can be concatenated into one cmd, right?

modules.d/50plymouth-ng/module-setup.sh
```diff
@@ -0,0 +1,150 @@
...
+depends() {
+    echo drm
+}
+
+# called by dracut
+install() {
+    PKGLIBDIR=$(pkglib_dir)
+    PLUGINDIR=$(plugin_dir)
+
+    inst_multiple readlink
```
pvalena, line 88: Shouldn't this rather be in `check()` with `require_binaries`?

modules.d/50plymouth-ng/module-setup.sh
```diff
@@ -0,0 +1,150 @@
...
+}
+
+# $1: Configuration file
+function set_configured_theme_path_from_file() {
+    [[ -z $CONFIGURED_THEME_DIR ]] && CONFIGURED_THEME_DIR=$(read_setting_from_file "$1" "ThemeDir")
+    [[ -z $CONFIGURED_THEME_DIR ]] && CONFIGURED_THEME_DIR=/usr/share/plymouth/themes
+    CONFIGURED_THEME_NAME=$(read_setting_from_file "$1" "Theme")
+}
+
+function set_theme_dir() {
```
pvalena, line 51: Why the use of `function`? It's not POSIX compatible AFAIK.

bdrung:
> > Frankly, I would just rename plymouth-ng back to plymouth and replace the old dracut module with the new one. That was my original intention.
> 
> Can you please upload the version that just renames plymouth-ng back to plymouth and replace the old dracut module with the new one ?

I agree as well. Since #442 was created as successor, I'll close this merge request.

PR #341, opened 2024-06-11
dracut-init: change lookup order for rules files

dtardon:
## Changes
Look for rules files in /etc/udev/rules.d first so that local modifications of system files are included.

## Checklist


LaszloGombos (approved):
LGTM.

`inst_simple` will not overwrite if the file is already installed.

PR #342, opened 2024-06-12
fix(90kernel-modules): install blk modules using symbol blk_alloc_disk

pvalena:
Corresponding kernel symbol blk_cleanup_disk is no longer used in the nvdimm driver and calls are made directly instead.

blk_alloc_disk is used to match the driver:
https://elixir.bootlin.com/linux/v6.1-rc8/source/drivers/nvdimm/pmem.c#L522

## Changes

Adds `blk_alloc_disk` to list of symbols which pull kernel modules to initrd.

## Checklist
- [x] I have tested it locally (previously used on RHEL-9)

Resolves: RHEL-32237

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -2,7 +2,7 @@
 
 # called by dracut
 installkernel() {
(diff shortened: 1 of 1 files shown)


LaszloGombos (approved):
thanks

PR #348, opened 2024-06-13
feat(test): support V=2 without logtee

bdrung:
## Changes

The Debian/Ubuntu autopkgtest runs the tests out of tree. In that environment `gcc` might not be installed, but it is needed to compile `logtee`.

So support running the tests with `V=2` but without `logtee`. Only use `logtee` if it is available.

## Checklist


LaszloGombos (approved):
lgtm

PR #354, opened 2024-06-13
fix: address shellcheck SC2319

bdrung:
## Changes

shellcheck complains:

```
In modules.d/80lvmthinpool-monitor/start-thinpool-monitor.sh line 12:
    [ -n "$_lvm2_thin_device" ] && return $?
                                          ^-- SC2319 (warning): This $? refers to a condition, not a command. Assign to a variable to avoid it being overwritten.
```

Let `is_lvm2_thinp_device` directly return the test exit code.

shellcheck complains:

```
In modules.d/95iscsi/iscsiroot.sh line 232:
    [ -z "$targets" ] && warn "Target discovery to $iscsi_target_ip:${iscsi_target_port:+$iscsi_target_port} failed with status $?" && return 1
                                                                                                                                ^-- SC2319 (warning): This $? refers to a condition, not a command. Assign to a variable to avoid it being overwritten.
```

The `while` loop that gets the output from `iscsiadm` via a pipe will exit with 0. Printing the exit code 0 does not help. So just drop it.

## Checklist

### DIFF
--- a/modules.d/95iscsi/iscsiroot.sh
+++ b/modules.d/95iscsi/iscsiroot.sh
@@ -229,7 +229,7 @@ handle_netroot() {
             echo "$target"
         done
     })
-    [ -z "$targets" ] && warn "Target discovery to $iscsi_target_ip:${iscsi_target_port:+$iscsi_target_port} failed with status $?" && return 1
+    [ -z "$targets" ] && warn "Target discovery to $iscsi_target_ip:${iscsi_target_port:+$iscsi_target_port} failed" && return 1
 
     found=
     for target in $targets; do
--- a/modules.d/80lvmthinpool-monitor/start-thinpool-monitor.sh
+++ b/modules.d/80lvmthinpool-monitor/start-thinpool-monitor.sh
@@ -9,7 +9,7 @@ is_lvm2_thinp_device() {
     _lvm2_thin_device=$(lvm lvs -S 'lv_layout=sparse && lv_layout=thin' \
         --nosuffix --noheadings -o vg_name,lv_name "$_device_path" 2> /dev/null)
 
-    [ -n "$_lvm2_thin_device" ] && return $?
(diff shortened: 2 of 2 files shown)


LaszloGombos (approved):

modules.d/95iscsi/iscsiroot.sh
```diff
@@ -229,7 +229,7 @@ handle_netroot() {
             echo "$target"
         done
     })
-    [ -z "$targets" ] && warn "Target discovery to $iscsi_target_ip:${iscsi_target_port:+$iscsi_target_port} failed with status $?" && return 1
+    [ -z "$targets" ] && warn "Target discovery to $iscsi_target_ip:${iscsi_target_port:+$iscsi_target_port} failed" && return 1
```
mwilck, line 232: I think that `&&` pipelines should be avoided, in particular for tests like this.
I suggest

```
[ ! -z "$targets" ] || { warn "blah"; return 1; }
```
or using an `if` clause.

bdrung: Why should && pipelines be avoided?

An `if` clause would probably the most readable code. Patches are welcome. :)

mwilck: > Why should && pipelines be avoided?

Because the pipeline's exit status is *failed* if any command in the pipeline fails. That may not hurt here, but it definitely does if you use `errexit`.

For example, people write code like this:

```bash
[ -f some_file ] && rm -f some_file
```
With `errexit` set, this will cause your script to abort if the file doesn't exist, which is probably not what you intended.
```bash
[ ! -f some_file ] || rm -f some_file
```
works though, and has exactly the same semantics as
```bash
if [ -f some_file ]; then rm -f some_file; fi
```
Which one is more readable is a matter of taste.

I for one use `errexit` a lot, and therefore I got used to avoiding `&&` pipelines completely.

PR #355, opened 2024-06-13
fix: quote single CTTY parameter and _ctty variable

bdrung:
## Changes

shellcheck complains about SC2086 (info):
> Double quote to prevent globbing and word splitting.

The variable `CTTY` contains either nothing or exactly one parameter. Use shell substitution to quote this single parameter in case it is not empty.

The variable `_ctty` contains a path and therefore is safe to quote.

## Checklist

### DIFF
--- a/modules.d/80test-root/test-init.sh
+++ b/modules.d/80test-root/test-init.sh
@@ -38,8 +38,7 @@ echo "made it to the rootfs!"
 
 if getargbool 0 rd.shell; then
     strstr "$(setsid --help)" "control" && CTTY="-c"
-    # shellcheck disable=SC2086
-    setsid $CTTY sh -i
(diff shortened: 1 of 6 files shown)


modules.d/80test-root/test-init.sh
```diff
@@ -39,7 +39,7 @@ echo "made it to the rootfs!"
 if getargbool 0 rd.shell; then
     strstr "$(setsid --help)" "control" && CTTY="-c"
     # shellcheck disable=SC2086
```
LaszloGombos, line 41: `shellcheck disable=SC2086` can be removed as part of this PR.

bdrung: removed.

LaszloGombos (approved):

PR #375, opened 2024-06-15
fix: disable SC2317 for calls by for_each_host_dev_and_slaves

bdrung:
## Changes

shellcheck 0.10 complains about SC2317 (info): Command appears to be unreachable. Check usage (or ignore if invoked indirectly). It fails to detect cases where a function is defined and later called by `for_each_host_dev_and_slaves`. Disable shellcheck in this case.

## Checklist


LaszloGombos (approved):
ok. CI is only on v0.9, but make sense to plan on upgrading eventually.

bdrung:
shellcheck 0.9 complains as well. I just checked with the newer version to see if that got smarter.

PR #376, opened 2024-06-16
Load the SELinux policy after switch_root. This fixes

gtrentalancia:
Load the SELinux policy after switch_root and remove the selinux-loadpolicy module.

This fixes the bootup process with recent kernels, as it was getting stuck on Permission Denied errors due to the early SELinux policy load.

---
 .github/labeler.yml                       |    4 -
 modules.d/98selinux/module-setup.sh       |   17 -------
 modules.d/98selinux/selinux-loadpolicy.sh |   70 ------------------------------
 modules.d/99base/init.sh                  |    5 ++
 4 files changed, 5 insertions(+), 91 deletions(-)


LaszloGombos:
Removing `selinux` module breaks compatibility. As far as I can tell removing this dracut module is not necessary for resolving this issue.

The solution should also honor e.g.`selinux=0` command line option.

gtrentalancia:
Yes, you are right, the patch can be improved by adding the selinux=0 option to disable SELinux. Thanks for pointing that out !

I can do that, as I believe such a patch would be the best option to resolve the issue (also please see the existing comments in the selinux-loadpolicy module).

Shall I proceed and create an improved version to accept the "selinux" boot option ?

gtrentalancia:
I have now submitted a revised and improved patch which should fix all the remaining issue. Can you please submit an updated review ?

LaszloGombos:
Please do not remove the `selinux` module as part of this PR.

gtrentalancia:
Can you please explain a bit more detailedly what problem are you expecting to happen if the existing selinux-loadpolicy module is removed ?

I am happy to leave the existing "selinux-loadpolicy" there, for backwards compatibility as you say, in light of the fact that SELinux "load_policy" would refuse to reload an already loaded policy when the "-i" option is used.

gtrentalancia:
Should be good to go now !

gtrentalancia:
By the way, why is the Commisery check failing ? It must be something to do with the title or description...

LaszloGombos:
Separation of concerns is important and  it would be desirable to avoid managing `selinux` in the `base` dracut module and instead try to encapsulate `selinux` logic into the `selinux` dracut module. This should also avoid the need for copying and repeating code.

This PR makes a change in `.github/labeler.yml`, which seems unnecessary. 

Ideally either @wgwoods or @daveyoung should review this PR.

gtrentalancia:
I have double-checked and thought again about the backwards-compatibility issue that you raised with removing the existing _selinux-loadpolicy_ module issue. To the best of my knowledge and understanding there is not such issue.

However, I am willing to leave most things as you suggested and as they currently are, in view of the fact that leaving the existing _selinux-loadpolicy_ module in place (and simply obsoleting it) does not pose a threat to the system security.

This should also compensate the fact that unfortunately I am only able to test on a limited number of setups: id est sysvinit and not other init daemons. I do very much welcome further testing by other users of alternative init daemons.

I am thinking of obsoleting it more clearly with a specific note in the top-level Documentation, so expect another change soon.

gtrentalancia:
No, I do not believe your opinion is combative, I just said what I believe the situation is, according to the testing that I have done using sysvinit.

I have now tried to omit the selinux module in the default configuration, but it is producing integration errors...

gtrentalancia:
Also, have not heard anything back from you since the last proposed version, so I was trying to further improve things...

If you have more feedback, please go on I am all ears !

gtrentalancia:
I have fixed the issue locally on my system, so I have nothing to gain in proposing the patch, other than trying to improve things to a larger audience, consider the change was already marked as pending by FIXME.

gtrentalancia:
If you have good comments to contribute, that is more than welcome. For example, about the more recent failures in the integration tests after changing the default configuration ?

I am not much into the test suite...

gtrentalancia:
The integration test failures are caused by https://github.com/dracut-ng/dracut-ng/pull/380

gtrentalancia:
This is now fixed in the [kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/security/selinux/ss/services.c?id=fde46f60f6c5138ee422087addbc5bf5b4968bf1)

PR #385, opened 2024-06-20
feat(dracut): search for zstd compressor first

bdrung:
## Changes

Searching for an available compressor was introduced in commit 693b7a32833b ("Cleanup compressor handling") but the order was not justified.

Ubuntu uses zstd for compressing the packages, the kernel modules, firmware files, and the initrd (when using initramfs-tools). zstd has a good compromise between compression ratio and speed. Therefore search for zstd first.

Further readings:

 * https://lists.ubuntu.com/archives/ubuntu-devel/2021-December/041726.html
 * analysis.ipynb on https://github.com/waveform80/compression

## Checklist


LaszloGombos (approved):
>  zstd has a good compromise between compression ratio and speed. Therefore search for zstd first.

This seems a good justification for this PR.

PR #386, opened 2024-06-20
refactor: replace long-time deprecated $tempnode with $devnode

aafeijoo-suse:
## Checklist

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -610,7 +610,7 @@ prepare_udev_rules() {
                     printf '%sIMPORT{builtin}="blkid"\n' "${line%%IMPORT BLKID}"
                 else
                     # shellcheck disable=SC2016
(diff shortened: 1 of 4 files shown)


LaszloGombos (approved):

bdrung (commented):
What about `tempnode` used in `modules.d/90mdraid/module-setup.sh`?

aafeijoo-suse:
> > What about `tempnode` used in `modules.d/90mdraid/module-setup.sh`?
> 
> It's a sed expression, not an actual udev rule, and it doesn't hurt IMO.

And btw this would break compatibility with mdadm < 3.3, 64-md-raid.rules uses $tempnode. So you can remove this part if you want to remove the mdadm < 3.3 code, but it's not my goal here.

bdrung (approved):

PR #388, opened 2024-06-21
systemd-networkd: add support for proper netroot invocation

### LINKED ISSUE #251: netroot=... fails with systemd-networkd
See this PR to reproduce - https://github.com/dracut-ng/dracut-ng/pull/238

BtbN:
This adds support for the networkd networking module to actually bring up netroots.
It's based on how the connman module does it, adapted for networkd.

## Changes

networkd (like dhclient, which gets told about them here: https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/35network-legacy/dhclient.conf#L8) needs to be told to request specific options. Like the route-path or its DHCPv6 equivalent (which are as of yet unused by dracut and unsupported by systemd-networkd).
Likewise, the dhclient config used to enable hardware/mac mode for the client-identifier(https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/35network-legacy/dhclient.conf#L4). This sets the equivalent networkd setting so people don't have to adjust their dhcp servers config when switching between the two modules, just cause they send a different client id.

The general flow of things is that the initqueue will wait for the network to be online, if the cmdline hook determines it neccesary.
Then a hookscript will iterate all interfaces, query networkd about their "onlineness" and if it confirms it, call the usual queue and netroot setup on the interface.

## Checklist

Fixes #251

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -109,14 +109,22 @@ jobs:
             matrix:
                 container: [
                         "arch",
+                        "gentoo",
                 ]
                 network: [
                         "systemd-networkd",
                 ]
                 test: [
+                        "20",
+                        "30",
                         "35",
                         "40",
                 ]
+                exclude:
+                  - container: "arch"
+                    test: "20"
+                  - container: "gentoo"
+                    test: "40"
             fail-fast: false
         container:
             image: ghcr.io/dracut-ng/${{ matrix.container }}
--- a/modules.d/01systemd-networkd/module-setup.sh
+++ b/modules.d/01systemd-networkd/module-setup.sh
@@ -57,7 +57,16 @@ install() {
         "$systemdsystemunitdir"/systemd-networkd-wait-online.service \
         "$systemdsystemunitdir"/systemd-networkd-wait-online@.service \
         "$systemdsystemunitdir"/systemd-network-generator.service \
-        ip
+        ip sed grep
+
+    inst_simple "$moddir"/99-wait-online-dracut.conf \
+        "$systemdsystemunitdir"/systemd-networkd-wait-online.service.d/99-dracut.conf
+
+    inst_simple "$moddir"/99-default.network \
+        "$systemdnetworkconfdir"/99-default.network
+
+    inst_hook cmdline 99 "$moddir"/networkd-config.sh
+    inst_hook initqueue/settled 99 "$moddir"/networkd-run.sh
 
     # Enable systemd type units
     for i in \
--- a/modules.d/01systemd-networkd/99-default.network
+++ b/modules.d/01systemd-networkd/99-default.network
@@ -1,0 +1,13 @@
+[Match]
+Kind=!*
+Type=!loopback
+
+[Network]
+DHCP=yes
+
+[DHCPv4]
+ClientIdentifier=mac
+RequestOptions=17
+
+[DHCPv6]
+RequestOptions=59 60
--- a/modules.d/01systemd-networkd/99-wait-online-dracut.conf
+++ b/modules.d/01systemd-networkd/99-wait-online-dracut.conf
@@ -1,0 +1,9 @@
+[Unit]
+Before=dracut-initqueue.service
+ConditionPathExists=/run/networkd/initrd/neednet
+
+[Service]
+TimeoutStartSec=120
+
+[Install]
+WantedBy=initrd.target
--- a/modules.d/01systemd-networkd/networkd-config.sh
+++ b/modules.d/01systemd-networkd/networkd-config.sh
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+type getcmdline > /dev/null 2>&1 || . /lib/dracut-lib.sh
+
+# Just in case we're running before it
+systemctl start systemd-network-generator.service
+
+# Customizations for systemd-network-generator generated networks.
+# We need to request certain DHCP options, and there is no way to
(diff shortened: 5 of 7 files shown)


BtbN:
This seems to be generally working okay now.
The only issue is that networkd has absolutely zero provisions to report on dhcp options it does not know/care about.
It just parses right over them. So "root=dhcp" and the like are impossible to implement, outside of doing crazy stuff like bringing in tcpdump.

But aside from that, this should allow networkd to be a more normal network module that actually works.

LaszloGombos:
CC @dustymabe @photos  

Some good ideas here: https://github.com/flatcar/bootengine/tree/flatcar-master/dracut/03flatcar-network

modules.d/01systemd-networkd/module-setup.sh
```diff
@@ -57,14 +57,23 @@ install() {
...
+    inst_hook cmdline 99 "$moddir"/networkd-config.sh
+    inst_hook initqueue/settled 99 "$moddir"/networkd-run.sh
 
     # Enable systemd type units
     for i in \
         systemd-networkd.service \
         systemd-networkd.socket \
         systemd-network-generator.service \
-        systemd-networkd-wait-online.service; do
+        networkd-wait-online-initrd.service; do
```
LaszloGombos, line 76: Could this be achieved as a systemd unit drop-in instead, similar to e.g. https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/01systemd-hostnamed/99-systemd-networkd-dracut.conf ? 

Just a thought.

BtbN: My way of thinking was that I want to avoid making the initqueue depend on the native networkd-wait-online service, just in case something else pulls in a service that also depends on it.
That could lead to the initqueue never running, cause there is no network, so it locks up forever.

BtbN: I'm also not sure if a drop-in can sensibly add to the Wanted-By section so it ends up in the right section.

BtbN:
This is generally complete now, and passed my tests locally.

modules.d/01systemd-networkd/module-setup.sh
LaszloGombos: We need to consider both `hostonly` and `non-hostonly` configurations.

In non-hostonly the user might also have some drop-ins that dracut specifically copies over to inside of the generated initrd, which as I understand will have no impact now - see https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/01systemd-networkd/module-setup.sh#L84 .

.github/workflows/integration.yml
```diff
@@ -114,6 +114,8 @@ jobs:
                         "systemd-networkd",
                 ]
                 test: [
+                        "20",
```
LaszloGombos, line 117: Test 20 has some other issues on Arch that needs similar changes to https://github.com/dracut-ng/dracut-ng/pull/381

You should either not add tests-20 yet or try to fix it please.

```
--- a/test/TEST-20-NFS/test.sh
+++ b/test/TEST-20-NFS/test.sh
@@ -366,7 +366,7 @@ test_setup() {
     # We do it this way so that we do not risk trashing the host mdraid
     # devices, volume groups, encrypted partitions, etc.
     "$DRACUT" -l -i "$TESTDIR"/server/overlay / \
-        -m "bash rootfs-block kernel-modules qemu" \
+        -a "bash rootfs-block kernel-modules qemu" \
         -d "piix ide-gd_mod ata_piix ext4 sd_mod" \
         --nomdadmconf \
         --no-hostonly-cmdline -N \
@@ -416,7 +416,7 @@ test_setup() {
     )
     # Make server's dracut image
     "$DRACUT" -l -i "$TESTDIR"/overlay / \
-        -m "bash rootfs-block debug kernel-modules watchdog qemu network-legacy" \
+        -a "bash rootfs-block debug kernel-modules watchdog qemu network-legacy" \
         -d "af_packet piix ide-gd_mod ata_piix ext4 sd_mod e1000 i6300esb" \
         --no-hostonly-cmdline -N \
         -f "$TESTDIR"/initramfs.server "$KVERSION" || return 1
```

BtbN: Yeah, that's for a different PR to address.
I'm tempted to add a section for gentoo + networkd, since that's where I have been testing most locally.

LaszloGombos:
> This is generally complete now, and passed my tests locally.

Can you please also add one more commit that comments out the failing sub-tests from test-40 to restore green CI - https://github.com/dracut-ng/dracut-ng/pull/395. Thanks !

BtbN:
Yeah, looking into that right now.
Was hoping they'd be easy to fix, but I can't immediately see what's broken.
So that might also be for the future.

modules.d/01systemd-networkd/module-setup.sh
BtbN: It does work fine as a drop-in from the looks of it, so gonna do that.

LaszloGombos:
Can you please also revisit the PR description (remove what is no longer correct) ? For example "This is NOT tested yet, for now I'm looking for feedback on the method." no longer applies.

BtbN:
Updated.
The one failing integration test passes locally for me in the same image. So I'm unsure what to do about that other than try to just re-run it, and then just disabling 35 on gentoo on CI, if it's still stubborn.

LaszloGombos (approved):
This looks great. Thanks !

PR #389, opened 2024-06-21
feat: add additional common configs

LaszloGombos:
## Changes

This PR makes it easier and consistent to package common configurations for distribution. 

## Checklist

### DIFF
--- a/dracut.conf.d/50-generic.conf
+++ b/dracut.conf.d/50-generic.conf
@@ -1,0 +1,2 @@
+# generic initrd works on other platforms than the one dracut was called
+hostonly="no"
--- a/dracut.conf.d/50-hostonly.conf
+++ b/dracut.conf.d/50-hostonly.conf
@@ -1,0 +1,2 @@
+# optimize initrd to be as small as possible for faster boot performance, tailored to the current host
+hostonly="yes"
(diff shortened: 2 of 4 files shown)


dracut.conf.d/50-generic.conf
```diff
@@ -0,0 +1,3 @@
+# generic initrd works on other platforms than the one dracut was called
+
```
bdrung, line 2: I would remove the empty line between the comment and the config value.

dracut.conf.d/50-hostonly.conf
```diff
@@ -0,0 +1,3 @@
+# as small as possible initrd for improved performance
+
```
bdrung, line 2: same here.

dracut.conf.d/50-generic.conf
LaszloGombos: ok, thanks.

dracut.conf.d/50-hostonly.conf
LaszloGombos: ok, thanks.

bdrung (approved):

Conan-Kudo (approved):

PR #392, opened 2024-06-21
Fix nbd tests

BtbN:
The NBD tests so far were near totally borked.
They were actually failing a lot, but due to the test_run() function ending with killing the server, which always succeeded, that was the results the tests actually returned.

This PR will probably break CI, but it just un-hides breakage that was there before, but counted as pass.


LaszloGombos (approved):
Thanks for fixing the test.

PR #393, opened 2024-06-21
fix: microcode initrd generation

SlowBrainDude:
CPU microcode patch files called `$name.initramfs` will be considered for inclusion to the early initrd.

This pull request changes...

.. the logic which tries to find a suitable CPU microcode patch file for inclusion in early boot.

Previously Dracut would look for a file called either `$name.early` or `$name.early.initramfs`. But at least on Debian the proper file is called `$name.initramfs`. This bug resulted in an initrd created by Dracut which did not include any CPU microcode update files, and no "early CPIO image" was created at all on my box.

(This may actually even be security relevant as CPU microcode updates may include security fixes.)

## Checklist

Sure. Without this fix no "early CPIO image" was created on my box. With this patch the "early CPIO image" containing the Intel CPU microcode patch gets created.

The affected logic isn't documented anywhere as far as I can tell. I'm not going to change that.

(But maybe you should consider documenting how this part works.)

Not really. To be honest: I've done already more than planed. I'm not going to write test for this. Do with this PR what you want but I won't put any more effort into it. Switching back to `initramfs-tools`.

This is a "drive by commit". May it be helpful, but I'm not really interested in Dracut at this point; I did the debugging already, so wasting a little bit more time to share the outcome seems bearable. But not more.

I found out just by sheer chance that my CPU didn't get any microcode update during boot any more. Debugging this was no fun. This is a shell script tire-fire. Should be rewritten in some sane programming language, imho. One does not write some multi 10kLOC programs in shell. Also just another example of the fact that mutable variables are devil's work. The intended logic could be expressed clearly and safely in a simple pattern match without any mutation, and this hard to spot bug would have never existed... Just my 2ct.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2140,9 +2140,13 @@ if [[ $early_microcode == yes ]]; then
                 if [[ $hostonly ]]; then
                     _src=$(get_ucode_file)
                     [[ $_src ]] || break
-                    [[ -r $_fwdir/$_fw/$_src ]] || _src="${_src}.early"
-                    [[ -r $_fwdir/$_fw/$_src ]] || _src="${_src}.initramfs"
-                    [[ -r $_fwdir/$_fw/$_src ]] || break
+                    [[ -r $_fwdir/$_fw/$_src ]] || _src_early_postfix="${_src}.early"
+                    [[ ! -d $_fwdir/$_fw/$_src_early_postfix && -r $_fwdir/$_fw/$_src_early_postfix ]] && _src=$_src_early_postfix || _src_initramfs_postfix="${_src}.initramfs"
+                    [[ ! -d $_fwdir/$_fw/$_src_initramfs_postfix && -r $_fwdir/$_fw/$_src_initramfs_postfix ]] && _src=$_src_initramfs_postfix || _src_early_initramfs_postfix="${_src}.early.initramfs"
+                    [[ ! -d $_fwdir/$_fw/$_src_early_initramfs_postfix && -r $_fwdir/$_fw/$_src_early_initramfs_postfix ]] && _src=$_src_early_initramfs_postfix
+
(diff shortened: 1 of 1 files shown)


SlowBrainDude:
Sorry for the force push series. I didn't test the failure case before, just the one that worked on my machine.

Now I've tested also the other cases locally. Seems to work properly now. At least I hope. Shell language is from hell...

SlowBrainDude:
As soon as the currently pending test-fixing PR is on "main" I'll rebase this one here once more. So you can merge it without needing to touch it yourself, if you see fit.

I think the whole microcode picking up stuff should have tests. But like said, that's a little bit too much for me.

bdrung:
Thanks for putting in the work. I fully agree that the shell is not a desirable language to program in and rewriting this part of dracut in a nicer language would be nice. initramfs-tools is written in shell as well. So they are both on the same level (except that dracut has some parts in C/Rust).

I tried to read the submitted diff and gave up. I tried to come up with a more readable fix for the wrong suffix: https://github.com/dracut-ng/dracut-ng/pull/411

SlowBrainDude:
I like the new version of the code fix much better. Thanks for coming up with it! :+1: 

(My fix was in the beginning just a one line change. Than it moved in a bad direction. I'm not good at shell scripting…)

I'm closing this here as it's obsolete.

I'm aware of the fact that `initramfs-tools` is also shell script monster. The implementation language was not the reason to switch back. Main reason is that `initramfs-tools` seems just better integrated in Debian. Also it creates much simpler and smaller initrds. I've used Dracut only because I was told that it's needed to build UKIs. This seems a false claim, I have now an UKI setup with an `initramfs-tools` built intird and it works fine.

I've seen you have a sane langue in the repo (Rust). Why not porting the shell monstrosity to it? I would bet even a literal rewrite would catch some more bugs like this here. But not my project. So I don't want to bother you guys. You know what you're doing I guess.

PR #394, opened 2024-06-21
test: use --kernel-cmdline to set only dracut arguments

LaszloGombos:
## Changes

Resolve regression from https://github.com/dracut-ng/dracut-ng/pull/252 .

Follow-up to https://github.com/dracut-ng/dracut-ng/commit/1993786f8aa183ce44705fce2a748cefdd2ed5c9

Linux kernel arguments should not be set with --kernel-cmdline, see #400

## Checklist

### DIFF
--- a/test/TEST-01-BASIC/test.sh
+++ b/test/TEST-01-BASIC/test.sh
@@ -14,7 +14,7 @@ test_run() {
     test_marker_reset
     "$testdir"/run-qemu \
         "${disk_args[@]}" \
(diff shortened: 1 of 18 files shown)


LaszloGombos:
Both @BtbN and @bdrung run into this trap.

bdrung:
The commit message lacks the explanation how this change resolves the regression for console output.

Conan-Kudo (approved):

PR #402, opened 2024-06-22
fix(dracut-functions): avoid awk in get_maj_min() 

LaszloGombos:
## Changes

The `get_maj_min()` cache lookup is commonly used across many flows. While `awk` should be available, some highly constrained environments may not have it. A second call to `grep` can provide the same behaviour without adding a dependency.

Lines in the cache will be of the form "/dev/sda2 8:2". `awk '{print $NF}'` returns the last word of a matching line. Since the initial matching regex is so specific a second call to grep can easily extract the last word.

## Checklist

Source: https://github.com/microsoft/azurelinux/pull/9210

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -243,7 +243,7 @@ get_maj_min() {
     local _out
 
     if [[ $get_maj_min_cache_file ]]; then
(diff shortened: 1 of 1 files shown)


LaszloGombos:
It feels justified in my opinion to make this change upstream as well and not use awk for such a core function.

Conan-Kudo (approved):

PR #407, opened 2024-06-23
fix(dbus-daemon): actually enable the dbus service and socket

### LINKED ISSUE #309: dbus-daemon fails to start
**Describe the bug**
The dbus-daemon fails to start, which leads to all kinds of issues.
The issue I was facing that lead me to discover this was netword failing to set the hostname, even though I explicitly added the hostnamed module.

**Distribution used**
Gentoo

**Dracut version**
060_pre20240104-r4, but the issue exists in current master

**Init system**
systemd

**To Reproduce**
Use systemd-networkd as networking plugin, have the dhcp server send a hostname, and see it not getting applied.

Or drop into the emergency shell in any initrd with hostnamed included, and try to run hostnamectl.
It won't work until you manually systemctl start dbus

**Expected behavior**
dbus.service/socket should start on its own.

**Additional context**
The solution to this exact problem has in fact already been applied to the dbus-broker module, but the dbus-daemon module still has the exact same issue: https://github.com/dracut-ng/dracut-ng/commit/7ed0461810602bfbd4d5492fc9ed82f15e57fa9f

I have for now worked around the issue by manually adding the following section to the install() function of the module I'm maintaining:

```
        if [ -e "$initdir$systemdsystemunitdir/dbus.service" ]; then
                inst_multiple -o \
                        "$systemdsystemunitdir"/multi-user.target.wants/dbus.service \
                        "$systemdsystemunitdir"/sockets.target.wants/dbus.socket
        fi
```

These two lines should probably be added to dbus-daemons install() function as well.

BtbN:
The dbus daemon not actually running has surprisingly little ill effects, since systemctl falls back to some built-in backup of pid1 to do its thing.
But one issue I ran into was that networkd was not setting the hostname it was sent via DHCP. Turns out this is due to there being no running dbus, so it fails to communicate the hostname-change to hostnamed.
It also makes debugging issues in an emergency shell a lot harder, since most systemd-commands outside of systemctl are inoperative.

## Changes
The two units don't have an Install-Section. They instead ship with respective links to them in the systems unit directory.
So those just need to be copied as well.

## Checklist

Fixes #309

### DIFF
--- a/modules.d/06dbus-daemon/module-setup.sh
+++ b/modules.d/06dbus-daemon/module-setup.sh
@@ -53,6 +53,8 @@ install() {
         "$systemdsystemunitdir"/dbus.service \
         "$systemdsystemunitdir"/dbus.socket \
         "$systemdsystemunitdir"/dbus.target.wants \
+        "$systemdsystemunitdir"/multi-user.target.wants/dbus.service \
+        "$systemdsystemunitdir"/sockets.target.wants/dbus.socket \
         busctl dbus-send dbus-daemon
 
     # Adjusting dependencies for initramfs in the dbus service unit.


BtbN:
The two failing tests are both passing locally.
Specially the arch tests should not be affected whatsoever, since arch uses dbus-broker.

Conan-Kudo (approved):

PR #409, opened 2024-06-23
fix(systemd-networkd): remove default network if at least one other w…

BtbN:
Otherwise, if the machine has other interfaces, networkd might pick them up and end up calling the netroot script on them.
In theory this should be fine, since it'll just fail and it will call it on the correct interface very soon.
But at least the iscsi 30 test cases, which have two network interfaces but not always use both of them, seem to get upset by it.
My guess is that their success depends on which interface gets a DHCP response first.

## Changes
This change simply removes the default network if any other network file was generated.
It also renames the default network file a bit, to make it less likely to accidentally collide with any hostonly user-provided file.

## Checklist

### DIFF
--- a/modules.d/01systemd-networkd/networkd-config.sh
+++ b/modules.d/01systemd-networkd/networkd-config.sh
@@ -18,6 +18,9 @@ for f in /run/systemd/network/*.network; do
         echo "[DHCPv6]"
         echo "RequestOptions=59 60"
     } >> "$f"
+
+    # Remove the default network if at least one was generated
+    rm -f "$systemdnetworkconfdir"/99-dracut-default.network
 done
 
 # Just in case networkd was already running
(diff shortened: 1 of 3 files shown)


modules.d/01systemd-networkd/networkd-config.sh
```diff
@@ -18,6 +18,9 @@ for f in /run/systemd/network/*.network; do
         echo "[DHCPv6]"
         echo "RequestOptions=59 60"
     } >> "$f"
+
+    # Remove the default network if at least one was generated
+    rm -f /etc/systemd/network/99-dracut-default.network
```
LaszloGombos, line 23: Seems inconsistent to use `"$systemdnetworkconfdir"` in one place and `/etc/systemd/network` in another ? Would it be possible to stay consistent ?

BtbN: Can I just use that there? I somehow thought it'd only be available at "build time".

BtbN: Found where to add it so it becomes available.

LaszloGombos: thanks

LaszloGombos (approved):

PR #435, opened 2024-06-30
ci: add hmaccalc, remove NetworkManager from openSUSE CI containers

LaszloGombos:
## Changes

hmaccalc package is needed by the fido dracut module.

Since openSUSE prefers network-legacy over NetworkManager, let's uninstall NetworkManager. This will help testing the network dracut module, as now it will select network-legacy backend without additional configuration.

## Checklist


Conan-Kudo:
openSUSE is using NetworkManager by default now, so I don't think this assumption is true anymore.

LaszloGombos:
@aafeijoo-suse what do you think about the how the CI should test networking for openSUSE ?

Conan-Kudo (approved):

PR #438, opened 2024-07-01
ci: run systemd-networkd tests on Ubuntu as well

LaszloGombos:
## Checklist

Fixes #

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -110,6 +110,7 @@ jobs:
                 container: [
                         "arch",
                         "gentoo",
+                        "ubuntu",
                 ]
                 network: [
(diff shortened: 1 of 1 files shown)


LaszloGombos:
@bdrung Do you prefer to test networking here upstream with systemd-networkd or network-manager or both ? 

Perhaps we can just test with systemd-networkd (and remove network-manager from the Ubuntu CI container) ?

Conan-Kudo:
Netplan configures networkd in some environments and NetworkManager in others, so I would probably say both need testing.

bdrung:
I prefer testing both.

bdrung (approved):

Conan-Kudo (approved):

Issue #193, opened 2024-04-17
Test 40  failing


LaszloGombos:
From the log

```
grep: /var/tmp/dracut-test.q2jsMX//var/tmp/dracut-test.q2jsMX/marker.img: No such file or directory
```

Seen it failing/timeout only on alpine and void (non-systemd) container.

LaszloGombos:
with NetworkManager it only passes on Fedora - see https://github.com/dracut-ng/dracut-ng/pull/440

LaszloGombos:
Only the following subtests seem to fail

>    # Encrypted root handling via LVM/LUKS over NBD                                                                                                                                                                                   
>    "NBD root=LABEL=dracut netroot=nbd:IP:port"

LaszloGombos:
Only failing on Gentoo and Ubuntu

LaszloGombos:
From Debian with network-legacy

```
[    5.164071] IPv6: ADDRCONF(NETDEV_CHANGE): enp0s1: link becomes ready
[   65.726366] dracut-initqueue[631]: Warning: dhcp for interface enp0s1 failed
[   73.654417] dracut-initqueue[468]: Warning: dracut-initqueue: timeout, still waiting for following initqueue hooks:
[   73.658093] dracut-initqueue[468]: Warning: /lib/dracut/hooks/initqueue/finished/nbdroot.sh: "nbd-client -check /dev/nbd0 > /dev/null 2>&1"
```

PR #440, opened 2024-07-01
fix(network): deprioritize connman for network selection

LaszloGombos:
## Changes

The overall goal of this PR is to enable more networking tests on the CI.

- test: re-enable test 30,35 with default networking
- test: use use --add instead of --modules to create test-makeroot, use USE_NETWORK consistently
- ci: run network tests on Arch, Debian, Gentoo and Ubuntu as well. Only run test 40 on Fedora otherwise it fails.
- fix(network): deprioritize connman support. This enables Arch to default to network-manager instead of connman and pass the network tests.
- ci: run systemd-networkd and network-legacy tests on Debian as well

There is an existing bug open to resolve the issues with test 40 - https://github.com/dracut-ng/dracut-ng/issues/193

## Checklist


LaszloGombos:
CC @zboszor (for connman module).

zboszor:
> CC @zboszor (for connman module).

The list of network was intended to be in alphabetical order but it does not really matter in my use case, as there is no other network manager daemon besides `connman` in the sysroot where the initramfs is created from.

Conan-Kudo (approved):

PR #442, opened 2024-07-01
feat(plymouth): install plymouth files via dracut-install

zboszor:
@LaszloGombos Alternative to #337 , just modifies the `plymouth` dracut module.

## Checklist

Fixes #


modules.d/50plymouth/module-setup.sh
```diff
@@ -33,22 +83,67 @@ depends() {
...
+        /usr/share/plymouth/plymouthd.defaults \
+        /etc/plymouth/plymouthd.conf
 
-    inst_multiple readlink
+    inst_multiple -o /usr/share/plymouth/themes/default.plymouth
+
+    inst_multiple -o "/usr/share/plymouth/*.png"
 
-    inst_multiple plymouthd plymouth plymouth-set-default-theme
+    inst_multiple -o /etc/system-release
```
LaszloGombos, line 121: Why is this needed ? Is this PR based on an earlier version of this module ? 

See also https://github.com/dracut-ng/dracut-ng/commit/d6cef3f28e15a834284786c79f108b7c42481f96

zboszor: It wad not based on an an earlier version of the `plymouth-ng` module, this is straight from #337 using `git rm -r 50plymouth ; git mv 50plymouth-ng 50plymouth`. I did not notice the quoted commit before.

modules.d/50plymouth/module-setup.sh
```diff
@@ -1,27 +1,77 @@
 #!/bin/bash
 
 pkglib_dir() {
-    local _dirs="/usr/lib/plymouth /usr/libexec/plymouth/"
+    local _dirs="/usr/lib/plymouth /usr/lib64/plymouth /usr/libexec/plymouth"
     if find_binary dpkg-architecture &> /dev/null; then
```
bdrung, line 5: This is an independent issue that I see during a quick review: `dpkg-architecture` is shipped by dpkg-dev and is not expected to be present on all systems. The arch directory should be derived via a different approach.

bdrung (commented):
A quick glance does not reveal any fundamental issues.

I plan to test and more closely review it, but I won't promise on a date for it.

modules.d/50plymouth/module-setup.sh
zboszor: `find_binary dpkg-architecture` succeeds only if the `dpkg-architecture` command is found.

dracut-init.sh
```diff
@@ -255,6 +255,21 @@ inst() {
...
+    local x
+    for x in "$dracutsysrootdir${1%/}"/*; do
+        x=${x#"$dracutsysrootdir"}
+        if [[ -d "$dracutsysrootdir$x" ]]; then
+            inst_dir "$x"
+            inst_recur "$x"
+        elif [[ -f "$dracutsysrootdir$x" ]]; then
+            inst "$x"
+        else
+            break
```
pvalena, line 268: This should never happen, right? If it ever happens, you probably want to issue a warning instead...

pvalena: Btw. in general, isn't `inst_dir` implicitly recursive? What is `inst_recur` exactly solving?

pvalena:
Other than that it LGTM, but I have not tested this in Fedora. Have you tried it on some OS?

modules.d/50plymouth/module-setup.sh
```diff
@@ -1,27 +1,77 @@
 #!/bin/bash
 
 pkglib_dir() {
-    local _dirs="/usr/lib/plymouth /usr/libexec/plymouth/"
+    local _dirs="/usr/lib/plymouth /usr/lib64/plymouth /usr/libexec/plymouth"
```
Conan-Kudo, line 4: Who is incorrectly shipping this in `/usr/lib64/plymouth`? I don't want to support that.

modules.d/50plymouth/module-setup.sh
```diff
@@ -1,27 +1,77 @@
...
+    fi
+    PLYMOUTH_THEME_NAME=${CONFIGURED_THEME_NAME:-spinner}
+}
+
 # called by dracut
 check() {
     [[ "$mount_needs" ]] && return 1
     [[ $(pkglib_dir) ]] || return 1
 
     require_binaries plymouthd plymouth plymouth-set-default-theme || return 1
```
bdrung, line 70:
```suggestion
    require_binaries plymouthd plymouth || return 1
```

plymouth-set-default-theme is not used any more.

Alphix:
I like the idea of not relying on [this script](https://gitlab.freedesktop.org/plymouth/plymouth/-/blob/main/scripts/plymouth-populate-initrd.in?ref_type=heads), as the current code does (if present)....

But (and I'm not sure if I've overlooked something) there seems to be no handling of the infrastructure necessary to support text output (needed for e.g. unlocking encrypted volumes)...as in, I see no fonts, no fontconfig, no `label-freetype.so` or `label-pango.so` references in the new installation logic...?

devkontrol:
Closing.

Issue #379, opened 2024-06-18
dracut-install - cp compatibility


LaszloGombos:
make dracut-install work with bsdutils not just with coreutils - see https://github.com/chimera-linux/cports/blob/master/main/dracut-install/patches/cp-compatibility.patch

Also in coreutils starting in v9.0 default is COW - https://github.com/coreutils/coreutils/commit/25725f9d41735d176d73a757430739fb71c7d043

CC @q66

LaszloGombos:
Closing based on the discussion https://github.com/dracut-ng/dracut-ng/pull/647 . Thanks @jozzsi @q66

PR #444, opened 2024-07-01
fix: default DRACUT_NO_XATTR to 1

### LINKED ISSUE #443: Copying xattrs fails when rootfs is btrfs but the temporary directory is tmpfs
**Describe the bug**
When rootfs is `brtfs` but the temporary directory is on `tmpfs` (with xattr support), `dracut` fails with lots of errors like:

```
cp: setting attributes for '/tmp/dracut.lD7eCG/initramfs/bin/dash': Operation not supported
dracut-install: ERROR: 'cp --reflink=auto --sparse=auto --preserve=mode,xattr,timestamps,ownership -fL /bin/dash /tmp/dracut.lD7eCG/ini
tramfs/bin/dash' failed with 1
dracut-install: ERROR: installing '/bin/sh'
dracut[E]: FAILED: /usr/lib/dracut/dracut-install -D /tmp/dracut.lD7eCG/initramfs /bin/sh
```

Running a `strace` on a similar `cp` invocation indicates that the problem is that it's trying to copy `btrfs.compression` xattr:

```
$ strace cp --reflink=auto --sparse=auto --preserve=mode,xattr,timestamps,ownership -fL /bin/dash /tmp/x
[...]
fgetxattr(3, "btrfs.compression", NULL, 0) = 3
fgetxattr(3, "btrfs.compression", "lzo", 3) = 3
fsetxattr(4, "btrfs.compression", "lzo", 3, 0) = -1 EOPNOTSUPP (Operation not supported)
[...]
```

To be honest, I'm not sure whether it's a bug in `dracut` or `cp`. `cp` is doing exactly as advised, i.e. trying to copy all xattrs — except you can't copy filesystem-specific xattrs across filesystems. If the goal is to preserve ACLs, then according to help `--preserve=mode` is sufficient for that. However, FWICS `--preserve=xattr` is actually needed to preserve capabilities.

**Distribution used**
Gentoo Linux

**Dracut version**
102

**Init system**
systemd

**To Reproduce**
1. Use `btrfs` with compression enabled as the root filesystem.
2. Use `tmpfs` for `TMPDIR`.
3. Run `dracut`.

**Expected behavior**
Dracut working, i.e. either not attempting to copy `btrfs.*` xattrs or ignoring failures from that.

**Additional context**
I can try reporting it to coreutils instead. Wanted to get your opinion first.

LaszloGombos:
## Changes

Follow-up to https://github.com/dracut-ng/dracut-ng/commit/93be2bc6ab00d40b304d2a3b1ce02fd8a3f31b31 .

## Checklist

Fixes #443

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1282,11 +1282,13 @@ if [[ -f $dracutbasedir/dracut-version.sh ]]; then
 fi
 
 if systemd-detect-virt -c &> /dev/null; then
-    export DRACUT_NO_MKNOD=1 DRACUT_NO_XATTR=1
+    export DRACUT_NO_MKNOD=1
     if [[ $hostonly ]]; then
         printf "%s\n" "dracut[W]: Running in hostonly mode in a container!" >&2
     fi
 fi
+
+export DRACUT_NO_XATTR=${DRACUT_NO_XATTR:-1}
 
 if [[ -f $dracutbasedir/dracut-init.sh ]]; then
     # shellcheck source=./dracut-init.sh
--- a/tools/test-github.sh
+++ b/tools/test-github.sh
@@ -30,7 +30,6 @@ else
             cd /lib/modules
             ls -1 | tail -1
         )" \
-        DRACUT_NO_XATTR=1 \
         TEST_RUN_ID="$RUN_ID" \
         ${TESTS:+TESTS="$TESTS"} \
         -k V=1 \


LaszloGombos:
CC @mgorny @q66 @Camelron

q66:
i don't see how this fixes #379?

LaszloGombos:
> i don't see how this fixes #379?

True. My bad. If you'd like to help with #379, feel free to upload a PR.

mgorny:
Hmm, I suppose this makes sense. We don't really need any xattrs in initramfs, since it's all executed as root anyway, correct?

Conan-Kudo:
That would break SELinux then, wouldn't it?

aafeijoo-suse (commented):
> ## Changes
> 
> Follow-up to [93be2bc](https://github.com/dracut-ng/dracut-ng/commit/93be2bc6ab00d40b304d2a3b1ce02fd8a3f31b31) .

Not a follow-up, that patch had a specific purpose.

> ## Checklist
> 
>     * [ ]  I have tested it locally
> 
>     * [ ]  I have reviewed and updated any documentation if relevant
> 
>     * [ ]  I am providing new code and test(s) for it

Change not tested.

> Fixes #443

It fixes #443 by hiding an issue in that system.

```
localhost:~ # grep /tmp /proc/mounts
tmpfs /tmp tmpfs rw,nosuid,nodev,size=2005424k,nr_inodes=1048576,inode64 0 0
localhost:~ # cp --reflink=auto --sparse=auto --preserve=mode,xattr,timestamps,ownership -fL /bin/bash /tmp/bash
localhost:~ # getfattr -m . -d /tmp/bash
getfattr: Removing leading '/' from absolute path names
# file: tmp/bash
security.ima=0sBARHEOScdCunmtSwmp1YrORCn3j+C5e2rNvdJ2ZkSSOFDw==

localhost:~ # cp --reflink=auto --sparse=auto --preserve=mode,timestamps,ownership -fL /bin/bash /tmp/bash2
localhost:~ # getfattr -m . -d /tmp/bash2
localhost:~ #
```

> fix: default DRACUT_NO_XATTR to 1

It does not fix anything, the other way around, this breaks IMA by default, and it will only be possible to avoid denials at boot by setting an environment variable.

mgorny:
FTR, I think the security context can be preserved via `--preserve=context` but that option fails immediately on non-SELinux kernels.

LaszloGombos:
@mgorny Have you had a chance to check if this resolves the issue for you? We need to discuss the other consequences of this change, but we should start confirming that at least in some scenarios this change is needed.

mgorny:
I haven't tested the patch itself but I've confirmed that setting `DRACUT_NO_XATTR=1` in the environment fixes the problem for me.

LaszloGombos:
> this breaks IMA by default

Thanks for the review @aafeijoo-suse 

This is not my area of expertise, but as I understand IMA is not on by default in many distributions that dracut supports. In openSUSE there is an extra with some extra dracut configuration for IMA - https://github.com/openSUSE/dracut-ng/blob/SUSE/102/suse/dracut.spec#L92 

The default should be the most safe option for the system to boot in my opinion. Sadly 'safe' is slightly different for each Linux distribution, but perhaps something as specific as IMA should be opt-in and not opt-out.

Would it make sense for e.g. dracut-ima package to set `DRACUT_NO_XATTR to 0` so that it is immune to changes like this PR ? 

For additional context, Azure Linux carries this patch: https://github.com/microsoft/azurelinux/blob/3.0/SPECS/dracut/0002-disable-xattr.patch

I understand that this PR is controversial and needs to be discussed. I do not have a strong opinion on this change. Since neither documentation nor the test suite covers 'DRACUT_NO_XATTR`, this PR the best way to get feedback. 

As you can see from the change 'DRACUT_NO_XATTR` is already set to 1 in some conditions (https://github.com/dracut-ng/dracut-ng/commit/7275c6f6a0f6808cd939ea5bdf1244c7bd13ba44).

LaszloGombos:
If the compatibility risk of this PR is too big of a concern, than I think the best alternative is to land https://github.com/dracut-ng/dracut-ng/pull/450 and make this config available to users and distributions.

pvalena:
> If the compatibility risk of this PR is too big of a concern, than I think the best alternative is to land #450 and make a config available to users and distributions to opt-out easily during packaging.

I prefer the alternative, as it also explicitly separates configuration from code.

LaszloGombos:
Closing for now to take it out from review queue and from the next release, but I hope we're continuing the discussion as users will continuity to run into https://github.com/dracut-ng/dracut-ng/issues/443

PR #445, opened 2024-07-01
ci: drop network-manager from Debian Docker image

LaszloGombos:
## Changes

The network-manager version in Debian stable does not yet support dracut's network-manager module.
Debian trixie resolved this issue, but for the CI we prefer to use Debian stable.

Ubuntu CI container already has the fix and we will be using Ubuntu to test network-manager  dracut module.

This PR will allow to run the test with the default network module, which now will default to systemd-networkd for Debian CI container.

Historical context: https://github.com/dracutdevs/dracut/pull/2123

## Checklist

CC @bdrung

### DIFF
--- a/test/container/Dockerfile-Debian
+++ b/test/container/Dockerfile-Debian
@@ -41,7 +41,6 @@ RUN apt-get update -y -qq && apt-get upgrade -y -qq && apt-get install -y -qq --
     multipath-tools \
     nbd-client \
     nbd-server \
-    network-manager \
     nfs-kernel-server \
     ntfs-3g \
     open-iscsi \


Conan-Kudo:
Why _do_ we prefer Debian stable? Does anyone backport the latest dracut to Debian stable?

LaszloGombos:
> Why _do_ we prefer Debian stable? Does anyone backport the latest dracut to Debian stable?

I think this is reasonable to reconsider and re-discuss, but from my point of view Debian stable CI gives the project the lowest common denominator that dracut still supports here upstream. 

When removing support for older dependencies (kernel, udev, systemd, etc) and the change does NOT break Debian Stable CI, that gives confidence to be able to make the change.

There is a bit of documentation what are the purpose of each CI containers - https://github.com/dracut-ng/dracut-ng/wiki/Dracut-development#test-containers .

Debian is one of the few that is not rolling.

LaszloGombos:
> Does anyone backport the latest dracut to Debian stable?

I think so - https://backports.debian.org/

LaszloGombos:
This is "only" an issue temporary (for one more year), so closing.

PR #447, opened 2024-07-02
test: do not disable selinux for tests

LaszloGombos:
## Changes

In general, CI should be testing what most users - which is usually the default configuration. 
In this case, it also make test coverage higher with selinux potentially enabled.

## Checklist


LaszloGombos:
@bdrung Do you see any concern with this PR ? Would this PR interfere with the testing as part of packaging ?

bdrung (approved):
Looks good. In case tests fail afterwards, the underlying reason should be fixed instead of disabling SELinux.

Conan-Kudo (approved):

PR #454, opened 2024-07-03
Create a documentation site

ianw:
This is a proposal to create a documentation site with Antora (https://antora.org/) which works with the existing asciidoc files.

As motivation; there is a somewhat difficult to navigate (due to length and no indexing) single .html file made already, and a wiki page that is kept separately and links back to several docs, which can be viewed via github's rendering.  I'm of the feeling that incorporating docs into the main workflow leads to a better outcome, especially taking advantage of pipelines to publish on commit ensuring things are kept up-to-date.

One change this seems to require is s/.asc/.adoc/ for antora to pick up everything correctly.  This is a bit annoying but workable.  Otherwise it's a little framework stuff and some minor syntax fixes to some of the files and "make doc_site" outputs the site in build/.  man/dracut.asc becomes the index page; everything else remains in place where it was.

I know little about nodejs and not found it onerous to figure out; "nodejs" installed by your distro and "npx antora" "just works"; it's very self-contained.

I've included the sample github workflow to publish it to github pages on commit taken from the Antora docs -- it may need tweaking by project admins.

The output is basic but functional.  The nice thing about a framework like this is the almost limitless options for improvement :)

I've put a sample @ https://people.redhat.com/~iwienand/dracut-web/dracut/


pvalena (approved):
LGTM, although https://people.redhat.com/~iwienand/dracut-web/dracut/ does not render for me, so I can't say it works.

LaszloGombos:
>  there is a somewhat difficult to navigate (due to length and no indexing) single .html file made already

This is a good PR to discuss if we should continue to maintain https://github.com/dracut-ng/dracut-ng/blob/main/dracut.html in the tree. This impacts packaging, so ideally we get some feedback on this from distributions.

This html is also currently hosted here https://dracut-ng.github.io/dracut.html

AndrewAmmerlaan (approved):
> Removing this html file would impacts packaging, so ideally we get some feedback on this from distributions.

The Gentoo ebuild installs that html file, but I don't think it is a huge issue to drop that and add an extra link to `HOMEPAGE` instead.

LaszloGombos:
If this PR land would the existing Makefile rule still work - https://github.com/dracut-ng/dracut-ng/blob/main/Makefile#L131 or it needs to be changed together with this PR ?

ianw:
> LGTM, although https://people.redhat.com/~iwienand/dracut-web/dracut/ does not render for me, so I can't say it works.
> 
> **EDIT: it did render in the end; looks fine**

Thanks, I feel certain that it's possible to have this render a sample document in the CI workflow for any changes, although exactly how I'm not sure :)

ianw:
> If this PR land would the existing Makefile rule still work - https://github.com/dracut-ng/dracut-ng/blob/main/Makefile#L131 or it needs to be changed together with this PR ?

This is a good point, I've left a non-working `dracut.html` target in the `Makefile`.  I'll push with that removed for clarity.

bdrung (approved):
Asciidoc seems to be a nice format and I have no objection of going into that direction.

jozzsi:
@ianw  - I am wondering what was hour reasoning to drop dracut.css ? You did not like the existing css styling or Antora does not have the ability to customize css the same way as the existing customization. 

I think dropping dracut.css is good idea as the default style respond better to different window sizes as the custom dracut.css, but wanted to also understand your motivation.

ianw:
> @ianw - I am wondering what was hour reasoning to drop dracut.css ? You did not like the existing css styling or Antora does not have the ability to customize css the same way as the existing customization.

TBH I didn't even really consider customising; I was just getting something that worked.

Antora can include custom CSS;  it's not 100% straight forward as you have to add it into the default site template, but doable.  The way I've done it before

Add

```
<link rel="stylesheet" href="{{{uiRootPath}}}/css/site.css">
<link rel="stylesheet" href="{{{siteRootPath}}}/css/local.css">
```

to ```supplemental-ui/partials/head-styles.hbs``` to source the default stylesheet and a custom one.

add

```yaml
static_files:
  - css/local.css
```

to ship the static file, and finally add ```supplemental-ui/css/local.css```

I'm a little behind at the moment but can look at fixing up merge conflicts and adding appropriate bits "soon"

ianw:
> @ianw - I am wondering what was hour reasoning to drop dracut.css ? You did not like the existing css styling or Antora does not have the ability to customize css the same way as the existing customization.

In the latest push I've very lightly styled it to use the same blue as the old site just for continuity.  I'm sure some of `docs/` might work in the site but I haven't moved any of that, yet.

Screenshot 2024-10-07 at 15-00-06 dracut 103 Dracut

README.md
```diff
@@ -29,8 +29,7 @@ subdirectory, and use functionality provided by dracut-functions to do their
 work.
 
 Documentation:
- - [Introduction](man/dracut.adoc)
- - [User Manual](man/dracut.usage.adoc)
+ - https://dracut-ng.github.io/dracut/
```
jozzsi, line 32: Can this be simply `https://dracut-ng.github.io/` instead ?

dracut.html
```diff
@@ -1,4202 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
jozzsi, line 1: I like @Nowa-Ammerlaan 's suggestion here - https://github.com/dracut-ng/dracut-ng/pull/454#pullrequestreview-2156383133

ianw: OK this would just be something for packagers to note?  The latest change does remove the .html to avoid confusion over what's the current documentation

README.md
ianw: p.s. updated site is now at simply

https://people.redhat.com/~iwienand/dracut-web/

dracut.html
ianw: Ok, per https://github.com/dracut-ng/dracut-ng/pull/454#issuecomment-2416778507 I've rebased this, retaining the removal of the `dracut.html` file

PR #455, opened 2024-07-03
docs: correct shutdown path disucssion

ianw:
Since 4f03054e540427682221e4c62b4d44f79424e38a, these shutdown steps have been in response to the ExecStop= called from the sysinit.target.

Correct the doc in the two places it mentions this.


LaszloGombos (approved):
LGTM

PR #457, opened 2024-07-04
fix(systemd): check for systemd-vconsole-setup.service

LaszloGombos:
## Changes

Make systemd consistent with [systemd-ask-password](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/01systemd-ask-password/module-setup.sh#L51)

Resolves the following error on Debian based systems

```
Failed to add dependency on unit, unit systemd-vconsole-setup.service does not exist.
Failed to add dependency on unit, unit systemd-vconsole-setup.service does not exist.
```

## Checklist


bdrung (approved):
LGTM

Conan-Kudo (approved):

PR #464, opened 2024-07-05
test: enable serial console for all test runs

bdrung:
## Changes

Enable serial console for all test runs to ease debugging.

## Checklist


bdrung:
Oh, drop that. All those cases use `TEST_KERNEL_CMDLINE` which sets the console. So this was fixed after the 102 release.

PR #467, opened 2024-07-05
feat(multipath): omit module if included with no multipath devices and no user conf

### LINKED ISSUE #465: multipath in no-hostonly mode
Consider this downstream openSUSE change upstream as well - https://github.com/openSUSE/dracut-ng/commit/03e801f4a27da27983f77d0e7d25779d21a822e1

@aafeijoo-suse @bdrung @mwilck

LaszloGombos:
## Changes

The multipath module is always included in no-hostonly mode, but including it in the initramfs might activate multipath on the root device, which could be undesirable. This behavior can be manually fixed by explicitly excluding the root device from multipath in /etc/multipath.conf.

Therefore, issue a warning in no-hostonly mode, if there is no multipath device and the user has not configured multipath.

EDIT: instead of warning, omit including multipath module.

bsc#1069169

## Checklist

Fixes #465

See:
 - https://github.com/openSUSE/dracut/pull/239 
 - https://github.com/openSUSE/dracut-ng/commit/03e801f4a27da27983f77d0e7d25779d21a822e1

### DIFF
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -21,13 +21,23 @@ majmin_to_mpath_dev() {
 
 # called by dracut
 check() {
+    local _any_mpath_dev
+
+    for_each_host_dev_and_slaves is_mpath
+    _any_mpath_dev=$?
+
     [[ $hostonly ]] || [[ $mount_needs ]] && {
-        for_each_host_dev_and_slaves is_mpath || return 255
+        [[ $_any_mpath_dev == 0 ]] || return 255
     }
 
     # if there's no multipath binary, no go.
     require_binaries multipath || return 1
     require_binaries kpartx || return 1
+
+    if [[ $_any_mpath_dev != 0 ]] && [[ ! -f /etc/multipath.conf ]]; then
+        dwarn "multipath: including module with no multipath devices and empty" \
+            "configuration, the root disk may be unintentionally multipathed."
+    fi
 
     return 0
 }
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -35,8 +35,7 @@ check() {
     require_binaries kpartx || return 1
 
     if [[ $_any_mpath_dev != 0 ]] && [[ ! -f /etc/multipath.conf ]]; then
-        dwarn "multipath: including module with no multipath devices and empty" \
-            "configuration, the root disk may be unintentionally multipathed."
+        return 1
     fi
 
     return 0


LaszloGombos:
This PR leads to the following warning on Arch, Gentoo, Debian and possibly others

> dracut[W]: multipath: including module with no multipath devices and empty configuration, the root disk may be unintentionally multipathed.

LaszloGombos:
An alternative to this PR would be to not include the module instead of the warning

```
    if [[ $_any_mpath_dev != 0 ]] && [[ ! -f /etc/multipath.conf ]]; then
        return 1
    fi
```

freswa (approved):

LaszloGombos:
Added an extra commit on top of the openSUSE commit to implement the safer alternative. I believe this will provide 'Principle of Least Surprise'.

Conan-Kudo (approved):

aafeijoo-suse (commented):
> ae1b1003305f1553c057970d07a917deb255b7ef fix(multipath): warn if included with no multipath devices and no user conf (bsc#1069169)

I don't know why the current upstream merged this if downstream authors didn't submit or approved it... FTR, this patch is false and misleading for most distros. e.g. Fedora/RHEL: https://github.com/dracutdevs/dracut/pull/2095#pullrequestreview-1197631628

> 4957ffa935ea8c393c8fed867f6051da4ca43d58 Added an extra commit on top of the openSUSE commit to implement the safer alternative. I believe this will provide 'Principle of Least Surprise'.

IMHO, it's quite surprising to make these kinds of changes to a specific module without review or approval from the experts on the specific topic (@mwilck or @bmarzins in this case). But on the other hand, it's not the first time I've seen this behavior in this repo, there is a sense of haste and improvisation in everything that is done here.

modules.d/90multipath/module-setup.sh
```diff
@@ -21,14 +21,23 @@ majmin_to_mpath_dev() {
...
     [[ $hostonly ]] || [[ $mount_needs ]] && {
-        for_each_host_dev_and_slaves is_mpath || return 255
+        [[ $_any_mpath_dev == 0 ]] || return 255
     }
 
     # if there's no multipath binary, no go.
     require_binaries multipath || return 1
     require_binaries kpartx || return 1
 
+    if [[ $_any_mpath_dev != 0 ]] && [[ ! -f /etc/multipath.conf ]]; then
```
pvalena, line 37: Just ftr it's better to use `-ne` for integers instead of `!=` in bash (the latter is inteded for strings).

modules.d/90multipath/module-setup.sh
```diff
@@ -21,14 +21,23 @@ majmin_to_mpath_dev() {
...
 # called by dracut
 check() {
+    local _any_mpath_dev
+
+    for_each_host_dev_and_slaves is_mpath
+    _any_mpath_dev=$?
+
     [[ $hostonly ]] || [[ $mount_needs ]] && {
-        for_each_host_dev_and_slaves is_mpath || return 255
+        [[ $_any_mpath_dev == 0 ]] || return 255
```
pvalena, line 30: (ad comment bellow; same here: `-eq`)

modules.d/90multipath/module-setup.sh
bdrung: Probably `-gt` since we do not want to act on a negative count of mpath devices. ;)

PR #469, opened 2024-07-05
fix(systemd-pcrphase): make tpm2-tss an optional dependency

### LINKED ISSUE #329: dracut[E]: Module 'systemd-pcrphase' depends on 'tpm2-tss', which can't be installed
**Describe the bug**
It seems like the systemd-pcrphase module always gets pulled into the initramfs even if it is not needed as of #256 

This results in the below message on Arch for most users.
```
dracut[E]: Module 'systemd-pcrphase' depends on 'tpm2-tss', which can't be installed
```

This was noticed by EndeavourOS user [here](https://forum.endeavouros.com/t/dracut-error/55999).  I have verified that it also occurs on a base Arch install.

This could be expected behavior but is somewhat unfortunate that most users see this message since it is probably somewhat uncommon to have `tpm2-tools` installed.

**Distribution used**
Arch Linux, EndeavourOS

**Dracut version**
102

**Init system**
systemd

**To Reproduce**
Run dracut in Arch.  Since systemd-pcrextend is part of the systemd package, it will always try to pull in this module.

LaszloGombos:
## Changes

Only make tpm2-tss a dependency if requirements to install tpm2-tss is met.

Follow-up to https://github.com/dracut-ng/dracut-ng/pull/256 ,

## Checklist

Fixes #329

CC @dalto8

### DIFF
--- a/modules.d/01systemd-pcrphase/module-setup.sh
+++ b/modules.d/01systemd-pcrphase/module-setup.sh
@@ -21,7 +21,17 @@ check() {
 # Module dependency requirements.
 depends() {
     # This module has external dependency on other module(s).
-    echo systemd tpm2-tss
+
+    local deps
+    deps="systemd"
+
+    # optional dependencies
+    module="tpm2-tss"
+    module_check $module > /dev/null 2>&1
+    if [[ $? == 255 ]]; then
+        deps+=" $module"
+    fi
+    echo "$deps"
 
     # Return 0 to include the dependent module(s) in the initramfs.
     return 0


dalto8:
I tested this as of commit 1a20358d763563e49023d33f720dff58133ba5ff.

With `tpm2-tools` installed, it pulled in the `systemd-pcrphase` dracut module.  The initramfs contained `systemd-pcrextend` and `tpm2`.

Without `tpm2-tools` installed, there was no error displayed.  The `systemd-pcrphase` dracut module was still pulled in and `systemd-pcrextend`  was still in the initramfs but `tpm2` was not.

LaszloGombos:
> Without `tpm2-tools` installed, there was no error displayed. The `systemd-pcrphase` dracut module was still pulled in and `systemd-pcrextend` was still in the initramfs but `tpm2` was not.

This is expected and this is matching dracut-101 behavior. 

Thanks for testing @dalto8 . Feel free to also give "official reviews" if the PR looks good to you.

dalto8 (approved):

bdrung (approved):
+1 from my side

Conan-Kudo (approved):

aafeijoo-suse (commented):
TIL that `systemd-pcrextend` (used to extend some TPM2 PCRs) works without `tpm2-tss` (the dracut module that adds the TPM2 drivers and libs to the initrd).

PR #472, opened 2024-07-06
fix(dracut-init.sh): clarify the error message

### LINKED ISSUE #329: dracut[E]: Module 'systemd-pcrphase' depends on 'tpm2-tss', which can't be installed

LaszloGombos:
## Changes

Clarify that dracut module is missing and not Linux command or package.

This will help clarify to the users what action to take as a response of this message - 

## Checklist

Fixes #329

CC @dalto8

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -932,7 +932,7 @@ check_mount() {
             && force_add_dracutmodules+=" $_moddep "
         # if a module we depend on fail, fail also
         if ! check_module "$_moddep"; then
-            derror "Module '$_mod' depends on '$_moddep', which can't be installed"
+            derror "Module '$_mod' depends on module '$_moddep', which can't be installed"
             return 1
         fi
     done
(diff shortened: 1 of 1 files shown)


dalto8:
I think the confusion is mostly caused by the fact that this is an error and not a warning.

LaszloGombos:
Looking at https://forum.endeavouros.com/t/dracut-error/55999

```
sudo pacman -S tmp2-tss
```

dalto8:
I see.  Yes, that person is trying to install the wrong thing.

Conan-Kudo:
Is this really `perf`?

Conan-Kudo (approved):

bdrung:
It would be even nicer if the error message would say which binary wasn't found on the system.

PR #474, opened 2024-07-06
ci: name the CI containers consistently

LaszloGombos:
## Changes

There is no longer a plan to have more than one container for the same Linux distribution.

## Checklist


Conan-Kudo:
IIRC we wanted to have both Fedora-latest and Fedora-Rawhide as well as openSUSE-Leap and openSUSE-Tumbleweed, no?

LaszloGombos:
Okey, let's keep it as it is now, but for e.g. Fedora-Rawhide we should really have someone helping with the container, like we have for e.g. Gentoo.

Issue #423, opened 2024-06-28
TEST-14-IMSM fails with systemd 256-1ubuntu1


bdrung:
**Describe the bug**

TEST-14-IMSM fails with systemd 256-1ubuntu1:

```
dracut[I]: Executing: ./dracut.sh -N -l --keep --tmpdir /var/tmp/dracut-test.Nh4CQJ -m test-root -f /var/tmp/dracut-test.Nh4CQJ/initramfs.root 6.8.0-31-generic
dracut[I]: *** Including module: test-root ***
dracut[I]: *** Including module: debug ***
dracut[I]: *** Including modules done ***
dracut[I]: *** Installing kernel module dependencies ***
dracut[I]: *** Installing kernel module dependencies done ***
dracut[I]: *** Resolving executable dependencies ***
dracut[I]: *** Resolving executable dependencies done ***
dracut[I]: *** Store current command line parameters ***
dracut[I]: *** Creating image file '/var/tmp/dracut-test.Nh4CQJ/initramfs.root' ***
dracut[I]: *** Creating initramfs image file '/var/tmp/dracut-test.Nh4CQJ/initramfs.root' done ***
Not removing /var/tmp/dracut-test.Nh4CQJ/dracut.G82OhJ.
dracut[I]: Executing: ./dracut.sh -N -l -i /var/tmp/dracut-test.Nh4CQJ/overlay / -a "test-makeroot bash lvm mdraid dmraid kernel-modules" -d "piix ide-gd_mod ata_piix ext4 sd_mod dm-multipath dm-crypt dm-round-robin faulty linear multipath raid0 raid10 raid1 raid456" -I "grep sfdisk realpath" -i ./create-root.sh /lib/dracut/hooks/initqueue/01-create-root.sh -f /var/tmp/dracut-test.Nh4CQJ/initramfs.makeroot 6.8.0-31-generic
dracut[I]: 62bluetooth: Could not find any command of '/usr/lib/bluetooth/bluetoothd /usr/libexec/bluetooth/bluetoothd'!
dracut[I]: *** Including module: bash ***
dracut[I]: *** Including module: dash ***
dracut[I]: *** Including module: systemd ***
dracut[I]: *** Including module: systemd-ask-password ***
dracut[I]: *** Including module: systemd-initrd ***
dracut[I]: *** Including module: systemd-journald ***
dracut[I]: *** Including module: systemd-modules-load ***
dracut[I]: *** Including module: systemd-pcrphase ***
[...]

PR #478, opened 2024-07-07
test: avoid writing to rootfs as it might be read-only

LaszloGombos:
## Changes

There is no need to remount rootfs read only right before shutdown.

## Checklist


Conan-Kudo (approved):

bdrung:
This commit fixes https://github.com/dracut-ng/dracut-ng/issues/423

PR #332, opened 2024-06-03
fix(dracut-install): fixes for fw_devlink supplier handling

alpernebbi:
## Changes

Some small changes to fw_devlinks supplier handling to include some dependencies it misses, and a faster method to find out the module name for a sysfs node:
- Add sysfs node parents' modules as dependencies (and parents of suppliers and so on)
- Do not limit supplier handling to the `platform` bus 
- Install modules for all suppliers of a supplier's module
- Use `driver/module` sysfs dirs for module name

Ends up cutting start-up time by half (~500ms) on ARM64. But it adds some start-up time (<100ms) on AMD64 because of devices outside `/sys/devices/platform` which this now handles.

<details><summary>Benchmarks on ARM64  (mt8183-kukui-jacuzzi-cozmo)</summary>

Before:
```
$ hyperfine -L count "250,125,50,10,1" './dracut-install -D "$(mktemp -d)" -o -m $(lsmod | cut -d" " -f1 | head -n {count})'
Benchmark 1: ./dracut-install -D "$(mktemp -d)" -o -m $(lsmod | cut -d" " -f1 | head -n 250)
  Time (mean ± σ):      3.710 s ±  0.030 s    [User: 2.299 s, System: 1.365 s]
  Range (min … max):    3.644 s …  3.750 s    10 runs
 
Benchmark 2: ./dracut-install -D "$(mktemp -d)" -o -m $(lsmod | cut -d" " -f1 | head -n 125)
  Time (mean ± σ):      2.947 s ±  0.015 s    [User: 1.930 s, System: 0.994 s]
  Range (min … max):    2.912 s …  2.964 s    10 runs
 
Benchmark 3: ./dracut-install -D "$(mktemp -d)" -o -m $(lsmod | cut -d" " -f1 | head -n 50)
  Time (mean ± σ):      2.394 s ±  0.026 s    [User: 1.621 s, System: 0.742 s]
  Range (min … max):    2.354 s …  2.437 s    10 runs
 
Benchmark 4: ./dracut-install -D "$(mktemp -d)" -o -m $(lsmod | cut -d" " -f1 | head -n 10)
  Time (mean ± σ):      1.255 s ±  0.009 s    [User: 1.004 s, System: 0.249 s]
  Range (min … max):    1.245 s …  1.274 s    10 runs
 
Benchmark 5: ./dracut-install -D "$(mktemp -d)" -o -m $(lsmod | cut -d" " -f1 | head -n 1)
[...]

Issue #316, opened 2024-05-24
performance regression in dracut-install 060


bdrung:
**Describe the bug**

When compared to Ubuntu 23.10 (dracut 059), creating intramfs files with update-initramfs in Ubuntu 24.04 (dracut 060) takes 2 to 5 times more time on ARM devices.

**Distribution used**
Ubuntu

**Dracut version**
060

**To Reproduce**
Run dracut-install on an ARM device.

**Additional context**

* The bug was reported on Ubuntu: https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/2065180
* Poor performance on ARM was mentioned in https://salsa.debian.org/kernel-team/initramfs-tools/-/merge_requests/102

bdrung:
From the Ubuntu bug:
> first bad commit 3de4c7313260fb600507c9b87f780390b874c870

alpernebbi:
I think dracut-install parses every fw_devlink supplier to get a dependency graph (resulting in a lot of `/sys` and `modules.alias.bin` reads), then uses that to install dependencies for modules given as arguments. Maybe it could parse suppliers for only the given modules? Or maybe it could cache the supplier dependency information across multiple runs on the same system?

FWIW, I get this on a RK3399 system:

```
$ time /usr/lib/dracut/dracut-install -D "$(mktemp -d)" --kerneldir "/lib/modules/$(uname -r)" -v -o -m nonexistent
dracut-install: Failed to find module 'nonexistent'

real    0m2.596s
user    0m1.728s
sys     0m0.864s
```

<details><summary>Some syscalls it does</summary>

```
$ strace -e trace=readlinkat,openat,newfstatat /usr/lib/dracut/dracut-install -D "$(mktemp -d)" --kerneldir "/lib/modules/$(uname -r)" -v -o -m nonexistent | uniq -c | sort -n | tail -20
   4984 newfstatat(4, "", {st_mode=S_IFDIR|0755, st_size=0, ...}, AT_EMPTY_PATH) = 0
   1673 readlinkat(AT_FDCWD, "/sys/devices/platform", 0xffffec4b0a40, 1023) = -1 EINVAL (Invalid argument)
   1429 openat(AT_FDCWD, "..", O_RDONLY)        = 4
   1272 readlinkat(AT_FDCWD, "/sys/devices", 0xffffec4b0a40, 1023) = -1 EINVAL (Invalid argument)
[...]

PR #479, opened 2024-07-07
perf(dracut-install): use driver/module sysfs dirs for module name

alpernebbi:
## Changes

To handle fw_devlink suppliers as dependencies, we traverse every device in /sys and collect a hashmap of module names to their suppliers before starting module installation work. The module names are found by reading and passing the "modalias" file to the `kmod_module_new_from_lookup()` function from libkmod. This function is unexpectedly slow on ARM64 systems with device-tree where this work is primarily necessary.

Usually (but not always), the running kernel has already looked up the modalias and bound a driver from the associated module. Get module names by resolving the "driver/module" symbolic link and taking its basename. If it doesn't exist, fall back to looking up the modalias.

It's possible that multiple modules match the modalias, in which case others will be ignored in favor of the one already chosen by the kernel.

## Checklist

Split from #332, further improves performance for #316.

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -1508,8 +1508,26 @@ static bool check_module_path(const char *path)
         return true;
 }
 
-static int find_kmod_module_from_sysfs_node(struct kmod_ctx *ctx, const char *sysfs_node, int sysfs_node_len,
-                                            struct kmod_list **modules)
+static int find_kmod_module_from_sysfs_driver(struct kmod_ctx *ctx, const char *sysfs_node, int sysfs_node_len,
+                                              struct kmod_module **module)
+{
+        char mod_path[PATH_MAX], mod_realpath[PATH_MAX];
+        const char *mod_name;
+        if (snprintf(mod_path, sizeof(mod_path), "%.*s/driver/module",
+                     sysfs_node_len, sysfs_node) >= sizeof(mod_path))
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

bdrung (approved):
That shows a nice performance improvement on a Raspberry Pi Zero 2W:
```
$ hyperfine --warmup 1 '/usr/lib/dracut/dracut-install -D /tmp/foo/ -m raid1' './dracut-install -D /tmp/foo/ -m raid1'
Benchmark 1: /usr/lib/dracut/dracut-install -D /tmp/foo/ -m raid1
  Time (mean ± σ):      1.041 s ±  0.004 s    [User: 0.652 s, System: 0.385 s]
  Range (min … max):    1.036 s …  1.053 s    10 runs
 
Benchmark 2: ./dracut-install -D /tmp/foo/ -m raid1
  Time (mean ± σ):     776.6 ms ±   4.6 ms    [User: 433.5 ms, System: 340.6 ms]
  Range (min … max):   771.3 ms … 786.5 ms    10 runs
 
Summary
  ./dracut-install -D /tmp/foo/ -m raid1 ran
    1.34 ± 0.01 times faster than /usr/lib/dracut/dracut-install -D /tmp/foo/ -m raid1
```

LaszloGombos (approved):

PR #485, opened 2024-07-08
fix(nfs): support rpcbind user named _rpc

bdrung:
## Changes

The Debian/Ubuntu package `rpcbind` creates a user named `_rpc`. Support running `rpcbind` with that user. Add a helper function `get_rpc_user` that derives the desired user from parsing `/etc/passwd`.

## Checklist


LaszloGombos:
Fixes https://github.com/dracut-ng/dracut-ng/issues/102

PR #486, opened 2024-07-08
ci: improve test coverage

LaszloGombos:
- ci: run extended tests on openSUSE as well
- test(FULL-SYSTEM): make sure systemd-pcrphase can be installed

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -56,6 +56,7 @@ jobs:
                         "debian",
                         "fedora",
                         "gentoo",
+                        "opensuse",
                         "ubuntu",
                         "void",
                 ]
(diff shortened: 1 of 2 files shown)


dracut.sh
```diff
@@ -1282,7 +1282,7 @@ if [[ -f $dracutbasedir/dracut-version.sh ]]; then
 fi
 
 if systemd-detect-virt -c &> /dev/null; then
-    export DRACUT_NO_MKNOD=1 DRACUT_NO_XATTR=1
+    export DRACUT_NO_MKNOD=1 DRACUT_NO_XATTR=${DRACUT_NO_XATTR:-1}
```
Conan-Kudo, line 1285: Why are we defaulting to NO_XATTR here?

Conan-Kudo (changes requested):

dracut.sh
LaszloGombos: Here is the PR where that change landed https://github.com/dracutdevs/dracut/pull/1499 and referenced by https://github.com/dracutdevs/dracut/issues/890 .

For now I will just take this change out of this PR as it seems this needs more discussion.

Conan-Kudo (approved):

Issue #481, opened 2024-07-07
systemd-pcrphase for Gentoo CI container


LaszloGombos:
https://github.com/dracut-ng/dracut-ng/issues/92 can not be reproduced on the Gentoo container.

I hope to at least confirm that systemd-pcrphase can be installed in the FULL_SYSTEMD test. Other systemd enabled containers already have support for systemd-pcrphase, except Gentoo.

CC @AndrewAmmerlaan

AndrewAmmerlaan:
I'm not sure I understand the issue. Is the problem that systemd-pcrphase is not included specific to Gentoo? What exactly should I test?

LaszloGombos:
Issue is just with the Gentoo Docker file, to make sure that dependent binaries are in the Docker image for https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/01systemd-pcrphase/module-setup.sh#L9 . Thanks !

PR #487, opened 2024-07-09
fix(Dockerfile-Gentoo): add requirements for systemd-pcrphase

AndrewAmmerlaan:
I think this should be enough for the `FULL_SYSTEMD` tests, but lets see what the CI says.

Closes: https://github.com/dracut-ng/dracut-ng/issues/481


AndrewAmmerlaan:
I think the test failure is not related to this change.

LaszloGombos (approved):
Thank you very much @AndrewAmmerlaan

PR #489, opened 2024-07-10
fix(multipath): omit module if included with no multipath devices

LaszloGombos:
## Changes

Follow-up to 4957ffa935. Return 255 instead of 1 and change the order of checks.

This change also allows removing the multipath workaround the the FULL-SYSTEMD test.

## Checklist

CC @mwilck @bmarzins @aafeijoo-suse

### DIFF
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -23,6 +23,10 @@ majmin_to_mpath_dev() {
 check() {
     local _any_mpath_dev
 
+    # if there's no multipath binary, no go.
+    require_binaries multipath || return 1
+    require_binaries kpartx || return 1
+
     for_each_host_dev_and_slaves is_mpath
     _any_mpath_dev=$?
 
@@ -30,12 +34,8 @@ check() {
         [[ $_any_mpath_dev == 0 ]] || return 255
     }
 
-    # if there's no multipath binary, no go.
-    require_binaries multipath || return 1
-    require_binaries kpartx || return 1
-
     if [[ $_any_mpath_dev != 0 ]] && [[ ! -f /etc/multipath.conf ]]; then
-        return 1
+        return 255
     fi
 
     return 0
--- a/test/TEST-04-FULL-SYSTEMD/test.sh
+++ b/test/TEST-04-FULL-SYSTEMD/test.sh
@@ -27,7 +27,7 @@ client_run() {
     test_marker_reset
     "$testdir"/run-qemu \
         "${disk_args[@]}" \
-        -append "$TEST_KERNEL_CMDLINE systemd.unit=testsuite.target systemd.mask=systemd-firstboot systemd.mask=systemd-vconsole-setup rd.multipath=0 root=LABEL=dracut mount.usr=LABEL=dracutusr mount.usrfstype=btrfs mount.usrflags=subvol=usr,ro $client_opts rd.retry=3 $DEBUGOUT" \
+        -append "$TEST_KERNEL_CMDLINE systemd.unit=testsuite.target systemd.mask=systemd-firstboot systemd.mask=systemd-vconsole-setup root=LABEL=dracut mount.usr=LABEL=dracutusr mount.usrfstype=btrfs mount.usrflags=subvol=usr,ro $client_opts rd.retry=3 $DEBUGOUT" \
         -initrd "$TESTDIR"/initramfs.testing || return 1
 
     if ! test_marker_check; then


LaszloGombos:
Related discussion - https://github.com/dracutdevs/dracut/pull/2095

> This is only necessary on SUSE or openSUSE distributions

We should discuss if this PR causes any harm/risk/failure in other distributions, and if it does not than perhaps is better to maintain it here upstream than downstream.

bdrung (approved):
LGTM

freswa (approved):
LGTM

tblume (commented):
Hm, I don't think it is the best approach to replace the warning in the original commit by quitting the multipath module.
That way we remove the possibility to switch to a multipath device after installation.
There might be other usecases when forcing multipathing into the initrd is desired.

Do we really want to remove this possibility just for the sake of a test?

pvalena (approved):
LGTM, makes sense.

aafeijoo-suse:
> Hm, I don't think it is the best approach to replace the warning in the original commit by quitting the multipath module. That way we remove the possibility to switch to a multipath device after installation. There might be other usecases when forcing multipathing into the initrd is desired.
> 
> Do we really want to remove this possibility just for the sake of a test?

@tblume ICYMI you can read https://github.com/dracut-ng/dracut-ng/pull/467#pullrequestreview-2165183262 for more background on this unnecessary mess.

LaszloGombos:
> Do we really want to remove this possibility just for the sake of a test?

The test case is is not just a test. It is a reproduction of a scenario that many users and distributions are running into. Here upstream we rely on the test suite to avoid regression. Test suite is also critical to maintain support for various Linux distributions and not just make decisions based on one distribution.  

> There might be other usecases when forcing multipathing into the initrd is desired.

Agreed. These use-case can be achieved by various ways even after this PR by e.g. ```--add multipath``` or a config file or making sure that ```/etc/multipath.conf``` exists.

Conan-Kudo (approved):

bmarzins:
Wait. Which users are running into problems when the multipath module is installed? I understand the SUSE specific issue here. But that shouldn't be possible unless distos modify how the upstream multipath-tools work. FWIW, this change shouldn't cause Fedora/RHEL distros any harm, since they already disable multipath if no multipath.conf file exists, but I don't know of any other distros that work this way. I'm not saying that this change is definitely going to cause anyone harm, but it seems unnecessary, since unless I'm missing something, the old rules should work find with upstream multipath-tools. So I just don't understand the benefit of changing something that works in upstream and could mess with distros, without any distro asking for it. What am I missing?

LaszloGombos:
@bmarzins the project hopes to test openSUSE here upstream to avoid regressing openSUSE. Here is an earlier attempt on testing openSUSE - https://github.com/dracutdevs/dracut/issues/2353 .

One could argue that upstream should encourage distributions not to fork the dracut project downstream so that it is easier to have a discussion here upstream (e.g. on how to reproduce the bug), so it is a benefit of the upstream project to support openSUSE here upstream.

mwilck:
This is an improvement over the previous attempts to "fix" the dracut multipath module, but still, it isn't right. 

It's against the spirit of "non-hostonly" to exclude a module only because it isn't in use while the initrd is being built. In my understanding, the in-use check is for hostonly mode only. Users shouldn't need to use `--add $MODULE` in non-hostonly mode.

The check for `multipath.conf` is bogus. It is not needed on modern distributions[^fedora]. OTOH, it might very well exist even if multipath is not enabled, e.g. as a leftover from previous distributions. If anything, you should check `/etc/multipath/wwids`. But that might be wrong as well[^systemd]. 

It seems that the original intention of this was to avoid enabling multipath in non-hostonly mode on systems that don't have multipath devices to follow the "principle of least surprise". I can see that a multipathed root disk may be a "surprise" to some. *But it usually doesn't do any harm*. There may be a performance penalty, but I expect it to be hardly noticeable, except maybe systems with slow CPU and very fast storage. On the contrary, not enabling multipath in the initrd on a system that does enable multipath in the booted system is highly dangerous, and likely to lead to boot failure.

Here's my suggestion: In non-hostonly mode, if no multipath devices are found and no `multipath.conf` exists (and no `/etc/multipath/conf.d/*.conf`, either), do include the module, but add a synthetic multipath.conf file that looks like this, possibly emitting a warning that it does so:
```
defaults {
    find_multipaths strict
}
```
This will have the effect that no multipath devices will be activated unless they're explicitly listed in `/etc/multipath/wwids`, which is how Fedora/RHEL behaves anyway (and other distros, too, if they stick with the upstream defaults). Users could still use the emergency shell to add WWIDs if they need to, because the necessary tools would be available in the initrd. This would not be the case if the entire module was omitted.

(open)SUSE users would need to create an empty `multipath.conf` file if they want to preserve the default behavior of `find_multipaths greedy`. But then, openSUSE uses hostonly mode by default, anyway, and we have educated our users to use `--add` and `--omit`. But as I said above, in non-hostonly mode, `--add` shouldn't be required for any module.

[^fedora]: Fedora/RHEL is would be no "modern" distribution according to this definition. I'm not sure why they always create `multipath.conf`. In general, it's definitely not necessary, and more often than not, harmful.
[^systemd]: Something like `systemctl is-enabled multipathd.service` might also work, but only with late multipath-tools releases that have disabled socket activation.

pvalena:
@mwilck thanks for nice write up.
Yes, I fully agree that is the best way to resolve this. I was actually planning to have a config in Fedora to include the module unconditionally (or based on RPM deps) to restore  the previous behaviour - but your way makes more sense.

pvalena:
> So, how to move on here? Am I expected to create a PR?

Yes please, if you can, if not, I'll just create an issue for it so we can track it.

aafeijoo-suse:
> So, how to move on here? Am I expected to create a PR?

I don't see why @mwilck has to waste his time here to fix the "fix" (377d52cb9867dc4074fe78965e1bfcdf80be14a1) of the "fix" (4957ffa935ea8c393c8fed867f6051da4ca43d58) of something that no one asked for and that was merged regardless of the contrary opinion of those who really know. If something is known to be wrong, the first and intermediate action should be a revert, but I haven't seen that yet. Even worse, this series of commits were included in the latest [103 release](https://github.com/dracut-ng/dracut-ng/releases/tag/103)...

PR #490, opened 2024-07-11
dmdquash-live-root: Add support for using erofs

bcl:
This adds support for rootfs compressed with erofs. Either as a plain erofs image or a LiveOS/rootfs.img ext4 filesystem compressed with erofs.

This patch does not make any attempt to change the squashfs directory naming (or variable names) in order to make these changes as small as possible and easy to review. It also does not make any attempt to support the multitude of available options other than what is needed by anaconda-dracut calling this script to setup the boot.iso root filesystem.

(which isn't to say it doesn't work, it just hasn't been tested and is outside the scope of this change).

## Changes

Add support for erofs to dmsquash-live-root for use with this lorax PR - https://github.com/weldr/lorax/pull/1403

## Checklist

Fixes #

### DIFF
--- a/modules.d/90dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/90dmsquash-live/dmsquash-live-root.sh
@@ -97,7 +97,6 @@ det_img_fs() {
     blkid -s TYPE -u noraid -o value "$1"
 }
 
-load_fstype squashfs
 CMDLINE=$(getcmdline)
 for arg in $CMDLINE; do
     case $arg in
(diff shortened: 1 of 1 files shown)


LaszloGombos:
Thanks. Shell lint and commisery fails, please check !

Not requiring it but wondering how hard would be to extend test 16 and perhaps make it easy to run that test with erofs instead of squashfs. 

This is where the squashfs is created - https://github.com/dracut-ng/dracut-ng/blob/main/test/TEST-16-DMSQUASH/create-root.sh#L28

Here is some basic info how to run test locally:  https://github.com/dracut-ng/dracut-ng/wiki/Dracut-development#running-tests-with-docker

 Alternatively you can just keep uploading versions of PR to run the test on the CI.
 
 If this turns out to be hard to do we can land it without the test changes as long as you tested it manually as the changes are relatively self contained. In that case please share how you tested.

pvalena (approved):
LGTM, although I have not tested it in any way.

Conan-Kudo:
FYI @schaefi, we will probably want to support this in kiwi too.

Conan-Kudo:
@bcl Can you please follow our commit style? `feat(dmsquash-live): add support for using erofs` should do the trick.

Conan-Kudo:
There are also some lint errors that you should fix.

bcl:
I'll see if I can come up with a test. The way anaconda uses this is a bit different, it has its own dracut module that calls dmsquash-live-root directly. See https://github.com/rhinstaller/anaconda/blob/master/dracut/anaconda-lib.sh#L147

Eventually this should be as complete as possible, but for the moment I don't want to get bogged down in making erofs exactly like squashfs.

LaszloGombos (approved):
Thanks !

Conan-Kudo:
This was merged with a typo in the commit message, FYI. `dmdquash` -> `dmsquash`

PR #494, opened 2024-07-12
fix(test): run test 14 with systemd again

bdrung:
## Changes

Commit 061a1068ba4c872e48605b9b0ecd282f4630ea8f ("test: avoid writing to rootfs as it might be read-only") fixes the test failure with systemd
256. So revert the changes to `test/TEST-14-IMSM` from commit 130f4dfce48b187944be9a0cacca794dd32428ad.

## Checklist


LaszloGombos (approved):
thanks

PR #496, opened 2024-07-12
Erofs tests

bcl:
## Changes

Adds testing for erofs support in dmsquash-live, makes sure the erofs kernel module is installed when using dmsquash-live.

## Checklist
- [X] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [X] I am providing new code and test(s) for it

PR #497, opened 2024-07-13
feat(test): add erofs-utils to the containers

bcl:
Add erofs-utils to all the containers for use in testing #496

### DIFF
--- a/test/container/Dockerfile-Arch
+++ b/test/container/Dockerfile-Arch
@@ -16,6 +16,7 @@ RUN pacman --noconfirm -Syu \
     dhcp \
     dmraid \
     elfutils \
+    erofs-utils \
     f2fs-tools \
     git \
     glibc \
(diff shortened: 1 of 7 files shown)


LaszloGombos:
Sadly Gentoo container  needs fixing.

LaszloGombos:
You could just not change Gentoo and when you make your changes to the test check if erofs-utils is available before testing it.. This is a good idea anyways and we can always fix Gentoo container later. 

Technically as long as at least one container is testing your feature, we will have coverage.

test/container/Dockerfile-Gentoo
```diff
@@ -46,6 +46,7 @@ RUN emerge --quiet --deep --autounmask-continue=y --with-bdeps=n \
     sys-fs/btrfs-progs \
     sys-fs/cryptsetup \
     sys-fs/dmraid \
+    sys-fs/erofs-utils \
```
LaszloGombos, line 49: I would just remove this change for now as I am not sure it is stable.

LaszloGombos: pushed a fix

LaszloGombos (approved):
Thanks

PR #499, opened 2024-07-14
fix(dracut): re-enable extended attributes in containers

LaszloGombos:
## Changes

Add common config to make it easier to disable xattr.
Running inside a container is not a good proxy for making decisions about filesystem properties.

Partial revert of https://github.com/dracutdevs/dracut/pull/1499

## Checklist

CC @cgwalters  @DaanDeMeyer for visibility

### DIFF
--- a/dracut.conf.d/50-no-xattr.conf.example
+++ b/dracut.conf.d/50-no-xattr.conf.example
@@ -1,0 +1,2 @@
+# disable xattr
+export DRACUT_NO_XATTR=1
--- a/dracut.sh
+++ b/dracut.sh
@@ -1282,7 +1282,7 @@ if [[ -f $dracutbasedir/dracut-version.sh ]]; then
 fi
 
 if systemd-detect-virt -c &> /dev/null; then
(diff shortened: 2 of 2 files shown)


pvalena (approved):
LGTM.

cgwalters:
Makes sense to me!

cgwalters (approved):

dracut.conf.d/50-no-xattr.conf.example
```diff
@@ -0,0 +1,2 @@
+# disable xattr
+DRACUT_NO_XATTR=1
```
aafeijoo-suse, line 2: Not to mention the convenience of this new approach of educating users to add environment variables as configuration options... I wonder how an [external C binary](https://github.com/dracut-ng/dracut-ng/blob/1e802f15fee3d6402d83e0efefef8bb88c5a33e3/src/install/dracut-install.c#L2299) can get this value without an `export`... because this variable is not being exported anywhere in `dracut.sh`. Does anyone tested that this _really_ works?

Conan-Kudo (approved):

PR #501, opened 2024-07-14
test(DMSQUASH): no need to partition root_erofs disk

LaszloGombos:
## Changes

Only one partition is needed, so need need to repartition. This makes the code easier to reason about and test runs a bit faster.

## Checklist


LaszloGombos:
@bcl I was wondering if you take a quick look as you have been working on this part. Thanks for helping with tests.

Conan-Kudo (approved):

PR #507, opened 2024-07-15
fix(systemd): move installation of libkmod to udev-rules module

puleglot:
## Changes
- move installation of libkmod.so from systemd module to udev-rules module

## Checklist

Downstream bug report: https://bugs.gentoo.org/935548

### DIFF
--- a/modules.d/95udev-rules/module-setup.sh
+++ b/modules.d/95udev-rules/module-setup.sh
@@ -93,7 +93,11 @@ install() {
         "${udevdir}"/usb_id \
         "${udevdir}"/v4l_id
 
-    inst_libdir_file "libnss_files*"
+    # Install required libraries.
+    _arch=${DRACUT_ARCH:-$(uname -m)}
+    inst_libdir_file \
+        {"tls/$_arch/",tls/,"$_arch/",}"libkmod.so*" \
+        {"tls/$_arch/",tls/,"$_arch/",}"libnss_files*"
 
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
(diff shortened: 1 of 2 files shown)


modules.d/01systemd-udevd/module-setup.sh
```diff
@@ -71,6 +71,8 @@ install() {
 
     # Install required libraries.
     _arch=${DRACUT_ARCH:-$(uname -m)}
-    inst_libdir_file {"tls/$_arch/",tls/,"$_arch/",}"libudev.so.*"
+    inst_libdir_file \
+        {"tls/$_arch/",tls/,"$_arch/",}"libudev.so.*" \
+        {"tls/$_arch/",tls/,"$_arch/",}"libkmod.so*"
```
LaszloGombos, line 76: I think a better solution would be to remove the change from systemd-udevd and unconditionally install libkmod.so* in udev-rules as systemd-udev has a dependency on udev-rules. Would you be able to make the change and retest please ?

modules.d/95udev-rules/module-setup.sh
```diff
@@ -93,7 +93,12 @@ install() {
         "${udevdir}"/usb_id \
         "${udevdir}"/v4l_id
 
-    inst_libdir_file "libnss_files*"
+    if ! dracut_module_included "systemd"; then
```
LaszloGombos, line 96: If the change removed from systemd-udevd, testing for ! dracut_module_included "systemd" not needed here.

modules.d/01systemd-udevd/module-setup.sh
puleglot: Indeed. For some reason I thought that systemd-udevd is also trying to install udev binary, but it's just checking its existence. So it's logical to install udev deps from udev-rules module.

AndrewAmmerlaan (approved):

LaszloGombos (approved):
lgtm

PR #509, opened 2024-07-16
fix(multipath): include module with "find_multipaths strict"

mwilck:
The previous commits 377d52c and 4957ffa excluded the multipath module in non-hostonly mode, with the purpose of avoiding to start multipath in the initramfs on systems that don't use multipath. This is against the idea of non-hostonly, which is to build an initramfs with all kernel modules and tools that may be necessary to bring up the system.

This patch takes a different approach. In non-hostonly mode, if there are no multipath devices and no multipath.conf, include the multipath module, but require "find_multipaths strict" policy. This will ensure that users that don't use dm-multipath won't be surprised by finding their root device under multipath control. But the multipath tools will still be included in the initramfs, allowing users to set up multipathing e.g. on the emergency shell if necessary. Also, if /etc/multipath/wwids exists, it will be included in the initramfs and will cause the WWIDs listed in that file to be multipathed.

Note that in multipath-tools upstream and many Linux distributions, "find_multipaths strict" is the default anyway. Users of distributions with different defaults (e.g. openSUSE) would just need to create /etc/multipath.conf (possibly with empty contents) to override this behavior if they intend to activate multipath in the initramfs.

Fixes: 377d52c ("fix(multipath): omit module if included with no multipath devices")
Fixes: 4957ffa ("fix(multipath): omit module if included with no multipath devices")

This pull request changes...

## Checklist

Fixes: #489

CC: @aafeijoo-suse @bmarzins @Conan-Kudo @pvalena @LaszloGombos


LaszloGombos (approved):
lgtm

In some distro's (e.g. Debian) this PR introduces the following log message, but tests seems to pass regardless of the log message.

```
[FAILED] Failed to start multipathd…apper Multipath Device Controller.
```

modules.d/90multipath/module-setup.sh
```diff
@@ -131,6 +122,19 @@ install() {
...
+    [[ $hostonly ]] || {
+        for_each_host_dev_and_slaves is_mpath \
+            || [[ -f /etc/multipath.conf ]] || {
+            cat > "${initdir}"/etc/multipath.conf << EOF
+# Minimal multipath.conf generated by dracut
+# Avoid any devices being multipathed by default
+defaults {
+         find_multipaths strict
+}
+EOF
```
pvalena, line 134: Since this is an install of the module, could we maybe keep a file `multipath.conf` and install it like in[ e.g.](https://github.com/dracut-ng/dracut-ng/blob/ef0972fe5349bdf6e821cb79a47cf412caf99059/modules.d/06rngd/module-setup.sh#L35)?

pvalena (approved):
Just a suggestion / RFE; otherwise LGTM, thanks!

modules.d/90multipath/module-setup.sh
mwilck: Feel free to do that. IMO this is simple enough to be kept inline.

bmarzins:
LGTM

modules.d/90multipath/module-setup.sh
LaszloGombos: A separate file does not seem to have many benefits here and can be considered a follow-up PR

jlebon:
Hmm, I think this conflicts with [multipathd-configure.service](https://github.com/dracut-ng/dracut-ng/blob/b0c37531092d3ccd20a1a867975b67c44aa8a5f9/modules.d/90multipath/multipathd-configure.service) as a way to auto-create `/etc/multipath.conf` in the initrd when `rd.multipath=default` is passed because now there'll be a default `multipath.conf` that will inhibit its `ConditionPathExists=!/etc/multipath.conf` conditional. And sadly the default here is different from the default from `mpathconf --enable` (`strict` vs `yes`).

Can't we just continue to not ship a `multipath.conf` by default in the initrd but still ship the tools in non-hostonly mode?

bmarzins:
> Hmm, I think this conflicts with [multipathd-configure.service](https://github.com/dracut-ng/dracut-ng/blob/b0c37531092d3ccd20a1a867975b67c44aa8a5f9/modules.d/90multipath/multipathd-configure.service) as a way to auto-create `/etc/multipath.conf` in the initrd when `rd.multipath=default` is passed because now there'll be a default `multipath.conf` that will inhibit its `ConditionPathExists=!/etc/multipath.conf` conditional. And sadly the default here is different from the default from `mpathconf --enable` (`strict` vs `yes`).
> 
> Can't we just continue to not ship a `multipath.conf` by default in the initrd but still ship the tools in non-hostonly mode?

You're right that this does conflict with multipathd-configure.service. The reason that @mwilck added the default config file has to do with distribution differences.  Only Centos and Fedora require a /etc/multipath.conf file in order for multipath to run (I think. Possibly other distros do this too). In upstream multipath, if you don't have /etc/multipath.conf installed, multipath and multipathd will run and just use the default config. This means that, on anything other than Centos and Fedora, if you install the multipath bits in dracut, it's going to run run multipath in the initramfs with the default config. In upstream multipath, this will set find_multipaths to "strict", so adding this file is unnecessary, since it doesn't change anything. However OpenSUSE (and possibly other distros) set find_multipaths to "greedy" by default. This means that if OpenSUSE includes the multipath bits, but not a config file, multipath will claim everything. This is obviously not what they want.

Perhaps an alternate solution could be to still install the multipath bits, but instead of installing a default /etc/multipath.conf file in this non-hostonly case, we could just disable multipathd.service. That will stop multipath from claiming any devices, and it should work for Centos/Fedora, OpenSUSE, and upstream. Users would still be able to start up multipathing from the dracut shell if necessary. How does that sound @mwilck?

mwilck:
Hm, I don't think that users expect to have to start multipathd.service manually from the dracut shell. That would be a pretty bad UI IMO, and much worse than having a root device unintentionally multipathed (see PR description above).

How does Fedora behave if `rd.multipath=default` is not set, and there is no `multipath.conf`? Could we just skip embedding `multipath.conf` in the initrd if `multipath-configure.service` is present / enabled?

bmarzins:
If there is no multipath.conf, multipathd won't start and multipath won't claim anything, regardless of rd.multipath. Not adding a multipath.conf if multipath-configure.service is getting installed makes sense.

mwilck:
@bmarzins, as this is already merged and needs a Fedora fix, would you mind creating a PR to fix it?

bmarzins:
Sure.

jlebon:
> Could we just skip embedding multipath.conf in the initrd if multipath-configure.service is present / enabled?

This SGTM. Though currently `multipathd-configure.service` is installed by default by the dracut module here as long as the `systemd` module is installed and `mpathconf` is available, which I had assumed is most cases. E.g. does OpenSUSE not have `multipathd-configure.service` in its initrd presently?

(But again, proposal WFM.)

bmarzins:
@jlebon mpathconf is also redhat only. upstream didn't want it, since it just uses the default configuration if there is no multipath.conf file, and distros that don't like the upstream defaults just change them directly.

PR #511, opened 2024-07-17
feat(systemd*): Include systemd config files from /usr/lib/systemd

pvalena:
and also use proper variables for the paths.

## Changes

The new systemd reads from both /etc and /usr/, so to accomodate this, I've added new paths to install configs from (I probably haven't covered all). This changes only hostonly behaviour; uses global variables:

```
# dracut.sh
set_global_var "systemd" "systemdutildir" "/lib/systemd:/lib/systemd/systemd-udevd" "/usr/lib/systemd:/usr/lib/systemd/systemd-udevd"
set_global_var "systemd" "systemdutilconfdir" "/etc/systemd"
```
(also defined in the config `/usr/lib/dracut/dracut.conf.d/01-dist.conf`)

Ref: https://issues.redhat.com/browse/RHEL-32506

## Checklist
```bash
# F41 (vagrant; cloud variant):
[root@localhost ~]# lsinitrd | grep system.conf
-rw-r--r--   1 root     root         2342 Jun  2 00:00 usr/lib/systemd/system.conf

# F40 (VM, server variant):
~ lsinitrd | grep system.conf
-rw-r--r--   1 root     root         2124 Mar 10 20:00 etc/systemd/system.conf
-rw-r--r--   1 root     root         2318 May  8 20:00 usr/lib/systemd/system.conf
drwxr-xr-x   2 root     root            0 Jun  3 20:00 usr/lib/systemd/system.conf.d
-rw-r--r--   1 root     root          217 Apr 11 20:00 usr/lib/systemd/system.conf.d/longer-default-shutdown-timeout.conf

# Rawhide (workstation):
$ sudo lsinitrd /boot/initramfs-6.10.0-64.fc41.x86_64.img | grep systemd\.conf
-rw-r--r--   1 root     root          124 Jun  4 02:00 etc/conf.d/systemd.conf
-rw-r--r--   1 root     root          773 Sep 27  2023 usr/lib/modprobe.d/systemd.conf
-rw-r--r--   1 root     root         2289 Sep 27  2023 usr/lib/tmpfiles.d/systemd.conf
```

### DIFF
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -42,6 +42,8 @@ install() {
         "$systemdutildir"/system-generators/systemd-debug-generator \
         "$systemdutildir"/system-generators/systemd-fstab-generator \
         "$systemdutildir"/system-generators/systemd-gpt-auto-generator \
+        "$systemdutildir"/system.conf \
+        "$systemdutildir"/system.conf.d/*.conf \
         "$systemdsystemunitdir"/debug-shell.service \
         "$systemdsystemunitdir"/cryptsetup.target \
         "$systemdsystemunitdir"/cryptsetup-pre.target \
@@ -94,8 +96,8 @@ install() {
 
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
-            /etc/systemd/system.conf \
-            /etc/systemd/system.conf.d/*.conf \
+            "$systemdutilconfdir"/system.conf \
+            "$systemdutilconfdir"/system.conf.d/*.conf \
             /etc/hosts \
             /etc/hostname \
             /etc/nsswitch.conf \
--- a/modules.d/01systemd-coredump/module-setup.sh
+++ b/modules.d/01systemd-coredump/module-setup.sh
@@ -35,6 +35,7 @@ install() {
     inst_multiple -o \
         "$sysctld"/50-coredump.conf \
         "$systemdutildir"/coredump.conf \
+        "$systemdutildir/coredump.conf.d/*.conf" \
         "$systemdutildir"/systemd-coredump \
         "$systemdsystemunitdir"/systemd-coredump.socket \
         "$systemdsystemunitdir"/systemd-coredump@.service \
@@ -52,7 +53,7 @@ install() {
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
             "$systemdutilconfdir"/coredump.conf \
-            "$systemdsystemconfdir/coredump.conf.d/*.conf" \
+            "$systemdutilconfdir/coredump.conf.d/*.conf" \
             "$systemdsystemconfdir"/systemd-coredump.socket \
             "$systemdsystemconfdir/systemd-coredump.socket.d/*.conf" \
             "$systemdsystemconfdir"/systemd-coredump@.service \
--- a/modules.d/01systemd-pstore/module-setup.sh
+++ b/modules.d/01systemd-pstore/module-setup.sh
@@ -34,6 +34,8 @@ install() {
     inst_dir /var/lib/systemd/pstore
     inst_multiple -o \
         "$tmpfilesdir/systemd-pstore.conf" \
+        "$systemdutildir"/pstore.conf \
+        "$systemdutildir/pstore.conf.d/*.conf" \
         "$systemdutildir"/systemd-pstore \
         "$systemdsystemunitdir"/systemd-pstore.service \
         "$systemdsystemunitdir/systemd-pstore.service.d/*.conf"
--- a/modules.d/01systemd-resolved/module-setup.sh
+++ b/modules.d/01systemd-resolved/module-setup.sh
@@ -50,6 +50,7 @@ install() {
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
+            "$systemdutilconfdir"/resolv.conf \
             "$systemdutilconfdir"/resolved.conf \
             "$systemdutilconfdir/resolved.conf.d/*.conf" \
             "$systemdsystemconfdir"/systemd-resolved.service \
--- a/modules.d/01systemd-timesyncd/module-setup.sh
+++ b/modules.d/01systemd-timesyncd/module-setup.sh
@@ -40,6 +40,7 @@ install() {
         "$systemdntpunits/*.list" \
         "$systemdutildir"/systemd-timesyncd \
         "$systemdutildir"/systemd-time-wait-sync \
+        "$systemdutildir"/timesyncd.conf \
         "$systemdutildir/timesyncd.conf.d/*.conf" \
         "$systemdsystemunitdir"/systemd-timesyncd.service \
         "$systemdsystemunitdir/systemd-timesyncd.service.d/*.conf" \


modules.d/01systemd-coredump/module-setup.sh
```diff
@@ -51,6 +51,7 @@ install() {
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
+            "$systemdutildir"/coredump.conf \
             "$systemdutilconfdir"/coredump.conf \
```
LaszloGombos, line 55: Is the following missing ? 

```
"$systemdutildir"/coredump.conf.d/*.conf" \
```

modules.d/01systemd-coredump/module-setup.sh
```diff
@@ -51,6 +51,7 @@ install() {
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
+            "$systemdutildir"/coredump.conf \
```
LaszloGombos, line 54: Is the following missing ? 

```
"$systemdutildir"/coredump.conf.d/*.conf" \
```

modules.d/00systemd/module-setup.sh
```diff
@@ -94,8 +94,10 @@ install() {
 
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
-            /etc/systemd/system.conf \
-            /etc/systemd/system.conf.d/*.conf \
+            "$systemdutildir"/system.conf \
+            "$systemdutildir"/system.conf.d/*.conf \
```
aafeijoo-suse, line 98: hostonly sections should include only files under /etc

modules.d/01systemd-coredump/module-setup.sh
aafeijoo-suse, line 54: > Is the following missing ?
> "$systemdutildir"/coredump.conf.d/*.conf" \

Yes, and the other is already added:
https://github.com/dracut-ng/dracut-ng/blob/1e802f15fee3d6402d83e0efefef8bb88c5a33e3/modules.d/01systemd-coredump/module-setup.sh#L37

modules.d/01systemd-journald/module-setup.sh
```diff
@@ -62,6 +62,8 @@ install() {
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
+            "$systemdutildir"/journald.conf \
+            "$systemdutildir/journald.conf.d/*.conf" \
```
aafeijoo-suse, line 66: Already added:
https://github.com/dracut-ng/dracut-ng/blob/1e802f15fee3d6402d83e0efefef8bb88c5a33e3/modules.d/01systemd-journald/module-setup.sh#L37-L38

modules.d/01systemd-networkd/module-setup.sh
```diff
@@ -80,6 +80,8 @@ install() {
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
+            "$systemdutildir"/networkd.conf \
+            "$systemdutildir/networkd.conf.d/*.conf" \
```
aafeijoo-suse, line 84: Already added:
https://github.com/dracut-ng/dracut-ng/blob/1e802f15fee3d6402d83e0efefef8bb88c5a33e3/modules.d/01systemd-networkd/module-setup.sh#L40-L41

modules.d/01systemd-pstore/module-setup.sh
```diff
@@ -44,6 +44,8 @@ install() {
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
+            "$systemdutildir"/pstore.conf \
+            "$systemdutildir/pstore.conf.d/*.conf" \
```
aafeijoo-suse, line 48: hostonly sections should include only files under /etc

modules.d/01systemd-resolved/module-setup.sh
```diff
@@ -51,7 +51,9 @@ install() {
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
             "$systemdutilconfdir"/resolved.conf \
+            "$systemdutildir"/resolved.conf \
             "$systemdutilconfdir/resolved.conf.d/*.conf" \
+            "$systemdutildir/resolved.conf.d/*.conf" \
```
aafeijoo-suse, line 56: Already added:
https://github.com/dracut-ng/dracut-ng/blob/1e802f15fee3d6402d83e0efefef8bb88c5a33e3/modules.d/01systemd-resolved/module-setup.sh#L40-L41

modules.d/01systemd-timesyncd/module-setup.sh
```diff
@@ -59,6 +59,8 @@ install() {
             "$systemdntpunitsconfdir/*.list" \
             "$systemdutilconfdir"/timesyncd.conf \
             "$systemdutilconfdir/timesyncd.conf.d/*.conf" \
+            "$systemdutildir"/timesyncd.conf \
```
aafeijoo-suse, line 62: hostonly sections should include only files under /etc

modules.d/01systemd-timesyncd/module-setup.sh
```diff
@@ -59,6 +59,8 @@ install() {
...
+            "$systemdutildir/timesyncd.conf.d/*.conf" \
```
aafeijoo-suse, line 63: Already added:
https://github.com/dracut-ng/dracut-ng/blob/1e802f15fee3d6402d83e0efefef8bb88c5a33e3/modules.d/01systemd-timesyncd/module-setup.sh#L43

modules.d/00systemd/module-setup.sh
pvalena: Why is that?
Systemd definitely takes host configuration from `/usr`... that's the point of this PR.
Why not to include it, when initrd conf. should ideally behave like the real system in host-only mode?

aafeijoo-suse: conf shipped by packages => `/usr` => non-hostonly (generic, not specific to any host)
conf created by some admin => `/etc` => hostonly (specific to the host)

I'm not saying that the conf in `/usr` should not be included, but always include it outside of the hostonly check.

lnykryn: I agree, the /usr file should be always included

modules.d/01systemd-coredump/module-setup.sh
lnykryn: I think the scheme with latest 256 systemd should be just similar. /usr stuff included always, /etc for hostonly.

But

At least in the rhel/centos stream, we set by default #Storage=external, which is, in my opinion, not a good default for initrd. The core-dump is stored in /var, which is effectively deleted after the switch root. Maybe storage=journal is a better default, and we should have our own default conf file?

pvalena:
Thanks for the review; fixed with: https://gist.github.com/pvalena/66fa9c500d52e3242af010e933d96e47

modules.d/01systemd-coredump/module-setup.sh
```diff
@@ -52,6 +53,7 @@ install() {
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
             "$systemdutilconfdir"/coredump.conf \
+            "$systemdutilconfdir/coredump.conf.d/*.conf" \
             "$systemdsystemconfdir/coredump.conf.d/*.conf" \
```
aafeijoo-suse, line 57: L57 is clearly a typo, it should be exactly as the line you're adding.

modules.d/01systemd-resolved/module-setup.sh
```diff
@@ -50,6 +50,7 @@ install() {
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
+            "$systemdutilconfdir"/resolv.conf \
```
aafeijoo-suse, line 53: I don't know anything about how `systemd-resolved` works, but does `/etc/systemd/resolv.conf` really exist? I thought the old standard was `/etc/resolv.conf`, and that is the only one I find on my Fedora vms, symlinked to `/run/systemd/resolve/stub-resolv.conf`.

pvalena: I have no idea. I've added it for compatibility purposes - i.e. make the approach standardized - everything to load from the same paths; as it has `        "$systemdutildir"/resolv.conf \` entry, I've simply translated it into the hostonly variant.

modules.d/01systemd-coredump/module-setup.sh
pvalena: Right.

aafeijoo-suse (commented):
@pvalena one last thing, I promise :) 
You know that since systemd-v256 udev also allow dropins: https://github.com/systemd/systemd/commit/967cd1712c8b2a352f8fa6700dcf1f8db70df3a0

I'm aware that `/usr/lib/udev/udev.conf.d/` is the place where openSUSE will ship custom distro-specific udev conf (maybe Fedora too), so you can also add the conf files under that path (and `/etc/udev/udev.conf.d/`) within the `systemd-udevd` module using this PR for completeness.

jozzsi:
`/etc/udev/udev.conf.d/` is already included in `https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/95udev-rules/module-setup.sh#L107` . `/usr/lib/udev/udev.conf.d/` should also be included in the `udev-rules` module 

Dracut should be able to also handle the case when `systemd-udevd module` is not included in the initrd but `udev-rules` module is included (I think this is the case for the Fedora live iso or when user omits systemd explicitly on Fedora).

aafeijoo-suse:
> `/etc/udev/udev.conf.d/` is already included in [udev-rules](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/95udev-rules/module-setup.sh#L107) .

That is not true, see the code you linked.

> `/usr/lib/udev/udev.conf.d/` should also be included in the `udev-rules` module.
>
> Dracut should be able to also handle the case when `systemd-udevd` module is not included in the initrd but `udev-rules` module is included (I think this is the case when user omits systemd explicitly on Fedora).

AFAIK, dropins are a systemd-specific feature.

jozzsi:
> AFAIK, dropins are a systemd-specific feature.

systemd is a large set of tools - including the `systemd-udevd` binary and the `systemd` binary. 

My understanding is that files read by `systemd-udevd` binary should be handled in the `udev-rules` module, since `systemd-udevd` gets included here [udev-rules](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/95udev-rules/module-setup.sh#L13) .

Files that are read by systemd binary (e.g. to understand when and how to start `systemd-udevd`, such as `systemd-udevd.service`) should be included in the `systemd-udevd` module. 

The name of the dracut modules are indeed confusing. 

If files under `/usr/lib/udev/udev.conf.d/` meant to configure `systemd-udevd` itself (and not how `systemd-udevd` gets started), than I think it is safer to include them in `systemd-udevd`.

From the linked commit (https://github.com/systemd/systemd/commit/967cd1712c8b2a352f8fa6700dcf1f8db70df3a0) it seems `/etc/udev/udev.conf` and `/usr/lib/udev/udev.conf.d/` are serving a similar purpose of providing input directly to  `systemd-udevd` binary.

modules.d/01systemd-resolved/module-setup.sh
jozzsi: I could not find a reference to `"$systemdutildir"/resolv.conf` in [systemd-resolved.service](https://github.com/systemd/systemd/blob/main/man/systemd-resolved.service.xml)

jozzsi: I could not find a reference to "$systemdutildir"/resolv.conf in systemd-resolved.service

jozzsi:
Learning more about dracut I just noticed that @puleglot and @AndrewAmmerlaan worked in this area of the code as part of https://github.com/dracut-ng/dracut-ng/pull/507 . Use case is that `systemd-udevd` binary is included (from `udev-rules` module) but systemd is not included (as `openrc` is used). I hope they can help review this PR as well.

modules.d/01systemd-resolved/module-setup.sh
pvalena: I suppose `/etc/resolv.conf` is used by the process itself - I would simply continue with this non-breaking change if you don't mind.

LaszloGombos (approved):
lgtm.

Thanks for the reviews @aafeijoo-suse .

PR #512, opened 2024-07-17
fix(lsinitrd): check skipcpio file directly

pvalena:
as the folder in /usr/lib/dracut/src might just exist; but the file might not.

## Changes

`lsinitrd` currently checks for directory existence, and assumes the file is just there; but that might not be the case. I have changed the check to explicitly check for the file in that folder. No behaviour change.

## Checklist

### DIFF
--- a/lsinitrd.sh
+++ b/lsinitrd.sh
@@ -347,7 +347,7 @@ case $bin in
                 echo "Early CPIO image"
                 list_files
             fi
-            if [[ -d "$dracutbasedir/src/skipcpio" ]]; then
+            if [[ -f "$dracutbasedir/src/skipcpio/skipcpio" ]]; then
                 SKIP="$dracutbasedir/src/skipcpio/skipcpio"
             else
                 SKIP="$dracutbasedir/skipcpio"


lsinitrd.sh
```diff
@@ -347,7 +347,7 @@ case $bin in
                 echo "Early CPIO image"
                 list_files
             fi
-            if [[ -d "$dracutbasedir/src/skipcpio" ]]; then
+            if [[ -f "$dracutbasedir/src/skipcpio/skipcpio" ]]; then
```
LaszloGombos, line 350: Perhaps `-x` is better here than `-f` ?

Conan-Kudo: well it depends on whether we only care about the presence of the file or if we also need to check for executable? I'm not sure it matters much in this case, so `-f` is fine?

LaszloGombos: Few lines below `-x` is checked for the same files

`if ! [[ -x $SKIP ]]; then1

Conan-Kudo: Ah, then yes, let's simplify here and use `-x`. It gives a _small_ speedup to simplify.

pvalena: I've opted for `-f` as there's already `-x` later in the code. I'd rather avoid refactoring :) - at least from my side.

LaszloGombos: My suggestion was simply use `-x` without additional refactoring (check for `-x` twice) to help to reason about the code, but the difference is marginal.

PR #515, opened 2024-07-20
chore!: drop support for ifcfg

### LINKED ISSUE #503: retire ifcfg module
This module is potentially used by Fedora and it is also used by some of the tests in the CI. 

The project should make steps towards having an agreement in place foe the timeline of this module and remove it from the upstream tree .

@pvalena

LaszloGombos:
## Checklist

Fixes #503


LaszloGombos:
CC @dtardon and @steffen-maier

Conan-Kudo:
I believe this module is still used by Mageia, OpenMandriva, and SUSE at the moment, though I think at least OpenMandriva and SUSE are in the process of migrating to NetworkManager.

FYI @berolinux @dracut-ng/opensuse-maint

berolinux:
The last OpenMandriva release that still uses ifcfg, 5.0, is about to go EOL. No problem from our side.

mwilck:
SUSE uses (and will continue to use) `35network-legacy`, but not `45ifcfg`.

@aafeijoo-suse, correct me if I'm wrong.

aafeijoo-suse:
> SUSE uses (and will continue to use) `35network-legacy`, but not `45ifcfg`.
> 
> @aafeijoo-suse, correct me if I'm wrong.

Yes, you're right.

Conan-Kudo:
My understanding is that SUSE Framework One / SUSE Linux 16 platform will be using NetworkManager, but either way, we have confirmation that it isn't being used by SUSE platforms anymore.

mwilck:
> My understanding is that SUSE Framework One / SUSE Linux 16 platform will be using NetworkManager

I was talking about current SLE.

Conan-Kudo:
@mwilck So you okay with us dropping this, or do you need it still?

mwilck:
ok for me.

Conan-Kudo (approved):

mikhailnov:
I do not understand why we need to pull such a complex thing as NetworkManager and dbus into initrd...

jozzsi:
> I do not understand why we need to pull such a complex thing as NetworkManager and dbus into initrd...

I find it hard to find out what action is expected to be taken on this comment here upstream. Can you please file a bug and include additional information (such us https://abf.io/import/dracut/blob/rosa2023.1/0026-Revert-chore-drop-support-for-ifcfg.patch) on what actions you expect from upstream maintainers who are reading this comment. Thanks @mikhailnov .

mwilck:
I interpret @mikhailnov's comment as an argument against this PR. Some people are trying to simplify the initrd, and not everyone uses systemd; for these users, the simplest possible networking module makes some sense.

OTOH, this module has been unmaintained for some time, and @mikhailnov, you can re-add it in your environment if you need it.
Therefore I still second this PR.

jozzsi:
> complex thing as NetworkManager and dbus ..

From this comment I led to believe that @mikhailnov believes that because `ifcfg` is now removed `NetworkManager` is the only remaining option for networking. This is not true. In addition to `NetworkManager`, `network-legacy`, `systemd-networkd` and perhaps also `connman` is still expected to work (and none of it has a hard dependency on `ifcfg`). See https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/40network/module-setup.sh#L14

PR #527, opened 2024-07-20
feat(shell-interpreter): meta package for improved shell selection

### LINKED ISSUE #522: base module should be dependent on one of the shell modules
**Describe the bug**
```
dracut -m crypt
```

leads to an invalid initrd as it does not install any shell module due to lack of dependencies.

```
dracut[I]: *** Including module: crypt ***
dracut[I]: *** Including module: dm ***
dracut[I]: *** Including module: rootfs-block ***
dracut[I]: *** Including module: udev-rules ***
dracut[I]: *** Including module: base ***
dracut[I]: *** Including module: fs-lib ***
dracut[I]: *** Including modules done ***
```

`base` module should be dependent on a shell module

LaszloGombos:
## Changes

cherry-pick the dracut module from https://github.com/dracutdevs/dracut/pull/2564 and use it to improve the CI and demonstrate the usefulness of this new dracut module and fix all the exposed bugs.

If dash and busybox is NOT installed or **bash module gets included explicitly (as it happens always when systemd gets included), this module should have no impact in the generated initramfs.** 

The module is independent and additive - a distribution might decide not to package the module with core dracut modules.

One notable side effect of this PR is that now the CI has test coverage for the `busybox` dracut module (without explicitly mentioning `busybox` in any of the test case), since the on Alpine `/bin/sh` points to `busybox`. Getting test coverage without adding a lot of code is very useful. As an example `01 on alpine` test now boots using busybox in the generated initrd.

**One valuable side effect of this change is that CI no longer requires `dash` so a distribution can decide not to package the `dash` dracut module and still run the CI tests downstream.** 

## Checklist

Fixes #522

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -554,6 +554,10 @@ pcmcia:
   - changed-files:
     - any-glob-to-any-file: 'modules.d/90pcmcia/*'
 
+shell-interpreter:
+  - changed-files:
+    - any-glob-to-any-file: 'modules.d/00shell-interpreter/*'
+
 test:
   - changed-files:
     - any-glob-to-any-file: ['test/*', 'test/**/*', 'modules.d/80test*', 'modules.d/80test*/*']
--- a/modules.d/00shell-interpreter/module-setup.sh
+++ b/modules.d/00shell-interpreter/module-setup.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Module dependency requirements.
+depends() {
+    # priority order
+    shells='bash dash busybox'
+
+    for shell in $shells; do
+        if dracut_module_included "$shell"; then
+            echo "$shell"
+            return 0
+        fi
+    done
+
+    shell=$(realpath /bin/sh)
+    shell=${shell##*/}
+
+    echo "$shell"
+    return 0
+}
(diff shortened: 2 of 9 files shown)


modules.d/00sh/module-setup.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+# This file is part of dracut.
+# SPDX-License-Identifier: GPL-2.0-or-later
+
+# Module dependency requirements.
+depends() {
+    # priority order
+    shells='dash bash busybox'
```
Conan-Kudo, line 8: I would prefer `bash` to be the most preferred option, followed by `dash` and then `busybox`.

Conan-Kudo (changes requested):
I think this is reasonable, but I don't think we should call this `sh`. Instead, it should be called `shell-interpreter`, since this governs the interpreter used for dracut.

Conan-Kudo (approved):

aafeijoo-suse (commented):
This PR was merged very quickly... I don't want to hurt any feelings so, just 2 remarks about 3975e26a84cd8645b15d0a863263a63da93650e0:
- What happens if `busybox` is compiled without setting `CONFIG_FEATURE_INSTALLER`?
- But the worst part is... where do you think `busybox` is effectively running `--install`? I'll give you a hint, see what the `ln_r` code does:

https://github.com/dracut-ng/dracut-ng/blob/381be95f1511b7d311f99b7c6cdcbf37084601e1/dracut-init.sh#L1093

PR #541, opened 2024-07-24
fix(man): update description of the --gzip option

floppym:
dracut no longer defaults to gzip. There is a list of compressors and the first available is used. See 9663307c and 693b7a32.

Add a warning about kernel support for gzip decompression, similar to all of the other compressor options.

### DIFF
--- a/man/dracut.8.asc
+++ b/man/dracut.8.asc
@@ -436,9 +436,13 @@ example:
     Install the space separated list of files into the initramfs, if they exist.
 
 **--gzip**::
-    Compress the generated initramfs using gzip. This will be done by default,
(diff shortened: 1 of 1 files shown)


chewi:
The contributor kindly did this after I reported my boot failure. Thank you! Perhaps we should explicitly mention that it tries to compress with whatever is available though? As it stands, I would now assume that it doesn't compress by default.

LaszloGombos (approved):

Conan-Kudo (approved):

bdrung:
+1 for checking the kernel support for the supported compression methods.

Issue #549, opened 2024-07-26
dracut-install fails to resolve dependencies when `--sysrootdir` is used


kubiko:
**Describe the bug**
When using `--sysrootdir` and installing file which is symbolic link, dracut tries to resole the real path, but get destination directory mistaken for sysrootdir related directory. The problem is caused by this line: [dracut-install.c#L831](https://github.com/dracutdevs/dracut/blob/master/src/install/dracut-install.c#L831)
We cannot assume that the passed path starts with `sysrootdir`

**Distribution used**
Ubuntu 24.04

**Dracut version**
060+5-1ubuntu3.1, but also reproducible from `main` branch

**Init system**
Ubuntu 24.04

**To Reproduce**
```
- setup sysroot in /tmp/mysysroot
mkdir -p /tmp/test.dracut/
/usr/lib/dracut/dracut-install --debug -D /tmp/test.dracut/ --ldd --all -r /tmp/mysysroot /usr/bin/kmod
/usr/lib/dracut/dracut-install --debug -D /tmp/test.dracut/ --ldd --all -r /tmp/mysysroot /usr/sbin/depmod
```
this errors with
```
dracut-install: Handle '/usr/sbin/depmod'
dracut-install: dracut_install('/usr/sbin/depmod', '/usr/sbin/depmod', 0, 1, 1)
dracut-install: dest dir '/tmp/test.dracut/usr/sbin' does not exist
dracut-install: dracut_install('/usr/sbin', '/usr/sbin', 1, 0, 1)
dracut-install: mkdir '/tmp/test.dracut/usr/sbin'
dracut-install: get_real_file('/tmp/mysysroot/usr/sbin/depmod')
dracut-install: get_real_file: readlink(/tmp/mysysroot/usr/sbin/depmod') returns '../bin/kmod'
dracut-install: get_real_file('/usr/sbin/depmod') => /tmp/mysysroot/usr/sbin/../bin/kmod'
dracut-install: dracut_install('/usr/sbin/../bin/kmod', '/usr/sbin/../bin/kmod', 0, 1, 1)
dracut-install: '/tmp/test.dracut/usr/sbin/../bin/kmod' already exists, but checking for any deps
dracut-install: get_real_file('/tmp/mysysroot/usr')
dracut-install: resolve_deps('usr') -> get_real_file('usr', true) = /tmp/mysysroot/usr'
dracut-install: dracut_install ret = -21
[...]

PR #550, opened 2024-07-26
fix(dracut-install): do not assume handled path starts with sysrootdir

kubiko:
When using `--sysrootdir` argument, we cannot assume `fulldstpath` and `fullsrcpath` always start with `sysrootdir`. When `dracut_install` is called on the `destination` directory, this can result in a passing pointer which is often beyond the valid buffer.

This pull request checks the handled path and only if it starts with `sysrootdir `, the pointer is incremented by the `sysrootdirlen` to get a path relative to the `sysrootdir`

Arguably current behaviour is dangerous, since if the passed buffer is shorter than `sysrootdirlen`, the resulting pointer, which is passed further, is to the random memory location.
This can be seen in debug logs when `resolve_deps('` log is either empty or has random garble printed.

## Changes
Check the path for `sysrootdir` before incrementing the pointer.

## Checklist
- [ x] I have reviewed and updated any documentation if relevant
- [ x] I am providing new code and test(s) for it

Fixes #
https://github.com/dracut-ng/dracut-ng/issues/549

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -885,7 +885,10 @@ static int dracut_install(const char *orig_src, const char *orig_dst, bool isdir
         if (ret == 0) {
                 if (resolvedeps && S_ISREG(sb.st_mode) && (sb.st_mode & (S_IXUSR | S_IXGRP | S_IXOTH))) {
                         log_debug("'%s' already exists, but checking for any deps", fulldstpath);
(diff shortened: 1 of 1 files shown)


LaszloGombos:
CC @jtornosm

LaszloGombos:
@kubiko the failing tests are unrelated to your PR..

LaszloGombos (approved):

PR #186, opened 2024-04-16
fix(dracut): link hooks directory from /lib/dracut/hooks to /var/lib/dracut/hooks

LaszloGombos:
## Changes

Since https://github.com/systemd/systemd/commit/ffc1ec73, /usr is mounted as read-only in the initramfs by default.

Fixes https://github.com/dracutdevs/dracut/issues/2588

Minimal viable subset of https://github.com/dracut-ng/dracut-ng/pull/114 . No documentation change to maintain backward compatibility.

PR #552, opened 2024-07-27
feat(btrfs): clean old stuff

dracutng:
## Changes

cherry-pick from https://github.com/aafeijoo-suse/dracut-mini/commit/c88676772fd9e38aea078776fdf8e6a086e59d09

## Checklist

Fixes #


Conan-Kudo:
ping @josefbacik

FYI: @davide125 @michel-slm

aafeijoo-suse (changes requested):
This is ridiculous, this patch should be invalid for an upstream project that aims to support non-systemd systems.

I should be shocked that someone with better things to do is monitoring personal github repos (I'm thinking about renaming it to `dracut-mini-cooper-mk1-16v`, I loved that car) and proactively creating PRs cherry-picking commits from there (btw, this "clever bot" removed the dependency from `systemd-udevd` from the original patch), but sadly the fact is that after all I've seen here I'm not surprised at all.

LaszloGombos:
I agree with @aafeijoo-suse , this PR would remove the capability to scan (and remount ro) btrfs when systemd is either not available or omitted.  Let's close this PR .

aafeijoo-suse:
> I agree with @aafeijoo-suse , this PR would remove the capability to scan (and remount ro) btrfs when systemd is either not available or omitted. CI does not have a test for this that would show the regression. Let's close this PR .

Ah, so this review should mean that you and @dracutng are not the same person. So, just to clarify for the wider audience, who is the person behind this fake @dracut-ng bot? Because it is clearly not automatic to find a **personal** downstream repository that has not been directly forked from any other dracut repository and modify certain parts of its commits to create PRs and push them upstream using the owner's name (FTR, this is the 2nd time it happened: https://github.com/dracut-ng/dracut-ng/pull/186#issuecomment-2061184241).

PR #556, opened 2024-07-29
Add an explicit --basedir

lu-zero:
I'd like to get dracut to assemble an initramfs for `target-arch` from `host-arch` by passing `--sysroot` but then it tries to look for its own files within the sysroot.

This is an attempt to fix it.

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -160,6 +160,8 @@ Creates initial ramdisk images for preloading modules
   --tmpdir [DIR]        Temporary directory to be used instead of default
                          ${TMPDIR:-/var/tmp}.
   -r, --sysroot [DIR]   Specify sysroot directory to collect files from.
+  -b, --basedir [DIR]   Specify basedir directory to use for dracut, by
+                         default it is {sysrootdir}/usr/lib/dracut.
   -l, --local           Local mode. Use modules from the current working
                          directory instead of the system-wide installed in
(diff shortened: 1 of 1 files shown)


chewi:
I can't see how `--sysroot` has ever worked without this. It otherwise tries to execute the target's dracut-install, which is not a native binary.

Even with this, I get errors all over the place unless Dracut is installed in the sysroot.

chewi:
Given my earlier test, I'm not confident it really works yet, but I've made a lot of changes since. Let me give it another try now.

chewi:
Nope, it still blows up all over the place. Lots of this:

```
dracut-install: ERROR: installing '/usr/lib/dracut/modules.d/98dracut-systemd/dracut-emergency.sh' to '/usr/bin/dracut-emergency'
dracut[E]: FAILED: /usr/lib/dracut/dracut-install -r /build/amd64-usr -D /build/amd64-usr/var/tmp/portage/sys-kernel/coreos-kernel-6.12.20/temp/dracut.dN6oHlb/initramfs -l /usr/lib/dracut/modules.d/98dracut-systemd/dracut-emergency.sh /usr/bin/dracut-emergency
```

I think dracut-install doesn't know how to install something from outside the sysroot. That might be fairly easy to fix, but yeah, it's not ready yet.

devkontrol:
Closing as this PR is incomplete.

PR #561, opened 2024-08-03
fix(dracut-install): do not copy SELinux context

cgzones:
SELinux is normally not enabled in the initramfs and thus the file contexts are not relevant.

On the other hand using the source file context requires dracut-install to have the setattr and unlink permissions on these source contexts, e.g. lib_t or modules_object_t, which only very few domains should have modification rights to, e.g. the system administrator or the package manager.

This pull request changes...

## Checklist

Fixes #


cgzones:
The extended attributes weren't copied in the Btrfs case until recently, see https://github.com/dracut-ng/dracut-ng/commit/3e1d0bc1df6fcc5649e38d1016f05712f634b2f5.

pvalena (approved):
LGTM.

jozzsi:
> SELinux is normally not enabled in the initramfs and thus the file contexts are not relevant.

Would the [`selinux`](https://github.com/dracut-ng/dracut-ng/tree/main/modules.d/98selinux) dracut module still function properly after this PR ?

Conan-Kudo:
I'm concerned about not copying SELinux contexts. People use dracut for pregenerated and signed initramfs builds too, and having SELinux policies enforceable at initramfs time is useful since there are interactive initramfs modes.

PR #562, opened 2024-08-04
fix(dracut.sh): use only compressor that kernel supports

### LINKED ISSUE #378: dracut does not check for kernel compression support
**Describe the bug**

When using zstd compression, dracut checks that the kernel supports zstd compressed initrds (`CONFIG_RD_ZSTD`). But dracut does not do this check for all the other possible compression formats. So you might end up with a xz or gzip compressed initrd that cannot be read by the kernel. Please also check those kernel configs:

```
CONFIG_RD_GZIP
CONFIG_RD_BZIP2
CONFIG_RD_LZMA
CONFIG_RD_XZ
CONFIG_RD_LZO
CONFIG_RD_LZ4
```

**Dracut version**
102

AndrewAmmerlaan:
Follows up on:
- https://github.com/dracut-ng/dracut-ng/pull/520
- https://github.com/dracut-ng/dracut-ng/pull/541

And fixes:
- https://github.com/dracut-ng/dracut-ng/issues/378

Note #520 contains a fix for the `check_kernel_config` function and should be merged first.

Closes: https://github.com/dracut-ng/dracut-ng/issues/378

## Checklist

Fixes #378

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1185,6 +1185,43 @@ if [[ $_no_compress_l == "$DRACUT_COMPRESS_CAT" ]]; then
     compress="$DRACUT_COMPRESS_CAT"
 fi
 
+check_kernel_compress_support() {
+    local kern_compress_config
+    case $1 in
+        "$DRACUT_COMPRESS_LBZIP2" | "$DRACUT_COMPRESS_BZIP2" | lbzip2 | bzip2 | */lbzip2 | */bzip2)
+            kern_compress_config=CONFIG_RD_BZIP2
+            ;;
+        "$DRACUT_COMPRESS_LZMA" | lzma | */lzma)
+            kern_compress_config=CONFIG_RD_LZMA
+            ;;
+        "$DRACUT_COMPRESS_XZ" | xz | */xz)
+            kern_compress_config=CONFIG_RD_XZ
+            ;;
+        "$DRACUT_COMPRESS_PIGZ" | "$DRACUT_COMPRESS_GZIP" | pigz | gzip | */pigz | */gzip)
+            kern_compress_config=CONFIG_RD_GZIP
+            ;;
+        "$DRACUT_COMPRESS_LZOP" | lzop | */lzop)
+            kern_compress_config=CONFIG_RD_LZO
+            ;;
+        "$DRACUT_COMPRESS_ZSTD" | zstd | */zstd)
+            kern_compress_config=CONFIG_RD_ZSTD
+            ;;
+        "$DRACUT_COMPRESS_LZ4" | lz4 | */lz4)
+            kern_compress_config=CONFIG_RD_LZ4
+            ;;
+        "$DRACUT_COMPRESS_CAT" | cat | */cat)
+            return 0
+            ;;
+        *)
+            derror "unknown kernel config option for compressor $1"
+            return 0
+            ;;
+    esac
+
+    check_kernel_config $kern_compress_config
+    return $?
+}
+
 [[ $hostonly == yes ]] && hostonly="-h"
 [[ $hostonly != "-h" ]] && unset hostonly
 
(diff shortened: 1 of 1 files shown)


dracut.sh
```diff
@@ -2335,23 +2335,59 @@ else
     DRACUT_KERNEL_RD_ZSTD=
```
bdrung, line 2335: Setting `DRACUT_KERNEL_RD_ZSTD` can be dropped.

dracut.sh
```diff
@@ -2335,23 +2335,59 @@ else
...
 fi
 
-if [[ $compress == $DRACUT_COMPRESS_ZSTD* && ! $DRACUT_KERNEL_RD_ZSTD ]]; then
-    dwarn "Kernel has no zstd support compiled in."
-    compress=
-fi
+
+check_kernel_compress_support() {
+    local kern_compress_config
+    case $1 in
```
bdrung, line 2341: issue: The `DRACUT_COMPRESS_*` variables could be set by the user. In that case this case statement would fail.

suggestion: Tie the `kern_compress_config` with the `DRACUT_COMPRESS_*` variables (probably at the `handle compression options` section to ensure they stay in sync if compressors are added/removed).

bdrung (changes requested):
Thanks for working on it!

dracut.sh
AndrewAmmerlaan: I'm not sure I understand how that resolves the issue. If the user sets e.g. `DRACUT_COMPRESS_BZIP2=foo` then there still is no way to know for sure which kernel config option should be checked. Probably it is `CONFIG_RD_BZIP2`, but the fact that this variable was set by the user does not necessarily mean that `compress=bzip2` as well.  And regardless of what compression type we are going to use, all of these variables will be populated so checking which ones are set does not help here.

This is why I made the `*)` case not fail fatally. If `compress=` is set to some command that we do not know, we simply warn but still try and hope for the best.

bdrung: The contract with the user is that `DRACUT_COMPRESS_BZIP2` points to a bzip2 compressor. This could be an uncommon path like `DRACUT_COMPRESS_BZIP2=foo/bar/bzip2` or some compatible command. So `CONFIG_RD_BZIP2` should be checked in that case.

AndrewAmmerlaan: How about this?

Now it checks for the correct kernel config option for:
- `DRACUT_COMPRESS_GZIP=/usr/bin/gzip  ./dracut.sh --kver 6.10.3-gentoo-dist /tmp/test.img -f --compress='/usr/bin/gzip -9 -n' `
- `DRACUT_COMPRESS_GZIP=/usr/bin/gzip  ./dracut.sh --kver 6.10.3-gentoo-dist /tmp/test.img -f --gzip `
- `./dracut.sh --kver 6.10.3-gentoo-dist /tmp/test.img -f --compress='/usr/bin/gzip -9 -n' `
- `./dracut.sh --kver 6.10.3-gentoo-dist /tmp/test.img -f --gzip `

That should cover all cases right?

pvalena (approved):
LGTM, although I've not tested the code.

dracut.sh
bdrung: Yes, this should cover most (the normal) cases.

bdrung (approved):
I would have preferred an associative array to just store the mapping:
```bash
declare -A compressor2config
compressor2config["$DRACUT_COMPRESS_BZIP2"] = CONFIG_RD_BZIP2
compressor2config["$DRACUT_COMPRESS_LBZIP2"] = CONFIG_RD_BZIP2
compressor2config["$DRACUT_COMPRESS_LZMA"] = CONFIG_RD_LZMA
[...]
    for i in $DRACUT_COMPRESS_ZSTD [...]; do
        check_kernel_config "compressor2config[$i]"
        [...]
```

But the user can specify an arbitrary compress command.

Conan-Kudo (approved):

PR #559, opened 2024-07-30
fix(dracut-functions.sh): skip dir in get_persistent_dev()

pfliu:
I observed the following directory layout under /dev/disk/by-path:

/dev/disk/by-path/
├── pci-0000:00:07.0 -> ../../vda
├── pci-0000:00:07.0-part
│   ├── by-partnum
│   │   ├── 1 -> ../../../../vda1
│   │   ├── 2 -> ../../../../vda2
│   │   └── 3 -> ../../../../vda3
│   ├── by-partuuid
│   │   ├── 102645f0-b6a7-47d5-93ae-60bd063ecbc5 -> ../../../../vda1 │   │   ├── 32dc3e8a-70cb-4afb-8642-d4f2e0a92941 -> ../../../../vda3 │   │   └── 46beb739-55c1-4746-ab90-b329a0c82ea5 -> ../../../../vda2 │   └── by-uuid
│       └── 11ea5cf3-7132-40d4-a114-a8d76edca51f -> ../../../../vda2
├── pci-0000:00:07.0-part1 -> ../../vda1
├── pci-0000:00:07.0-part2 -> ../../vda2
├── pci-0000:00:07.0-part3 -> ../../vda3
├── virtio-pci-0000:00:07.0 -> ../../vda
├── virtio-pci-0000:00:07.0-part1 -> ../../vda1
├── virtio-pci-0000:00:07.0-part2 -> ../../vda2
└── virtio-pci-0000:00:07.0-part3 -> ../../vda3

As showed above, 'pci-0000:00:07.0-part' is a directory, so get_maj_min() returns '0:0' for it.

This will raise an issue in the case that kdump dumps the vmcore through nfs. In that case, no block device is passed to
get_persistent_dev(), so _dev=$(get_maj_min "$1") also equals '0:0'. Finally, it wrongly fits into the match:
        if [ "$_tmp" = "$_dev" ]; then
            printf -- "%s" "$i"
            return
        fi

And kdump kernel boots up and wait for 'pci-0000:00:07.0-part' indefinitely.

This pull request changes...

## Checklist
- [ ] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

Fixes #

PR #565, opened 2024-08-05
fix(dracut-functions.sh): Only return block devices from get_persistent_dev

Vogtinator:
With udev 256, there are now directories such as
/dev/disk/by-path/pci-0000:02:00.0-nvme-1-part/ which match here.

In case a nonexisting file/device was passed to get_persistent_dev, it returned the first directory it looked at because both have maj:min 0:0. This accidental conversion from garbage to a sensible looking path leads to weird behaviour later.

Instead of filtering out directories explicitly switch the check to only return block devices, which also takes care of the character special /dev/mapper/control.

## Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -294,8 +294,7 @@ get_persistent_dev() {
         /dev/disk/by-partlabel/* \
         /dev/disk/by-id/* \
         /dev/disk/by-path/*; do
-        [[ -e $i ]] || continue
-        [[ $i == /dev/mapper/control ]] && continue
+        [[ -b $i ]] || continue
         [[ $i == /dev/mapper/mpath* ]] && continue
         _tmp=$(get_maj_min "$i")
         if [ "$_tmp" = "$_dev" ]; then


aafeijoo-suse (commented):
Better than #559, but it should not fail with the current code in main.
See:
- https://github.com/dracutdevs/dracut/issues/2592
- #36
- #133
- #155

BTW, @Vogtinator you shouldn't have run into this issue using the latest version of dracut shipped with openSUSE, have you?

Vogtinator:
> BTW, @Vogtinator you shouldn't have run into this issue using the latest version of dracut shipped with openSUSE, have you?

@DimStar77 did, he had a stale `add_device+=" UUID=... "` which was invalid syntax there which only broke with udev 256.

LaszloGombos (approved):

pvalena:
LGTM.

I've also checked that `-b` is able to resolve symlinks :).

pvalena (approved):

Issue #110, opened 2024-04-05
Do not re-compress compressed kernel modules and firmware files


bdrung:
Ubuntu ships the kernel modules and firmware files zstd compressed. Dracut puts those compressed files in the initramfs and compresses the cpio archive containing them.

Please put compressed kernel modules and firmware files in an uncompressed cpio instead and place that cpio before the compressed cpio one.

See the similar change in initramfs-tools in Ubuntu: https://bugs.launchpad.net/ubuntu/+source/initramfs-tools/+bug/2028567

bdrung:
initramfs-tools behaved similar to https://github.com/OpenMandrivaAssociation/dracut/blob/master/dracut-044-dont-compress-kernel-modules-within-initramfs.patch before we implemented https://bugs.launchpad.net/ubuntu/+source/initramfs-tools/+bug/2028567

bdrung:
I just need to port the code change from https://salsa.debian.org/kernel-team/initramfs-tools/-/merge_requests/128 to dracut. Expect me to do that in the near future.

PR #568, opened 2024-08-06
feat(dracut-install.c): try decompressing kernel modules and firmware

### LINKED ISSUE #560: Should dracut uncompress compressed kernel modules before installing into the initramfs?
A kernel that is configured to compress modules (for example `CONFIG_MODULE_COMPRESS_XZ=y`) still supports loading uncompressed modules. Given that it is size-wise more efficient to compress a bunch of uncompressed files, compared to compressing a bunch of files that are individually already compressed. It might make sense for Dracut to decompress modules it finds before including them in the initramfs.

The same holds for firmware files (`CONFIG_FW_LOADER_COMPRESS_XZ=y`).

To put this in numbers, an UKI for the same kernel with the same user space utilities, but with pre-xz-compressed kernel modules and firmware is 356.352 kb larger on my system.

Nowa-Ammerlaan:
Compressing uncompressed files is more efficient then compressing pre-compressed files, as also indicated in the help text for the `CONFIG_MODULE_COMPRESS_*` kernel configuration options.

I am far from an expert when it comes to C, so there may be a more elegant way to do this. But nonetheless, this works on my system and as a rough benchmark my UKI with hostonly initramfs is 539.648 kilobytes (1.3%) smaller after this change. This result obviously scales with the amount of included modules and firmware so for a non-hostonly UKI I would expect a bigger difference.

Something that is still puzzling me is that the UKI is now actually slower to boot, I would have expected that it would be faster since after the initramfs is decompressed, no additional decompression is required to load the kernel modules or firmware:

Before this change:
```
Startup finished in 21.598s (firmware) + 1.836s (loader) + 1.467s (kernel) + 3.529s (initrd) + 8.326s (userspace) = 36.758s
graphical.target reached after 8.266s in userspace.
```
And after this change:
```
Startup finished in 21.742s (firmware) + 1.810s (loader) + 1.770s (kernel) + 8.630s (initrd) + 8.308s (userspace) = 42.262s
graphical.target reached after 8.264s in userspace.
```

That's 5 seconds! Maybe someone knows why this result is opposite to what I expected, and if it can be remedied? At boot it did not really show any obvious points where it was hanging, nor are the any failures in the logs.

Closes: https://github.com/dracut-ng/dracut-ng/issues/560

## Checklist

Fixes #560

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -97,7 +97,8 @@ static bool arg_mod_filter_symbol = false;
 static bool arg_mod_filter_nosymbol = false;
 static bool arg_mod_filter_noname = false;
 
-static int dracut_install(const char *src, const char *dst, bool isdir, bool resolvedeps, bool hashdst);
+static int dracut_install(const char *src, const char *dst, bool isdir, bool resolvedeps, bool hashdst,
+                          bool decompr);
 static int install_dependent_modules(struct kmod_ctx *ctx, struct kmod_list *modlist, Hashmap *suppliers_paths);
 
 static void item_free(char *i)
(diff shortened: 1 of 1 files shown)


AndrewAmmerlaan:
The complaint from CodeQL is not something new, it is already applicable to the code in `main`.

AndrewAmmerlaan:
> Something that is still puzzling me is that the UKI is now actually slower to boot, I would have expected that it would be faster since after the initramfs is decompressed, no additional decompression is required to load the kernel modules or firmware:

Hmm, actually, if I install the modules on root in the uncompressed form as well, and use the code as it is in `main`, then I get the same result from `systemd-analyze`. So the extra 5 seconds I observe is not actually caused by this change.

I suppose that it takes longer for the kernel to decompress the initramfs precisely because it was more efficiently compressed.

AndrewAmmerlaan:
Here's the same comparison using zstd instead of xz for kernel, module, firmware and initramfs compression:

Before this change:
```
Startup finished in 21.335s (firmware) + 1.806s (loader) + 1.052s (kernel) + 3.706s (initrd) + 8.261s (userspace) = 36.162s
graphical.target reached after 8.234s in userspace.
```

And after this change:
```
Startup finished in 21.785s (firmware) + 1.774s (loader) + 993ms (kernel) + 7.981s (initrd) + 8.230s (userspace) = 40.765s
graphical.target reached after 8.198s in userspace.
```

Again, slower, but the difference is slightly smaller then before.

The resulting hostonly UKI is 1692.672 kilobytes (4.0%) smaller, that is a more then three times bigger difference compared to xz.

src/install/dracut-install.c
```diff
@@ -426,14 +427,82 @@ static int cp(const char *src, const char *dst)
...
+                if (endswith(src, ".ko")) {
+                        log_debug("not compressed: %s", src);
+                } else {
+                        log_warning("unknown compression: %s", src);
+                }
+                // Not compressed or unknown compressor, fallback to copying
+                goto fallback;
+        }
+
+        int pid;
```
marcosfrm, line 455: Not a big matter, but this should be `pid_t` (needs change in `cp()` too).

src/install/dracut-install.c
```diff
@@ -426,14 +427,82 @@ static int cp(const char *src, const char *dst)
...
+static int decompress_cp(const char *src, const char *dst)
+{
+        char decompressor[5];
+        char *dest_base = strdup(dst);
+        if (endswith(src, ".xz")) {
+                strcpy(decompressor, "xz");
+                dest_base[strlen(dest_base) - 3] = 0;
+
+        } else if (endswith(src, ".gz")) {
+                strcpy(decompressor, "gz");
```
marcosfrm, line 439: The tool is `gzip`.

src/install/dracut-install.c
```diff
@@ -426,14 +427,82 @@ static int cp(const char *src, const char *dst)
...
+        int dest_desc =
+                open(dest_base, O_WRONLY | O_CREAT | O_CLOEXEC | O_TRUNC,
+                     (sb.st_mode) & (S_ISUID | S_ISGID | S_ISVTX | S_IRWXU | S_IRWXG | S_IRWXO));
+        if (dest_desc < 0)
+                goto fallback;
+
+        pid = fork();
+        if (pid == 0) {
+                dup2(dest_desc, 1);
+                execlp(decompressor, decompressor, "--keep", "--decompress", src, "--stdout", "--force", NULL);
```
marcosfrm, line 471: Not a big matter, but `gzip` an `xz` man pages put the source file in the end of the command line.

src/install/dracut-install.c
```diff
@@ -426,14 +427,82 @@ static int cp(const char *src, const char *dst)
...
+                        goto fallback;
+                }
+        }
+        ret = WIFSIGNALED(ret) ? 128 + WTERMSIG(ret) : WEXITSTATUS(ret);
+        log_debug("decompress ret = %d", ret);
+        if (ret != 0) {
+                log_warning("'%s --keep --decompress %s --stdout --force' failed with %d", decompressor, src, ret);
+                // Something went wrong, fall back to copying, remove the empty
+                // file we touched in open().
+                remove(dest_base);
```
marcosfrm, line 488: It is always a file. Just call `unlink()`.

src/install/dracut-install.c
```diff
@@ -426,14 +427,82 @@ static int cp(const char *src, const char *dst)
...
+
+        int dest_desc =
+                open(dest_base, O_WRONLY | O_CREAT | O_CLOEXEC | O_TRUNC,
+                     (sb.st_mode) & (S_ISUID | S_ISGID | S_ISVTX | S_IRWXU | S_IRWXG | S_IRWXO));
+        if (dest_desc < 0)
+                goto fallback;
+
+        pid = fork();
+        if (pid == 0) {
+                dup2(dest_desc, 1);
```
marcosfrm, line 470: Close `dest_desc` on the child process after duplicating it.

bdrung:
The better approach will be https://github.com/dracut-ng/dracut-ng/issues/110 which will reduce the boot time.

marcosfrm:
Forget it, it's not worth it, because there's a serious problem: all firmware symbolic links will be broken when decompressed and will need to be recreated. This is too much complexity for little gain. The idea of putting the compressed files in a separate cpio is better.

PR #570, opened 2024-08-08
fix(systemd-initrd): Add base as dependency

PrincessDoreena:
For systemd to run properly inside initramfs, /etc/initrd-release is required to exist. [1] 
/etc/initrd-release is only written in the base module.

[1]: https://systemd.io/INITRD_INTERFACE/#using-systemd-inside-an-initrd

This pull request changes...

## Changes
Dependencies of the `systemd-initrd` module

## Checklist
- [ x ] I have tested it locally
- [ x ] I have reviewed and updated any documentation if relevant
- [ x ] I am providing new code and test(s) for it


jozzsi:
I think we should take a closer look of which of the many`*systemd*` modules should add the base module dependency. Not only `systemd` module but also the `systemd-initrd` and `dracut-systemd` modules are good candidates for adding the base dependency.

This PR break 6 tests as it seems the meaning of `systemd` dracut module is also used to build not only initrd's but also rootfs (without `/etc/initrd-release` file).

Based on this observations, I think it is better to add the `base` dependency to the `systemd-initrd` - which is a module specific to build initrds.

Hopefully moving this change to `systemd-initrd` from `systemd` will fix the CI failures as well.

modules.d/00systemd/module-setup.sh
```diff
@@ -13,7 +13,7 @@ check() {
 
 # called by dracut
 depends() {
-    return 0
+    echo base
```
LaszloGombos, line 16: @PrincessDoreena the suggestion from @jozzsi makes sense for me. This change should be in the `systemd-initrd` dracut module instead. Would you be able to make this change in this PR or should someone else take it over from here ?

LaszloGombos (changes requested):
This change should be in the `systemd-initrd` module.

PrincessDoreena:
I will try fixing it today.

Laszlo Gombos ***@***.***> ezt írta (időpont: 2024. aug. 18.,
V 17:41):

PrincessDoreena:
Verbose log for the test failure on arch linux:
[test.log](https://github.com/user-attachments/files/16670217/test.log)

LaszloGombos (approved):
thanks. It has been a while since we've gained a new test. :-)

PR #573, opened 2024-08-09
fix(nfs): include also entries from /usr/lib/{passwd,group}

pvalena:
as those paths are used by bootc instead of the /etc ones.

## Changes

Handle  /usr/lib/{passwd,group} configs additionally to /etc/ variants in nfs module.

## Checklist

### DIFF
--- a/modules.d/95nfs/module-setup.sh
+++ b/modules.d/95nfs/module-setup.sh
@@ -125,8 +125,16 @@ install() {
 
     # Rather than copy the passwd file in, just set a user for rpcbind
     # We'll save the state and restart the daemon from the root anyway
-    grep -E '^(nfsnobody|_rpc|rpc|rpcuser):' "$dracutsysrootdir"/etc/passwd >> "$initdir/etc/passwd"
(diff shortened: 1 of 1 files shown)


lnykryn (approved):
LGTM

modules.d/95nfs/module-setup.sh
```diff
@@ -125,8 +125,21 @@ install() {
...
-    grep -E '^nogroup:|^rpc:|^nobody:' "$dracutsysrootdir"/etc/group >> "$initdir/etc/group"
+
+    local _confdir
+    local _filepath
+    for _confdir in etc usr/lib; do
+
+      _filepath="${dracutsysrootdir}/${_confdir}/passwd"
+
+      [[ -r "${_filepath}" ]] \
+        && grep -E '^(nfsnobody|_rpc|rpc|rpcuser):' "${_filepath}" >> "$initdir/${_confdir}/passwd"
```
pvalena, line 136: simplification:

```suggestion
      grep -sE '^(nfsnobody|_rpc|rpc|rpcuser):' "${_filepath}" >> "$initdir/${_confdir}/passwd"
```

PR #574, opened 2024-08-09
feat(fips): add support for UKIs

vittyvk:
Kernel integrity check in FIPS module is incompatible with UKIs as neither /boot/vmlinuz-`uname-r` nor /boot/.vmlinuz-`uname-r`.hmac are present. UKI is placed to $ESP\EFI\Linux\<install-tag>-<uname-r>.efi and if a .hmac file is present next to it, it is possible to do similar check.

Note, UKIs have a 'one size fits all' command line and 'boot=' is not expected to be set. Luckily, if the UKI is systemd-stub based then we can expect 'LoaderDevicePartUUID' variable containing PARTUUID of the ESP to be set. Mount it to /boot using the existing logic.

## Changes
Support for UKI integrity checking added to the FIPS module.

## Checklist
- [ X ] I am providing new code. Testing this outside of a real UKI based environment is problematic.

References:
Fedora/ARK kernel MR (creating .hmac for UKI): https://gitlab.com/cki-project/kernel-ark/-/merge_requests/3314
virt-firmware MR (installing .hmac to the ESP): https://gitlab.com/kraxel/virt-firmware/-/merge_requests/17

### DIFF
--- a/modules.d/01fips/fips.sh
+++ b/modules.d/01fips/fips.sh
@@ -14,8 +14,21 @@ else
     }
 fi
 
+# Checks if a systemd-based UKI is running and ESP UUID is set
+is_uki() {
+    [ -f /sys/firmware/efi/efivars/StubFeatures-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f ] \
+        && [ -f /sys/firmware/efi/efivars/LoaderDevicePartUUID-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f ]
+}
+
 mount_boot() {
     boot=$(getarg boot=)
+
+    if is_uki && [ -z "$boot" ]; then
+        # efivar file has 4 bytes header and contain UCS-2 data. Note, 'cat' is required
+        # as sys/firmware/efi/efivars/ files are 'special' and don't allow 'seeking'.
+        # shellcheck disable=SC2002
+        boot="PARTUUID=$(cat /sys/firmware/efi/efivars/LoaderDevicePartUUID-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f | tail -c +5 | tr -d '\0' | tr 'A-F' 'a-f')"
+    fi
 
     if [ -n "$boot" ]; then
         if [ -d /boot ] && ismounted /boot; then
@@ -81,6 +94,41 @@ do_rhevh_check() {
     return 0
 }
 
+do_uki_check() {
+    local KVER
+    local uki_checked=0
+
+    KVER="$(uname -r)"
+    # UKI are placed in $ESP\EFI\Linux\<intall-tag>-<uname-r>.efi
+    if ! [ "$FIPS_MOUNTED_BOOT" = 1 ]; then
+        warn "Failed to mount ESP for doing UKI integrity check"
+        return 1
+    fi
+
+    for UKIpath in /boot/EFI/Linux/*-"$KVER".efi; do
+        # UKIs are installed to $ESP/EFI/Linux/<entry-token-or-machine-id>-<uname-r>.efi
+        # and in some cases (e.g. when the image is used as a template for creating new
+        # VMs) entry-token-or-machine-id can change. To make sure the running UKI is
+        # always checked, check all UKIs which match the 'uname -r' of the running kernel
+        # and fail the whole check if any of the matching UKIs are corrupted.
+
+        [ -r "$UKIpath" ] || break
+
+        local UKI="${UKIpath##*/}"
+        local UKIHMAC=."$UKI".hmac
+
+        fips_info "checking $UKIHMAC"
+        (cd /boot/EFI/Linux/ && sha512hmac -c "$UKIHMAC") || return 1
+        uki_checked=1
+    done
+
+    if [ "$uki_checked" = 0 ]; then
+        warn "Failed for find UKI for checking"
+        return 1
+    fi
+    return 0
+}
+
 nonfatal_modprobe() {
     modprobe "$1" 2>&1 > /dev/stdout \
         | while read -r line || [ -n "$line" ]; do
@@ -133,6 +181,9 @@ do_fips() {
         elif [ -e "/run/install/repo/images/pxeboot/vmlinuz" ]; then
             # This is a boot.iso with the .hmac inside the install.img
             do_rhevh_check /run/install/repo/images/pxeboot/vmlinuz || return 1
+        elif is_uki; then
+            # This is a UKI
+            do_uki_check || return 1
         else
             BOOT_IMAGE="$(getarg BOOT_IMAGE)"
 
--- a/modules.d/01fips/module-setup.sh
+++ b/modules.d/01fips/module-setup.sh
@@ -59,7 +59,7 @@ install() {
     inst_hook pre-udev 01 "$moddir/fips-load-crypto.sh"
     inst_script "$moddir/fips.sh" /sbin/fips.sh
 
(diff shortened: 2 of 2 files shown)


modules.d/01fips/fips.sh
```diff
@@ -81,6 +92,35 @@
...
+do_uki_check() {
+    KVER=$(uname -r)
+    # UKI are placed in $ESP\EFI\Linux\<intall-tag>-<uname-r>.efi
+    if ! [ "$FIPS_MOUNTED_BOOT" = 1 ]; then
+        warn "Failed to mount ESP for doing UKI integrity check"
+        return 1
+    fi
+
+    uki_checked=0
+    pushd /boot/EFI/Linux/
```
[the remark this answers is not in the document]
vittyvk: Will take a look and probably throw away pushd/popd shortly.

modules.d/01fips/fips.sh
```diff
@@ -14,9 +14,20 @@
...
+    [ -f /sys/firmware/efi/efivars/StubFeatures-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f ] \
+        && [ -f /sys/firmware/efi/efivars/LoaderDevicePartUUID-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f ]
+}
+
 mount_boot() {
     boot=$(getarg boot=)
 
+    if is_uki && [ -z "$boot" ]; then
+        # efivar file has 4 bytes header and contain UCS-2 data
+        boot="PARTUUID=$(cat /sys/firmware/efi/efivars/LoaderDevicePartUUID-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f | tail -c +5 | tr -d '\0' | tr [A-Z] [a-z])"
```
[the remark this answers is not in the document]
vittyvk: Not useless, actually. /sys/firmware/efi/efivars/ files are 'special' so neither:
```
tail -c +5 < /sys/firmware/efi/efivars/LoaderDevicePartUUID-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f | tr -d '\0' | tr '[A-Z]' '[a-z]'
```
nor
```
tail -c +5 /sys/firmware/efi/efivars/LoaderDevicePartUUID-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f | tr -d '\0' | tr '[A-Z]' '[a-z]'
```
work ("cannot seek to relative offset 4: Illegal seek").

modules.d/01fips/fips.sh
```diff
@@ -14,9 +14,22 @@
...
+}
+
 mount_boot() {
     boot=$(getarg boot=)
 
+    if is_uki && [ -z "$boot" ]; then
+        # efivar file has 4 bytes header and contain UCS-2 data. Note, 'cat' is required
+        # as sys/firmware/efi/efivars/ files are 'special' and don't allow 'seeking'.
+        # shellcheck disable=SC2002
+        boot="PARTUUID=$(cat /sys/firmware/efi/efivars/LoaderDevicePartUUID-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f | tail -c +5 | tr -d '\0' | tr 'A-Z' 'a-z')"
```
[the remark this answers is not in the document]
vittyvk: These are hex literals, actually. A-F -> a-f should do fine.

modules.d/01fips/module-setup.sh
```diff
@@ -64,7 +64,7 @@ install() {
     inst_hook pre-udev 01 "$moddir/fips-load-crypto.sh"
     inst_script "$moddir/fips.sh" /sbin/fips.sh
 
-    inst_multiple sha512hmac rmmod insmod mount uname umount grep sed cut find sort
+    inst_multiple sha512hmac rmmod insmod mount uname umount grep sed cut find sort tail tr
```
neverpanic, line 67: Shouldn't you also add `cat` here, just to make sure it actually exists?

vittyvk: Nice catch, will add!

vittyvk:
@pvalena @aafeijoo-suse could you please take a look? Thanks!

modules.d/01fips/fips.sh
```diff
@@ -81,6 +94,33 @@ do_rhevh_check() {
...
+    KVER=$(uname -r)
+    # UKI are placed in $ESP\EFI\Linux\<intall-tag>-<uname-r>.efi
+    if ! [ "$FIPS_MOUNTED_BOOT" = 1 ]; then
+        warn "Failed to mount ESP for doing UKI integrity check"
+        return 1
+    fi
+
+    uki_checked=0
+    for UKIpath in /boot/EFI/Linux/*-"$KVER".efi; do
+        [ -f "$UKIpath" ] || break
```
pvalena, line 107: What is this check for?

Maybe let's use `-r` to check it's readable (i.e. we have permissions to read it)?

modules.d/01fips/fips.sh
```diff
@@ -81,6 +94,33 @@ do_rhevh_check() {
...
+
+        UKI=${UKIpath##*/}
+        UKIHMAC=."$UKI".hmac
+
+        fips_info "checking $UKIHMAC"
+        (cd /boot/EFI/Linux/ && sha512hmac -c "$UKIHMAC") || return 1
```
pvalena, line 113: Please note that this will end on first unsuccesful check. Is that intentional? Are multiple UKIpaths (un)expected?

If we just need to check one, we could simply `return 0` here instead, and `return 1` bellow and drop the `uki_checked` variable.

modules.d/01fips/fips.sh
```diff
@@ -81,6 +94,33 @@ do_rhevh_check() {
     return 0
 }
 
+do_uki_check() {
+    KVER=$(uname -r)
```
pvalena, line 98: Shouldn't all local variables be declared with `local` to ensure there're no var collision?

pvalena (changes requested):
Overall it looks good, but it might need a small cleanup.

modules.d/01fips/fips.sh
pvalena: * this applies to all functions in this change

modules.d/01fips/fips.sh
vittyvk: This is intentional. UKIs are installed to $ESP/EFI/Linux/<machine-id>-<uname-r>.efi but e.g. when machine id changes (and the image is moved to a different host) + kernel is reinstalled (sounds already very unlikely, but still) we may get two UKIs of the same version. As the kernel is already booted, we can't say for sure which one was it, especially when a bootloader with menu (e.g. sb-boot) was used. So for FIPS integrity check, I decided that we can be on the safe side and check **all** UKIs of the same version.

modules.d/01fips/fips.sh
vittyvk: This is mostly for the case, when there are no UKIs matching '/boot/EFI/Linux/*-"$KVER".efi' and we actually get a string with '*' :-) Switching to '-r' is nevertheless a good idea, will do!

modules.d/01fips/fips.sh
pvalena: Just to clear thing up (probably deserves a comment), you want all of them pass? (That's what it currently does.)

vittyvk: Yes, this is the intention: to be on the safe side, let's make them ALL pass the check. I've added a comment explaining the decision.

pvalena (approved):
LGTM, thanks!

PR #576, opened 2024-08-13
feat(fips-crypto-policies): make c-p follow FIPS mode automatically

neverpanic:
(Supersedes #575, @t184256 opened that in my absence to speed up the review process; I've made the change requested in https://github.com/dracut-ng/dracut-ng/pull/575#pullrequestreview-2231750738)

For a system that uses crypto-policies to be switched to FIPS mode correctly, it needs to be

- booted with `fips=1` on the kernel command line
- switched to the FIPS crypto-policy (or a policy derived from it)
- have the fips dracut module enabled

On older systems, there were additional steps, for example, creating `/etc/system-fips`.

We have repeatedly seen inconsistencies between those different toggles, either because the user space tooling to switch between those does not (for reliability, maintainability, and compliance reasons) undo some of the steps it does when disabling FIPS mode, or because other installation methods (bootc, containers, image builder) independently do some of those steps. Eventually, all of these ended with user confusion.

We can avoid this situation by eliminating the difference by treating the `fips=1` kernel command line switch as a single source of truth, and making all others follow automatically. This module provides this for crypto-policies, by adding bind-mounts before pivot if the system has not already been switched to a FIPS-based crypto-policy.

This requires some support from the crypto-policies package (because it needs to deal with the bind mounts when a user calls `update-crypto-policies --set`), so make it a no-op unless

 - `fips=1` is on the kernel command line
 - crypto-policies is installed
 - crypto-policies supports the bind-mounts (indicated by the presence of the `default-fips-config` file)
 - the policy isn't already FIPS

These checks should make this safe to add to the initramfs on all current systems.

The bind-mounts also need to happen in the initramfs already, because systemd links against OpenSSL, and doing them later means that systemd will start with an OpenSSL configuration that isn't tailored for FIPS.

See also [1], which adds the user space support to crypto-policies, along with a systemd service that does the same steps in case dracut hasn't already done them (which is useful for environments that don't use an initramfs like containers).

  [1]: https://gitlab.com/redhat-crypto/fedora-crypto-policies/-/merge_requests/191

This pull request changes...

## Checklist
- [x] I have tested it locally (using clean Fedora rawhide VMs with this PR on top of dracut package)
- [x] I am providing new code…
- [ ] …and test(s) for it

Can you point me to documentation on how to write a dracut test? I'd be happy to provide tests.

### DIFF
--- a/modules.d/01fips-crypto-policies/fips-crypto-policies.sh
+++ b/modules.d/01fips-crypto-policies/fips-crypto-policies.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/sh
+
+type getarg > /dev/null 2>&1 || . /lib/dracut-lib.sh
+
+if ! fipsmode=$(getarg fips) || [ "$fipsmode" = "0" ] || [ -z "$fipsmode" ]; then
+    # Do nothing if not in FIPS mode
+    return 0
+fi
+
+policyfile=/etc/crypto-policies/config
+fipspolicyfile=/usr/share/crypto-policies/default-fips-config
+backends=/etc/crypto-policies/back-ends
+fipsbackends=/usr/share/crypto-policies/back-ends/FIPS
+
+# When in FIPS mode, check the active crypto policy by reading the
+# $root/etc/crypto-policies/config file. If it is not "FIPS", or does not start
+# with "FIPS:", automatically switch to the FIPS policy by creating
+# bind-mounts.
+
+if ! [ -r "${NEWROOT}${policyfile}" ]; then
+    # No crypto-policies configured, possibly not a system that uses
+    # crypto-policies?
+    return 0
+fi
+
+if ! [ -f "${NEWROOT}${fipspolicyfile}" ]; then
+    # crypto-policies is too old to deal with automatic bind-mounting of the
+    # FIPS policy over the normal policy, do not attempt to do the bind-mount.
+    return 0
+fi
+
+policy=$(cat "${NEWROOT}${policyfile}")
+
+# Remove the largest suffix pattern matching ":*" from the string (i.e., the
+# complete list of active policy modules), then check for FIPS. This is part of
+# POSIX sh (https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02).
+if [ "${policy%%:*}" = "FIPS" ]; then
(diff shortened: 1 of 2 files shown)


neverpanic:
The failing integration test 16 seems unrelated.


```
modprobe: FATAL: Module ib700wdt not found in directory /lib/modules/6.10.3-200.fc40.x86_64
```
suggests the driver for the watchdog device could not be loaded. As a consequence, it seems dracut fails to kick the watchdog:
```
[    1.290745] dracut: Triggering watchdog
[    1.291087] watchdog: watchdog0: watchdog did not stop!
```
This happens multiple times, and I guess eventually causes the system to shut down?

LaszloGombos:
> The failing integration test 16 seems unrelated.

It looks to me that the PR contributes to those failures as these failures do not happen to other PRs. 

You suspect that this can be confirmed with the following change 
```
check() {
    # only enable on systems that use crypto-policies
-    [ -d "$dracutsysrootdir/etc/crypto-policies" ] && return 0
+   [ -d "$dracutsysrootdir/etc/crypto-policies" ] && return 255

    # include when something else depends on it or it is explicitly requested
    return 255
}
```

This condition is probably only true on Fedora and openSUSE containers, which is why that is the only two containers failing.

neverpanic:
> I suspect that this can be confirmed with the following change
> 
> ```
> check() {
>     # only enable on systems that use crypto-policies
> -   [ -d "$dracutsysrootdir/etc/crypto-policies" ] && return 0
> +   [ -d "$dracutsysrootdir/etc/crypto-policies" ] && return 255
> 
>     # include when something else depends on it or it is explicitly requested
>     return 255
> }
> ```

Indeed, the test passes with this change.

I have no idea why, but I'll do some debugging.

neverpanic:
Latest force-push should fix it; I used `exit 0` in the module, but apparently it needs to be `return 0`, otherwise the shell is stopped completely.

neverpanic:
I am not able to reproduce the integration test failures for 30 on fedora (where this PR possibly makes a difference) or 30 on gentoo (where it shouldn't make a difference) locally.

I'll wait for the automatic re-trigger to see if it was a temporary issue.

pvalena (approved):
LGTM.

Conan-Kudo (approved):

PR #577, opened 2024-08-14
fix(test): always install kernel modules

jozzsi:
Always install Linux kernel modules regardless of the value of hostonly.

## Checklist


LaszloGombos (approved):
thanks

PR #582, opened 2024-08-16
feat(lsinitrd.sh): look for initrd in /usr/lib/modules/

pvalena:
Introduce new path for lsinitrd.sh to look into:

/usr/lib/modules/$kver/initramfs.img

Which is valid on all ostree-based systems, and also other image based
systems with pre-generated initramfs.

Ref: https://issues.redhat.com/browse/RHEL-35890


cgwalters:
I can confirm this works on a booted/deployed ostree/bootc host system, but it still fails when run in a container image (in the general case, where the host kernel is different from the one embedded in the container).

This is actually a big general papercut with dracut right now in that it always assumes the kernel it's running on is the kernel to target, and one we work around as you can see in https://docs.fedoraproject.org/en-US/bootc/initramfs/#_regenerating_the_initrd

It would be quite helpful for dracut to detect it's in a container (e.g. `systemd-detect-virt -c`) to start, and check for a single `/usr/lib/modules/$kver` directory and automatically use *that* kernel (and I guess this actually then kind of snowballs a bit into learning to *write* to `/usr/lib/modules/$kver/initramfs.img` by default, if it exists).

cgwalters:
BTW, testing the in-container case couldn't be easier (IMO), just e.g. `podman run --pull=newer --rm -ti quay.io/centos-bootc/centos-bootc:stream9 bash` and in there do e.g.

```
# curl -k -L --remote-name-all https://kojihub.stream.rdu2.redhat.com/kojifiles/work/tasks/5597/4535597/dracut-{squash-,network-,}057-68.git20240816.el9.x86_64.rpm
# rpm -Uvh *.rpm
```

(The separate curl is because the kojihub uses the RHT cert chain and unless you project that into your container you won't get it...which can also be fixed by bind mounting in `/etc/pki/ca-trust`)

Testing the *deployed* host...if you have a few minutes feel free to give https://docs.fedoraproject.org/en-US/bootc/podman-bootc-cli/ a try...I think it makes spinning up a bootc host from a container pretty easy. In there, it's just `bootc usroverlay` to make a transient overlay, then the exact same instructions as above to test changes.

LaszloGombos:
@pvalena @cgwalters  - what do you think about sharing the functionality (e.g. `get_default_initramfs_image`) by moving it to a sharable location - similar to how it is done in this commit - https://github.com/aafeijoo-suse/dracutla/commit/11fd40025bf6dbb2933185f256c598e250ff6d3d  ?

pvalena:
> @pvalena @cgwalters - what do you think about sharing the functionality (e.g. `get_default_initramfs_image`) by moving it to a sharable location - similar to how it is done in this commit - [aafeijoo-suse/dracutla@11fd400](https://github.com/aafeijoo-suse/dracutla/commit/11fd40025bf6dbb2933185f256c598e250ff6d3d) ?

Sure, although I don't think refactoring the code should be part of this PR.

pvalena:
> BTW, testing the in-container case couldn't be easier (IMO), just e.g. `podman run --pull=newer --rm -ti quay.io/centos-bootc/centos-bootc:stream9 bash` and in there do e.g.
> 
> ```
> # curl -k -L --remote-name-all https://kojihub.stream.rdu2.redhat.com/kojifiles/work/tasks/5597/4535597/dracut-{squash-,network-,}057-68.git20240816.el9.x86_64.rpm
> # rpm -Uvh *.rpm
> ```
> 

Thanks, I didn't get around to bootc yet. Will do that next time! I kind-of doubted it would work in the container... it's somehow a bit counterintuitive, but it might be quite useful in CI / testing :).

pvalena:
> I can confirm this works on a booted/deployed ostree/bootc host system, but it still fails when run in a container image (in the general case, where the host kernel is different from the one embedded in the container).
> 
> This is actually a big general papercut with dracut right now in that it always assumes the kernel it's running on is the kernel to target, and one we work around as you can see in https://docs.fedoraproject.org/en-US/bootc/initramfs/#_regenerating_the_initrd
> 
> It would be quite helpful for dracut to detect it's in a container (e.g. `systemd-detect-virt -c`) to start, and check for a single `/usr/lib/modules/$kver` directory and automatically use _that_ kernel (and I guess this actually then kind of snowballs a bit into learning to _write_ to `/usr/lib/modules/$kver/initramfs.img` by default, if it exists).

@cgwalters 
The paths are handled not from dracut directly, but from install.d scriptlet (we also intend to move it to grub scriplet instead, as it's responsible for creating boot menu entries in the first place...) as this can vary on different from OSs, it's just guess work on lsinitrd part - for convenience. The paths aren't fixed, or mandatory, and what is default/selected in `dracut` is mostly irrelevant, as we expect everyone to use `install.d` scriplet instead, on all Fedora-derived OSs. It doesn't make sense to me, to change the default behaviour at all - better to preserve backwards compatibility, and handle everyhing in the [scriptlets](https://github.com/dracut-ng/dracut-ng/blob/main/install.d/50-dracut.install).

PR #583, opened 2024-08-16
fix(nfs): set correct ownership and permissions for statd directory

pvalena:
## Changes

The directory ownership for the statd directory should be rpcuser:rpcuser.

## Checklist

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1924950

### DIFF
--- a/modules.d/95nfs/module-setup.sh
+++ b/modules.d/95nfs/module-setup.sh
@@ -120,8 +120,13 @@ install() {
     mkdir -m 0755 -p "$initdir/var/lib/nfs"
     mkdir -m 0755 -p "$initdir/var/lib/nfs/rpc_pipefs"
     mkdir -m 0770 -p "$initdir/var/lib/rpcbind"
-    [ -d "/var/lib/nfs/statd/sm" ] && mkdir -m 0755 -p "$initdir/var/lib/nfs/statd/sm"
-    [ -d "/var/lib/nfs/sm" ] && mkdir -m 0755 -p "$initdir/var/lib/nfs/sm"
+    [ -d "$dracutsysrootdir/var/lib/nfs/statd/sm" ] \
+        && mkdir -m 0700 -p "$initdir/var/lib/nfs/statd" \
+        && mkdir -m 0755 -p "$initdir/var/lib/nfs/statd/sm" \
+        && chown -R rpcuser:rpcuser "$initdir/var/lib/nfs/statd"
+    [ -d "$dracutsysrootdir/var/lib/nfs/sm" ] \
+        && mkdir -m 0755 -p "$initdir/var/lib/nfs/sm" \
+        && chown -R rpcuser:rpcuser "$initdir/var/lib/nfs/sm"
 
     # Rather than copy the passwd file in, just set a user for rpcbind
     # We'll save the state and restart the daemon from the root anyway


modules.d/95nfs/module-setup.sh
```diff
@@ -120,8 +120,13 @@ install() {
     mkdir -m 0755 -p "$initdir/var/lib/nfs"
     mkdir -m 0755 -p "$initdir/var/lib/nfs/rpc_pipefs"
     mkdir -m 0770 -p "$initdir/var/lib/rpcbind"
-    [ -d "/var/lib/nfs/statd/sm" ] && mkdir -m 0755 -p "$initdir/var/lib/nfs/statd/sm"
-    [ -d "/var/lib/nfs/sm" ] && mkdir -m 0755 -p "$initdir/var/lib/nfs/sm"
+    [ -d "$dracutsysrootdir/var/lib/nfs/statd/sm" ] \
+        && mkdir -m 0700 -p "$initdir/var/lib/nfs/statd" \
+        && mkdir -m 0755 -p "$initdir/var/lib/nfs/statd/sm" \
+        && chown -R rpcuser:rpcuser "$initdir/var/lib/nfs/statd"
```
jozzsi, line 126: `rpcuser` might be user for other distributions. 
Can we perhaps use `inst_dir` here instead which would pick up the correct user automatically from the host ?

pvalena: How do you mean to use `inst_dir`? To grep for permissions?

FTR this is not changing anything outside the generated initramfs. Maybe there would be better a way to detect that with grep, like here: https://github.com/dracut-ng/dracut-ng/blob/ed5ab5787177f2be8a620a8d2d63a9ad26fbf149/modules.d/95nfs/module-setup.sh#L137

WDYT?

modules.d/95nfs/module-setup.sh
```diff
@@ -120,8 +120,13 @@ install() {
     mkdir -m 0755 -p "$initdir/var/lib/nfs"
     mkdir -m 0755 -p "$initdir/var/lib/nfs/rpc_pipefs"
     mkdir -m 0770 -p "$initdir/var/lib/rpcbind"
-    [ -d "/var/lib/nfs/statd/sm" ] && mkdir -m 0755 -p "$initdir/var/lib/nfs/statd/sm"
-    [ -d "/var/lib/nfs/sm" ] && mkdir -m 0755 -p "$initdir/var/lib/nfs/sm"
+    [ -d "$dracutsysrootdir/var/lib/nfs/statd/sm" ] \
```
aafeijoo-suse, line 123: This directory is not provided upstream, afaik it's RH/Fedora specific. See `man rpc.statd(8)` or http://git.linux-nfs.org/?p=steved/nfs-utils.git;a=blob;f=utils/statd/statd.man;h=7441ffde2687e2a541bdea78d04a30358cccb558;hb=refs/heads/master#l434

modules.d/95nfs/module-setup.sh
```diff
@@ -120,8 +120,13 @@ install() {
...
+        && mkdir -m 0700 -p "$initdir/var/lib/nfs/statd" \
+        && mkdir -m 0755 -p "$initdir/var/lib/nfs/statd/sm" \
+        && chown -R rpcuser:rpcuser "$initdir/var/lib/nfs/statd"
+    [ -d "$dracutsysrootdir/var/lib/nfs/sm" ] \
+        && mkdir -m 0755 -p "$initdir/var/lib/nfs/sm" \
+        && chown -R rpcuser:rpcuser "$initdir/var/lib/nfs/sm"
```
aafeijoo-suse, line 129: `rpcuser:rpcuser` does not exist on SUSE distros, it is `statd:statd`

PR #584, opened 2024-08-16
feat(dracut-init.sh): give --force-add precedence over --omit

pvalena:
## Changes

This gives precedence of force_add_dracutmodules to omit_dracutmodules, as there is not other way to override omit_dracutmodules list, and users would expect it to be overriden from command line.

Ref: https://github.com/dracut-ng/dracut-ng/pull/569

This way, `--add` retains it behaviour, and `--force-add` gains additional functionality in non-hostonly mode. The module may still be skipped if the module check returns 1, but it should throw error (as I'd expect for `--force-add`).

## Checklist

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -938,8 +938,10 @@ check_module() {
     [[ $2 ]] || mods_checked_as_dep+=" $_mod "
 
     if [[ " $omit_dracutmodules " == *\ $_mod\ * ]]; then
-        ddebug "Module '$_mod' will not be installed, because it's in the list to be omitted!"
-        return 1
+        if [[ " $force_add_dracutmodules " != *\ $_mod\ * ]]; then
+            ddebug "Module '$_mod' will not be installed, because it's in the list to be omitted!"
+            return 1
+        fi
     fi
 
     if [[ " $dracutmodules $add_dracutmodules $force_add_dracutmodules" == *\ $_mod\ * ]]; then


LaszloGombos:
I think this is a good direction and I would prefer this PR over https://github.com/dracut-ng/dracut-ng/pull/569 . I acknowledge this PR is still in draft state, but wanted to note that documenting that `--force-add` now has precedence over over `--omit` would be important here https://github.com/dracut-ng/dracut-ng/blob/main/man/dracut.8.asc?plain=1#L80

pvalena:
Tested: https://gist.github.com/pvalena/f8cdb9aab623e2f713d823df1824fa72

Should I change/clarify `--help` or man pages; e.g. the `--force-add` option or the `dracut.conf` docs?

pvalena:
@LaszloGombos sorry, I've missed your previous comment -- like this, or some other wording?
https://github.com/dracut-ng/dracut-ng/commit/3cda7e55afe343f3f55456089a17efa772ce697b

LaszloGombos:
> [3cda7e5](https://github.com/dracut-ng/dracut-ng/commit/3cda7e55afe343f3f55456089a17efa772ce697b)

This language works for me.

pvalena:
> @pvalena is this still draft PR ? Thanks !

I didn't get around to writing the docs change, sorry :).

ianw:
> @ianw Can you perhaps help @pvalena with the doc on this small change ? I can make sure the changes land together (so that it does not have to be the same PR). Thank you both.

here's my best guess : https://github.com/dracut-ng/dracut-ng/pull/802 :)

PR #585, opened 2024-08-17
fix(crypt): install dm_crypt module in non-hostonly mode as well

jozzsi:
Without dm_crypt Linux module, the dracut module will not be fully functional.

## Checklist

### DIFF
--- a/modules.d/90crypt/module-setup.sh
+++ b/modules.d/90crypt/module-setup.sh
@@ -23,8 +23,7 @@ depends() {
 
 # called by dracut
 installkernel() {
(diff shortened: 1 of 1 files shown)


LaszloGombos:
@jozzsi This change looks good to me. Can you please share how did you found this issue, so that we can perhaps understand better the motivation ?

jozzsi:
> Can you please share how did you found this issue, so that we can perhaps understand better the motivation ?

I was working on https://github.com/dracut-ng/dracut-ng/pull/589, which turns on hostonly for Alpine and exposed this issue.

PR #587, opened 2024-08-17
test(FULL-SYSTEM): simplify and make it easier to maintain

jozzsi:
Remove packaging and distribution specific systemd service files.
    
No need to explicitly check if /usr is mounted as without it Switch Root service would fail with the following error:
'Failed to start initrd-switch-root.service - Switch Root.'

## Checklist

### DIFF
--- a/test/TEST-04-FULL-SYSTEMD/test-init.sh
+++ b/test/TEST-04-FULL-SYSTEMD/test-init.sh
@@ -1,21 +1,7 @@
 #!/bin/sh
 : > /dev/watchdog
-
-export PATH=/usr/sbin:/usr/bin:/sbin:/bin
-command -v plymouth > /dev/null 2>&1 && plymouth --quit
 exec > /dev/console 2>&1
 
-systemctl --failed --no-legend --no-pager > /run/failed
-
-ismounted() {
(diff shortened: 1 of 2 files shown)


test/TEST-04-FULL-SYSTEMD/test-init.sh
```diff
@@ -1,21 +1,7 @@
...
-    findmnt "$1" > /dev/null 2>&1
-}
-
-if ! ismounted /usr; then
-    echo "**************************FAILED**************************"
-    echo "/usr not mounted!!"
-    cat /proc/mounts >> /run/failed
-    echo "**************************FAILED**************************"
-fi
+usr/bin/systemctl --failed --no-legend --no-pager > /run/failed
```
LaszloGombos, line 5: I think this should be `/usr/bin/systemctl'.

LaszloGombos (changes requested):
Thanks for helping with tests and contributing to dracut. Other than that one nit, it looks good to me. I think it is a good idea to remove this leftover code.

test/TEST-04-FULL-SYSTEMD/test-init.sh
jozzsi: Good catch.

PR #590, opened 2024-08-17
revert(dracut-init.sh): add module to mods_to_load before checking dependencies

jozzsi:
This reverts commit d0f8fde5668cfd7fda1d15824e268b4949b4fd04.

This PR resolves the issue where during initrd generation dracut modules where dependencies are not met are included in the initrd (e.g. rngd for Alpine).

## Checklist

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -893,9 +893,6 @@ check_mount() {
         fi
     fi
 
-    [[ " $mods_to_load " == *\ $_mod\ * ]] \
-        || mods_to_load+=" $_mod "
-
     for _moddep in $(module_depends "$_mod" "$_moddir"); do
         # handle deps as if they were manually added
         [[ " $dracutmodules " == *\ $_mod\ * ]] \
@@ -913,6 +910,9 @@ check_mount() {
             return 1
         fi
     done
+
+    [[ " $mods_to_load " == *\ $_mod\ * ]] \
+        || mods_to_load+=" $_mod "
 
     return 0
 }
@@ -968,9 +968,6 @@ check_module() {
         fi
     fi
 
-    [[ " $mods_to_load " == *\ $_mod\ * ]] \
-        || mods_to_load+=" $_mod "
-
     for _moddep in $(module_depends "$_mod" "$_moddir"); do
         # handle deps as if they were manually added
         [[ " $dracutmodules " == *\ $_mod\ * ]] \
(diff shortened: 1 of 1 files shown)


LaszloGombos:
CC @prudo1 for visibility and suggestions.

prudo1:
@LaszloGombos thanks for the hint!

@jozzsi could you give some more information on your setup. Thing is that I tested this scenario while implementing the PR. In those tests everything worked as expected, esp. the initrd build failed when a dependency could not be meat.  But there's always a chance that I missed a case while testing.

jozzsi:
@prudo1 The regression is for the following scenario:

Dracut module `A` has a dependency on dracut module `B`. The `check()` function for module `A` passes, but for module `B` it fails. In this scenario not only module `B` should not be installed, but also module `A` should not get installed.

pvalena:
Oh, I'm unsure how I've missed this regression :). @jozzsi thanks for the explanation - we should include it in the comments.

@prudo1 yes, as the default (`auto`) is to include all modules, but only if the modules dependencies are included as well. Otherwise, the initrd is misconfigured. I do not think currently rich-dependencies (e.g. `or`) are supported, but there is `dracut_module_included` which could be utilized in check - but that relies on the correct module inclusion ordering (e.g. which are included first) - which I'm unsure dracut does.

Please let me know if you have some fix prepared.

LGTM with the revert, and `squash` needs to be fixed in some other way (e.g. by reversing the dependency direction & adjusting `check()` instead).

prudo1:
Hi @pvalena,
alright, when it's intended to drop modules, when their dependencies are not met, then we need to revert the commit.

The problem is that you cannot rely on `dracut_modules_included` within `check()` or `depends()`. For example, if you have three modules `A1`, `A2`, and `B`, where both `A1` and `A2` have a dependency on `B`. Then you cannot use `dracut_modules_included` in `B` to check which of the `Ax` modules tries to include it as the `Ax` modules will only be added to `modules_to_load` **after** `check()` from `B` has been executed. That's the situation for `99 squash` and the reason why I included the commit in the first place.

Anyway, I'll take a look on how to solve the dependency mess for `99squash`. The idea I'm having is to make `99squash` a pure meta module and move all the common code into a new `99squash-common` module. That would at least break the cyclic dependency the modules have at the moment.

prudo1:
I created a PR to fix the issue in 99squash. I also included the commit from this PR and improved the commit message. IMHO this PR can be closed.

@jozzsi I hope that is ok with you. If not, please let me know. I can drop the commit again.

jozzsi:
closing as https://github.com/dracut-ng/dracut-ng/pull/616 landed instead.

PR #596, opened 2024-08-21
fix(udev-rules): move *-persistent-storage.rules to rootfs-block

jozzsi:
Move rules that are not part of the udev upstream project and are dracut specific out of the `udev` dracut module for better module organization.

## Checklist

Fixes #


LaszloGombos (approved):
ok. This should make it easier to maintain `udev-rules` module.

Maybe later on we can even move these rules to more specific module (e.g. multipath).

Conan-Kudo (approved):

PR #598, opened 2024-08-21
fix(rescue): make rescue always no-hostonly

jozzsi:
Currently the rescue image is only set to no-hostonly if it is generated via dracut-rescue.install. Make the rescue image generation more predictable and consistent by setting no-hostonly right in the rescue module.

## Checklist

Fixes #


modules.d/03rescue/module-setup.sh
```diff
@@ -8,6 +8,8 @@ check() {
 
 # called by dracut
 install() {
+    export hostonly="no"
```
pvalena, line 11: Doesn't this override the `hostonly` variable even outside the module? It should probably be stored and then restored after the install.

pvalena: In any case, I don't see why would you set it here, as it has no effect AFAIK.

jozzsi: I got this idea from https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/80test/module-setup.sh#L26 .

I changed it to a rescue config file as we discussed at https://github.com/dracut-ng/dracut-ng/pull/569

pvalena (approved):
LGTM, although I've not tested the change. Do we have any test for rescue initrd?

freswa (approved):
Look reasonable to me.

LaszloGombos (approved):
LGTM

It seems soon one UKI can serve both a regular and rescue boot - see https://github.com/uapi-group/specifications/pull/108

I think this PR fits right into this use-case.

PR #600, opened 2024-08-24
ci(Debian): make clang the default C compiler

### LINKED ISSUE #593: add clang to CI
**Describe the enhancement**
In addition to gcc, CI should test with clang as well

jozzsi:
Debian will allow to test clang both amd64 on and on arm64

## Checklist

Fixes #593

### DIFF
--- a/test/container/Dockerfile-Debian
+++ b/test/container/Dockerfile-Debian
@@ -13,6 +13,7 @@ RUN apt-get update -y -qq && apt-get upgrade -y -qq && apt-get install -y -qq --
     bzip2 \
     ca-certificates \
     cargo \
+    clang \
     console-setup \
     cpio \
     cryptsetup \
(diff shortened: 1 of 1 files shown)


LaszloGombos (approved):
ok. this would fix https://github.com/dracut-ng/dracut-ng/issues/593

Conan-Kudo (changes requested):
Debian builds officially with GCC and does not use Clang as the default compiler. If we want clang tests, we should make a separate test image for it.

LaszloGombos:
> Debian builds officially with GCC and does not use Clang as the default compiler. If we want clang tests, we should make a separate test image for it.

@bdrung what are your thoughts on this - specifically for the Ubuntu CI container ? 

Clang would only be used to build dracut source code as other packages are build from source. Any interest in clang from your perspective ?

bdrung:
Clang is not the default on Debian/Ubuntu. So there is probably no need to test clang both on Debian and Ubuntu. So having one *additional* test with clang on either Debian or Ubuntu would be fine.

jozzsi:
It seems there is not enough interest in clang to make it a default in any of the containers. There are certainly other ways to test clang - one of which is to test it outside of the upstream CI. Closing this PR for now.

PR #601, opened 2024-08-24
test(test.sh): script to run tests in podman 

jozzsi:
Updated `HACKING.md` to point to the new test.sh script.

## Checklist

Fixes #

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -230,17 +230,7 @@ init has the following hook points to inject scripts:
 
 ```console
 $ cd <DRACUT_SOURCE>
-$ podman pull [CONTAINER]
-$ podman run --rm -it \
-    --cap-add=SYS_PTRACE --user 0 \
-    -v /dev:/dev -v ./:/dracut:z \
-    [CONTAINER] \
-    bash -l
-# cd /dracut
-# ./configure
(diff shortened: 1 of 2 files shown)


LaszloGombos (approved):
Thanks. After landing, I can also update the wiki - https://github.com/dracut-ng/dracut-ng/wiki/Dracut-development#running-tests-with-docker .

tools/test.sh
```diff
@@ -0,0 +1,13 @@
...
+
+TESTS=$1
+CONTAINER=$2
+
+[ -z "$CONTAINER" ] && CONTAINER='ghcr.io/dracut-ng/fedora'
+
+# if podman is available prefer podman over docker
+[ -x "$(command -v podman)" ] && alias docker=podman
+
+docker run --rm -it --device=/dev/kvm -v "$PWD"/:/z "$CONTAINER" /z/tools/test-github.sh id "$TESTS"
```
Conan-Kudo, line 13: I would prefer we only officially support Podman rather than both. There are subtle differences between the two and I don't want us to make people think we are capable of dealing with issues from Docker.

Conan-Kudo (changes requested):

pvalena (approved):

Conan-Kudo (approved):

PR #608, opened 2024-08-26
fix(base): init from base is not needed when systemd is enabled

jozzsi:
## Checklist

Fixes #


LaszloGombos (approved):
lgtm

Conan-Kudo (approved):

PR #612, opened 2024-08-26
fix(dracut-init.sh): remove unused functions

jozzsi:
## Checklist

Fixes #


Conan-Kudo (changes requested):
I'm not sure we can remove these, as third-party modules might be using them.

jozzsi:
Perhaps we should re-discuss this in a future release, but let's close it for now.

PR #613, opened 2024-08-26
chore: remove deprecated command line options

jozzsi:
Remove handling for the following command line options

- rdbreak
- rd.ccw
- rd_CCW
- rd_DASD_MOD
- rd_DASD
- rdinitdebug rdnetdebug
- rd_DM_UUID
- rdblacklist
- rdinsmodpost
- rdloaddriver
- rd_NO_FSTAB
- rdinfo
- check
- rdlivedebug
- live_dir
- liveimg
- overlay
- readonly_overlay
- reset_overlay
- live_ram
- rd_NO_CRYPTTAB
- rd_LUKS_KEYDEV_UUID
- rd_LUKS_KEYPATH
- rd_NO_LUKS
- rd_LUKS_UUID
- rd_NO_LVMCONF
- rd_LVM_LV
- rd_NO_LVM
- rd_LVM_VG
- rd_NFS_DOMAIN
- rd_NO_PLYMOUTH
- rd_retry
- rdshell
- rd_NO_SPLASH
- rd_NO_ZFCPCONF
- rd_ZFCP
- rd_ZNET

## Checklist

Fixes #

### DIFF
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -1431,8 +1431,6 @@ rd_LUKS_KEYDEV_UUID:: rd.luks.keydev.uuid
 
 rd_LUKS_KEYPATH:: rd.luks.keypath
 
-rd_NO_LUKS:: rd.luks=0
-
 rd_LUKS_UUID:: rd.luks.uuid
 
 rd_NO_LVMCONF:: rd.lvm.conf
--- a/man/dracut.cmdline.7.asc
+++ b/man/dracut.cmdline.7.asc
@@ -1391,8 +1391,6 @@ Deprecated, renamed Options
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 Here is a list of options and their new replacement.
 
-rd.ccw:: rd.znet
-
 rd_NO_DM:: rd.dm=0
 
 rdblacklist:: rd.driver.blacklist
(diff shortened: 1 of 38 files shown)


lnykryn:
Honestly, I have a bad feeling about this, but unfortunately, that's all I have. We don't have any hard data about what people use on the kernel cmdline.
I tried to go through the code of some dependent components, namely Anaconda and Lorax, and I don't think they use any of these, but I would be careful around "liveimg" and "overlay".
On the other hand, we had a deprecation warning there for a long time, so if we remove those, I don't think users can blame us if we remove them and something breaks.

LaszloGombos:
CC also @AdamWill for input on compat.

mwilck:
IMO this is fair. I haven't seen any of these options being used for years.

AdamWill:
yeah...I kinda agree with both @lnykryn and @mwilck , it feels pretty much inevitable that this will break *someone*, but I can't find anything it breaks in the obvious places. On the whole, I'd say let's try and it and see. It's easy enough to revert something if necessary.

pvalena:
LGTM.

pvalena (approved):

LaszloGombos (approved):
Thanks everyone for helping with the reviews and checking the usage of these arguments.

Conan-Kudo (approved):
This gives me the heebie-jeebies, but let's try it anyway...

PR #616, opened 2024-08-27
fix(squash): fix regression introduced when adding erofs support

prudo1:
This series reverts commit d0f8fde5 ("fix(dracut-init.sh): add module to mods_to_load before checking dependencies") that introduced a regression reported in #590 and adjusts 99squash to work without the commit. In addition the series contains two small fixes/improvements concerning the module.

Note: While testing the series I ran into a problem when including 95squash-erofs together with erofs-utils-1.8.1-1.fc40 on Fedora. Combining the right dracut options caused the boot to fail as file `/var/lib/dracut/hooks/initqueue/finished/devexists-\\x2fdev\\x2fdisk\\x2fby-uuid\\x2fb775d138-7db8-4e2b-8b74-6e859c1e8d45.sh` got corrupted. Downgrading to erofs-utils-1.7.1-3.fc40 fixed the problem. So I don't believe it is a problem caused by this series.

## Checklist

Fixes #
#590


LaszloGombos:
@prudo1 Thanks for the follow-up. I will abandon - https://github.com/dracut-ng/dracut-ng/pull/590.

How would you feel about `squash-lib` instead of `squash-common` ? I think `squash-lib` fits better with the existing dracut module naming schema.

prudo1:
@LaszloGombos no problem. After all I messed it up in the first place.

True, `squash-lib` fits better in the current naming convention. I've renamed the module and updated the commit messages. Hopefully I didn't miss anything.

LaszloGombos (approved):
lgtm

Conan-Kudo (approved):

PR #619, opened 2024-08-28
fix(systemd-networkd): make sure default network is always last

### LINKED ISSUE #618: Updated `systemd-networkd` module in v103 brings in unwanted DHCP config
**Describe the bug**
After updating Dracut to version 103 on one of the servers I manage, the server is unable to bring up its network interfaces properly inside initramfs.

**Distribution used**
Current Gentoo (Hardened, no-multilib, systemd).

**Dracut version**
[`sys-kernel/dracut-103-r2`](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-kernel/dracut/dracut-103-r2.ebuild) from Gentoo Portage tree.

**Init system**
[`sys-apps/systemd-255.7-r1`](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-apps/systemd/systemd-255.7-r1.ebuild) from Gentoo Portage tree.

**To Reproduce**
1. Have a Gentoo node that uses [`dracut-crypt-ssh`](https://github.com/dracut-crypt-ssh/dracut-crypt-ssh/) for remote LUKS unlock;
2. Have a network config that requires a bonding interface to be brought up with a static IP configuration in order to be reachable;
3. Have no `ip=` or similar options in kernel/initramfs cmdline, all network configuration is copied over from the root filesystems' `/etc/systemd/network` directory;
4. Upgrade Dracut from version [`060_pre20240104-r4`](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-kernel/dracut/dracut-060_pre20240104-r4.ebuild) (the previous "stable" version) or from version [`102`](https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-kernel/dracut/dracut-102.ebuild) (currently marked as "unstable" in Gentoo) to version `103-r1` and rebuild the initramfs;
5. Reboot. Because of the extra configuration files added by the new `systemd-networkd` module, the bonding interface fails to go online and the kernel logs various errors related to this interface, like it having more physical interfaces tied to it than it's supposed to have, etc.

**Expected behavior**
The `systemd-networkd` module should only add these new config files and hooks when needed and/or when asked by the system administrator.

**Additional context**
Here's how the servers' `/etc/systemd/network` directory looks like:
```
some-node ~ # ls -lah /etc/systemd/network
total 41K
drwxr-xr-x 2 root root   9 Aug 28 11:18 .
drwxr-xr-x 6 root root  17 Aug 28 04:43 ..
-rw-r--r-- 1 root root  73 Aug 28 11:18 bond0-members.network
-rw-r--r-- 1 root root 148 Apr 23 18:07 bond0.netdev
-rw-r--r-- 1 root root 339 Apr 23 22:14 bond0.network
-rw-r--r-- 1 root root 321 May 28 00:20 eno1.network
```

And here's the config (with IP addresses redacted):
```ini
## bond0-members.network
[Match]
Name=enp6s0f0 enp6s0f1

[Link]
RequiredForOnline=no

[Network]
Bond=bond0
####

## bond0.netdev
[NetDev]
Name=bond0
Kind=bond

[Bond]
Mode=802.3ad
MIIMonitorSec=1s
LACPTransmitRate=fast
UpDelaySec=2s
DownDelaySec=8s
TransmitHashPolicy=layer2+3
####

## bond0.network
# External world-reachable network
[Match]
Name=bond0

[Network]
Gateway=192.168.XXX.YYY
# Disable any dynamic addressing
# for both IPv4 and IPv6:
DHCP=no
IPv6AcceptRA=no
LinkLocalAddressing=no

[Address]
Address=192.168.XXX.ZZZ/24
Scope=link

[Address]
Address=A.B.C.D/32
Scope=global
####

## eno1.network
# Site-local OOB network
[Match]
Name=eno1

[Network]
Address=192.168.XXX.YYY/24
# Disable any dynamic addressing
# for both IPv4 and IPv6:
DHCP=no
IPv6AcceptRA=no
LinkLocalAddressing=no

[Route]
Destination=192.168.XXX.0/24
Gateway=192.168.XXX.ZZZ
Metric=200
####
```

On Dracut versions `060` and `102` the `etc/systemd/network` directory inside the generated initramfs is identical to the same directory on the root filesystem. Version `103` adds a new file, `99-dracut-default.network`, which seems to somehow mess up with the existing config. Maybe these new config files and hooks should only be added if a certain option is present in the cmdline (e.g. `netroot=<...>`)? Or maybe it would be better to extract this new functionality into a new module?

BtbN:
## Changes

Renames the default network file so it has 4 leading 'z', to make it as unlikely as possible for it to be anything but the last file, even if the user has files not following the common naming scheme of "00-abc.network".

## Checklist

Fixes #618

### DIFF
--- a/modules.d/01systemd-networkd/module-setup.sh
+++ b/modules.d/01systemd-networkd/module-setup.sh
@@ -63,7 +63,7 @@ install() {
         "$systemdsystemunitdir"/systemd-networkd-wait-online.service.d/99-dracut.conf
 
     inst_simple "$moddir"/99-default.network \
-        "$systemdnetworkconfdir"/99-dracut-default.network
+        "$systemdnetworkconfdir"/zzzz-dracut-default.network
 
     inst_hook cmdline 99 "$moddir"/networkd-config.sh
     inst_hook initqueue/settled 99 "$moddir"/networkd-run.sh
--- a/modules.d/01systemd-networkd/networkd-config.sh
+++ b/modules.d/01systemd-networkd/networkd-config.sh
@@ -20,7 +20,7 @@ for f in /run/systemd/network/*.network; do
     } >> "$f"
 
     # Remove the default network if at least one was generated
-    rm -f "$systemdnetworkconfdir"/99-dracut-default.network
+    rm -f "$systemdnetworkconfdir"/zzzz-dracut-default.network
 done
 
 # Just in case networkd was already running


LaszloGombos (approved):
Thanks !

Conan-Kudo (approved):

PR #622, opened 2024-08-29
fix(fips-crypto-policies): make it depend on fips dracut module

jozzsi:
## Checklist

Fixes #

### DIFF
--- a/modules.d/01fips-crypto-policies/module-setup.sh
+++ b/modules.d/01fips-crypto-policies/module-setup.sh
@@ -10,6 +10,12 @@ check() {
 }
 
 # called by dracut
+depends() {
+    echo fips
+    return 0
+}
+
+# called by dracut
 install() {
     inst_hook pre-pivot 01 "$moddir/fips-crypto-policies.sh"
 


LaszloGombos:
CC @neverpanic to help with the review. Thanks !

modules.d/01fips-crypto-policies/module-setup.sh
```diff
@@ -22,6 +23,4 @@ installkernel() {
 # called by dracut
 install() {
     inst_hook pre-pivot 01 "$moddir/fips-crypto-policies.sh"
-
-    inst_multiple mount
```
neverpanic, line 26: Not sure what dracut's policies on duplicate `inst_multiple` calls are — I realize you're removing this because the `fips` module already includes it, but personally, I'd have left it to document the dependency.

modules.d/01fips-crypto-policies/module-setup.sh
```diff
@@ -11,6 +11,7 @@ check() {
 
 # called by dracut
 depends() {
+    echo fips
```
neverpanic, line 14: In general, this looks good to me.

There's one pitfall, that we should consider but probably ignore as a misconfiguration:
- users with `fips=1` on the kernel command line, and
- `/boot` on a separate filesystem, and
- no `boot=UUID=$fs_uuid` parameter, and
- the `fips` dracut module not enabled

Those configurations would start to fail to boot when the `fips` module is enabled. Personally, I believe this is a misconfiguration, since it skips the kernel integrity checks that would be required by FIPS, so it's better to make this explicitly fail rather than continue to silently ignore it.

neverpanic (approved):

modules.d/01fips-crypto-policies/module-setup.sh
jozzsi: ok. `base` module also installs `mount`. As I understand there is a slight performance impact as the actual copying of the binary will happen several times.

LaszloGombos:
@jozzsi can you please rebase this PR ? Thanks !

LaszloGombos (approved):
lgtm

PR #623, opened 2024-08-29
test(RAID-DEG): reenable systemd for this test

jozzsi:
e0e5424a7b5e387ccb70e47ffea5a59716bf7b76 enabled this fix as well.

## Checklist

Fixes #


LaszloGombos:
@jozzsi It seems this test still fails on Ubuntu (only).

PR #626, opened 2024-08-30
chore: remove boilerplate

jozzsi:
It is not required to redefine empty functions.

## Checklist

Fixes #

### DIFF
--- a/modules.d/00bash/module-setup.sh
+++ b/modules.d/00bash/module-setup.sh
@@ -4,29 +4,17 @@
 
 # Prerequisite check(s) for module.
 check() {
-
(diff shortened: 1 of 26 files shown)


LaszloGombos (approved):
ok

Conan-Kudo:
Do we have default behaviors when these aren't defined?

jozzsi:
> Do we have default behaviors when these aren't defined?

Yes. For example for `check`, see https://github.com/dracut-ng/dracut-ng/blob/main/dracut-init.sh#L747 . Also for this amount of change, tests would have failed if there are no defaults.

Conan-Kudo (approved):

PR #627, opened 2024-08-30
fix(systemd): remove steps not required inside initrd

jozzsi:
- fix(systemd): /sbin/init is not required inside initrd
- fix(systemd): do not set unused target as the default 
- test(NBD): fix the way rootfs is built for test

## Checklist

Fixes #

### DIFF
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -137,7 +137,6 @@ EOF
         unset _wrapper
     fi
     ln_r "$_systemdbinary" "/init"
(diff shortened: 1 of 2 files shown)


LaszloGombos:
@jozzsi I think the change in the test case introduces an issue that was fixed by https://github.com/dracut-ng/dracut-ng/pull/629 . Can you please do the same approach here in this PR ?  Thanks !

LaszloGombos (approved):
lgtm

Conan-Kudo (approved):

PR #629, opened 2024-08-31
test: network dracut package needs to be added with '-a'

jozzsi:
'-m network' does not really work as it is not able to include the actual networking backend as [`check_module`](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/40network/module-setup.sh#L20) call with always fail.

## Checklist

Fixes #


LaszloGombos (approved):
lgtm

PR #633, opened 2024-08-31
test(FULL-SYSTEM): remove non-systemd dracut modules from the test

jozzsi:
i18n is not longer required after 238378ad2b7367900a0050bb9fc5b9eab5294cba.

## Checklist

Fixes #


LaszloGombos (approved):
ok. 

Seems this create an `mkosi-initrd` type of initrd. It is good to have a test for that with dracut.

PR #635, opened 2024-09-01
test(BASIC): allow to run the test with various fstypes

jozzsi:
Default fstype (when test allows it) is now xfs for Arch and btrfs for openSUSE.   

Tested with the following fstypes:
 - ext2
 - ext3
 - ext4
 - btrfs
 - jfs
 - xfs
 
XFS filesystem labels can be at most 12 characters long, so move the test case for the '  rdinit=/bin/sh' label to TEST-02 instead and keep TEST-01 simple.

XFS filesystem must be larger than 300MB.

## Checklist

Fixes #

### DIFF
--- a/test/TEST-01-BASIC/create-root.sh
+++ b/test/TEST-01-BASIC/create-root.sh
@@ -15,7 +15,7 @@ set -ex
 # populate TEST_FSTYPE
 . /env
 
-eval "mkfs.${TEST_FSTYPE} -q -L '  rdinit=/bin/sh' /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_root"
+eval "mkfs.${TEST_FSTYPE} -q -L dracut /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_root"
 mkdir -p /root
 mount -t "${TEST_FSTYPE}" /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_root /root
 cp -a -t /root /source/*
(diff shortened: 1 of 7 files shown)


LaszloGombos (approved):
nice.

Perhaps we should really keep BASCI test basic and move some of the 'non-basic' tests steps out of BASIC into another tests (e.g. unusual label with spaces test, unusual driver name tests).

Conan-Kudo:
I'm not sure we want to introduce tests for bcachefs right now, the kernel and userspace is still experimental and not stable.

jozzsi:
> I'm not sure we want to introduce tests for bcachefs right now, the kernel and userspace is still experimental and not stable.

This change allows testing `jfs` and the test does not really tests bcachefs - as in bcachefs is not mentioned in the code, it simply allows (outside of the tree) to test with bcachefs also.

If it helps, perhaps I can also improve the test to support xfs (which I think needs bit more changes).

Conan-Kudo:
I think we're going to want to test ext4; xfs; and btrfs, as those are the three commonly used filesystems.

jozzsi:
> I think we're going to want to test ext4; xfs; and btrfs, as those are the three commonly used filesystems.

@Conan-Kudo The easiest would be to have a 'preferred' filesystem type for each distribution/CI container that individual tests can overrule. 

Some tests (e.g. TEST-04) would force btrfs, but most tests would just default to the 'preferred' filesystem to somewhat minimize the number of tests we run.

Currently for all CI container the 'preferred' filesystem is always ext4. 

What would be a good CI container to 'prefer' btrfs or xfs ? Fedora ? openSUSE ? Arch ?

pvalena (approved):

Issue #145, opened 2024-04-08
Support bcachefs as well as bcachefs encrypted root unlocking


notmentaloutlaw:
Bcachefs is a new fs built into the kernel since 6.7 it will also be included in fedora 40 when it releases.
It supports filesystem level encryption which has its benefits over normal luks. More information can be found at https://bcachefs.org/Encryption/.

Just like how "crypt" module opens luks encrypted root via cryptsetup. It would be great if dracut could unlock bcachefs encrypted partitions a similar way. In userspace the way to unlock a bcachefs encrypted drive and mount it is:

```
bcachefs unlock /dev/sdaX
# type password
mount /dev/sdaX /mountpoint
```

Again there is more of an explanation on their website https://bcachefs.org/Encryption/ as well as resources such as https://wiki.archlinux.org/title/Bcachefs, https://wiki.gentoo.org/wiki/Bcachefs.

necrose99:
https://chatgpt.com/share/68e86caa-d66c-8005-8b9b-f30cf5af07c2

was meddling with own dracut modules ...
However the more dynamic root detection is good... better than mine  inital prototype... 
  both sets could be merged perhaps... 

Anyway I had concepted gpg , tmps2,  yubikey oath as means to unseal.... and pipe cmd 
Mines more in draft 0.1 RFC  but anyone cares to run with it.. 

Force drivers in , Include tpm 2 module 
Include gpg , yubikey  option/s 

Dkms module is in guru , for kmod , 6.18 etc..
Or force patching kernel 
Gpo.zugaina.org....  

Refind or grub , could use efi stub drivers of bcachefs...

PR #636, opened 2024-09-01
ci: add more fstypes support (xfs, jfs, f2fs)

jozzsi:
This PR is just a container change, no tests are added or removed.

## Checklist

Fixes #

### DIFF
--- a/test/container/Dockerfile-Ubuntu
+++ b/test/container/Dockerfile-Ubuntu
@@ -25,6 +25,7 @@ RUN apt-get update -y -qq && apt-get upgrade -y -qq && \
     fcoe-utils \
     fdisk \
     file \
+    f2fs-tools \
     g++ \
     gawk \
     git \
@@ -34,6 +35,7 @@ RUN apt-get update -y -qq && apt-get upgrade -y -qq && \
     isc-dhcp-client \
     isc-dhcp-server \
     iscsiuio \
+    jfsutils \
     jq \
     kmod \
     less \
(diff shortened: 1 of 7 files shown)


test/container/Dockerfile-Ubuntu
```diff
@@ -6,6 +6,7 @@ RUN apt-get update -y -qq && apt-get upgrade -y -qq && \
     DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends -o Dpkg::Use-Pty=0 \
     asciidoc \
     astyle \
+    bcachefs-tools \
```
bdrung, line 9: bcachefs-tools was removed from Debian unstable and the future is unclear.

jozzsi: @bdrung what do you suggest we do ? 

As you can see I did not add it to the Debian container and I certainly hope we can have at least one container with bcachefs tools (it does not have to be Ubuntu).

If the CI container no longer builds due some distro packaging changes, it is trivial to remove.

There seems to be interest for a dedicated bcachefs dracut module also - https://github.com/dracut-ng/dracut-ng/issues/145

Conan-Kudo: I agree here, as it is, bcachefs is experimental and churning too much for use to test. I would leave it out for now. 
I think we're going to want to test ext4; xfs; and btrfs, as those are the three commonly used filesystems.

bdrung: I am fine with adding bcachefs-tools to the Ubuntu container, since it can be easily dropped when needed.

jozzsi: For now I removed bcachefs in the interest of making progress first with other fs types, like xfs .

LaszloGombos (approved):
thanks

PR #637, opened 2024-09-02
fix(systemd-udevd): make systemd-sysctl, systemd-modules-load optional

### LINKED ISSUE #510: Dracut now has many dependencies on systemd-modules-load, requiring building modules support into systemd.
**Describe the bug**
With the recent rework/expansion of systemd integration in dracut-10x,
quite a few dependencies now tie in systemd-modules-load unconditionally.
Which in turn requires module support compiled into systemd.
I use dracut for UKIs with monolithic kernels, it would be nice to have those handled.
The ramdisk's job is to unlock the root file system using a pcr measurement + pin.

That said, I'm aware of the ancient workaround monolithic kernels require (for hostonly) anyway
https://github.com/dracutdevs/dracut/issues/56
so this may well not have priority and rather be a feature request/maintenance load/...

I'll probably file a request https://github.com/gentoo to have dracut depend on systemd[kmod]
if the dep is here to stay :) Thanks!

**Distribution used**
Gentoo

**Dracut version**
102

**Init system**
 systemd-256.1

**To Reproduce**
Using systemd-256.1 without support for kernel modules compiled in,
creating the UKI fails due to unsatisfied dependencies.

**Expected behavior**
Even more complex environment tests at runtime?

**Additional context**
I originally filed this as https://github.com/dracutdevs/dracut/issues/2664 in error, and was redirected here.

jozzsi:
systemd-modules-load is not a hard dependency for systemd-sysctl or for systemd-udevd.

## Checklist

Fixes #510


jozzsi:
@AndreMullinger @AndrewAmmerlaan I was wondering if one you could help testing or reviewing this PR ? 

I run some manual tests and also changed one of the CI test to test for this condition.

`systemd-modules-load` not installed (but rest of systemd is installed) seems to be a condition that would likely only happen only on Gentoo with `with systemd[-kmod]` . Thanks !

modules.d/01systemd-udevd/module-setup.sh
```diff
@@ -16,9 +16,18 @@ check() {
 
 # Module dependency requirements.
 depends() {
-    # This module has external dependency on other module(s).
-    echo udev-rules systemd systemd-sysctl systemd-modules-load
-    # Return 0 to include the dependent module(s) in the initramfs.
+    local deps
+    deps="udev-rules systemd"
+
+    # install optional dependencies unless there are omitted
```
AndreMullinger, line 22: s/there/they

jozzsi: fixed. Thanks !

AndreMullinger: I applied your patches (minus the third patch modifying the test) to the latest git version and built a ramdisk against systemd[-kmod]. It builds and boots without issue, the boot log looks good.

I don't feel qualified to review the patches, really. I can follow their intent and the logic looks right, fwiw.

LaszloGombos (approved):
Thanks @jozzsi and @AndreMullinger . Test case looks great as well !

I noticed similar PR to check optional dependencies are made for network meta module.

PR #638, opened 2024-09-02
fix(network): handle '-m network'

### LINKED ISSUE #634: --modules dracut argument is not compatible with meta modules such as network meta module
**Describe the bug**
On the same host the following two commands leads to different set of dracut modules installed. This is unexpected.

`dracut -a network` - works as expected 
`dracut -m network` - always selects network-legacy 

If the meta module is included via `-m` `check_module` fails unexpectedly.

jozzsi:
If the meta module is included via `-m` `check_module` fails unexpectedly.

Call `module_check` directly instead of `check_module`.

Use this as an opportunity to refactor the network module code for improved readability.

From the CI test log it can now be observed that some tests are switched to `network-manager` from `network-legacy` (e.g. on arch) when `USE_NETWORK` is set to `network`.

This change expose some additional issue with the tests themselves. For some tests we need to continue to force network-legacy for server as most test server code not yet support booting into systemd and instead it just starts the server code from /sbin/init.

## Checklist

Fixes #634

### DIFF
--- a/modules.d/40network/module-setup.sh
+++ b/modules.d/40network/module-setup.sh
@@ -7,26 +7,26 @@ check() {
 
 # called by dracut
 depends() {
+    echo -n "net-lib kernel-network-modules "
+
     is_qemu_virtualized && echo -n "qemu-net "
 
-    for module in connman network-manager network-legacy systemd-networkd; do
+    for module in network-manager systemd-networkd connman network-legacy; do
         if dracut_module_included "$module"; then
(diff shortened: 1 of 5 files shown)


LaszloGombos (approved):
ok by me. 

Seems there is a lot more work here to make the CI pass only on `network-manager` without `network-legacy`. 

@jozzsi do you know why TEST-20 (NFS) does not have the same issue as other networking tests ?

Issue #154, opened 2024-04-11
[RFC] Automatically generate deps for dlopen() deps?


poettering:
Many of systemd's library dependencies were turned into dlopen() based dependencies in the recent years/months. For classic initrd generators these are hard to automatically handle, because they are not mentioned in the ELF header.

In this issue over on the systemd project I am proposing to change something about it: https://github.com/systemd/systemd/pull/32234

I was wondering if there's any interest in this from the dracut/dracut-ng camp?

The concept is generic, but systemd would be the first provider of this information I guess. You can extract this kind of "weak" dlopen library dependencies via `readelf -p .note.uapi.dlopen` or an equivalent command from ELF.

Anyway, would love your input on that issue, in case you are interested in this.

chewi:
Note to others, it ended up being `readelf -p .note.dlopen`.

Dracut doesn't seem to define a minimum systemd version, and obviously we can't require v256+ now, so we'd still need to keep the existing workarounds for a long while, only skipping them if systemd is new enough. We could check the version with pkg-config as we already get other values from there.

chewi:
[json-c](https://github.com/json-c/json-c) is quite widely used, including by cryptsetup, so it is likely to be present on systems running Dracut. @jozzsi, you asked me to look into this, so please could I get your opinion?

chewi:
Alternatively, I've just noticed that jq is an optional Dracut dependency already. It seems more useful to the wider project than json-c.

jozzsi:
@chewi in some modules dracut is using [jq](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/95nvmf/module-setup.sh#L5). 

Would  that be an option here ? 

I would however do not introduce a hard mandatory dependency on jq in core systemd dracut modules - use it optionally to optimize if available, otherwise fall back to safe (and include libraries even if they might not be necessary needed).

chewi:
Heh, snap. Yes, that should work.

chewi:
It's been a few months, but I've now come back around to this after a long detour. It's nearly working using sd-json (rather than jq), but I need to turn the sonames into actual paths.

This can't be done using ldd. It can sort of be done with ldconfig, but I want to do it in a cross-friendly manner. I figured that if I could make that work, then I could also make the ldd stuff cross-friendly without having to rely on alternative scripts. This is all proving rather messy though.

[...]

PR #639, opened 2024-09-04
fix(systemd, systemd-journal): only install libgcrypt where actually needed

chewi:
systemd v256 started loading libgcrypt with dlopen, so the systemd module started installing it unconditionally. However, the only part of systemd that uses it without explicitly linking it is the journal forward secure sealing feature. This module does not install journald or anything else that is likely to need libgcrypt. Even if some other part did need it, it would be pulled in using the usual ldd mechanism.

This change also benefits older systemd versions because systemd can be built without any libgcrypt dependency at all.

Following that, the systemd-journal module should check whether libgcrypt is actually needed. systemd-resolved explicitly links against it, so check using ldd in a similar manner to the crypt-gpg module.

CC @aafeijoo-suse

## Checklist

### DIFF
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -155,6 +155,5 @@ EOF
     # Install library file(s)
     _arch=${DRACUT_ARCH:-$(uname -m)}
     inst_libdir_file \
-        {"tls/$_arch/",tls/,"$_arch/",}"libgcrypt.so*" \
         {"tls/$_arch/",tls/,"$_arch/",}"libnss_*"
 }
--- a/modules.d/01systemd-journald/module-setup.sh
+++ b/modules.d/01systemd-journald/module-setup.sh
@@ -54,10 +54,17 @@ install() {
     # Install library file(s)
     _arch=${DRACUT_ARCH:-$(uname -m)}
     inst_libdir_file \
-        {"tls/$_arch/",tls/,"$_arch/",}"libgcrypt.so*" \
         {"tls/$_arch/",tls/,"$_arch/",}"liblz4.so.*" \
         {"tls/$_arch/",tls/,"$_arch/",}"liblzma.so.*" \
         {"tls/$_arch/",tls/,"$_arch/",}"libzstd.so.*"
+
+    # The journal forward secure sealing feature loads libgcrypt with dlopen
+    # rather than explicitly linking against it, so we need to explicitly
+    # install it. However, systemd can be built without any libgcrypt
+    # dependency, so check whether systemd-resolved explicitly links against it.
+    if $DRACUT_LDD "${dracutsysrootdir}${systemdutildir}/systemd-resolved" | grep -q libgcrypt; then
+        inst_libdir_file {"tls/$_arch/",tls/,"$_arch/",}"libgcrypt.so*"
+    fi
 
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then


jozzsi:
> Following that, the systemd-journal module should check whether libgcrypt is actually needed. systemd-resolved explicitly links against it, so check using ldd in a similar manner to the crypt-gpg module.

It would be good to avoid inspecting (with ldd) systemd-resolved (which might not exists) inside the systemd-journal module . If we really end up doing this than a better place to do it would be the systemd-resolved dracut module itself.

Would it be sufficient just install libgcrypt.so* optionally (-o) but unconditionally ?

chewi:
libgcrypt.so is likely to be on the wider system for other reasons, as is the case for us, but we're desperately trying to keep the size of the initrd down. How about this compromise: if systemd-resolved is missing, then assume libgcrypt is needed?

jozzsi:
> How about this compromise: if systemd-resolved is missing, then assume libgcrypt is needed?

I find the current PR really hard to reason about (e.g. years from now on without a test case) and error prone (e.g. calling ldd inside --sysroot).

> we're desperately trying to keep the size of the initrd down

Could you just omit `systemd-journal` module entirely so that it does not pull-in `libgcrypt.so` ? This would require a PR to make systemd-journal` optional similar to https://github.com/dracut-ng/dracut-ng/pull/637

chewi:
I don't think calling ldd like this is controversial. crypt-gpg already does it like this, and ldd gets called against all the installed sysroot binaries behind the scenes.

Having now seen the Gentoo Dockerfile, I could add a test by disabling gcrypt there and checking that the library is omitted vs present in the other cases. That's assuming something else doesn't pull it in.

We don't want to omit the journal entirely. We just don't need this feature.

jozzsi:
CC @AndrewAmmerlaan 

> Having now seen the Gentoo Dockerfile, I could add a test by disabling gcrypt there and checking that the library is omitted vs present in the other cases. That's assuming something else doesn't pull it in.

Perhaps you can add a step to TEST-05, but please try to find a way to keep the test distro independent and piut the distro dependent bits into the Gento0 Dockerfile.

> We don't want to omit the journal entirely. We just don't need this feature.

Could you perhaps take advantage of https://github.com/dracut-ng/dracut-ng/issues/154 instead of inspecting `systemd-resolved` ?

jozzsi:
> I thought that sounded interesting, so I started looking into it, but there are one or two challenges.

Similar work in a different project - https://github.com/osboot/make-initrd/commit/48d87e552a4d707d6718189849b8ad9f6e451e4a

chewi:
Interesting, but they've taken quite a different approach, actually calling `dlopen` (which will break cross) and using json-c rather than jq.

chewi:
Oh, that's handy because the I/O handling around spawning jq was the ugliest part of what I'd written so far. I haven't forgotten about this either, just had other priorities lately.

keszybz:
The gcrypt dep is only used for "forward secure sealing". For two things: setting up a key to sign future journald and to check existing journals with a key that was stored externally. The first is not something that'd be done in an initrd. The second possibly could be done *from* an initrd, but in general, one would just use a live boot or move the disk to a different system for investigation. So overall, I don't think this feature is needed in the initrd and I'd suggest just dropping the library altogether.

chewi:
While we don't use forward secure sealing, we do a lot of other things in our initrd that would usually be done later, so I don't want to be a hypocrite by asking for it to be dropped. It's been a few months, but I've now come back around to this after a long detour. It's nearly working using sd-json, but I need to turn the sonames into actual paths. This can't be done using ldd. It can sort of be done with ldconfig, but I want to do it in a cross-friendly manner. Following that, I might as well change the ldd code to make that cross-friendly too, since the current approach of using alternative ldd scripts is not much fun.

chewi:
Hmmm. Finding the libraries is messier than I'd anticipated.

There's no proper way to determine the paths that ldconfig would look in. There is /etc/ld.so.conf, of course, but it may not include the "standard" locations, leaving us to guess. Guessing is essentially what the dracut-functions.sh libdirs code does, but that's only used to find specific libraries, not their dependencies. I feel like this needs to be more accurate. There are also weird edge cases like `$ORIGIN` that would need to be handled.

That makes me reconsider ldconfig, which works as long as the endianness is the same. I could live with that, but it doesn't exist for musl. For musl, we would have to read `/etc/ld-musl-${LDSO_ARCH}.path`, where `LDSO_ARCH` is almost impossible to determine. We'd have to glob in the hope of finding the right file and/or use the default locations. That's all doable, but it won't be pretty.

chewi:
I'm going to close this and repost some of the above in #154.

PR #607, opened 2024-08-25
test: document and fix extra verbose mode (V=2)

jozzsi:
Set 'rd.debug' for all tests  in 'V=2' mode to make it more verbose than 'V=1' test mode.

## Checklist
- [ ] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

Fixes #

PR #640, opened 2024-09-05
feat(test-root): only include debug module if V is set to 2

jozzsi:
This PR enables CI to run faster by default (V=1).

When debugging (V=2) debug module will be included as before.

Follow-up to https://github.com/dracut-ng/dracut-ng/pull/607 .

## Checklist

Fixes #

CC @bdrung

### DIFF
--- a/modules.d/80test-root/module-setup.sh
+++ b/modules.d/80test-root/module-setup.sh
@@ -6,7 +6,11 @@ check() {
 }
 
 depends() {
-    echo debug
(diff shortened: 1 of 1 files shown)


LaszloGombos:
@jozzsi Thanks. 

Perhaps you can have a larger PR for commits that are only improving tests and CI. I think this would help with the reviews and managing a review queue.

LaszloGombos (approved):

PR #641, opened 2024-09-06
fix(nvmf): require NVMeoF modules

mwilck:
Make sure the necessary kernel modules for NVMeoF boot support are included.

## Checklist

cc: @aafeijoo-suse @tbzatek @johnmeneghini @Douglas-Farley

### DIFF
--- a/modules.d/95nvmf/module-setup.sh
+++ b/modules.d/95nvmf/module-setup.sh
@@ -2,7 +2,10 @@
 
 # called by dracut
 check() {
+    local -A nvmf_trtypes
+
     require_binaries nvme jq || return 1
+    require_kernel_modules nvme_fabrics || return 1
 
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (commented):
The new check looks good, but this made me think about the current `installkernel()` function:

https://github.com/dracut-ng/dracut-ng/blob/5aa1c46618478064cb390b67f552ddfd3e0c1b17/modules.d/95nvmf/module-setup.sh#L72-L73

- why `nvme_rdma` can be required but is never installed?
- why `nvme_tcp` is always installed in hostonly mode if not is always required on the `check()` function?

mwilck:
I've pushed another commit to address @aafeijoo-suse's issues.

tbzatek (approved):
No objections, makes sense.

LaszloGombos (approved):
lgtm

PR #642, opened 2024-09-06
fix(plymouth): make plymouth-set-default-theme optional

jozzsi:
Some distributions (for example Ubuntu) does not package plymouth-set-default-theme.

## Checklist

Fixes #

CC @bdrung @zboszor

### DIFF
--- a/modules.d/50plymouth/module-setup.sh
+++ b/modules.d/50plymouth/module-setup.sh
@@ -20,7 +20,7 @@ check() {
     [[ "$mount_needs" ]] && return 1
     [[ $(pkglib_dir) ]] || return 1
 
-    require_binaries plymouthd plymouth plymouth-set-default-theme || return 1
+    require_binaries plymouthd plymouth || return 1
 
     return 0
 }
@@ -44,9 +44,21 @@ install() {
 
     inst_hook emergency 50 "$moddir"/plymouth-emergency.sh
 
-    inst_multiple readlink
+    inst_multiple \
+        readlink \
+        plymouth \
+        plymouthd
 
-    inst_multiple plymouthd plymouth plymouth-set-default-theme
+    inst_multiple -o \
+        plymouth-set-default-theme \
+        /usr/share/plymouth/plymouthd.defaults \
+        /usr/share/plymouth/themes/default.plymouth
+
+    # Install the hosts local user configurations if enabled.
+    if [[ $hostonly ]]; then
+        inst_multiple -H -o \
+            /etc/plymouth/plymouthd.conf
+    fi
 
     if ! dracut_module_included "systemd"; then
         inst_hook pre-trigger 10 "$moddir"/plymouth-pretrigger.sh


LaszloGombos:
@jozzsi I think you need to replace the plymouth-set-default-theme functionality with installing the default as part of the dracut module (see https://gitlab.freedesktop.org/plymouth/plymouth/-/blob/main/scripts/plymouth-set-default-theme.in?ref_type=heads#L150 )
 - /usr/share/plymouth/themes/default.plymouth
 - /etc/plymouth/plymouthd.conf
 - /usr/share/plymouth/plymouthd.defaults

jozzsi:
Since now `plymouth` module is included on the Ubuntu CI, it exposed a potential Ubuntu issue that Ubuntu is packaging `plymouth-populate-initrd` which calls into `plymouth-set-default-theme` which is not packaged on Ubuntu.

See the log from test runs on the CI:

```
dracut[I]: *** Including module: plymouth ***
/usr/libexec/plymouth//plymouth-populate-initrd: line 20: plymouth-set-default-theme: command not found
```

Longer term solution is probably continue to discuss https://github.com/dracut-ng/dracut-ng/pull/442 . This PR is a small step in that direction and it enables more testing on the CI as `plymouth` module is changing.

bdrung:
We have bug https://bugs.launchpad.net/ubuntu/+source/plymouth/+bug/1596220 and/or https://bugs.launchpad.net/ubuntu/+source/plymouth/+bug/2049222 for that.

jozzsi:
> We have bug https://bugs.launchpad.net/ubuntu/+source/plymouth/+bug/1596220 and/or https://bugs.launchpad.net/ubuntu/+source/plymouth/+bug/2049222 for that.

Thats good that Ubuntu has efforts to align with other distributions, but there is a larger questions here.

PR https://github.com/dracut-ng/dracut-ng/pull/442/ (back from July) is trying to eliminate the need for plymouth-set-default-theme (essentially reimplementing plymouth-set-default-theme inside dracut) - see 
https://github.com/dracut-ng/dracut-ng/pull/442/files#diff-fa3c4908c59513d6872eb05f2c32fef0432ffb6a6b859e623da8c0a5a50fa982L49 

> Can we ask @bdrung if Ubuntu could ship the binary?

Would this suggest that we are no longer pursuing PR https://github.com/dracut-ng/dracut-ng/pull/442/

Conan-Kudo:
I would prefer not to merge either this or #442 if we don't need to. People packaging plymouth wrong is a problem from their end that they need to fix.

jozzsi:
> I would prefer not to merge either this or #442 if we don't need to. People packaging plymouth wrong is a problem from their end that they need to fix.

I will close this PR based on this feedback from @Conan-Kudo and lets continue discussing #442 which had both positive and negative feedback.

bdrung:
I have some work in a branch to push https://github.com/dracut-ng/dracut-ng/pull/442 over the finish line.

Conan-Kudo:
I would prefer that this not be merged. I would like Ubuntu to consider no longer stripping `plymouth-set-default-theme` instead.

PR #646, opened 2024-09-07
chore(dracut-install): fix clang build

jozzsi:
Fix the following build error with clang

```
src/install/dracut-install.c:1643:58: error: unused parameter 'ctx' [-Werror,-Wunused-parameter]
 1643 | static void find_suppliers_for_sys_node(struct kmod_ctx *ctx, Hashmap *suppliers, const char *node_path_raw,
      |                                                          ^
1 error generated.
```

## Checklist

Fixes #

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -1640,7 +1640,7 @@ static int find_modules_from_sysfs_node(struct kmod_ctx *ctx, const char *sysfs_
         return 0;
 }
 
-static void find_suppliers_for_sys_node(struct kmod_ctx *ctx, Hashmap *suppliers, const char *node_path_raw,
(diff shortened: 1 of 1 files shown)


LaszloGombos (approved):
thanks.

@jozzsi  can you please share how have you created an environment for clang ?

jozzsi:
> @jozzsi can you please share how have you created an environment for clang ?

I used Alpine. I uploaded the container changes here - https://github.com/dracut-ng/dracut-ng/pull/649 .

@LaszloGombos what do you think of having Alpine for clang ? I tried to use Debian based container before (https://github.com/dracut-ng/dracut-ng/pull/600), but it seems for those containers it was more important to test the usual stable default distro configuration.

PR #647, opened 2024-09-07
chore: improve portability for calling `cp`

### LINKED ISSUE #379: dracut-install - cp compatibility

jozzsi:
`--sparse=auto` is the default behavior in coreutils.
`--reflink=auto` is the default behavior in coreutils since v9.0.  
All CI containers are using newer versions of cp than v9.0.

## Checklist

Fixes #379 (partially)

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -20,9 +20,9 @@
 export LC_MESSAGES=C
 
 if [[ $EUID == "0" ]] && ! [[ $DRACUT_NO_XATTR ]]; then
-    export DRACUT_CP="cp --sparse=auto --preserve=mode,timestamps,xattr,links -dfr"
+    export DRACUT_CP="cp --preserve=mode,timestamps,xattr,links -dfr"
 else
-    export DRACUT_CP="cp --sparse=auto --preserve=mode,timestamps,links -dfr"
+    export DRACUT_CP="cp --preserve=mode,timestamps,links -dfr"
 fi
 
 # is_func <command>
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -20,9 +20,9 @@
 export LC_MESSAGES=C
 
 if [[ $EUID == "0" ]] && ! [[ $DRACUT_NO_XATTR ]]; then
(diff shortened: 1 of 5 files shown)


jozzsi:
@marcosfrm @bdrung any concern with this change ? 

I tent to like removing code when possible.

Thanks !

marcosfrm:
They'll still need to patch things downstream, right? If yes, I don't see much advantage in merging this.

jozzsi:
@q66 are you able to chime in ? would this PR help in its current form ?

bdrung (approved):
LGTM

q66:
> @q66 are you able to chime in ? would this PR help in its current form ?

it's about the same (regardless i'd have to patch it in the same way)

q66:
that said i don't think you need to worry about this too much, dracut won't work on alternative userlands like our freebsd one regardless (too many assumptions elsewhere)

we are patching/packaging dracut-install because initramfs-tools needs it now, but the downstream patch we have is small and doesn't really add much work

if you wanted to be actually portable, the right thing would be to implement the necessary subset of cp logic in C and not depend on exec'ing cp at all, but that's a lot more involved in general...

bdrung:
...or not use `cp` at all. The initrd could be generated from the files at their original place.

q66:
every implementation has somewhat different behavior of the flags that don't conform to posix so i don't think portability is achievable (and chimerautils cp is actually slightly different from freebsd, since it preserves xattrs with `-a` implicitly when it can and ignores when it can't - i implemented that in order to have the xattr preservation functionality without adding new flags or significantly altering the behaviors, upstream freebsd does not do any xattr preservation, and the `-a` flag is effectively the same as `-RpP`; other than that it implicitly uses reflinks for both freebsd upstream and ours)

jozzsi:
Thanks for the feedback @q66 @marcosfrm @LaszloGombos ! Given the feedback I am no longer planning to pursue this idea further.

PR #649, opened 2024-09-07
ci(alpine): use clang instead of gcc for more test coverage

jozzsi:
CI container to reproduce clang build issues.

## Checklist

Fixes #


LaszloGombos (approved):
make sense to me . The primary purpose for CI containers is to increase test coverage .
It is good to keep the number of test containers small.

PR #650, opened 2024-09-08
ci: improve tests

jozzsi:
Create one PR for all the CI improvements to keep the review queue easier to manage.
- use ukify if available. Create test coverage for using dracut with ukify
- search for bash in $PATH to improve portability
- set rd.retry consistently by only setting it in one place
- do not force rootfstype, let the initrd figure it out
- remove less from the required binaries for tests
- remove dracut-lib.sh dependency from real rootfs
- ifcfg dracut module is no longer available
- tcpdump is not required to run the tests
- showmount is not required to run the tests
- move driver regression testing out to a different test
- consolidate TEST_KERNEL_CMDLINE
- move test scripts under test/
- create a config for test test-makeroot test-root
- move more code out from tests into common test functions
- always keep the generated initrd to help writing assertions
- remove unnecessary intermediate cleanups
- move test dracut modules under the test/ directory
- prepare for testing with more than one kernel command line
- move dracut.conf.d conf dir for tests under $TESTDIR
- move test scripts under test/


LaszloGombos (approved):
@jozzsi Thanks for this work.

I find it is a good approach to continue to wok on the CI - with a large PR but small commits within that PR to keep the review queue small.

The commits are small so it is easy to understand the changes, but please try to add more information in the comment messages (not just commit title) in general. I noticed there is a few, but not many.

I like the idea of moving all CI related files under test/ directory.

Looks like CI is still green, so I trust there is no regression.

Thanks again !

PR #653, opened 2024-09-10
fix(dracut): do not add all lib subdirs to `LD_LIBRARY_PATH` with `--sysroot`

aafeijoo-suse:
On systems with too many library subdirectories, the current approach causes `Argument list too long` errors, and the initrd also fails to build.

This patch does not fix the original issue by itself (i.e., libsystemd libs not found), so all the dracut modules that require libraries stored in subdirectories must explicitly install them using `inst_libdir_file`, which already successfully handles the `--sysroot` prefix.

## Checklist

Fixes e0b876823d9c608db7132cab9e5edd62543a27ae


LaszloGombos (approved):
thanks

CC @chewi for potential additional feedback.

chewi:
I take it these additional directories are normally found via the rpath? When cross-compiling, a special ldd script is needed to handle the non-native binaries, but the script is also expected to handle the sysroot. Perhaps such a wrapper should also be used when using `--sysroot` against a native target, then none of this special handling would be needed.

jozzsi:
related discussion - https://github.com/dracut-ng/dracut-ng/pull/339 (but should not hold up landing this PR).

aafeijoo-suse:
> I take it these additional directories are normally found via the rpath? When cross-compiling, a special ldd script is needed to handle the non-native binaries, but the script is also expected to handle the sysroot. Perhaps such a wrapper should also be used when using `--sysroot` against a native target, then none of this special handling would be needed.

FTR, an openSUSE package is using `--sysroot` to build the initrd of a specific btrfs snapshot (same host, same arch, same system). The best option in our case is not a complex ldd wrapper, but to stop using `--sysroot`. The good news is that will happen very soon, so my life will be easier.

chewi:
In #154, I am looking to use the libc's elf.h to find discover dlopen dependencies. It doesn't seem like a huge stretch to take this further and discover the regular `DT_NEEDED` dependencies in the same way, removing the need for ldd and setting `LD_LIBRARY_PATH`.

aafeijoo-suse:
> In #154, I am looking to use the libc's elf.h to find discover dlopen dependencies. It doesn't seem like a huge stretch to take this further and discover the regular `DT_NEEDED` dependencies in the same way, removing the need for ldd and setting `LD_LIBRARY_PATH`.

That would mean reverting e0b876823d9c608db7132cab9e5edd62543a27ae / d0c82322ce1b44e72539b6dd537711afb68f081e, which is ok from my side. But IMO 921792f201e954de461d3b551e01b5369d666db8 should stay, as other dracut modules also explicitly install their required libraries, but systemd was missing theirs.

PR #660, opened 2024-09-11
chore: improve modularity by moving steps into the proper modules

jozzsi:
## Checklist

Fixes #


LaszloGombos (changes requested):
@jozzsi this seems like a very small win. If the change can not be all moved to the `systemd` module (with no change in the `base` module) than I do not think we should land this PR.

PR #663, opened 2024-09-12
feat(systemd-coredump): persist coredumps to /sysroot

aafeijoo-suse:
Otherwise they are lost after switching root.

## Checklist


jozzsi:
/run (and journal) survives switching root and /sysroot might not support rw. 

I wonder how other systemd based initrd's deal with this scenario (e.g. mkosi) and if this should be handled by systemd instead of dracut.

PR #665, opened 2024-09-13
revert(dracut-emergency): optionally print fs help

jozzsi:
It does not look like this has been ever used.

This is the only commit that is attempt to use /run/dracut (instead of /run/initramfs).

Original commit: https://github.com/dracutdevs/dracut/pull/338

## Checklist

Fixes #

CC @markfasheh @danimo


aafeijoo-suse (commented):
> It does not look like this has been ever used.

- https://build.opensuse.org/projects/openSUSE:Factory/packages/btrfsprogs/files/module-setup.sh.in?expand=1
- https://build.opensuse.org/projects/openSUSE:Factory/packages/xfsprogs/files/module-setup.sh.in?expand=1

jozzsi:
> > It does not look like this has been ever used.
> 
> * https://build.opensuse.org/projects/openSUSE:Factory/packages/btrfsprogs/files/module-setup.sh.in?expand=1
> * https://build.opensuse.org/projects/openSUSE:Factory/packages/xfsprogs/files/module-setup.sh.in?expand=1

Thank you @aafeijoo-suse  . I mean that it does not look like this has been ever used in this repository (upstream). 

Thanks for pointing out the out of tree usage in openSUSE.

Not only we should than keep this functionality, but perhaps actually use it here upstream to provide at least some minimal test coverage and more importantly some default quality of life benefits to all dracut users regardless of the Linux distribution.

jozzsi:
systemd should mount /run as a tmpfs over what is actually stored in the /run directory in the initrd.

When [/run/dracut](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/98dracut-systemd/dracut-emergency.sh#L17) gets executed /run/dracut would not be visible anymore is it is mounted over.

@aafeijoo-suse can you please help us understand if this feature is currently working in openSUSE ?

LaszloGombos:
@jozzsi  I am wondering if it would be possible for the `btrfs` dracut module to install an `emergency` hook so that `inst_fsck_help` function and the change in `dracut-systemd/dracut-emergency.sh` is no longer needed. This would perhaps also make it function properly in the non-systemd case.

PR #666, opened 2024-09-13
fix(systemd-sysext): make non-hostonly non-host

jozzsi:
Follow-up to https://github.com/dracut-ng/dracut-ng/pull/125

Not sure about the policy of picking up host extensions and automatically assuming that they are needed in the initrd, but regardless extensions under /etc and /var by definition are host specific.

## Checklist

Fixes #

### DIFF
--- a/modules.d/01systemd-sysext/module-setup.sh
+++ b/modules.d/01systemd-sysext/module-setup.sh
@@ -36,9 +36,6 @@ install() {
 
     inst_multiple -o \
         "/usr/lib/confexts/*.raw" \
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (commented):
I'm tempted to comment on the errors in this PR, but since it was clearly not tested by the submitter, I won't.

LaszloGombos:
@jozzsi would you be able to add a test case for this change ? Thanks !

LaszloGombos (approved):

bdrung (approved):
LGTM as well.

PR #667, opened 2024-09-13
fix(dracut): don't apply aggressive strip to kernel modules

ryncsn:
Unlike ordinary binaries, kernel module will be unusable if stripped with "-p". Fix this by always use "-g" only.

So far it didn't cause many issues since most kernels have their modules signed or compressed so this is skipped.

Signed-off-by: Kairui Song <kasong@tencent.com>

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2177,6 +2177,7 @@ if [[ $do_strip == yes ]]; then
         fi
     done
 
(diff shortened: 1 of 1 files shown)


aafeijoo-suse (approved):

LaszloGombos (approved):
lgtm

PR #670, opened 2024-09-13
fix(dracut): --list-modules should imply --no-kernel as well

jozzsi:
This will ensure that 'dracut --list-modules' works well in a container.

## Checklist

Fixes #


LaszloGombos (approved):
thanks

PR #672, opened 2024-09-13
perf(base): /etc/initrd-release is only for systemd

jozzsi:
## Checklist

Fixes #


LaszloGombos:
@jozzsi  can you please break this PR out into two separate PRs as I think largely the two commits are independent. This would help accelerate the reviews as we can review both PRs independently. Thanks !

PR #673, opened 2024-09-14
chore(systemd): move steps into more specific dracut modules

jozzsi:
## Checklist

Fixes #


LaszloGombos:
@jozzsi As part of this work, it would be help to document the difference in scope between `systemd` and `dracut-systemd` and non-`systemd` dracut modules.

PR #676, opened 2024-09-18
perf(systemd-initrd): do not depend on base module

jozzsi:
Make sure /etc/initrd exist without depending on the base module.

Test SYSTEMD-INITRD confirms no regression as this tests now no longer includes the base module and it still passes the boot tests with systemd (hence the perf tag).

## Checklist

Fixes #


LaszloGombos (approved):
CC @PrincessDoreena 

I also agree that base module is a bit of a heavy dependency just for /etc/initrd-release.

I tink the biggest win here is still the passing test case (05).

PR #679, opened 2024-09-19
chore: create a dedicated directory for each reference config

jozzsi:
This change helps with downstream packaging as packaging will no no longer needs to rename the file.

## Checklist

Fixes #


LaszloGombos (approved):
ok

PR #681, opened 2024-09-20
fix(network): call both check_module and module_check

jozzsi:
Add back testing the module with check_module as well as module_check.

Follow-up to c4b57722fb

## Checklist

Fixes #

### DIFF
--- a/modules.d/40network/module-setup.sh
+++ b/modules.d/40network/module-setup.sh
@@ -21,7 +21,7 @@ depends() {
     for module in network-manager systemd-networkd connman; do
         # install the first viable module, unless there omitted
         module_check $module > /dev/null 2>&1
-        if [[ $? == 255 ]] && ! [[ " $omit_dracutmodules " == *\ $module\ * ]]; then
+        if [[ $? == 255 ]] && ! [[ " $omit_dracutmodules " == *\ $module\ * ]] && check_module "$module"; then
             echo "$module"
             return 0
         fi


jozzsi:
> @jozzsi I realize this is still a draft PR. Please include a description on how to reproduce the problem that you're resolving with this PR - especially as it seems to be some back and forth changes in this area. Thanks !

Without this PR, one some Linux distributions (e.g. Void) adding `network` meta module no longer works and fails with the following error

```
dracut[D]: Module 'dbus-broker' will not be installed, because command 'busctl' could not be found!
dracut[D]: Module 'dbus-daemon' will not be installed, because command 'busctl' could not be found!
dracut[E]: Module 'dbus' depends on module 'dbus-daemon', which can't be installed
dracut[E]: Module 'network-manager' depends on module 'dbus', which can't be installed
dracut[E]: Module 'network' depends on module 'network-manager', which can't be installed
```

LaszloGombos (approved):
thanks

sergio-correia:
> > @jozzsi I realize this is still a draft PR. Please include a description on how to reproduce the problem that you're resolving with this PR - especially as it seems to be some back and forth changes in this area. Thanks !
> 
> Without this PR, on some Linux distributions (e.g. Void) adding `network` meta module no longer works and fails with the following error
> 
> ```
> dracut[D]: Module 'dbus-broker' will not be installed, because command 'busctl' could not be found!
> dracut[D]: Module 'dbus-daemon' will not be installed, because command 'busctl' could not be found!
> dracut[E]: Module 'dbus' depends on module 'dbus-daemon', which can't be installed
> dracut[E]: Module 'network-manager' depends on module 'dbus', which can't be installed
> dracut[E]: Module 'network' depends on module 'network-manager', which can't be installed
> ```

Interestingly, with this change I experience the opposite on Fedora: with it, I am unable to add the `network` module, as the added `check_module()` call will fail.

sergio-correia:
> @sergio-correia - if/when you have a chance, it would be best to test v106 release - sadly it is not yet released on Fedora - see https://src.fedoraproject.org/rpms/dracut/pull-request/74

Hi there, thanks for the pointer. I just tested a local build of 106 now and the results do not change.

The test is fairly simple:
1) `dracut -f --add network` <- this works
2)  If we have a config file instead, it does not work:
```
cat /etc/dracut.conf.d/01-network.conf
dracutmodules+=" network "
```
```
dracut -f
dracut[E]: Module 'network' depends on module 'network-legacy', which can't be installed
dracut[E]: Module 'network' cannot be installed.
```

If I revert this patch (i.e., remove the `&& check_module "$module"` bit in `40network/module-setup.sh`), it works. 

Would you like me to open a separate issue to track this?

PR #682, opened 2024-09-20
fix: make sure both /lib64 and /lib are included

jozzsi:
Fixes https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1078545

This PR resolves the warning seen on CI (e.g. test 01-BASIC) both for Debian and Ubuntu containers.

```
[FAILED] Failed to start multipathd.service…Mapper Multipath Device Controller.
```

## Checklist

Fixes #

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -87,7 +87,7 @@ DRACUT_LDCONFIG=${DRACUT_LDCONFIG:-ldconfig}
 
 # Detect lib paths
 if ! [[ $libdirs ]]; then
-    if [[ $("$DRACUT_LDD" "$dracutsysrootdir$DRACUT_TESTBIN") == */lib64/* ]] &> /dev/null \
+    if [[ $("$DRACUT_LDD" "$dracutsysrootdir$DRACUT_TESTBIN") == *"=> "*/lib64/* ]] &> /dev/null \
         && [[ -d $dracutsysrootdir/lib64 ]]; then
         libdirs+=" /lib64"
         [[ -d $dracutsysrootdir/usr/lib64 ]] && libdirs+=" /usr/lib64"


bdrung:
What is the failure message of multipathd.service?

```
$ ls /lib/multipath/ -1
libcheckcciss_tur.so
libcheckdirectio.so
libcheckemc_clariion.so
libcheckhp_sw.so
libcheckrdac.so
libcheckreadsector0.so
libchecktur.so
libforeign-nvme.so
libprioalua.so
libprioconst.so
libpriodatacore.so
libprioemc.so
libpriohds.so
libpriohp_sw.so
libprioiet.so
libprioontap.so
libpriopath_latency.so
libpriorandom.so
libpriordac.so
libpriosysfs.so
libprioweightedpath.so
```

Which binary or library needs to load libraries from `/lib/multipath/`?

jozzsi:
> Which binary or library needs to load libraries from `/lib/multipath/`?

Not sure, but I can confirm that e.g. on Fedora the location of these files is `/lib64/multipath/`

https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/90multipath/module-setup.sh#L140 suppose to install all files under `multipath`, but somehow it does not work as intended on Debian based systems. It seems the intention is to include all files under that directory (and assume that all files might be needed).

mwilck (changes requested):
Nack.

Installation of the `.so` files is achieved via the `inst_libdir_file` command further down, as you already figured.
`inst_libdir_file` must be fixed for Debian.

Including a hardcoded path which may or may not match is not the way to proceed.

bdrung:
`libdirs` should contain `/lib /usr/lib` on Debian. `inst_libdir_file "multipath/*"` should end up iterating over `/lib/multipath/*` and therefore find the files.

jozzsi:
> `libdirs` should contain `/lib /usr/lib` on Debian. `inst_libdir_file "multipath/*"` should end up iterating over `/lib/multipath/*` and therefore find the files.

Sadly not..

libdirs is set to '/lib64 /usr/lib64 /lib/x86_64-linux-gnu' currently on the main branch.

Uploading a new PR that would change libdirs to 

'/lib64 /usr/lib64 /lib /usr/lib /lib/x86_64-linux-gnu'

bdrung:
That culprit code smells and the proposed change does not make this part better. Let me explain:

`[ $("$DRACUT_LDD" "$dracutsysrootdir$DRACUT_TESTBIN") == */lib64/* ]` is evaluated to `[ ldd /bin/sh == */lib64/* ]`.

On Debian or Ubuntu on amd64:

```
$ ldd /bin/sh
	linux-vdso.so.1 (0x00007ffcc46ff000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x0000712473807000)
	/lib64/ld-linux-x86-64.so.2 (0x0000712473a19000)
```

Shouldn't we query ld.so to get the list of shared library search path? See `ld.so --help` or:

```
$ ld.so --list-diagnostics | grep system_dirs
path.system_dirs[0x0]="/lib/x86_64-linux-gnu/"
path.system_dirs[0x1]="/usr/lib/x86_64-linux-gnu/"
path.system_dirs[0x2]="/lib/"
path.system_dirs[0x3]="/usr/lib/"
```

mwilck:
> That culprit code smells and the proposed change does not make this part better. Let me explain:
> 
> `[ $("$DRACUT_LDD" "$dracutsysrootdir$DRACUT_TESTBIN") == */lib64/* ]` is evaluated to `[ ldd /bin/sh == */lib64/* ]`.
> 

Perhaps this should be changed to something like this:
```
[[ $("$DRACUT_LDD" "$dracutsysrootdir$DRACUT_TESTBIN") == *"=> "*/lib64/* ]]
```
still looks like a hack but should do what's expected.

mwilck:
Or this, somewhat cleaner:
```
is64=
while read -r _a _b _c; do 
    [[ "$_b" == "=>" ]] || continue
    [[ "$_c" == */lib65/* ]] && { 
        is64=1
        break
    }
done < <(ldd /bin/sh)
echo "$is64"
```

zeha:
> Or this, somewhat cleaner:
> 
> ```
> is64=
> while read -r _a _b _c; do 
>     [[ "$_b" == "=>" ]] || continue
>     [[ "$_c" == */lib64/* ]] && { 
>         is64=1
>         break
>     }
> done < <(ldd /bin/sh)
> echo "$is64"
> ```

I don't know what `is64` should mean, but I'll note that on arm64 Debian doesn't use /lib64/ anywhere.

mwilck:
> I don't know what is64 should mean
It was just a code sample to demonstrate how to possibly improve the code to determine `libdirs`.

> but I'll note that on arm64 Debian doesn't use /lib64/
That's known. But other distributions do. What exactly do you want to convey here?

zeha:
> > but I'll note that on arm64 Debian doesn't use /lib64/
> 
> That's known. But other distributions do. What exactly do you want to convey here?

Just that checking for /lib64/ is a distribution-specific thing, and can only be classified as a hack. Please don't put it into shipping code. systemd caused us a lot of headache with them hardcoding /lib64 in some places.

mwilck:
> Just that checking for /lib64/ is a distribution-specific thing, and can only be classified as a hack. Please don't put it into shipping code. systemd caused us a lot of headache with them hardcoding /lib64 in some places.

I find it very hard not to be sarcastic about this comment. AFAIR, neither systemd nor dracut have been enthusiastically endorsed by the Debian community for a long time. They have been developed with different OS design in mind. There's nothing wrong with that. Debian developers are welcome to point out the incompatibilities, but calling anything that doesn't match the Debian design a "hack" is certainly not helpful. If anything, the original patch suggested in this PR can be called a hack.

I made a proposal how to fix the broken code that sets `libdir` such that it works on Debian. And all you can think of is call it a hack? Why not be constructive instead and make a better proposal?

Please kindly explain what harm my proposed solution does to Debian-based distributions.

jozzsi:
> `[[ $("$DRACUT_LDD" "$dracutsysrootdir$DRACUT_TESTBIN") == *"=> "*/lib64/* ]]`

Thanks @mwilck for the suggestions. This idea does not seem to work, based on my quick testing.

> I don't know what is64 should mean, but I'll note that on arm64 Debian doesn't use /lib64/ anywhere.

@zeha thanks for helping out. It would have been more actionable to mention the issue for more context and less chance to misunderstand the intention of your comment.

I have no plans to work on this PR any further, I hope someone more familiar with Debian/Ubuntu will uploads a PR that at least move this forward.

I filed a bug for someone else to pick it up - https://github.com/dracut-ng/dracut-ng/issues/685

PR #692, opened 2024-10-01
feat(fips): include openssl's fips.so and openssl.cnf

pvalena:
## Changes

Include default `openssl.cnf` and `fips.so` if openssl is detected.

Ref: https://github.com/redhat-plumbers/dracut-rhel9/pull/67

## Checklist

### DIFF
--- a/modules.d/01fips/module-setup.sh
+++ b/modules.d/01fips/module-setup.sh
@@ -62,4 +62,17 @@ install() {
     inst_multiple sha512hmac rmmod insmod mount uname umount grep sed cut find sort cat tail tr
 
     inst_simple /etc/system-fips
+
+    # if we have openssl we need to install their fips library and configuration
+    [ -x /usr/bin/openssl ] && {
+        read -r _ conf < <(openssl version -d)
+        conf=${conf#\"}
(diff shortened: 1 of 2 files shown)


pvalena:
I'm unsure how this affects another distros.

modules.d/01fips/module-setup.sh
```diff
@@ -62,4 +62,17 @@ install() {
...
+    [ -x /usr/bin/openssl ] && {
+        read -r _ conf < <(openssl version -d)
+        conf=${conf#\"}
+        conf=${conf%\"}
+        inst_simple "${moddir}/openssl.cnf" "$conf/openssl.cnf"
+
+        read -r _ mod < <(openssl version -m)
+        mod=${mod#\"}
+        mod=${mod%\"}
+        inst_simple "$mod/fips.so"
```
jozzsi, line 76: In the current code (crypt module), we do the following

```
    _arch=${DRACUT_ARCH:-$(uname -m)}
    inst_libdir_file \
        {"tls/$_arch/",tls/,"$_arch/",}"/ossl-modules/fips.so" \
```

PR #705, opened 2024-10-11
fix(dmsquash-live): checkisomd5 is installed into /usr/bin

jozzsi:
From https://abf.io/import/dracut/blob/rosa2023.1/0012-fix-dmsquash-live-set-correct-path-to-checkisomd5.patch

CC @mikhailnov @pvalena 

## Checklist

### DIFF
--- a/modules.d/90dmsquash-live/checkisomd5@.service
+++ b/modules.d/90dmsquash-live/checkisomd5@.service
@@ -6,7 +6,7 @@ Before=shutdown.target
 [Service]
 Type=oneshot
 RemainAfterExit=no
-ExecStart=/bin/checkisomd5 --verbose %f
+ExecStart=/usr/bin/checkisomd5 --verbose %f
 StandardInput=tty-force
 StandardOutput=inherit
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
This looks like something incomplete in ROSA, there should be a symlink for `/bin` to `/usr/bin`.

jozzsi:
`isomd5sum` (https://github.com/rhinstaller/isomd5sum) installs the binary into `/usr/bin`, not sure why should we keep the `.service` file to point to `/bin`, regardless of how ROSA has `/bin` arranged.

mikhailnov:
@jozzsi thanks for upstreamizing our pathes, I was going to do this, thanks for help!
@Conan-Kudo We do have a symlink /usr/bin -> /bin in rosa2023.1, but we did not have it in rosa2021.1, but I left this patch in rosa2023.1 to make it point to a correct location without symlinks.

pvalena:
LGTM.

Conan-Kudo (approved):

PR #711, opened 2024-10-12
fix(systemd-cryptsetup): change the ordering for consistency

jozzsi:
All other systemd modules are numbered as 01.

## Checklist

Fixes #


Cornelicorn:
> I just realized this could also help with [latchset/clevis#493](https://github.com/latchset/clevis/pull/493) . Let's land it.

Yes, just tested it out, solves it.

PR #742, opened 2024-10-18
docs: move antora-playbook.yaml under doc_site

jozzsi:
Declutter root directory and simplify GitHub Action rules.

CC @ianw 
 
## Checklist


ianw:
I think this will need more changes to live here.  Inside it it the `supplemental-ui` will need to be updated and the start path to the docs module.

PR #744, opened 2024-10-18
docs: split guide.adoc into separate documents

jozzsi:
To make it easier to manage and read going forward..

CC @ianw


ianw:
Nice idea, I was thinking of this when I imported it :)  I think you want to update the `nav.adoc` to update the navigation on the left, but otherwise LGTM

ianw:
LGTM!

PR #749, opened 2024-10-18
docs: move a section from the README to the generated docs

jozzsi:
I think we should keep the README small and drive people to the generated docs site instead.

@ianw what do you think ?


ianw:
This LGTM 😊. I was thinking this is better in the generated docs, I’m glad others agree to 😂

PR #768, opened 2024-10-21
fix(lsinitrd): improve KERNEL_VERSION detection

jozzsi:
## Checklist

Fixes #

### DIFF
--- a/lsinitrd.sh
+++ b/lsinitrd.sh
@@ -96,7 +96,15 @@ while (($# > 0)); do
     shift
 done
 
-[[ $KERNEL_VERSION ]] || KERNEL_VERSION="$(uname -r)"
+if ! [[ $KERNEL_VERSION ]]; then
+    if type -P systemd-detect-virt &> /dev/null && systemd-detect-virt -c &> /dev/null; then
+        # shellcheck disable=SC2012
+        KERNEL_VERSION="$(cd /lib/modules && ls -1 | tail -1)"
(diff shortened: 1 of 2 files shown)


lsinitrd.sh
```diff
@@ -96,7 +96,16 @@ while (($# > 0)); do
     shift
 done
 
-[[ $KERNEL_VERSION ]] || KERNEL_VERSION="$(uname -r)"
+if ! [[ $KERNEL_VERSION ]]; then
+    if type -P systemd-detect-virt > /dev/null 2>&1 && systemd-detect-virt -c &> /dev/null; then
```
cgwalters, line 100: Why not consistently `&> /dev/null`?

lsinitrd.sh
```diff
@@ -96,7 +96,16 @@ while (($# > 0)); do
...
+        # shellcheck disable=SC2012
+        KERNEL_VERSION="$(cd /lib/modules && ls -1 | tail -1)"
+        # shellcheck disable=SC2012
+        KERNEL_VERSION || KERNEL_VERSION="$(cd /usr/lib/modules && ls -1 | tail -1)"
+    else
+        KERNEL_VERSION="$(uname -r)"
```
cgwalters, line 106: I think we want this fallback unconditionally - I can imagine cases today that are intentionally running dracut in a container but targeting the host kernel.

cgwalters:
Thanks for working on this!

PR #770, opened 2024-10-21
test: make documentation as part of test runs

jozzsi:
Do not disable documentation for basic test run. This will build manpages for each CI container (only once per CI run)..

## Checklist

Fixes #


ianw:
LGTM, probably generating the man pages on each platform is useful for package sanity checks, even if it is a bit repetitive.  `doc_site` has remained as a separate target, and isn't part of `all` which I think is right?

PR #777, opened 2024-10-21
ci: test both opensuse leap and tumbleweed

jozzsi:
For Fedora (Rawhide), Debian (sid), Ubuntu (rolling) we have both a stable and next-stable container available - see https://github.com/dracut-ng/dracut-ng/blob/main/.github/workflows/container-extra.yml#L36 .

This PR makes a similar change for openSUSE.

## Checklist

Fixes #


Conan-Kudo:
This doesn't make sense yet to do, since Leap 15 is never getting dracut-ng. I would prefer that the default be set to Tumbleweed and then we can add leap 16 once the appropriate artifacts are in place.

Conan-Kudo:
Also, unlike the other rolling distributions, Tumbleweed is a rolling release that is productionized for users, so we should be continuously testing it.

PR #782, opened 2024-10-22
docs: switch to asciidoctor for man generation, with fallback

ianw:
## Changes

This switches the man page generator to asciidoctor, with a fallback option to configure to use the existing asciidoc/docbook path.

The reason to do this is to consolidate the published site generation and man page generation around asciidoctor.  With that in place, we can use some asciidoctor features like extensions to better link the man pages together, or provide other tweaks to the documentation.

This switches the default generation to asciidoctor, but adds a --disable-asciidoctor switch as a temporary escape hatch for packagers.

The testing is converted to asciidoctor, with an additional test of the asciidoc toolchain just to avoid major regressions there.

Documentation is updated.

## Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -48,7 +48,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo apt -y update
-          sudo apt -y install asciidoc gcc libkmod-dev libsystemd-dev pkg-config
+          sudo apt -y install asciidoctor gcc libkmod-dev libsystemd-dev pkg-config
 
       - name: Autobuild
         uses: github/codeql-action/autobuild@v3
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -46,6 +46,20 @@ jobs:
                 uses: actions/checkout@v4
             -   name: "${{ matrix.container }} TEST-${{ matrix.test }}"
                 run: ./test/test-github.sh "TEST-${{ matrix.test }}" ${{ matrix.test }}
+
+    # Test deprecated asciidoc man generation path
+    doc_asciidoc:
+      name: asciidoc man generation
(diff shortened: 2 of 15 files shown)


ianw:
> Perhaps you can upload the container changes as a separate PR and keep both asciidoctor and asciidoc in the containers for now for the transition period.

That was https://github.com/dracut-ng/dracut-ng/pull/783 :)

I do think that making the default asciidoctor is probably best; hopefully all packagers need to do is change a build-dep, but if there is an issue they can add the flag and report an issue here and we can consider the path forward?

Conan-Kudo:
I would actually rather not use `asciidoctor` at all, as that involves pulling in a ruby stack too. Not to mention, `asciidoctor` is not available in RHEL but `asciidoc` is. So if we're going to use adoc format, let's stick exclusively with `asciidoc`.

ianw:
> I would actually rather not use `asciidoctor` at all, as that involves pulling in a ruby stack too. Not to mention, `asciidoctor` is not available in RHEL but `asciidoc` is. So if we're going to use adoc format, let's stick exclusively with `asciidoc`.

It doesn't feel like you want to discuss this, but I would say that `git` builds on fedora and el9 with `asciidoctor` (https://src.fedoraproject.org/rpms/git/blob/rawhide/f/git.spec#_24) and it is on all tested platforms here; so it I woudl say that shows it is not particularly onerous.

The reason to not leave the flag for `asciidoc` indefinitely was two-fold; firstly to consolidate the manual building with the generated-site to the same toolchain, just for overall ongoing sanity.

Secondly this enables things like custom macros to cross-link man pages that work with html and groff output.  This is modeled on the `git` documentation.  They do support a global `asciidoc` switch as well, but that leads to having extensions written three times in python, ruby and javascript to maintain compatibility between asciidoc, asciidoctor and antora.

PR #783, opened 2024-10-22
docs: also install asciidoctor in container

ianw:
## Changes

This is a pre-requirement change to enable building man pages with an asciidoctor toolchain in CI.

## Checklist


ianw:
See https://github.com/dracut-ng/dracut-ng/pull/782 which uses this package to make the man pages

ianw:
@LaszloGombos it seems that the run cancelled?  https://github.com/dracut-ng/dracut-ng/actions/runs/11471178983 

I think this means some of the containers aren't updated with the asciidoctor dependency

Conan-Kudo:
I would prefer this to be reverted, as we should not be using asciidoctor for man pages.

PR #808, opened 2024-10-26
fix(systemd): add vmw_vsock_virtio_transport kernel module

jozzsi:
NotifyAccess through sd-notify

As an example [multipathd.service](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/90multipath/multipathd.service) is using NotifyAccess .

Systemd PR: https://github.com/systemd/systemd/pull/27806

CC @mwilck 

## Checklist

Fixes #

### DIFF
--- a/modules.d/00systemd/module-setup.sh
+++ b/modules.d/00systemd/module-setup.sh
@@ -15,7 +15,7 @@ check() {
 }
 
 installkernel() {
-    hostonly='' instmods autofs4 ipv6 dmi-sysfs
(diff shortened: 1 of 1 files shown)


mwilck (changes requested):
AFAIK, `vmw_vsock` is a virtual socket module, the purpose of which is communication [between a VM and the hypervisor it's running on](https://github.com/torvalds/linux/blob/master/net/vmw_vsock/Kconfig). Worse, it's specific to VMware.

This isn't needed for `type=notify`, which simply uses a local AF_UNIX socket inside the VM.

So, NACK.

PR #818, opened 2024-10-28
fix(dracut): rework timeout for devices added via --mount and --add-device

aafeijoo-suse:
Currently, dracut adds a systemd dropin that sets an infinite timeout _only_ for the underlying persistent devices of each host device that must be active in the initrd. E.g.:

```
localhost:~ # dracut -f --stdlog 3 --install "/mnt" --mount "/dev/sda1 /mnt btrfs rw,relatime"
localhost:~ # lsinitrd | grep timeout.conf
-rw-r--r--   2 root     root            0 Oct 28 14:21 etc/systemd/system/dev-disk-by\x2duuid-0d1b24c2\x2df112\x2d48ef\x2d8442\x2d1001cffc92f0.device.d/timeout.conf
localhost:~ # ls -l /dev/disk/by-uuid/0d1b24c2-f112-48ef-8442-1001cffc92f0
lrwxrwxrwx 1 root root 10 Oct 28 11:43 /dev/disk/by-uuid/0d1b24c2-f112-48ef-8442-1001cffc92f0 -> ../../sda1
```

It can be verified in the emergency shell:

```
sh-5.2# systemctl show --property JobRunningTimeoutUSec /dev/disk/by-uuid/0d1b24c2-f112-48ef-8442-1001cffc92f0
JobRunningTimeoutUSec=infinity
sh-5.2# systemctl show --property JobRunningTimeoutUSec /dev/sda1
JobRunningTimeoutUSec=1min 30s
```

This can cause a problem if a user application expects to mount /dev/sda1 in the initrd (that's why it was explicitly added `--mount "/dev/sda1 ..."`), but for some reason the link is not created within 1min 30s.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -328,6 +328,16 @@ push_host_devs() {
     done
 }
 
+# Fills up user_devs stack variable and makes sure there are no duplicates
+push_user_devs() {
+    local _dev
+    for _dev in "$@"; do
+        [[ -z $_dev ]] && continue
+        [[ " ${user_devs[*]} " == *" $_dev "* ]] && return
+        user_devs+=("$_dev")
+    done
+}
+
 check_conf_file() {
     if grep -H -e '^[^#]*[+]=\("[^ ]\|.*[^ ]"\)' "$@"; then
         printf '\ndracut[W]: <key>+=" <values> ": <values> should have surrounding white spaces!\n' >&2
@@ -1698,7 +1708,7 @@ for line in "${fstab_lines[@]}"; do
             push_host_devs "$mp"
         done
     fi
-    push_host_devs "$dev"
+    push_user_devs "$dev"
     host_fs_types["$dev"]="$3"
 done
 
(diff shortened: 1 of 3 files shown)


dracut.sh
```diff
@@ -1685,7 +1696,7 @@ for line in "${fstab_lines[@]}"; do
             push_host_devs "$mp"
         done
     fi
-    push_host_devs "$dev"
+    push_user_devs "$dev"
     host_fs_types["$dev"]="$3"
```
mwilck, line 1700: If you push these to `user_devs` rather than `host_devs`, these devices won't be taken into account in `for_each_host_dev_and_slaves_all()`, which might lead to missing modules in hostonly mode. Or am I missing something?

modules.d/99base/module-setup.sh
```diff
@@ -105,6 +105,22 @@ install() {
...
+
+                for _dev in "${user_devs[@]}"; do
+
+                    case "$_dev" in
+                        /dev/?*) wait_for_dev "$_dev" 0 ;;
+                        *) ;;
+                    esac
+
+                    _pdev=$(get_persistent_dev "$_dev")
+                    [[ $_dev == "$_pdev" ]] && continue
```
mwilck, line 117: Why do you need to do this for both `_dev` and `_pdev`?

dracut.sh
aafeijoo-suse: No, you stand correct, I was missing that. Thanks.

Conan-Kudo (approved):

modules.d/99base/module-setup.sh
Conan-Kudo: While I can't see the bug, the logic seems reasonable to me.

PR #827, opened 2024-10-30
fix(dm): remove 59-persistent-storage-dm.rules

mwilck:
## Changes

59-persistent-storage-dm.rules has been obsolete for a long time. The blkid builtin is run from 13-dm.rules. The "dmsetup info" call in this file is actually harmful, because it overrides properties that have already been set in 10-dm.rules, in particular `DM_SUSPENDED`, to which it assigns a string like "Active" or "Suspended", whereas the stock DM rules and systemd rules expect the property to have the value "0", or "1", respectively.

In general, it seems wrong to read a property like `DM_SUSPENDED` multiple times while processing a uevent.

## Checklist


mwilck:
@prajnoha for information.

aafeijoo-suse (approved):
> (The dmsetup info call also retrieved some properties that 10-dm.rules does not set, like readonly, open, tables_loaded, and using_dev, but AFAICS these properties aren't used anywhere).

I couldn't find any use of these properties in any udev rules either, so it should be ok.

prajnoha:
> @prajnoha for information.

Oh, wow! Unbelievable this got unnoticed for so long, thanks for fixing that!

PR #829, opened 2024-10-30
fix(systemd-networkd): depend on net-lib

### LINKED ISSUE #825: dracut/systemd-networkd: no such file or directory: /sbin/netroot
**Describe the bug**
when using dracut module systemd-networkd a file not found error comes up on boot and in the logs

**Distribution used**
endeavour os /arch

**Dracut version**
dracut-ng 105

**Init system**
systemd

**To Reproduce**
enable systemd-networkd

**Expected behavior**
no not found error

**Additional context**
I think the networkd dracut module might need 45net-lib as a dependency because it's 45net-lib that copies the netroot.sh script to /sbin/netroot. But I'm not sure what else 45net-lib does so not sure.

jozzsi:
## Checklist

CC @BtbN @gmileka 

Fixes #825

### DIFF
--- a/modules.d/01systemd-networkd/module-setup.sh
+++ b/modules.d/01systemd-networkd/module-setup.sh
@@ -22,7 +22,7 @@ check() {
 depends() {
 
     # This module has external dependency on other module(s).
-    echo kernel-network-modules systemd-sysusers systemd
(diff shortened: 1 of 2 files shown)


BtbN:
Looks sensible to me. I guess it's usually pulled in by something else, and I/most people never noticed it depends on this.

Issue #833, opened 2024-10-31
--kernel-image ...  --uefi fails (sometimes) if the kernel image is a relative path


Nowa-Ammerlaan:
**Describe the bug**
This:
```
dracut --kver XXXX --kernel-image arch/x86/boot/bzImage --uefi --hostonly --fstab --kmoddir XXXX
```
fails, while this:
```
dracut --kver XXXX --kernel-image $(pwd)/arch/x86/boot/bzImage --uefi --hostonly --fstab --kmoddir XXXX
```
works.

The error is:
```
objcopy: 'arch/x86/boot/bzImage': No such file
sed: can't read /var/tmp/dracut.xxxx/uefi/kernel.sbat: no such file or directory
stat: cannot statx 'arch/x86/boot/bzImage': no such file or directory
/usr/bin/dracut: line 2551: offs + : syntax error: operand expected (error token is "+ ")
```
But the file does exist.

I thought I had reported this upstream, but it seems I forgot. The issue effects multiple downstream users in Gentoo, though I cannot reproduce it myself. I suspect it is somehow related to the file system.

See also: https://forums.gentoo.org/viewtopic-t-1165582.html

**Distribution used**
Gentoo

**Dracut version**
103+

**Init system**
not applicable

**To Reproduce**
Call `make install` in the kernel source directory. The kernel makefiles pass a relative path to `/sbin/installkernel` which passes on this relative path to the dracut kernel installation hooks. Or call dracut manually with a relative path for the kernel image. `--uefi` is required for the problem to occur.

**Expected behavior**

It works regardless of whether the path is relative or absolute.

**Additional context**

I cannot reproduce on an ext4 file system.

PR #834, opened 2024-10-31
dracut.sh: ensure abs path for objcopy args

Nowa-Ammerlaan:
Still needs verification that this actually fixes the reported problem.

## Changes

Add new function to convert relative paths to absolute paths, use this function for all variables that are involved in the creation of an UKI.

## Checklist

Fixes #

Closes: https://github.com/dracut-ng/dracut-ng/issues/833


Nowa-Ammerlaan:
This has now been confirmed to fix the problem (though I still don't understand why some users hit this while I cannot reproduce it on my end).

PR #842, opened 2024-10-31
docs: revert installing asciidoctor in CI container

jozzsi:
Based on feedback from @Conan-Kudo. See also the discussion at 

https://github.com/dracut-ng/dracut-ng/pull/783
https://github.com/dracut-ng/dracut-ng/pull/724

## Checklist

Fixes #

### DIFF
--- a/test/container/Dockerfile-alpine
+++ b/test/container/Dockerfile-alpine
@@ -19,7 +19,6 @@ ENV CC=clang
 
 RUN apk add --no-cache \
     asciidoc \
-    asciidoctor \
     bash \
     binutils \
     blkid \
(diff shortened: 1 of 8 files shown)


mwilck:
I didn't make any comment about this commit. I just opposed changing the default.

ianw:
I think we can probably have the links *and* maintain `asciidoc` with another extension.  It's just like 3 separate paths to build the documentation.

This kind of implies the documentation site can't be built on some platforms -- which is fine because it's published which is the way to access it and it really only needs to build in CI.  In a way, we should probably distribute `man` files so not everything needs a docbook toolchain too, it's just people are used to that :)

I'm not particularly opposed to reverting, I just thought it led to a simpler solution in the end

PR #875, opened 2024-11-05
fix(rngd): do not include the module if we can not start the service

jozzsi:
Some distributions do not have this service file present (e.g. openSUSE).

Follow-up to a952820 and https://github.com/dracut-ng/dracut-ng/pull/290

## Checklist

Fixes #


Nowa-Ammerlaan:
This causes a regression because `require_binaries` checks for executable permissions, but systemd service files do not have those permissions. Resolved in: https://github.com/dracut-ng/dracut-ng/pull/1207

PR #904, opened 2024-11-08
fix(dracut-functions.sh): check for modules in --kmoddir, not in --sysroot

Nowa-Ammerlaan:
Modules are installed from the directory specified by `--kmoddir`, but currently the `check_kernel_module()` function is checking for the module in `--sysroot/lib/modules`. This is notably not the same when kernels packages are being built inside some docker container. We should check for the modules existence in the directory we are actually going to install it from.

## Checklist

Fixes # Downstream discovered issue


cgwalters:
I don't know this code enough to give more than a superficial ACK sorry

PR #910, opened 2024-11-09
feat: remove --local option

jozzsi:
BREAKING CHANGE: this option was used only for the upstream testsuite, but not needed any more.

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -176,10 +176,6 @@ Creates initial ramdisk images for preloading modules
   --tmpdir [DIR]        Temporary directory to be used instead of default
                          ${TMPDIR:-/var/tmp}.
   -r, --sysroot [DIR]   Specify sysroot directory to collect files from.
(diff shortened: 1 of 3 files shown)


Nowa-Ammerlaan:
> LGTM
> 
> @Nowa-Ammerlaan  - I saw you're looking into running these tests downstream. 
> Would this PR be OK with you.

As long as make check still works and uses the to-be installed modules I'm happy.

Nowa-Ammerlaan:
> 
> > As long as make check still works and uses the to-be installed modules I'm happy.
> 
> I think now you would need to install dracut and modules first in some kind of sandbox, _before_ you could test them.
> build -> install -> test -> packing up.

That would require changing some things in the ebuild, a minor inconvenience, but doable.

Conan-Kudo:
Is it all that bad to retain the ability to run the test suite without installing it?

PR #914, opened 2024-11-09
ci(gentoo): optimize container

jozzsi:
- Combine some container layers.
- Allows to opt-out of optional packages.
- Splitting the package installation into two bigger layers allows the container runtime to download them in parallel.

Locally a minimal gentoo container can be built with the following command (without making local changes to the Dockerfile).

`docker build -f test/container/Dockerfile-gentoo --build-arg EXTRAPACKAGES=no1

## Checklist

Fixes #

### DIFF
--- a/test/container/Dockerfile-gentoo
+++ b/test/container/Dockerfile-gentoo
@@ -9,43 +9,64 @@
 # - rng-tools (to increase coverage)
 # - ntfs3g (to keep container small)
 
+ARG EXTRAPACKAGES=yes
+
 FROM docker.io/gentoo/portage:latest AS portage
 
 FROM docker.io/gentoo/stage3:systemd
 COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo
 
-# Speed-up using binpkgs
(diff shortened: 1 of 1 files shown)


Nowa-Ammerlaan:
> Splitting the package installation into two bigger layers allows the container runtime to download them in parallel.

This does the opposite, it will be slower because the dependency resolution is doubled and it also opens up the possibility of installing the same package twice if it is a dependency of both jobs. Note that FEATURES=parrelel-fetch and FEATURES=parallel-install are already enabled so fetching is already parallelized.

Otherwise the PR looks good to me

Nowa-Ammerlaan:
> Thanks @Nowa-Ammerlaan  . We might have misunderstood each other. I was referring to using the resulting container image (e.g. docker pull) and not building the image (with e.g. docker build). During a typical CI run the image gets pulled several times. Thanks for the review.
> 

Aah right, sorry, I thought you meant emerge was running in parallel. 

LGTM then, thanks!

Nowa-Ammerlaan (approved):

PR #917, opened 2024-11-09
feat(dmsquash-live): add support for rd.live.overlay.nouserconfirmprompt

jozzsi:
From https://github.com/microsoft/azurelinux/blob/3.0/SPECS/dracut/allow-liveos-overlay-no-user-confirmation-prompt.patch

It seems useful for all Linux distributions . 

CC @gmileka 

## Checklist

Fixes #

### DIFF
--- a/man/dracut.cmdline.7.adoc
+++ b/man/dracut.cmdline.7.adoc
@@ -1272,6 +1272,11 @@ of the base root filesystem and the persistent overlay, or
 root filesystem, and `/run/overlayfs` becomes the temporary, writable, upper
 directory overlay, to complete the bootable root filesystem.
 
+**rd.live.overlay.nouserconfirmprompt=**::
+Supresses the 'Using temporary overlay' blocking prompt that asks for a
+user confirmation before proceeding during boot time. This is to allow
+the boot process to continue to completion without user interation.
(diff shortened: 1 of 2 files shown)


FGrose:
The bothersome prompt is designed to only appear if a _specified_ persistent overlay is not available at boot time and only a temporary overlay is possible (alerting the user to the reality that root filesystem changes will not persist as requested).  If non-persistence is the intention—no persistent overlay should be specified.  (That feature may not be clearly documented or understood.)

Please verify that the command line is not requesting persistence.
If the unwelcome prompt appears without the specification of persistence, there is a bug in the code logic that should be fixed.  The proposed 'remedy' seems to be misguided, as it is a redundant way to request non-persistence.

Please clarify the perceived need.

Conan-Kudo:
Well, we need it because the prompt has been breaking the attempt to do the Modernize Live Media Fedora Change for a few cycles now. I'm happy someone beat me to getting to this. 😄

Conan-Kudo:
However, this patch doesn't seem to preserve the original attribution...

Conan-Kudo:
Replaced with #934 (which uses the original committer and patch metadata).

PR #919, opened 2024-11-09
revert(dracut-systemd): include systemd-cryptsetup module when needed

floppym:
This reverts commit e0e5424a7b5e387ccb70e47ffea5a59716bf7b76.

Depending on systemd-cryptsetup causes the dracut-systemd module to be excluded from the initrd when users have cryptsetup installed but systemd has been built with cryptsetup disabled.

This makes life harder for users with a default dracut config.

Bug: https://bugs.gentoo.org/943035

### DIFF
--- a/modules.d/98dracut-systemd/module-setup.sh
+++ b/modules.d/98dracut-systemd/module-setup.sh
@@ -12,17 +12,7 @@ check() {
 
 # called by dracut
 depends() {
-    local deps
-    deps="systemd-initrd systemd-ask-password"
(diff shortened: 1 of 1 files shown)


floppym:
cc @jozzsi

floppym:
Not sure why a few of the integration tests are failing. I see a timeout, so could just be flakey.

floppym:
Thanks, I have asked the reporter of the Gentoo bug to test it.

PR #921, opened 2024-11-09
fix(dracut-systemd): check systemd-cryptsetup before including

jozzsi:
On Gentoo, depending on systemd-cryptsetup causes the dracut-systemd module to be excluded from the initrd when users have cryptsetup installed but systemd has been built with cryptsetup disabled.

This makes life harder for Gentoo users with a default dracut config.

Bug: https://bugs.gentoo.org/943035

## Checklist

Fixes #


jayofdoom:
As the original submitter of the bug above; I've tested this patch and found that it resolves the issue. Thanks! +1 LGTM

PR #928, opened 2024-11-10
feat(iscsi): remove support for legacy iscsiroot= option

jozzsi:
Inspired by 

https://github.com/aafeijoo-suse/dracut-core/commit/a618102

CC @mwilck 

## Checklist

Fixes #


mwilck:
AFAICS, you've made exactly the same changes that @afeioo-suse did. Why don't you wait for him to submit a PR?

Anyway, I wouldn't even think about approving a PR about iSCSI that has the "I have tested it locally" checkbox not checked. Changes like this need intensive testing on different setups, with and without iBFT and iSCSI offload cards of different types.

PR #934, opened 2024-11-10
feat(dmsquash-live): add support for rd.live.overlay.nouserconfirmprompt

Conan-Kudo:
From: @gmileka

Dracut allows the creation of a LiveOS using a read-only squashfs and a read-write overlay on top.

If the read-write overlay is backed by a ramdisk, Dracut halts and prompts the user to confirm whether to continue or not.

This interaction during the boot process is not desired in all cases.

This change introduces a new flag (rd.live.overlay.nouserconfirmprompt) that when defined, it suppresses the prompt and allows the boot process to continue to completion without user interation.

There is no impact to existing configurations and their associated behavior. The new behavior only takes effect only when the new switch is explicitly defined by the image build as a kernel parameter.

### DIFF
--- a/man/dracut.cmdline.7.adoc
+++ b/man/dracut.cmdline.7.adoc
@@ -1272,6 +1272,11 @@ of the base root filesystem and the persistent overlay, or
 root filesystem, and `/run/overlayfs` becomes the temporary, writable, upper
 directory overlay, to complete the bootable root filesystem.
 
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
This patch will make it possible to avoid the hangs that occur when trying to run auto-persistence by default and there's nowhere to do persistence (which came up with https://github.com/weldr/lorax/pull/1360#issuecomment-2103493001 by @AdamWill).

PR #939, opened 2024-11-10
ci: bring back musl/openrc option for Gentoo

jozzsi:
CC @Nowa-Ammerlaan 

## Checklist

Fixes #


Nowa-Ammerlaan (approved):
I think we officially don't support musl anymore, but I don't really mind adding tests for it either. So LGTM.

PR #950, opened 2024-11-13
fix(dracut-init): add compatibility with Debian/Ubuntu for libdirs detection

### LINKED ISSUE #685: multipath does not work on Debian/Ubuntu based distributions
See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1078545 and https://github.com/dracut-ng/dracut-ng/pull/682 (for a failed attempt to resolve it).

See the warning seen on CI (e.g. test 01-BASIC) both for Debian and Ubuntu containers.

```
[FAILED] Failed to start multipathd.service…Mapper Multipath Device Controller.
```

`libdirs` is set to /lib64 /usr/lib64 /lib/x86_64-linux-gnu currently on the main branch for Debian and many of the multipath .so's are located under the `/lib/` directory instead.

jozzsi:
Make sure if /lib/ is included in libdirs when needed.                                                                                                                                                                                
                                                                                                                                                                                                                                      
This change is required for multipath dracut module to work properly on Debian/Ubuntu based distributions.    

libdir before this PR
- Ubuntu 
`/lib64 /usr/lib64 /lib/x86_64-linux-gnu`

- Fedora
`/lib64 /usr/lib64 /usr/lib64/iscsi`

- openSUSE
`/lib64 /usr/lib64`

libdir after this PR
- Ubuntu
`/lib64 /usr/lib64 /lib /usr/lib /lib/x86_64-linux-gnu`

- Fedora
`/lib64 /usr/lib64 /lib /usr/lib /usr/lib64/iscsi`

- openSUSE
`/lib64 /usr/lib64 /lib /usr/lib`

## Checklist

Fixes #685

CC @bdrung @mwilck @zeha

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -91,7 +91,10 @@ if ! [[ $libdirs ]]; then
         && [[ -d $dracutsysrootdir/lib64 ]]; then
         libdirs+=" /lib64"
         [[ -d $dracutsysrootdir/usr/lib64 ]] && libdirs+=" /usr/lib64"
-    else
+
+    fi
+
+    if [[ -d $dracutsysrootdir/lib ]]; then
         libdirs+=" /lib"
         [[ -d $dracutsysrootdir/usr/lib ]] && libdirs+=" /usr/lib"
     fi
--- a/test/TEST-41-FULL-SYSTEMD/test.sh
+++ b/test/TEST-41-FULL-SYSTEMD/test.sh
@@ -138,9 +138,8 @@ EOF
 
     grep -F -a -m 1 ID_FS_UUID "$TESTDIR"/marker.img > "$TESTDIR"/luks.uuid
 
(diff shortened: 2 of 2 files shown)


bdrung (commented):
The diff looks good. Can you document the reasoning for this change in the commit plus referencing https://github.com/dracut-ng/dracut-ng/issues/685?

bdrung (approved):
Oh, I forgot to respond. I agree that this PR is a step in the right direction and that we should merge it as is. Further improvements are better made in a follow-up PR.

On recent Ubuntu version `/lib` should be a symlink to `/usr/lib`. So only having `/lib/x86_64-linux-gnu` should not be a problem. But for not /usr-merged systems this could be an issue.

mwilck:
Late to the party, but this is not a clean solution IMO. 

I thought we agreed in #682 that the right solution would be to fix the way `inst_libdir_file()` parses the output of `ldd`, rather than hard-code an additional path element.

Issue #722, opened 2024-10-16
dracut-ng 060+ boots aarch64 laptops incorrectly


classabbyamp:
**Describe the bug**

on dracut-ng 060 and higher (tested up to 105), my aarch64 laptop (lenovo x13s) does not load many essential modules during the initramfs phase of boot, leading to many parts of the hardware misbehaving or not working, like wlan, battery, audio, and bluetooth.

bisection found:

> 3de4c7313260fb600507c9b87f780390b874c870 is the first bad commit

and by reverting that commit (and several later commits that build off of it: 8de0258d71dc5600d715d7534471e35b2b75c7be, 07e2c4926780b672922563a6ea0bf1bd4bcfcd9f, d71bec4aa444d92820e428c0629d0e75e268c815, 6500e95494175819b382acbac8eafcdf72fabd6d, 131822e26d76a3ce2028e9a545be2af066805629), the issue goes away.

**Distribution used**

Void Linux (kernel 6.11.3)

**Dracut version**

060 to 105

**Init system**

runit

**To Reproduce**

1. add the following configuration:
```bash
force_drivers+=" nvme phy_qcom_qmp_pcie "
# keyboard
force_drivers+=" i2c_hid_of i2c_qcom_geni "
# display
force_drivers+=" leds_qcom_lpg pwm_bl qrtr pmic_glink_altmode gpio_sbu_mux panel_edp msm "
force_drivers+=" phy_qcom_qmp_combo pinctrl_sc8280xp gpucc_sc8280xp dispcc_sc8280xp phy_qcom_edp "
```
2. generate an initramfs
3. reboot

**Expected behavior**

normal boot, with a working system

**Additional context**

- [verbose initramfs build log (dracut-ng 105 with revert)](https://github.com/user-attachments/files/17401710/dracut059.log)
- [verbose initramfs build log (dracut-ng 105)](https://github.com/user-attachments/files/17401723/dracut105.log)
- [dmesg (dracut-ng 105 with revert)](https://github.com/user-attachments/files/17401867/dmesg-broken.txt)
- [dmesg (dracut-ng 105)](https://github.com/user-attachments/files/17401869/dmesg-fixed.txt)

alpernebbi:
> does not load many essential modules during the initramfs phase of boot,

[...]

PR #952, opened 2024-11-13
fix(dracut-install): revert add fw_devlink suppliers as module dependencies

jozzsi:
Original discussion - https://github.com/dracutdevs/dracut/pull/2075

see also https://github.com/dracut-ng/dracut-ng/issues/316 and #722

Void Linux reverted these changes downstream: see https://raw.githubusercontent.com/void-linux/void-packages/refs/heads/master/srcpkgs/dracut/patches/revert-fw_devlink.patch .

There is a history of performance and regression concern around this area of code. Note that many distributions (except of course Void) are shipping this change - including Fedora, Gentoo, Arch and Debian. 

It would be useful to have some kind of test scenario where the impact of this change could be better understood. The project should not have a fragmentation between distributions in a core area like this without understanding the benefit and having a test case if possible.

The CI now has some basic arm64 test infrastructure (not just amd64). 

This reverts commit 8de0258d71dc5600d715d7534471e35b2b75c7be. 
This reverts commit 07e2c4926780b672922563a6ea0bf1bd4bcfcd9f. 
This reverts commit d71bec4aa444d92820e428c0629d0e75e268c815. 
This reverts commit 6500e95494175819b382acbac8eafcdf72fabd6d. 
This reverts commit 131822e26d76a3ce2028e9a545be2af066805629. 
This reverts commit 3de4c7313260fb600507c9b87f780390b874c870.

## Checklist

CC @classabbyamp @bdrung @alpernebbi

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -1616,7 +1616,7 @@ static int install_dependent_module(struct kmod_module *mod, int *err)
                 if (*err == 0) {
                         *err = kmod_module_get_weakdeps(mod, &modweak);
                         if (*err == 0)
-                                *err = install_dependent_modules(ctx, modweak, NULL);
+                                *err = install_dependent_modules(modweak);
                 }
 #endif
         } else {
@@ -1710,7 +1710,7 @@ static int install_module(struct kmod_module *mod)
         if (ret == 0) {
                 ret = kmod_module_get_weakdeps(mod, &modweak);
                 if (ret == 0)
-                        ret = install_dependent_modules(ctx, modweak, NULL);
+                        ret = install_dependent_modules(modweak);
         }
 #endif
 


AdamWill:
@nullr0ute @pcdubs do you guys know what this was for, specifically? is it needed on some SBCs?

alpernebbi:
Sorry for the short reply. I have been trying to test all this on whatever Chromebooks I have and will try to write in more detail later, but #316 and #322 are also very relevant.

nullr0ute:
> There is a history of performance and regression concern around this area of code. Note that many distributions (except of course Void) are shipping this change - including Fedora, Gentoo, Arch and Debian.

Can you link to some of the history you're referring to?

AdamWill:
I think it's the links in the first post. But it would sure be nice to know what this is *for*. I did try and follow it out on the Fedora/RH side, but couldn't find any specific references to exactly what these changes were meant to fix on what hardware.

nullr0ute:
> I think it's the links in the first post. But it would sure be nice to know what this is _for_. I did try and follow it out on the Fedora/RH side, but couldn't find any specific references to exactly what these changes were meant to fix on what hardware.

I'll dig into it later in the week when I get some spare cycles.

alpernebbi:
The problem this code solves is: there are dynamic kernel module runtime dependencies only expressed in boot firmware. For example, a display panel usually needs a backlight, but it might be `gpio_backlight` or `pwm_bl` or something else depending on hardware. The panel driver doesn't statically depend on either, IIUC because an error on the one it doesn't need shouldn't stop the panel from working with the one it needs. This problem also recurses, e.g. `pwm_bl` might need any one PWM driver, `gpio_backlight` might need any one GPIO driver, which might need any one pinctrl driver.

These dependencies are usually in the device-tree descriptions of each hardware. Most of these are included in the kernel tree, but the modern idea is that the firmware should carry it and provide it to the kernel. The information is then parsed by the kernel for it to probe modules in dependency order to avoid deferred probes. Otherwise, e.g. panel drivers can defer probing until the specific backlight for that hardware becomes available, which can defer until a specific PWM is available, and so on. And if we don't recursively include the dependencies for storage-related modules, we won't be able to mount the root and get them at all.

This mechanism is known as "fw_devlink", with "consumer" and "supplier" devices being linked in sysfs. The code here ends up recursively parsing those links to add the supplier modules when a consumer module is `dracut-install`ed. All those modules are necessary to probe the requested driver on that hardware. The newly added suppliers can also be consumers, so things get complicated.

However, parts of this code is flawed, which I've tried to fix in #332. For example on my RK3399 Chromebook, `pwm_bl` has `pwm-cros-ec` as a fw_link dependency. That PWM is a child of a `cros-ec` device, which has `cros-ec-spi` as a supplier, which has `spi_rockchip` as a supplier; all of these are missing when running `dracut-install -m panel-edp`.

So far I've been thinking of this just in terms of `dracut-install`. But for `dracut`'s own initramfs generation, it doesn't seem to add much in the end. Here is a tarball of data I've collected along with the scripts I wrote: [dracut-pr952-data.tar.gz](https://github.com/user-attachments/files/17918667/dracut-pr952-data.tar.gz). I've tested with v105-2 from Debian by invoking `dracut` with `DRACUT_INSTALL=` pointing to the different binaries, and by directly calling those binaries for each loaded module.

To sum up dracut's own initramfs, reverting the code as in this PR:

- `rockchip_saradc`, `industrialio-triggered-buffer` and `kfifo_buf` are no longer in the initramfs on some Rockchip systems. I don't know if this would break anything.
- `mtk_iommu` is no longer in the initramfs on MediaTek systems. This definitely breaks a lot of things including the display.

With #332 applied instead:

- I see no change in modules included on Rockchip systems. (In part thanks to my e69e413)
- `mtk_scp`, `mtk_scp_ipi`, `mtu3` (but not on hostonly-mode=strict, should've been included), `udc-core` (only on hostonly-mode=sloppy, weird) may be included in the initramfs on MediaTek systems.

I guess most of the dependencies are already included manually or with whatever patterns used for hostonly modules. We can extend those for new undeclared dependencies that pop up. (Or, rather, I should write a long email to kernel mailing lists about listing potential suppliers in `weakdep`s, but I'm slow with social stuff, sorry.)

So I think it's a question of what you are promising `dracut-install` does for downstreams. On Debian's initramfs-tools I have already implemented this fw_devlink handling for hardware-specific initramfs builds, so reverting shouldn't be a problem there.

(FWIW, I would appreciate if #332 could be merged before reverting it all, if only for software archaeology reasons)

classabbyamp:
the big difference between fedora and void would be systemd/systemd-udev vs eudev

alpernebbi:
> but still the regression does not seem to be either confirmed (reproduced) or denied.

I wouldn't call it regression. Use of `force_drivers` means it wasn't working before. It continues to not work, in a slightly different way, that needs to be accounted for differently. It's unfortunate that people have relied on this config (e.g. also in [Gentoo wiki](https://wiki.gentoo.org/wiki/Lenovo_ThinkPad_X13s#Dracut)) to workaround the breakage, instead of fixing it in upstream projects.

Relying on `add_drivers`, `force_drivers` etc. long-term is bad, they should only be used temporarily to diagnose what makes things work or break. Things working on Fedora 41 without these is some evidence that it is a configuration issue, but it might also be working due to the affected modules being built-in for example, if they are so. If it's a matter of missing modules in the initramfs, that should be fixed in dracut, perhaps in 90kernel-modules. If it's a module not probing automatically, probably in udev or kernel. If a module breaks because it's loaded before/after a specific module, then that should be fixed in kernel.

I think the root of the problem is, vendors on the ARM side have been carrying over a culture of embedded development practices to de-facto general purpose devices. Things like: developing kernel drivers as built-in only, or simply not testing them as modules after someone does that work, not declaring their dependencies properly, erroring out instead of deferring probe, relying on custom OS images with device-specific configuration to make things work instead of making generic OS images work, and so on. 

But anyway, solving what is wrong with that configuration needs someone who has access to the device and is motivated enough to investigate. I had asked for more info on that issue in order to help, but now I have much less free time.

classabbyamp:
sorry for not adding more info, i've been busy with many things, but it remains in my todo list

AdamWill:
@ausil has an x13s and was trying to run Fedora on it, I believe. If that's a system potentially affected by this whole thing, he might be able to run test scenarios to see how it affects that HW.

alpernebbi:
I had rebased #332 to avoid conflicts with 8de0258d71dc5 and 07e2c4926780b then. I rebased it again and tested on one Qualcomm SC7180 Chromebook now, should be good to merge. It may not solve every module dependency problem (we have some ["hidden dependency" handling in initramfs-tools](https://salsa.debian.org/kernel-team/initramfs-tools/-/blob/6d0c7921025a117b1748871e5c557cf335fe3831/hook-functions#L844-919) for example), but we can still choose to revert all of this even after merging it, so it's strictly an improvement in my eyes.

PR #954, opened 2024-11-13
docs: warn against the usage of -n, -o, -d

jozzsi:
These are common footguns. We should do a better job warning the users against these usually untested configurations.

CC @ianw 

## Checklist

Fixes #

### DIFF
--- a/man/dracut.8.adoc
+++ b/man/dracut.8.adoc
@@ -96,6 +96,10 @@ example:
 **-o, --omit** _<list of dracut modules>_::
     Omit a space-separated list of dracut modules. This parameter can be
     specified multiple times.
+    Warning: Avoid manually omitting dracut modules, as you may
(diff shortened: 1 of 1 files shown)


bdrung (approved):
LGTM. The description of `--omit` does not say why it is not recommended and what should be used instead.

man/dracut.8.adoc
```diff
@@ -96,6 +96,10 @@ example:
 **-o, --omit** _<list of dracut modules>_::
     Omit a space-separated list of dracut modules. This parameter can be
     specified multiple times.
+    Best to avoid omitting dracut modules as a user as you might be
```
bdrung, line 99: The phrase "Best to avoid omitting" made me stumble reading it - multiple times (due to the double negation). Could that maybe rephrased?

Issue #953, opened 2024-11-13
NFS test fails on s390x


bdrung:
**Describe the bug**

The dracut 105 made progress and TEST-16-DMSQUASH ran successful on Ubuntu plucky s390x (the upstream-dracut-live autopkgtest). The TEST-40-NBD (upstream-dracut-network autopkgtest) fails:

```
dracut[I]: Executing: /usr/bin/dracut -N -l -i /tmp/autopkgtest.Wm3NU8/autopkgtest_tmp/dracut-test.MiNImQ/overlay / -a "test rootfs-block debug kernel-modules network-legacy" -d "af_packet piix ide-gd_mod ata_piix ext4 sd_mod e1000 drbg" -i ./server.link /etc/systemd/network/01-server.link -i ./wait-if-server.sh /lib/dracut/hooks/pre-mount/99-wait-if-server.sh -f /tmp/autopkgtest.Wm3NU8/autopkgtest_tmp/dracut-test.MiNImQ/initramfs.server 6.11.0-8-generic
dracut[I]: 62bluetooth: Could not find any command of '/usr/lib/bluetooth/bluetoothd /usr/libexec/bluetooth/bluetoothd'!
dracut[I]: 95nfs: Could not find any command of 'rpcbind portmap'!
[...]
[...]

PR #985, opened 2024-11-17
test(MULTINIC): fix virtio regression

jozzsi:
Currently it is skipped on all containers, as none of the containers have `/etc/sysconfig/network-scripts` directory any more.

Since the tests do not run on CI, they are easily regress - seems https://github.com/dracut-ng/dracut-ng/issues/953 already caused a bit of regression.

This is just a bit of patch work - someone with more networking knowledge should take a look at how to stabilize all the networking tests on all distros.

CC @bdrung 

## Checklist

Fixes #

### DIFF
--- a/test/TEST-61-MULTINIC/test.sh
+++ b/test/TEST-61-MULTINIC/test.sh
@@ -75,8 +75,8 @@ client_test() {
         -net nic,macaddr=52:54:00:12:34:"$mac3",model=virtio \
         -netdev hubport,id=n1,hubid=1 \
         -netdev hubport,id=n2,hubid=2 \
-        -device virtio,netdev=n1,mac=52:54:00:12:34:98 \
-        -device virtio,netdev=n2,mac=52:54:00:12:34:99 \
+        -device virtio-net-pci,netdev=n1,mac=52:54:00:12:34:98 \
+        -device virtio-net-pci,netdev=n2,mac=52:54:00:12:34:99 \
         -append "$TEST_KERNEL_CMDLINE $cmdline ro init=/sbin/init systemd.log_target=console" \
         -initrd "$TESTDIR"/initramfs.testing || return 1
 


test/TEST-61-MULTINIC/test.sh
```diff
@@ -15,7 +15,7 @@ test_check() {
         return 1
     fi
 
-    test -d /etc/sysconfig/network-scripts
+    command -v exportfs &> /dev/null
```
bdrung, line 18: I was working on making this test run against the Ubuntu package and had this exact change staged locally. :laughing:

test/TEST-61-MULTINIC/test.sh
```diff
@@ -80,8 +80,8 @@ client_test() {
         -net nic,macaddr=52:54:00:12:34:"$mac3",model=virtio \
         -netdev hubport,id=n1,hubid=1 \
         -netdev hubport,id=n2,hubid=2 \
-        -device virtio,netdev=n1,mac=52:54:00:12:34:98 \
-        -device virtio,netdev=n2,mac=52:54:00:12:34:99 \
+        -device virtio-net-pci,netdev=n1,mac=52:54:00:12:34:98 \
+        -device virtio-net-pci,netdev=n2,mac=52:54:00:12:34:99 \
```
bdrung, line 84: Can you put this change into a separate commit and reference the commit that broke it?

test/TEST-62-BONDBRIDGEVLAN/test.sh
```diff
@@ -16,8 +16,8 @@ test_check() {
         echo "Test needs dhclient for server networking... Skipping"
         return 1
     fi
-    test -d /etc/sysconfig/network-scripts
-    module_check ifcfg
+
+    command -v exportfs &> /dev/null
```
bdrung, line 20: Can you put this fix into a separate commit?

bdrung (commented):
These changes look like a step in the right direction. Test 61 does still not work: Starting the DHCP/NFS server fails. "watchdog: watchdog0: watchdog did not stop" and then the kernel stays in a "Rebooting in 1 seconds.." loop. I am investigating...

bdrung:
Found the culprit in the server log:
```
/usr/bin/env: 'bash': No such file or directory
[    8.026059] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00007f00
```
I'll fix that one.

bdrung:
The strange thing is: bash is included in the server initrd:

```
3cpio -t /var/tmp/dracut-test.y6wMAi/initramfs.server | grep bash
usr/bin/bash
```

test/TEST-61-MULTINIC/test.sh
bdrung: The "test(MULTINIC): virtio regression" commit does not introduce a regression, but fixes one. How about adding a "fix" there?

bdrung:
Found the issue: `/usr/bin/env` is not inside the server initrd.

.github/workflows/integration.yml
```diff
@@ -196,8 +196,6 @@ jobs:
                     - ubuntu
                 test:
                     - "60"
-                    - "61"
-                    - "62"
```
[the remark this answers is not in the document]
bdrung: Ah, now I understand it. So I suggest to defer this commit until all other bits landed to make the CI green with these tests - unless we fail to make these test succeed in a reasonable amount of time.

bdrung:
@jozzsi can you put the virtio fix in a separate merge request to get this merged?

PR #988, opened 2024-11-18
ci(debian): install dracut-network

jozzsi:
This should install more downstream dependencies and brings upstream closer to downstream.

CC @bdrung 

## Checklist

Fixes #


bdrung:
It depends what you want to test. The downstream test install only the needed packages for each test (e.g. dracut-network is only installed for network-related tests). You could decide that you want to install all packages, but please exclude the dracut-config-* packages.

PR #991, opened 2024-11-19
docs: warn against the usage of omit, drivers, dracutmodules

jozzsi:
These are common footguns. We should do a better job warning the users against these usually untested configurations.

Follow-up to eb1ae6c

## Checklist

Fixes #


bdrung (approved):
You might want to reference commit eb1ae6c4f83663bfe87c835feaa649f272d780bb

PR #993, opened 2024-11-19
fix(iscsi): include /usr/lib/open-iscsi/startup-checks.sh

bdrung:
## Changes

On Debian/Ubuntu iscsid.service (from open-iscsi 2.1.10-1) fails to start in the initrd, because its `ExecStartPre` script is not included in the initrd:

```
$ grep ExecStartPre /usr/lib/systemd/system/iscsid.service
ExecStartPre=/usr/lib/open-iscsi/startup-checks.sh
```

The TEST-70-ISCSI test will succeed on Ubuntu 25.04 (plucky) with that patch plus `iscsiuio.socket` included in open-iscsi (see https://launchpad.net/bugs/2072484).

## Checklist


mwilck (changes requested):
Nack, this is Debian-only.

If you need it on Debian, make it optional. 

In general, changes like this should *never* been made without involving the maintainers of the respective upstream projects (here, open-iscsi, of which I am not a maintainer).

modules.d/95iscsi/module-setup.sh
```diff
@@ -198,6 +198,9 @@ install() {
         "$systemdsystemunitdir"/iscsiuio.socket \
         "$systemdsystemunitdir"/sockets.target.wants/iscsid.socket \
         "$systemdsystemunitdir"/sockets.target.wants/iscsiuio.socket
+    if grep -q '^ExecStartPre=/usr/lib/open-iscsi/startup-checks.sh$' "$systemdsystemunitdir/iscsid.service"; then
+        inst_simple /usr/lib/open-iscsi/startup-checks.sh
```
mwilck, line 202: It might be sufficient to just use `inst_simple -o` (`-o`: "If SOURCE does not exist, do not fail").

bdrung: The current approach is more strict. It will fail in case /usr/lib/open-iscsi/startup-checks.sh was found in the service file but fails to install.

bdrung:
I would have expected the CI to complain for the distributions that do not have `/usr/lib/open-iscsi/startup-checks.sh`. I made it conditional and extended the commit description:
> The upstream open-iscsi project ships an iscsid.service [1] that does not have an ExecStartPre entry, but Debian/Ubuntu uses their own service file [2].
>
> [1] https://github.com/open-iscsi/open-iscsi/blob/master/etc/systemd/iscsid.service.template
> [2] https://salsa.debian.org/linux-blocks-team/open-iscsi/-/blob/master/debian/iscsid.service

modules.d/95iscsi/module-setup.sh
mwilck: ok, fine with me.

mwilck (approved):
LGTM

bdrung:
> Lastly, perhaps we should consider having an `inst_service` function that takes the service file name, parses the service file and installs all of its dependencies. If we would have a function like that we would not even having this discussion as only the service file name would be in the dracut module file. At some point one of the test cases had a similar approach - see https://github.com/dracutdevs/dracut/blob/master/test/TEST-04-FULL-SYSTEMD/test.sh#L80

Here you go: #1031

PR #995, opened 2024-11-19
ci: switch from ubuntu:rolling to ubuntu:devel

jozzsi:
The goal is to test with latest Ubuntu before a dracut release.

When I enabled ubuntu:rolling I simply missed that :devel is the latest version not :rolling.

CC @bdrung 

## Checklist

Fixes #


bdrung (approved):
I agree that we should test ubuntu:devel since that is the target for new dracut upstream releases. Maybe we want to keep ubuntu:rolling to also test the latest release.

bdrung:
ubuntu:devel is comparable with Debian testing regarding stability (both are "protected" by autopkgtest). Debian unstable is comparable with ubuntu:devel where you enable -proposed.

PR #1002, opened 2024-11-19
test: enable TEST 61 on fedora

jozzsi:
This change will enable us to drive at least with one eye open :-)

## Checklist

Fixes #


bdrung:
The change looks good, but please split the change into atomic commits.

PR #1020, opened 2024-11-22
feat(nvmf): enable other shells (dash) not just bash

jozzsi:
I do not see anywhere where bash would be required for the `nvmf` module.

@bdrung @mwilck what do you think ? Am I missing something ? 

## Checklist

Fixes #

### DIFF
--- a/modules.d/95nvmf/module-setup.sh
+++ b/modules.d/95nvmf/module-setup.sh
@@ -63,7 +63,7 @@ check() {
 
 # called by dracut
 depends() {
-    echo bash rootfs-block network
+    echo rootfs-block network
     return 0
 }
 


bdrung:
Yes, I see no reason for hardcoding bash. Commit 2f03d69f9dfdc0c905c4d9cb69ee93a5256123cb introduced bash. All shell scripts in the module (excluding module-setup.sh) do not use bashisms (checked with shellcheck).

Can you add `Fixes: 2f03d69f9df ("95nvmf: add module for NVMe-oF")` to the commit and mention that the module scripts have no bashisms?

mwilck (approved):

bdrung (approved):

PR #1021, opened 2024-11-22
chore(iscsi): remove duplicate inst_multiple call

bdrung:
## Changes

The iscsid and iscsiuio services are installed by an `inst_multiple` two times. Remove the already installed services from the second call.

## Checklist


bdrung:
Oh, right.

PR #1028, opened 2024-11-22
fix(base): fallback when shell-interpreter is not included

### LINKED ISSUE #648: shell-interpreter assumes that /bin/sh is always a symbolic link
**Describe the bug**
[shell-interpreter](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/00shell-interpreter/module-setup.sh#L17) assumes that /bin/sh is always a symbolic link . In some Linux distributions /bin/sh is the actual shell binary and not a symbolic link.

jozzsi:
Follow-up to e139edb

This would also resolve #648

## Checklist

Fixes #

### DIFF
--- a/modules.d/99base/module-setup.sh
+++ b/modules.d/99base/module-setup.sh
@@ -19,6 +19,9 @@ install() {
     ln -s dracut-util "${initdir}/usr/bin/dracut-getarg"
     ln -s dracut-util "${initdir}/usr/bin/dracut-getargs"
 
+    # fallback when shell-interpreter is not included
+    [ ! -e "${initdir}/bin/sh" ] && inst_simple "${initdir}/bin/sh" "/bin/sh"
+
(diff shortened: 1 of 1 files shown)


modules.d/99base/module-setup.sh
```diff
@@ -19,6 +19,9 @@ install() {
     ln -s dracut-util "${initdir}/usr/bin/dracut-getarg"
     ln -s dracut-util "${initdir}/usr/bin/dracut-getargs"
 
+    # fallback when shell-interpreter is not included
+    [ ! -e "${initdir}/bin/sh" ] && inst_simple "${initdir}/bin/sh" "/bin/sh"
```
bdrung, line 23: Shouldn't that live in shell-interpreter?

bdrung: Thanks. That's a valid reason. Can we have this explanation in the commit message? Then I will approve this merge request.

PR #1029, opened 2024-11-22
fix(dracut-functions): do not assume btrfs is installed

jozzsi:
Fix the following warning

```
dracut-functions.sh: line 816: btrfs: command not found
```

## Checklist

Fixes #

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -813,12 +813,14 @@ lvm_internal_dev() {
 
 btrfs_devs() {
     local _mp="$1"
-    btrfs device usage "$_mp" \
-        | while read -r _dev _; do
-            str_starts "$_dev" "/" || continue
-            _dev=${_dev%,}
-            printf -- "%s\n" "$_dev"
-        done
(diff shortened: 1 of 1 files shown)


bdrung (commented):
Is that the right approach? Shouldn't we print a warning in that case?

bdrung:
> I can not think of a condition that would warrant the warning. Currently we call btrfs even if there is no btrfs partition on the system and btrfs dracut module is not included. I certainly do not want to falsely alarm users in this condition.

Really? I see 3 `btrfs_devs` calls in the code. All three are checking for btrfs before:

```sh
    if [[ $3 == btrfs ]]; then
[...]
            for i in $(btrfs_devs "$mp"); do
```

```sh
        if [[ $(find_mp_fstype "$mp") == btrfs ]]; then
            for i in $(btrfs_devs "$mp"); do
```

```sh
            if [[ $_t == btrfs ]]; then
                for i in $(btrfs_devs "$_m"); do
```

So I fail to see how you can hit this code path without having a btrfs partition.

PR #1031, opened 2024-11-22
feat: add inst_service for installing service files

bdrung:
## Changes

Systemd service file can define commands in `ExecStart`, `ExecStartPre`, `ExecStartPost`, `ExecStop`, etc.

Add an `inst_service` helper function to install service files together with all the referenced executables. This allows dropping the workaround from commit 7fe7fa943748 for iSCSI on Debian/Ubuntu.

Needed before merging:
* check if the initrd content is still the same
* add some test cases. Useful real world examples:
```
/usr/lib/systemd/system/apt-daily.service:ExecStartPre=-/usr/lib/apt/apt-helper wait-online
/usr/lib/systemd/system/initrd-parse-etc.service:ExecStart=@/usr/lib/systemd/system-generators/systemd-fstab-generator systemd-sysroot-fstab-check
/usr/lib/systemd/system/man-db.service:ExecStart=+/usr/bin/install -d -o man -g man -m 0755 /var/cache/man
/usr/lib/systemd/system/systemd-networkd.service:ExecStart=!!/usr/lib/systemd/systemd-networkd
```

## Checklist

Fixes #

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -368,6 +368,31 @@ inst_script() {
     fi
 }
 
+# Install the specified systemd service files and all referenced programs in
+# ExecStart*, ExecStop*, ExecReload*.
+inst_service() {
+    local _optional
+    local -a needed_files service_files
+
+    if [[ $1 == "-o" ]]; then
+        _optional="-o"
+        shift
+        for service_file in "$@"; do
+            if test -e "$service_file"; then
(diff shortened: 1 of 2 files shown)


devkontrol:
For additional inspiration - https://github.com/archlinux/mkinitcpio/blob/master/install/systemd

PR #1033, opened 2024-11-22
ci: release Dracut quarterly

bdrung:
## Changes

Release Dracut quarterly to synchronize the Dracut release with the Ubuntu release. Ubuntu XX.04 has its feature freeze mid to end of February. Ubuntu XX.10 has its feature freeze mid to end of August. So release Dracut on the beginning of February and August to have some time left to get it landed in the distribution.

To get into sync, we should do a release on Dec-1. Then do the follow-up release on Feb-1.

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,8 +3,8 @@ name: Release
 
 on:  # yamllint disable-line rule:truthy
   schedule:
-    ## Schedule the job to run on Apr-1, Aug-1, Dec-1
-    - cron: '0 0 1 APR,AUG,DEC *'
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
We need this to be shifted back a few weeks so it's also aligned with Fedora schedule.

Conan-Kudo:
We need this to be shifted back a few weeks so it's also aligned with Fedora schedule. Can we shift this to January (and thus follow regular quarters) so that releases can be integrated into Fedora and openSUSE Leap?

Conan-Kudo:
Fedora's soft deadline is February 4, the hard deadline is February 18. openSUSE's is sometime near the end of January, @lkocman would know more exact dates.

PR #1045, opened 2024-12-05
ci(alpine): dhclient is no longer packaged in alpine

jozzsi:
Resolves the following error:

```
dhclient (no such package):
```

## Checklist

Fixes #


test/container/Dockerfile-alpine
```diff
@@ -26,10 +26,10 @@ ARG DISTRIBUTION
 # Packages to pass the BASIC test
 RUN apk add --no-cache \
     asciidoc \
-    dracut \
```
bdrung, line 29: Why do you remove dracut?

PR #1063, opened 2024-12-24
fix(dracut-systemd): suppress gpt-auto-root devices if not needed

### LINKED ISSUE #1062: Dracut hostonly image unable to boot if `root=` is not found on kernel command line due to systemd-gpt-auto-generator
**Describe the bug**

When using `dracut --hostonly --hostonly-cmdline` to build initramfs/initrd image it's possible to boot using generated image into system without a need to specify `root=` parameter on kernel command line. However if `root=` is not specified on kernel command line, `systemd-gpt-auto-generator` also comes into play. As a result systemd configuration generated by dracut conflicts with systemd generation generated by systemd-gpt-auto-generator and lead to a broken boot.

**Distribution used** / **Dracut version** 

Fedora 41 (dracut 103-3.fc41)
Ubuntu 24.10 (103-1ubuntu3)
Debian testing (105-2)

**Init system**

systemd

**To Reproduce**

1. Install Fedora Linux using LVM+encryption setup
2. Generate dracut image using --hostonly --hostonly-cmdline
3. Optionally use systemd-cryptenroll to enroll tpm2 based key to avoid typing LUKS key every time
4. Configure bootloader entry that uses generated hostonly image
5. Ensure there are no root= or systemd.gpt_auto=0 parameters passed directly to kernel by bootloader.
6. Empty cmdline/options should actually work fine with hostonly generated image
7. Reboot

**Expected behavior**

System boots using `root=` value embedded into generated initramfs/initrd image

**Observed behavior**

System is unable to boot

**Additional context**

The issue is not new and was reported and discussed previously, in particular:

https://github.com/dracutdevs/dracut/pull/1722
https://github.com/dracut-ng/dracut-ng/issues/1038

Here is relevant systemd issue:

https://github.com/systemd/systemd/issues/35628

Also there is a similar issue opened by me:

https://github.com/dracut-ng/dracut-ng/issues/1052

Despite issues look similar they have different scope/focus.
That issue is focused on broken emergency shell, whereas this issue is focused on broken boot.

RomanValov:
The `systemd-gpt-auto-generator` comes into play if there is no `root=` on kernel command line defined. However dracut-generated image may embed `root=` inside the image (`/etc/cmdline`, `/etc/cmdline.d`). Then dracut's `rootfs-generator` attempts to boot using "embedded" `root=`. As a result both generators fall into conflict and cause a broken boot.

## Changes

The patch suppresses gpt-auto-root devices if `root=` is known but is not found on kernel command line.

## Checklist

Fixes #1062

### DIFF
--- a/modules.d/98dracut-systemd/rootfs-generator.sh
+++ b/modules.d/98dracut-systemd/rootfs-generator.sh
@@ -50,6 +50,12 @@ generator_mount_rootfs() {
         [ -d "$GENERATOR_DIR"/initrd-root-fs.target.requires ] || mkdir -p "$GENERATOR_DIR"/initrd-root-fs.target.requires
         ln -s ../sysroot.mount "$GENERATOR_DIR"/initrd-root-fs.target.requires/sysroot.mount
     fi
+    if ! [ -L "$GENERATOR_DIR"/dev-gpt\\x2dauto\\x2droot.device ]; then
+        ln -s /dev/null "$GENERATOR_DIR"/dev-gpt\\x2dauto\\x2droot.device
+    fi
+    if ! [ -L "$GENERATOR_DIR"/dev-gpt\\x2dauto\\x2droot\\x2dluks.device ]; then
+        ln -s /dev/null "$GENERATOR_DIR"/dev-gpt\\x2dauto\\x2droot\\x2dluks.device
+    fi
 }
 
 generator_fsck_after_pre_mount() {


savyajha:
I think the discussion has passed beyond what I can comfortably offer to help with (with the limits of my knowledge). I also do not use dracut on any of my current systems so may not be of much help now.

RomanValov:
> @RomanValov , instead of suppressing systemd-gpt-auto-generator, what do you think of suppressing dracut's generator ?

The systemd-gpt-auto-generator is tailed to a particular use cases (possibly widely used though).
Same time it has a lot of restrictions and requirements to be usable, in particular:

* It does not work with LVM devices
* It does not (correctly) work with multiboot drives
* It does not support legacy non-UEFI / non-GPT / non-BLI systems

Personally I've bumped into mutliboot drive scenario

Sorry for the delayed follow up

RomanValov:
Personally I wouldn't consider this to be a fix but actually a workaround. Maybe the real fix should be on systemd' behalf with the check of root device from sources other then kernel parameters. I.e. systemd-fstab-generator does this by looking initrd's /etc/fstab for `/sysroot` mount

Maybe it also makes sense to adopt such way of detecting root (and generating sysroot mount units) instead of using dracut's rootfs-generator. For me it looks like `/sysroot` entry processing is newer way of generating sysroot mount units (see systemd/systemd#3551) and dracut's rootfs-generator was written when there were no such thing. It does not directly solve the issue in question but if needed I can create a ticket for this

Issue #1057, opened 2024-12-17
TEST-26-ENC-RAID-LVM asks for luks password when using --hostonly mode


bdrung:
**Describe the bug**

The dracut 105-2ubuntu4 upstream-dracut-core autopkgtest hangs. The TEST-26-ENC-RAID-LVM asks for luks password:

```
TEST: root filesystem on an encrypted LVM PV on a RAID-5
[...]
[ 0.000000] Command line: rd.retry=10 rd.info rd.shell=0 panic=1 oops=panic softlockup_panic=1 systemd.crash_reboot console=ttyS0,115200n81 root=/dev/dracut/root rd.auto
[...]
Starting systemd-cryptsetup@luks\x…309a-5a90-4ae8-acdc-ee47b42a6ee8...

Please enter passphrase for disk luks-b51f309a-5a90-4ae8-acdc-ee47b42a6ee8: (press TAB for no echo)
```

Full log: https://autopkgtest.ubuntu.com/results/autopkgtest-plucky/plucky/amd64/d/dracut/20241205_232205_fdba1@/log.gz

The delta between dracut 105-2ubuntu3 (with successful autopkgtest) and 105-2ubuntu4 is small: https://launchpadlibrarian.net/760654984/dracut_105-2ubuntu3_105-2ubuntu4.diff.gz

Adding --no-hostonly to the test_dracut call will make the test green again.

**Distribution used**
Ubuntu

**Dracut version**
105 and main

**Init system**
systemd

bdrung:
I reverted commit ac9cbe0d1c1ec4795dada3b802727dca431bbd88 and tested again, but it still fails the same way.

Once dracut 105-2ubuntu4 migrates from plucky-proposed to plucky the configuration will be covered by our CI.

PR #1066, opened 2024-12-26
feat(crypt): remove empty /etc/crypttab

jozzsi:
crypt module should not install an empty /etc/crypttab in hostonly mode.

Some distributions ship with an /etc/crypttab without entries (just with comments). In those distributions installing /etc/crypttab as part of --install command line option in hostonly mode would not work as expected.

## Checklist

Fixes: https://github.com/dracut-ng/dracut-ng/issues/1057

### DIFF
--- a/modules.d/90crypt/module-setup.sh
+++ b/modules.d/90crypt/module-setup.sh
@@ -136,7 +136,13 @@ install() {
                 done
             fi
         done < "$dracutsysrootdir"/etc/crypttab > "$initdir"/etc/crypttab
-        mark_hostonly /etc/crypttab
+
+        # Remove empty /etc/crypttab to allow creating it later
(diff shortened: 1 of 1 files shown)


modules.d/90crypt/module-setup.sh
```diff
@@ -136,7 +136,13 @@ install() {
                 done
             fi
         done < "$dracutsysrootdir"/etc/crypttab > "$initdir"/etc/crypttab
-        mark_hostonly /etc/crypttab
+
+        # Remove empty /etc/crypttab to allow creating it later
+        if [ -s "$initdir"/etc/crypttab ]; then
+            mark_hostonly /etc/crypttab
+        else
+            rm -rf "$initdir"/etc/crypttab
```
bdrung, line 144: Using `-r` (recursive) on removing a file looks odd to me.

PR #1097, opened 2025-01-04
feat(systemd-sysusers): run systemd-sysusers as part of the build process

jozzsi:
This PR makes the boot process faster and the generated initrd smaller.

## Checklist

Fixes #


bdrung:
> This PR makes the boot process faster and the generated initrd smaller.

Can you back up that with some numbers?

bdrung:
The change looks good. I have no concerns. @jozzsi can you add your explained reason to the git commit description?

marc-hb:
Regression #1242 bisected to  this commit.

marc-hb:
@jozzsi can you please take a look at regression #1242 caused by this change?

PR #1117, opened 2025-01-12
fix(systemd-sysuser): add support for Gentoo

### LINKED ISSUE #274: tpm2-tss - unknown group tss reported
**Describe the bug**

From downstream Bug: https://bugs.gentoo.org/925524

After building an initrd with Dracut with systemd (which pulls in tpm2-tss module), prior to switching sysroot, these non-fatal errors will appear:

/usr/lib/tmpfiles.d/tpm2-tss-fapi.conf:2: Failed to resolve user 'tss': No such process
/usr/lib/tmpfiles.d/tpm2-tss-fapi.conf:4: Failed to resolve user 'tss': No such process
/usr/lib/tmpfiles.d/tpm2-tss-fapi.conf:6: Failed to resolve group 'tss': No such process
/usr/lib/tmpfiles.d/tpm2-tss-fapi.conf:7: Failed to resolve group 'tss': No such process
/usr/lib/udev/rules.d/60-tpm-udev.rules:3 Unknown user 'tss', ignoring.
/usr/lib/udev/rules.d/60-tpm-udev.rules:4 Unknown group 'tss', ignoring.

The tpm2-tss module does attempt to create at least the tss user via a file in sysusers.d:

https://github.com/dracutdevs/dracut/blob/master/modules.d/91tpm2-tss/module-setup.sh#L35C21-L35C35

The problem is this file tpm2-tss.conf does not exist on Gentoo machines. It is normally installed by tpm2-tss but we do not because we have acct-user/tss and the group packages.

I think the best solution to send a patch upstream so that the acct-user/acct-group files are considered as well as the standard file.

**Distribution used**
Gentoo Linux

**Dracut version**
59 and up

**Init system**
Systemd

**To Reproduce**
Build a dracut image on a Gentoo Linux system and include the tpm2-tss module

**Expected behavior**
The required user exists in the initramfs

**Additional context**
See also: https://github.com/dracutdevs/dracut/issues/1700

jozzsi:
## Changes

systemd-sysuser dracut module uses inst_sysusers function to install configuration files. These configuration files are named differently in Gentoo, which impacts upstream Gentoo CI as well.

Partially fixes https://github.com/dracut-ng/dracut-ng/issues/274

Based on Gentoo downstream patch https://github.com/gentoo/gentoo/blob/master/sys-kernel/dracut/files/dracut-103-acct-user-group-gentoo.patch by Nowa Ammerlaan <nowa@gentoo.org>

This PR is required to restore green CI.

CC @Nowa-Ammerlaan 

## Checklist

Fixes #274

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -636,10 +636,10 @@ inst_libdir_file() {
 
 # install sysusers files
 inst_sysusers() {
-    inst_multiple -o "$sysusers/$*"
+    inst_multiple -o "$sysusers/$*" "$sysusers/acct-*-$*"
 
     if [[ $hostonly ]]; then
-        inst_multiple -H -o "$sysusersconfdir/$*"
+        inst_multiple -H -o "$sysusersconfdir/$*" "$sysusers/acct-*-$*"
     fi
 }
 


Nowa-Ammerlaan (approved):
I'm not opposed to this change, but please note that this is not guaranteed to fix everything because the user/group files are named after the user/group, not the owning package. See for example `tpm2-tss` which on our end is simply `tss`.

Conan-Kudo (approved):

PR #1121, opened 2025-01-15
Fixes for booting from iSCSI offload with bnx2i

mwilck:
## Changes

Booting from iSCSI offload is a scenario that is rarely tested because it requires special hardware. Moreover, different offload devices behave in slightly different ways.

Currently dracut handles booting from such devices rather poorly. It uses code that has been written with iSCSI over TCP in mind. But many iSCSI offload cards, in particular bnx2i, do not require any regular Linux network interfaces to be configured. The configuration is usually done entirly in firmware. (Configuration via iscsid and iscsiuio is also possible, but hardly ever used in boot-from-iSCSI setups).

Trying to bring up the network in the initramfs in a setup like this will in the best case delay booting unnecessarily, and in the worst case break booting entirely. The latter can happen if the delays sum up such that they  cause a device timeout, or if certain network setup operations jeopardize the setup of the iSCSI offload connection (changing the MTU on a related network device is known to have this effect, for example).

This patch set changes the behavior for booting over bnx2i iSCSI offload cards such that network bringup is skipped entirely if possible. iSCSI connection setup is attempted early, even before any network interface has been brought up. For the offload case, this is likely to succeed, and if it fails, the "normal" configuration path is taken later anyway.

According to our testing in the SUSE Labs, both reliability and speed of booting from bnx2i, with and without multipath, has been improved significantly by this patch set. We have also done tests with regular iSCSI over TCP/IP and seen no regressions.

So far I've applied the changes to bnx2i only. It's likely that other offload cards like cnic would also benefit, but because we're lacking the hardware for testing, we couldn't verify that. Testing with such hardware (after modifying the first patch accordingly) would be highly welcome.

## Checklist

Adding @aafeijoo-suse, @gonzoleeman for information.


aafeijoo-suse:
Thanks for working on this Martin.

> Trying to bring up the network in the initramfs in a setup like this will in the best case delay booting unnecessarily, and in the worst case break booting entirely.

So, imagine a pure systemd scheme (without all the dracut hackery around), does this mean that there are some iSCSI offload cards that do not need to be configured `After=network-online.target` and do not `Require=network-online.target`, i.e., `iscsiadm` does not need the traditional network setup to login to the iSCSI target?

aafeijoo-suse:
> a session to a target T can be established when there's connectivity to T, but no sooner. For TCP/IP that means that a route to T exists, which is not implied by network-online.target. The route could be up before network-online.target is reached, or vice versa, it could still be missing after network-online.target

Thanks, understood then. So we are missing a `"there is connectivity".target`.

> Offload cards are only special in the sense that with them, iSCSI connectivity may not even need a single Linux network interface.

This "_may not even need_" means that some cards need a network interface and others don't, right? And that's why `rd.iscsi.firmware` is now with your patch independent of `rd.iscsi.transport`, not setting `$netroot` with the latter. So, in this case `parse-iscsiroot.sh` should be returning 1:

https://github.com/dracut-ng/dracut-ng/blob/1c5f45a24b34e3fb81e29f44de451bd1f9638203/modules.d/95iscsi/parse-iscsiroot.sh#L141-L143

but since this return value is ignored in `source_hook` we're all good.

Could you share what specific kernel command line options are you using with these cards? you're not using `root=? netroot=iscsi rd.iscsi.firmware=1` as `man dracut.cmdline(7)` states, right? Otherwise `$netroot` would not be unset in the line you changed.

mwilck:
> This "may not even need" means that some cards need a network interface and others don't, right? 

I can't answer that with confidence. I used the word "may" because I just don't know the details for the other cards. Many of them are rather old and don't matter much any more these days. We tried to activate some old hardware in our lab for testing them but most of these systems were so old that we couldn't bring them up any more at all. `bnx2i` is actively maintained and regularly tested by Marvell though. Therefore this patch set focuses on it. As I said, I'd very grateful for for test results with other cards.

Here's the command line of my main testbed:
```
BOOT_IMAGE=/boot/vmlinuz-6.4.0-150600.23.25-default root=UUID=7fed5ce8-e24f-4b67-b2a8-f0317365fbba iomem=relaxed mitigations=auto security=apparmor crashkernel=328M,high crashkernel=72M,low splash=verbose console=tty0 systemd.show_status=0 console=ttyS0,57600n8
```
`etc/cmdline.d` in the initrd image contains:
```
# lsinitrd /boot/initrd etc/cmdline.d/95iscsi.conf
rd.iscsi.firmware=1 rd.iscsi.offload=bnx2i
```
That's it as far as iSCSI is concerned.
The parameters for booting are set in the firmware (iBFT table).

mwilck:
> Thanks, understood then. So we are missing a `"there is connectivity".target`.

Not quite. there may be multiple iSCSI targets connected via different transports and networks. Connectivity to one doesn't imply connectivity to others.

aafeijoo-suse:
> ```
> # lsinitrd /boot/initrd etc/cmdline.d/95iscsi.conf
> rd.iscsi.firmware=1 rd.iscsi.offload=bnx2i
> ```

Ahh so `netroot=iscsi` is not necessary in this case, the doc is outdated: https://github.com/dracut-ng/dracut-ng/commit/c3f7567a65b72eb5195f3293ac881489a8d4a4b3

aafeijoo-suse:
> > Thanks, understood then. So we are missing a `"there is connectivity".target`.
> 
> Not quite. there may be multiple iSCSI targets connected via different transports and networks. Connectivity to one doesn't imply connectivity to others.
> 
> But that's no topic for this PR, not for dracut in general and not even for systemd.

I agree, I just wanted to better understand how this works.

aafeijoo-suse (approved):

mwilck:
> Ahh so netroot=iscsi is not necessary in this case, the doc is outdated: 

I don't disagree. There are lots of iscsi-related parameters, and it's not easy to understand how they interact. I'd like to sort it out but that would be another PR and a considerable amount of effort.

Let me just mention one thing: there is the parameter `rd.iscsi.mp=1`: "_Configure all iBFT interfaces, not only used for booting (multipath)_". 

In my opinion this parameter should be removed or at least deprecated. There is absolutely no point in waiting for all paths being up, _in the initramfs_, in a multipath scenario. The very point of multipathing is that it makes your system work if only some of the paths are available. This parameter would do the opposite, it would cause booting to fail if just one path was unavaiable, and thus increase the likelihood of errors compared to a non-multipath setup. IMO that's obvious nonsense.

aafeijoo-suse:
> Let me just mention one thing: there is the parameter `rd.iscsi.mp=1`: "_Configure all iBFT interfaces, not only used for booting (multipath)_".
> 
> In my opinion this parameter should be removed or at least deprecated. There is absolutely no point in waiting for all paths being up, _in the initramfs_, in a multipath scenario. The very point of multipathing is that it makes your system work if only some of the paths are available. This parameter would do the opposite, it would cause booting to fail if just one path was unavaiable, and thus increase the likelihood of errors compared to a non-multipath setup. IMO that's obvious nonsense.

Thanks, good to know. At least this only affects `network-legacy`.

Issue #215, opened 2024-04-21
LLVM's objdump do not work in generating uki


FlysoftBeta:
**Describe the bug**
 Generating UKI using LLVM's objdump would result in the following error:

```
dracut[F]: Failed to get the size of /usr/lib/systemd/boot/efi/linuxx64.efi.stub to create UEFI image file
```

**Distribution used**
Gentoo (using stage3-amd64-musl-llvm-20240414T161904Z.tar.tz)

**Dracut version**
sys-kernel/dracut-060_pre20240104-r4

**Init system**
OpenRC

**To Reproduce**
1. Set up a Gentoo installation with musl libc and llvm
2. `echo "sys-kernel/installkernel dracut uki" > /etc/portage/package.use/installkernel`
3. `emerge sys-kernel/installkernel && emerge sys-kernel/gentoo-kernel-bin`

**Additional context**

Run `objdump -h /usr/lib/systemd/boot/efi/linuxx64.efi.stub` manually(https://github.com/dracut-ng/dracut-ng/blob/ffeb32b2d2caecc2111bfd93b6919dc4f09d3c2d/dracut.sh#L2508C14-L2508C15), it seems there are LMA, file offset and align column missing:

### LLVM:

```
flysoft / # objdump -h /usr/lib/systemd/boot/efi/linuxx64.efi.stub

/usr/lib/systemd/boot/efi/linuxx64.efi.stub:    file format coff-x86-64

Sections:
Idx Name          Size     VMA              Type
  0 .text         0000edb6 000000014df91000 TEXT
  1 .rodata       0000292c 000000014dfa0000 DATA
  2 .data         000002d8 000000014dfa3000 DATA
  3 .sdmagic      00000027 000000014dfa4000 DATA
  4 .sbat         000000cb 000000014dfa5000 DATA
  5 .reloc        00000080 000000014dfa6000 DATA

```

### GNU binutils:

```
/usr/lib/systemd/boot/efi/linuxx64.efi.stub:     file format pei-x86-64

Sections:
Idx Name          Size      VMA               LMA               File off  Algn
  0 .text         0000edb6  000000014df91000  000000014df91000  00000400  2**4
                  CONTENTS, ALLOC, LOAD, READONLY, CODE
  1 .rodata       0000292c  000000014dfa0000  000000014dfa0000  0000f200  2**2
                  CONTENTS, ALLOC, LOAD, READONLY, DATA
[...]

PR #1122, opened 2025-01-15
feat(UKI): use ukify when available to generate UKI

jozzsi:
## Changes

Prefer to use ukify over the current objcopy and awk based solution to generate UKI.

Potentially fixes https://github.com/dracut-ng/dracut-ng/issues/215

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2632,30 +2632,21 @@ if [[ $uefi == yes ]]; then
     cp "$uefi_stub" "$tmp_uefi_stub"
     objcopy --remove-section .sbat "$tmp_uefi_stub" &> /dev/null
 
-    if objcopy \
-        ${DRACUT_REPRODUCIBLE:+--enable-deterministic-archives --preserve-dates} \
-        ${uefi_osrelease:+--add-section .osrel="$uefi_osrelease" --change-section-vma .osrel=$(printf 0x%x "$uefi_osrelease_offs")} \
-        ${uefi_cmdline:+--add-section .cmdline="$uefi_cmdline" --change-section-vma .cmdline=$(printf 0x%x "$uefi_cmdline_offs")} \
-        ${uefi_splash_image:+--add-section .splash="$uefi_splash_image" --change-section-vma .splash=$(printf 0x%x "$uefi_splash_offs")} \
-        --add-section .sbat="$sbat_out" --change-section-vma .sbat="$(printf 0x%x "$uefi_sbat_offs")" \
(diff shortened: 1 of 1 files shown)


Nowa-Ammerlaan:
Personally I don't really see the point since ukify ships with its own kernel installation hook that can be called after dracut has done its job, so wrapping ukify in dracut does not really create new possibilities (at least from a Gentoo perspective). But since the old code remains as the fallback option I am also not strongly opposed to this change.

I do think there should be some switch to disable this offloading of the uki generation even if the ukify command is present.

dracut.sh
```diff
@@ -2632,42 +2632,71 @@ if [[ $uefi == yes ]]; then
...
-                && sbverify --cert "${uefi_secureboot_cert}" "$outfile" > /dev/null 2>&1; then
-                dinfo "*** Creating signed UEFI image file '$outfile' done ***"
-            else
-                rm -f -- "$outfile"
-                dfatal "*** Creating signed UEFI image file '$outfile' failed ***"
-                exit 1
-            fi
-        else
+    if command -v ukify &> /dev/null; then
+        echo "Using ukify to create UKI"
```
bdrung, line 2636: Is it intentional to use `echo` here instead of a logging command?

dalto8:
I took a brief look at this.  

First, here is my understanding, please correct me if I am wrong.
* This change is only applied if `uefi` is set to `yes` in the config
* If it is set to yes, and `ukify` is installed, it uses Ukify to create the UKI

If that is so, I see no issues here.

We do use kernel-install.  However, if someone wanted kernel-install to be responsible for the UKI creation, they would not also set the dracut config to generate a UKI.  I don't see this as conflicting.

Nowa-Ammerlaan:
> Has anybody been testing uki on aarch64 ?
> 
> See also [systemd/systemd#34780](https://github.com/systemd/systemd/issues/34780)
> 
> Using dracut's current uki generation on leads to the following error using UEFI test
> 
> ```
> objcopy: /boot/vmlinuz-XX: file format not recognized
> sed: can't read /var/tmp/dracut-test.iqLJDA/initrd/dracut.GFt4Nj/uefi/kernel.sbat: No such file or directory
> ```

Does this kernel have ZBOOT enabled? I think this is required for UKIs to work properly on arm64.

Conan-Kudo (approved):

Nowa-Ammerlaan:
We need a way to disable this, it is breaking stuff: https://bugs.gentoo.org/961525

PR #1125, opened 2025-01-17
ci: switch to native ARM GitHub runners

jozzsi:
## Changes

See https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

Fedora arm container building time reduced from 40 min to 3 min with this PR.

## Checklist

Fixes #


bdrung (approved):
Great. With these ARM runners, we should add Ubuntu into the mix as well.

PR #1152, opened 2025-01-22
fix(dracut-install): install compressed blobs that match wildcard fwpath

### LINKED ISSUE #1150: dracut-install does not install compressed wildcard blobs
**Describe the bug**

Take kernel module intel-ish-ipc for example, it has two firmware declarations:

```
$ modinfo intel-ish-ipc | grep ^firmware
firmware: intel/ish/ish_*.bin
firmware: intel/ish/ish_lnlm.bin
```

And with additional custom ISH firmware installed from [bug 2094768](https://bugs.launchpad.net/bugs/2094768), update-initramefs gives:

```
$ update-initramefs -v -u -k $(uname -r)
...
dracut-install: Missing firmware intel/ish/ish_*.bin for kernel module intel_ish_ipc
dracut-install: mkdir '/var/tmp/mkinitramfs_9uk7FI/lib/firmware/intel/ish'
dracut-install: cp '/lib/firmware/intel/ish/ish_lnlm_39ceeaf8_58.7724.2.bin.zst' '/var/tmp/mkinitramfs_9uk7FI/lib/firmware/intel/ish/ish_lnlm_39ceeaf8_58.7724.2.bin.zst'
dracut-install: ln -s 'ish_lnlm_39ceeaf8_58.7724.2.bin.zst' '/var/tmp/mkinitramfs_9uk7FI/lib/firmware/intel/ish/ish_lnlm.bin.zst'
...
```

In source code for `/usr/lib/dracut/dracut-install`, that is, source package dracut, file `src/install/dracut-install.c`, function `install_firmware`, it tries to invoke `glob()` with full path expanded from "intel/ish/ish_*.bin", but while all the binaries were compressed, this matches no file and none of the custom ISH firmware blobs will be installed.

**Distribution used**

Ubuntu

**Dracut version**

This was first found in Ubuntu/Noble, which has 060, but from the git history and the differences, this should be reproducible with all versions of dracut.

**Init system**

Systemd.

**To Reproduce**

1. touch `/lib/firmware/intel/ish/ish_foo.bin.zst`
2. `update-initramefs -v -u -k $(uname -r)`, and check if above file was added.

**Expected behavior**

Any file under `/lib/firmware/intel/ish` matches `intel/ish/ish_*.bin`, compressed (.zst, .xz) or uncompressed (.bin), should be added into the newly created initramefs.

**Additional context**

Launchpad bug: https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/2095518

vicamo:
## Changes

Call `install_firmware_fullpath` based on glob matched, compressed or uncompressed fwpath.

## Checklist

Fixes #1150

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -1437,12 +1437,15 @@ static int install_all(int argc, char **argv)
         return r;
 }
 
-static int install_firmware_fullpath(const char *fwpath)
+static int install_firmware_fullpath(const char *fwpath, bool maybe_compressed)
 {
         const char *fw = fwpath;
         _cleanup_free_ char *fwpath_compressed = NULL;
         int ret;
         if (access(fwpath, F_OK) != 0) {
+                if (!maybe_compressed)
+                        return 1;
+
                 _asprintf(&fwpath_compressed, "%s.zst", fwpath);
                 if (access(fwpath_compressed, F_OK) != 0) {
                         strcpy(fwpath_compressed + strlen(fwpath) + 1, "xz");
@@ -1458,6 +1461,23 @@ static int install_firmware_fullpath(const char *fwpath)
                 log_debug("dracut_install '%s' OK", fwpath);
         }
         return ret;
+}
+
+static bool install_firmware_glob(const char *fwpath)
+{
+        size_t i;
+        _cleanup_globfree_ glob_t globbuf;
+        bool found = false;
+        int ret;
+
+        glob(fwpath, 0, NULL, &globbuf);
+        for (i = 0; i < globbuf.gl_pathc; i++) {
+                ret = install_firmware_fullpath(globbuf.gl_pathv[i], false);
+                if (ret == 0)
+                        found = true;
+        }
+
+        return found;
 }
 
 static int install_firmware(struct kmod_module *mod)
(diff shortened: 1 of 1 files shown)


schopin-pro:
Despite it not being documented anywhere in the kernel doc, diving into the code confirms that it's a generic feature rather than something implemented by specific drivers: https://github.com/torvalds/linux/blob/v6.13/drivers/base/firmware_loader/main.c#L922

Personally I'm not a huge fan of the `fwpath_compressed` re-use, I think it's a footgun waiting to go off on an unsuspecting contributor wanting to add another compression algorithm.

The approach here should mostly work, except for one edge case: the firmware files shipped are .zst and .xz (no .bin) but the kernel doesn't have ZSTD support. However, I don't know if there's an easy way to get that information from there :/

bdrung (approved):
I successfully tested this change. First I searched for an affected kernel module:

```
for module in $(find /lib/modules/$(uname -r) -name \*.ko.zst); do m=$(basename "$module"); m=${m%%.*}; if modinfo $m|grep '^firmware:.*\*'; then echo $module; fi; done
```

I found `brcmfmac`. This has the firmware glob `brcm/brcmfmac*-pcie.*.bin`:

```
$ ls /usr/lib/firmware/brcm/brcmfmac*-pcie.*.bin*
/usr/lib/firmware/brcm/brcmfmac43602-pcie.ap.bin.zst
```

dracut-install from Ubuntu 24.04:

```
$ /usr/lib/dracut/dracut-install -D ~/tmpfs/output-old -m brcmfmac
$ find ~/tmpfs/output-old -name 'brcmfmac*-pcie.*.bin*'
```

Then this proposed patch:

```
$ ./src/install/dracut-install -D ~/tmpfs/output-new -m brcmfmac
$ find ~/tmpfs/output-new -name 'brcmfmac*-pcie.*.bin*'
~/tmpfs/output-new/usr/lib/firmware/brcm/brcmfmac43602-pcie.ap.bin.zst
```

The code looks good (enough). We should merge it and can improve on it in a following merge request.

Suggestions for code improvements:
* A return early pattern could be used for the try, try with ".zst" suffix, try with ".xz" suffix logic. This would reduce the indentation level.
* Implement the try, try with ".zst" suffix, try with ".xz" suffix logic is currently done twice. The code could do this logic first and then do the is glob differentiation afterwards.

Conan-Kudo (approved):

PR #990, opened 2024-11-19
ci(gentoo): force hostonly mode as is recommended by Gentoo wiki

jozzsi:
See https://wiki.gentoo.org/wiki/Dracut.

Following the discussion at https://github.com/dracut-ng/dracut-ng/issues/674 - with permission from @Nowa-Ammerlaan  - I'd like to propose switching the CI for Gentoo to run in hostonly mode to increase confidence that this mode is well supported on Gentoo.

CC @Nowa-Ammerlaan 

## Checklist
- [ ] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

Fixes #

Issue #748, opened 2024-10-18
Please support hostonly mode with full portability


bdrung:
**Describe the enhancement**

In Ubuntu, initramfs-tools builds portable initrds by default. We want that user can replace hardware (e.g. the graphics card, keyboard) and still be able to boot. Taking a disk from one machine, plug it into a different machine, and boot should work out of the box.

That was the reason Ubuntu sets `hostonly=false`. Pull request #746 and issue #674 revealed that I misunderstood the `no-hostonly` mode and I had a lack of knowledge. The man page says in "Boot parameters":

> An initramfs generated without the "hostonly" mode, does not contain any system configuration files (except for some special exceptions), so the configuration has to be done on the kernel command line.

But as you can see in #746 we want to honor the system configuration files in Ubuntu.

The usage section of the man page says:

> If you want to create lighter, smaller initramfs images, you may want to specify the --hostonly or -H option. Using this option, the resulting image will contain only those dracut modules, kernel modules and filesystems, which are needed to boot this specific machine. This has the drawback, that you can’t put the disk on another controller or machine, and that you can’t switch to another root filesystem, without recreating the initramfs image. The usage of the --hostonly option is only for experts and you will have to keep the broken pieces.

As mentioned above Ubuntu wants to support putting the disk on another controller or machine.

There is `--hostonly-mode` that can be `strict` or `sloppy`.

> In "sloppy" host-only mode, extra drivers and modules will be installed, so minor hardware change won’t make the image unbootable (e.g. changed keyboard), and the image is still portable among similar hosts.

I am proposing to add another mode, e.g. `full` that will support major hardware changes. Then Ubuntu will switch to this mode.

bdrung:
> One way forward is to invoke dracut two times.
> 
> First invocation with `hostonly=false` to compute kernel modules (with `--kernel-only`) - for "full kernel module portability".
> 
> Second invocation with `hostonly=true` (with `--no-kernel`) to compute the rest of the initrd (which now will include any system configuration files).
> 
> I personally quite excited about this usage of dracut - similar usage is covered by one of the tests - [SYSTEMD-INITRD](https://github.com/dracut-ng/dracut-ng/blob/main/test/TEST-42-SYSTEMD-INITRD/test.sh#L77)

I like to have mode that will be equivalent to that approach.

bdrung:
The documentation of dracut is slightly contradictory:

https://dracut-ng.github.io/dracut-ng/ says in "Host v Default mode":

[...]

PR #1158, opened 2025-01-23
feat(dracut.install): force hostonly for kernel-install plugin

jozzsi:
Distributions that have adapted and integrated dracut and adopted the dracut kernel install plugin should default to hostonly mode initrd.

This PR also makes the two plugins (50-dracut.install  51-dracut-rescue.install) a bit more aligned and symmetric as 51-dracut-rescue.install) is already forcing non-hostonly/generic.

Follow-up to 9fc9128a08e6262393e6b7ae5ade040795ab6668 and https://github.com/dracut-ng/dracut-ng/pull/990 .

CC @Nowa-Ammerlaan

## Checklist

Fixes #

### DIFF
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -71,6 +71,7 @@ done
 
 # shellcheck disable=SC2046
 dracut -f \
+    --add-confdir hostonly \
     ${noimageifnotneeded:+--noimageifnotneeded} \
     $([[ $KERNEL_INSTALL_VERBOSE == 1 ]] && echo --verbose) \
     $([[ -n $KERNEL_IMAGE ]] && echo --kernel-image "$KERNEL_IMAGE") \


Nowa-Ammerlaan:
I fear this would create more confusion since the default when `dracut` is invoked directly would now be different then the default if `dracut` is invoked via `kernel-install`. That being said, this could be a first step towards changing the default behaviour in general.

This change would require a news item downstream, plus adjustments for the legacy `installkernel` [hook](https://github.com/projg2/installkernel-gentoo/blob/master/hooks/50-dracut.install) that we maintain downstream (to restore symmetry between systemd and openrc), but that is something I can manage.

dalto8:
I continue to think the default should be [this](https://github.com/dracut-ng/dracut-ng/issues/748), however, of the currently available options, hostonly is the only sensible one I think.

That being said, it would not be harmful to us to change the kernel-install default to hostonly so I am not opposed to this change.

bdrung:
Instead of changing the default setting in the kernel-install script, the default should be changed instead. For Ubuntu we only use hostonly until https://github.com/dracut-ng/dracut-ng/issues/748 is implemented.

PR #1175, opened 2025-01-26
dracut-init: strip out unused/unlikely AMDGPU firmware

MingcongBai:
This pull request changes...

## Changes

Introduce logic to remove unlikely or impossible firmware on a platform- specific basis to save space - this is the largest set of firmware in the linux-firmware.git tree.

To update the list of prefixes:

1. Look for (grep) MODULE_FIRMWARE in /drivers/gpu/drm/amd (please note that AMD does not always use the file name to the firmware as parameter to this macro.
2. Reference "Misc AMDGPU driver information"[^1], Wikipedia, and TechPowerUp to determine which class certain set(s) of firmware belong out of the list below.
3. Refer to manufacturers and/or OEM contacts for state of support (some should be obvious - there is no non-x86 APUs - yet?).

[^1]: https://docs.kernel.org/gpu/amdgpu/driver-misc.html

## Checklist

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -985,6 +985,70 @@ for_each_module_dir() {
 dracut_kernel_post() {
     local dstdir="${dstdir:-"$initdir"}"
 
+    if [ -d "${dstdir}"/lib/firmware/amdgpu/ ]; then
+        # AMDGPU firmware stripping: Remove unlikely or impossible firmware on
+        # a platform-specific basis to save space - this is the largest set of
+        # firmware in the linux-firmware.git tree.
+        #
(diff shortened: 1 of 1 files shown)


MingcongBai:
> @MingcongBai Thanks for the PR. This project runs automatic CI test for each PR and for this PR the test is failing. Can you please check ?

Hi, thanks for the notice. All tests now pass.

Conan-Kudo (approved):
I can't see a reason for this to be wrong off the bat, so let's take it.

Conan-Kudo:
Something similar might eventually be needed for NVIDIA firmware too...

PR #1176, opened 2025-01-26
dracut-init: decompress kernel modules and firmware before final compression

MingcongBai:
This pull request changes...

## Changes

Decompress all kernel modules and firmware (if compressed), this allows for better compression ratio as we make the compressed initramfs.

This may save over 10MiB post-XZ.

## Checklist

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -1072,6 +1072,10 @@ dracut_kernel_post() {
         fi
     fi
 
+    if [[ $handle_precompress == decomp* ]]; then
+        dracut_kernel_post_decomp
+    fi
+
     for _f in modules.builtin modules.builtin.alias modules.builtin.modinfo modules.order; do
         [[ -e $srcmods/$_f ]] && inst_simple "$srcmods/$_f" "/lib/modules/$kernel/$_f"
     done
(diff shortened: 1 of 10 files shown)


MingcongBai:
> @MingcongBai Thanks for the PR. This project runs automatic CI test for each PR and for this PR the test is failing. Can you please check ?

Hi, sorry for the delay - I was on new year's vacation here. I will handle this (and my other PR) tomorrow.

MingcongBai:
@LaszloGombos Hi there, I was able to eliminate all lint/commit issues but I'm confused as to how the Integration Tests were failing, could you please advise on what to do next?

bdrung:
In Debian/Ubuntu we did this (extracting the kernel modules) in initramfs-tools as well, but we changed to putting compressed kernel modules into a separate, uncompressed cpio. See https://discourse.ubuntu.com/t/reduce-initramfs-size-and-speed-up-the-generation-in-ubuntu-23-10/38972

So I suggest doing the same in dracut as well.

Artoria2e5:
> In Debian/Ubuntu we did this (extracting the kernel modules) in initramfs-tools as well, but we changed to putting compressed kernel modules into a separate, uncompressed cpio. See https://discourse.ubuntu.com/t/reduce-initramfs-size-and-speed-up-the-generation-in-ubuntu-23-10/38972
> 
> 
> 
> So I suggest doing the same in dracut as well.

The two approaches do not replace each other. Not recompressing the modules helps generation speed and does not hurt compressed size, but the one thing it does not provide is 10 MB of more compression.

Artoria2e5:
welp. let me run more actions on the local branch and leave the PR one alone.

Artoria2e5:
This is usually a better idea unless the maximum size of the initramfs is somehow the bottleneck. That said now that it's a separate function it would be easy to switch.

I wouldn't want to overload --compress though. Might add a switch for "what to do with compressed kernel modules and firmware" in general, just to leave the "don't compress it" option open as well (it would produce a time saving with basically no size increase)

Artoria2e5:
This is usually a better idea unless the maximum size of the initramfs is somehow the bottleneck. That said now that it's a separate function it would be easy to switch.

I wouldn't want to overload --compress though. Might add a switch for "what to do with compressed kernel modules and firmware" in general, just to leave the "don't compress it" option open as well (it would produce a time saving with basically no size increase)

Nowa-Ammerlaan:
See also: https://github.com/dracut-ng/dracut-ng/pull/568

I tried this before but we settled on solving the problem differently by putting the compressed modules in a separate initramfs and cating them together like we do for the microcode.

bdrung:
Thanks @Nowa-Ammerlaan for pointing to the previous work. I prefer the approach taken in https://github.com/dracut-ng/dracut-ng/pull/568 to add the decompression to dracut-install. That way one file copy can be skipped. So the initrd generation should be faster.

My recommendation is to pick up https://github.com/dracut-ng/dracut-ng/pull/568 and implement https://github.com/dracut-ng/dracut-ng/pull/568#issuecomment-2341464032 and handle the firmware symbolic links (https://github.com/dracut-ng/dracut-ng/pull/568#issuecomment-2341878027).

Artoria2e5:
All the failures use gzip compression. Probably a clue here, will dig deeper.

The differential shellcheck isn't really my fault -- the existing code looks exactly like that.

Lint issues will be fixed with another substantial update later. Don't want to waste action runs -- I've wasted enough!

dracut-init.sh
```diff
@@ -1083,6 +1087,86 @@ dracut_kernel_post() {
...
+            ddebug "Fixing up decompressed firmware: $fwlink ..."
+            fwtarget=$(readlink "$fwlink")
+            rm "$fwlink"
+            ln -s "${fwtarget%.*}" "${fwlink%.*}"
+        done
+}
+
+dracut_kernel_post_split() {
+    ddebug "Splitting kernel modules and firmware into a non-compressed directory ..."
+    mkdir -p "$no_compress_dir"/d/lib/modules/"$kernel"/
```
Nowa-Ammerlaan: This might need a check to verify that the modules/firmware actually are compressed.

Artoria2e5: Yeah, I agree.

bdrung:
I'll port the "Put compressed kernel modules and firmware in an uncompressed cpio" patch from initramfs-tools to dracut. So that you do not have to reinvent the wheel for that part.

Artoria2e5:
Uhm, sure. The Debian approach still needs an `early_cpio` marker to make our `lsinitrd` work independently of `skipcpio`, assuming we don't rewrite the `skipcpio` routine in shell like they do with `unmkinitranfs`.

Artoria2e5:
Oh come the heck on alpine, what are you doing?

Artoria2e5:
finally!

devkontrol:
@MingcongBai @Artoria2e5 -  apologies for the unresponsiveness on this PR. Landing this large PR will take time, but let's try to make progress. 

Noticed some unrelated commits in this PR - https://github.com/dracut-ng/dracut/pull/1176/changes/f343821aeb4913440b5488598bcbed618528b80b 

Would you be able to factor those out, or would you be OK if we're factor those out into separate PR ? That should make this PR smaller and easier to review.

devkontrol:
This PR has merge conflicts. I recommend break this PR into sub-PRs to speed up review and landing it.

Marking it draft for now to take it out from the review queue.

PR #1190, opened 2025-01-29
feat(systemd-battery-check): always include the module if possible

pvalena:
This pull request changes...

## Checklist

Fixes #


pvalena:
Sorry, wanted to know if CI passed, and to link the the corresponding discussion as a reason to include it :). I do not have HW to test it I'm afraid -- but we can test in Fedora, I think, and can let you know if it goes well.

pvalena:
* (And I didn't have time to do so yesterday; hence the draft.)

pvalena:
> In general I am in favor, but I hope change like this should be tested.

FTR 
I'm not changing any code behaviour, just the module inclusion; so I hope it was actually tested when it was created - even if it was included only manually, it should not be broken or malicious :).

pvalena:
Little bit of smoke-testing: https://github.com/redhat-plumbers/dracut-fedora/pull/47 (see the lsinitrd diffs)

PR #1203, opened 2025-02-07
test(FULL-SYSTEMD): skip systemd-network-management if missing

bdrung:
The Debian/Ubuntu package runs the FULL-SYSTEMD test without dracut-network installed. Therefore the test fails due to the missing `systemd-network-management` module:

```
TEST: Full systemd serialization/deserialization test with /usr mount
dracut[E]: Module 'systemd-network-management' cannot be found.
```

Fixes: 75f3a191b675 ("test(FULL-SYSTEMD): when systemd-networkd is available, install it")

### DIFF
--- a/test/TEST-41-FULL-SYSTEMD/test.sh
+++ b/test/TEST-41-FULL-SYSTEMD/test.sh
@@ -72,7 +72,7 @@ test_setup() {
         dracut_modules="$dracut_modules dbus-daemon"
     fi
 
-    if [ -f /usr/lib/systemd/systemd-networkd ]; then
+    if [ -f /usr/lib/systemd/systemd-networkd ] && [ -e "${PKGLIBDIR}/modules.d/00systemd-network-management/module-setup.sh" ]; then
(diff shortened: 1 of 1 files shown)


bdrung:
PKGLIBDIR is used in other places as well.

`systemd-networkd` is part of the `systemd` binary package on Debian/Ubuntu. So checking for it would always succeed.

Nowa-Ammerlaan:
> PKGLIBDIR is used in other places as well.

It's used in the tests, and once in the plymouth module, but there it is initialized as well. So I'm not sure if this resolves to anything at runtime.

Conan-Kudo:
> `systemd-networkd` is part of the `systemd` binary package on Debian/Ubuntu. So checking for it would always succeed.

It is not shipped in RHEL or SLE/openSUSE in the `systemd` package. For RHEL, it is available as `systemd-networkd` in EPEL for RHEL and `systemd-network` in openSUSE/SLE. It is not available by default in either distribution.

Conan-Kudo (approved):

PR #1204, opened 2025-02-08
Allow for option DTB section when building UEFI images

sdleitch:
As per Unified Kernal Image specifications, the PE executable EFI images can contain an optional `.dtb` section to supply a device tree to a kernel, through the EFI stub. A number of EFI stubs, such as systemd-boot, support this. This adds that functionality to dracut.


sdleitch:
> Thanks @sdleitch . An alternative is #1122 (and use a conf file for ukify). What do you think of that approach vs this PR (when it comes to your use-case) ?

I am not familiar with this `ukify` tool, but it does look as though it could fill this requirement fairly effectively. My one thought is that it seems less widely used that `objdump`. If you're hesitant to include this functionality while looking into if #1122 fills this requirement I can test that option out.

sdleitch:
> Thanks. Let's also see what others thinks.
> 
> The `objdump` based solution seems to have some issues that are unresolved and maybe fixed with `ukify` - e.g. #215
> 
> Some distributions might also not package `ukify` (yet), was not sure if that is an issue for you.
> 
> Would be great if you could report back if ukify actually works for this case. Note that you do not need #1122 patch to try ukify, if you ask dracut just to create an initrd.

@jozzsi Apologies for the delay, but I have tested ukify in place of objdump here and it works great.

I've actually gone so far as to break my use case out into using it standalone separate from the initrd generation following in dracut invocation. Feel free to close this if you think, or in some way work it in as a fallback for environments/distros lacking ukify.

Conan-Kudo:
I think we don't need this, given that `ukify` handles this case well.

Issue #1195, opened 2025-02-01
Remove "Creating user/group" messages when --quiet used


gene-git:
Tag 106 at  commit 956c08774074ddc45b2f975e13d5c13d1fc36eff

dracut now prints 20 lines when creating users and groups even with --quiet option.
It would be nice to have these silenced when quiet option is used.

Sample output:
```
Creating group 'nobody' with GID 65534.
Creating group 'audio' with GID 997.
Creating group 'disk' with GID 995.
Creating group 'input' with GID 994.
Creating group 'kmem' with GID 993.
Creating group 'kvm' with GID 992.
Creating group 'lp' with GID 991.
Creating group 'optical' with GID 990.
Creating group 'render' with GID 989.
Creating group 'sgx' with GID 988.
Creating group 'storage' with GID 987.
Creating group 'tty' with GID 5.
Creating group 'uucp' with GID 986.
Creating group 'video' with GID 985.
Creating group 'users' with GID 984.
Creating group 'systemd-journal' with GID 983.
Creating user 'root' (Super User) with UID 0 and GID 0.
Creating user 'nobody' (Kernel Overflow User) with UID 65534 and GID 65534.
Creating group 'nobody' with GID 65534.
Creating group 'audio' with GID 997.
```

moritz1000:
I would say this issue needs the `bug` tag and not the `enhancement` one.

gene-git:
@moritz1000 

that's a fair point. I wavered on which way to go before taking the approach that it was not a functionality change and landed on enhancement. Don't feel strongly either way,  and also happy if dev(s)  make a choice.

volklord:
This (debug) stuff is printed on **stderr**. So, redirecting stderr (>/dev/null) would suppress any other errors too. But yeah, dracut SHOULD NOT print this if **--quiet** was given on the command line.

bdrung:
I still see these log messages on Ubuntu when #1205 is applied. Looks like systemd-sysusers writes to stderr.

bdrung:
I agree that we should open a bug against systemd-sysusers to have an option to disable those messages.

The short term fix is some bash magic: redirect stderr and exclude these messages (`grep -v '^Creating '`). The magic part is to only pipe stderr into the grep command and to still preserve error codes.

PR #1205, opened 2025-02-08
fix(systemd-sysusers): always silence stdout

jozzsi:
## Changes

systemd-sysusers does not have quiet option, so always silence stdout (but not stderr).

Fixes: https://github.com/dracut-ng/dracut-ng/issues/1195

## Checklist

Fixes #


bdrung (approved):
It would be nice if redirecting to `/dev/null` is only done when not run in debug mode.

PR #1206, opened 2025-02-08
fix(release): tagging and release generation is no longer automated

LaszloGombos:
Create release commit on relese branch automatically, but leave the rest of the release creation process up to the project admins to allow manual review and intervention.

We do not have a valid email address for the release automation, so let's make it obvious that it is just a placeholder email address.

@bdrung @Conan-Kudo

### DIFF
--- a/tools/release.sh
+++ b/tools/release.sh
@@ -58,7 +58,7 @@ if [ -z "$(git config --get user.name)" ]; then
 fi
 
 if [ -z "$(git config --get user.email)" ]; then
-    git config user.email "<dracutng@gombos.dev>"
+    git config user.email "<dracutng@localhost.localdomain>"
 fi
 
 # Check in AUTHORS and NEWS.md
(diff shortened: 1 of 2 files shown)


tools/release.sh
```diff
@@ -58,7 +58,7 @@ if [ -z "$(git config --get user.name)" ]; then
 fi
 
 if [ -z "$(git config --get user.email)" ]; then
-    git config user.email "<dracutng@gombos.dev>"
+    git config user.email "<dracut>"
```
Conan-Kudo, line 61:
```suggestion
    git config user.email "<dracutng@localhost.localdomain>"
```

Conan-Kudo (changes requested):

tools/release.sh
bdrung: Don't we have any contact address that we could use here?

tools/release.sh
```diff
@@ -58,7 +58,7 @@ if [ -z "$(git config --get user.name)" ]; then
 fi
 
 if [ -z "$(git config --get user.email)" ]; then
-    git config user.email "<dracutng@gombos.dev>"
+    git config user.email "<dracutng@localhost.localdomain"
```
Conan-Kudo, line 61: Missing closing `>`.

Conan-Kudo (changes requested):

Conan-Kudo (approved):

bdrung (approved):
We should really come up with a reliable contact address.

PR #1207, opened 2025-02-09
fix(rngd): do not check for +x perms

Nowa-Ammerlaan:
require_binaries checks for an executable binary, but systemd service files are usually not executable nor binary.

resolves an issue discovered on gentoo where the rngd module would refuse to be included due to a missing file that does in fact exist

This pull request changes...

## Checklist

Fixes #


Nowa-Ammerlaan:
> small nit: I think we usually do not include the module number in the title - fix(rngd)

Fixed!

Conan-Kudo (approved):

PR #1209, opened 2025-02-09
fix(base): only create nobody user for nfs dracut module

jozzsi:
## Changes

Usual local initrd does not need nobody user.

## Checklist

Fixes #


bdrung:
Please rebase.

PR #1210, opened 2025-02-09
ci(opensuse): install dracut-sshd

jozzsi:
Make sure that sshd dracut module does not interfere with upstream dracut modules.

## Checklist

Fixes #

### DIFF
--- a/test/container/Dockerfile-opensuse
+++ b/test/container/Dockerfile-opensuse
@@ -4,7 +4,7 @@
 # - network-legacy
 # - mkosi-initrd
 # - hmaccalc (fido)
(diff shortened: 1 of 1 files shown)


bdrung:
From a technical perspective: A dracut module should either live in the dracut repository or in the corresponding upstream repository. I see no benefit of standalone modules. They need to track dracut development and their upstream project development.

Integrating dracut-ssh into dracut makes totally sense to me.

Conan-Kudo:
I don't really want to test out of tree modules. If we want this functionality, we'll need to bring it into dracut itself.

Issue #1199, opened 2025-02-06
Duplicate `root` entries in /etc/shadow cause OpenSSH to reject login


pyromaniac2k:
**Describe the bug**
Running dracut-ng ends up generating two lines of root password in the `/etc/shadow` file. The first one is most likely generated by ` 60systemd-sysusers`, while the second one is added by `90base`.

I do not know if this issue is new, or if OpenSSH got stricter recently, but while using https://github.com/gsauthof/dracut-sshd to unlock my LUKS disks during boot, SSH login gets rejected since my last update.

**Distribution used**
Arch

**Dracut version**
dracut-ng 106

**Init system**
systemd

**To Reproduce**
just run dracut and look at the created `/etc/shadow`

**Expected behavior**

- only one line per user in `shadow`
- `OpenSSH` works

**Additional context**
We shortly discussed that in Matrix yesterday. `90base`  needs to be changed to replace the line matching `^root:` in `shadow` instead of just appending it.

PR #1212, opened 2025-02-10
fix(systemd-sysusers): make sure tss user for tpm2 is created

jozzsi:
## Changes

systemd-sysuser module should be orders afer all modules that depends on it.

This is to make sure that all inst_sysusers calls are in place before systemd-sysusers is called.

Not only this fixes  tpm2-tss module, this should also fix https://github.com/dracut-ng/dracut-ng/issues/1199

CC @pyromaniac2k

## Checklist

Fixes #

### DIFF
--- a/modules.d/99systemd-sysusers/module-setup.sh
+++ b/modules.d/99systemd-sysusers/module-setup.sh
@@ -2,6 +2,9 @@
 # This file is part of dracut.
 # SPDX-License-Identifier: GPL-2.0-or-later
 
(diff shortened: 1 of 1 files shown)


pyromaniac2k:
This fixes #1199

MartinX3 (approved):

Conan-Kudo:
This change doesn't do anything?

Conan-Kudo:
Oh I see, the module number changed.

Conan-Kudo (approved):

PR #1214, opened 2025-02-10
btrfs: also install btrfstune

maharmstone:
btrfs-progs also includes the program btrfstune; can we please include this in dracut?

This controls a few things that require the filesystem to be unmounted, so having it in dracut would be useful for anything that affects the root partition. My use case was that I needed the block group tree to be enabled for my root partition, but as btrfstune wasn't there I couldn't use the rescue console for this.

Thanks


Conan-Kudo (approved):

Conan-Kudo (approved):

Conan-Kudo:
Yup, thanks!

PR #1218, opened 2025-02-11
test: use "set -eu" in all tests

bdrung:
## Changes

To make all tests more robust, use `set -eu` in all tests. Then the `|| return 1` constructs can be removed.

## Checklist

Fixes #


bdrung:
Split into #1220 and #1221.

PR #1233, opened 2025-02-26
More iSCSI offload fixes

mwilck:
Fix issues with iSCSI offload using the qedi driver:

- speed up and stabilized iSCSI setup in the initrd by avoiding unnecessary network interface bringup. This just extends #1121 for the qedi driver, which I hadn't done previously because of a lack of a suitable test bed, which I now have. I have reasons to believe that the same would also be helpful for other iSCSI offload engines, if not for all, but I'm lacking hardware to test the remaining ones.
- avoid frequent boot failure with just a single iSCSI path, which is caused by a race condition between iscsiuio shutdown and switching root (details in the commit message). This problem is not qedi-specific, but it has been observed on a system with qedi hardware. I think that we've seen it with bnx2i previously without realizing that. See also https://github.com/systemd/systemd/issues/3436.

## Checklist

CC: @aafeijoo-suse, @gonzoleeman, @njavali


aafeijoo-suse (approved):
You're one of the few people with access to iSCSI offload cards using the `qedi` driver. Thanks for working on this.

mwilck:
TBH, no, I have run the tests with the code from https://github.com/openSUSE/dracut/pull/405. But the code is 100% identical, it was a clean cherry-pick.

If this is necessary I need to see what I can do.

mwilck:
One problem is that (open)SUSE needs network-legacy, but that can't be installed with this version of dracut because it requires dhclient (and we use wicked-dhclient instead).

I suppose @aafeijoo-suse has patches to fix this. It would be nice to be able to use unmodified upstream dracut on (open)SUSE.

mwilck:
I've tested with this code base now. I just had to remove `dhclient` from the list of prerequisites and the list of binaries to be installed in `35network-legacy`. Note that that will probably make the network-legacy module non-functional (SUSEs own dracut fork has additional changes substituting `dhclient` by something from the `wicked` toolset), but that doesn't matter because with the changes from this PR, the system at hand boots without network setup in the initrd.

**However** I had to copy the SUSE dracut.conf.d contents to my local dracut.conf.d to make it work. The reason is that the 95iscsi module (not only with this patch, but traditionally, and like a few other storage-related modules I'm aware of) relies heavily on the `cmdline()` functionality. (open)SUSE sets `hostonly_cmdline=yes` by default, which is what makes this patch set effective. Without it, the user would need to construct the command line herself:

```
rd.iscsi.firmware=1 rd.iscsi.transport=qedi
```

Conan-Kudo (approved):

PR #1234, opened 2025-02-27
(feat) Introduce create-missing-root module

esposem:
The idea of this module is basically based on @poettering blogpost "Fitting everything together": https://0pointer.net/blog/fitting-everything-together.html 

The goal is to be able to build a full root filesystem from scratch by only providing the /usr partition and optionally the verity partition that verifies /usr. 

The main difference with the blogpost is that it also requires /etc to be stored in /usr/etc, otherwise all the units that need files in /etc fail. 

The choice of keeping /etc is also to be able to wipe out completely an existing system (assuming /usr can be written to), and keep the same users credentials, ssh keys and so on.

This goial is achieved by using systemd-tmpfiles and systemd-sysusers for the most part (which IMO is amazing!).

This module supports /usr as well as verity-backed /usr, creating ext4,btrfs,xfs roots that can be plaintext or LUKS encrypted.

It is configurable via kernel cmdline but also when providing a config file /etc/create-missing-root.conf when building initramfs.

The only thing that misses is selinux policy. Unfortunately the root fs that systemd creates is fully unlabeled, and I haven't found a clean way to label it at first boot.

To learn more about cmdline options and how it exactly works, refer to the first commit message.

what is left to do:
- the copy of /usr/etc into /etc could be made as optional
- alternatively, instead of copying we could bind mount or add an overlay
- currently the module finds out which disk to create the root to by looking at the ESP location. If there is no ESP, it won't be able to find a disk
- selinux. This one is tough.

## Changes
Add the new create-missing-root module

## Checklist

Fixes #


pvalena:
LGTM. Although I have not tested it yet. Will give it a try on Fedora and let you know the results.

esposem:
Hi @jozzsi 
> One area (in addition to lack of test case and lack of documentation) that this PR is lacking

If you can redirect me to the tests and documentation, that would be good. I am too struggling to find any documentation for any of the other modules installed in dracut. For example, regarding selinux, I was trying to understand what https://github.com/dracut-ng/dracut-ng/tree/main/modules.d/98selinux is doing but I found 0 documentation, and the only [link](https://docs.kernel.org/admin-guide/LSM/SELinux.html) under the core modules list is pretty vague.
Also as documentation, do you mean updating a line in https://github.com/dracut-ng/dracut-ng/blob/main/doc_site/modules/ROOT/pages/modules/core.adoc ?

Same applies to tests. I see no test being developed for any module, to be honest...

> As an example where the functionality is not clear is, what is the expectation if both create_root.overlay and rd.live.overlay.overlayfs is set on the commend line ? This is difficult to answer because I find that this new dracut module overlaps in functionality with existing dracut modules.

I don't really believe there is an overlap in functionalities, to be honest. The `overlayfs` module just overlays root, while `finish-root.sh` (btw I am pretty open into changing module & file names) simply overlays /usr. 
The question whether they have conflicts during execution is a valid question indeed.
I gave a look at `overlayfs` and I am not sure how it works that itself. I honestly don't understand the point of `mount --bind "$NEWROOT" /run/rootfsbase`  (https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/90overlayfs/prepare-overlayfs.sh#L11), especially if run on pre-mount where sysroot is just an empty initramfs folder. 
But even with that, supposing it is pointing to the real root, would that be an issue? And if it isn't, that wouldn't change the current behavior of what it is doing now. 
For the rest, I guess whether `prepare-root.sh` is working on an overlay or not, is not much of a problem. 
BTW I think it makes no sense of mixing `overlayfs` with my module, maybe I should change module_setup to make sure they are not both loaded in the same config?

keszybz:
> The main difference with the blogpost is that it also requires /etc to be stored in /usr/etc

I don't think you can use `/usr/etc` like this. Suse already uses `/usr/etc`, see https://en.opensuse.org/openSUSE:Packaging_UsrEtc.

---

Systemd upstream is doing a lot of work in the same area. Can you explain how that interacts with those features, and what is actually missing from systemd to cover those use cases? If systemd is not providing some functionality, it might make sense to add it there… Let's at least discuss what is missing to find the best home for any new features.

esposem:
I don't believe there is an actual conflict, this feature could be orthogonal to it. I am aware that /usr/etc should be indeed used for configs coming from package and /etc as host configs (with higher priority). 
But practically, if you get rid of the root to get a new one, you need to make a choice to see if you want to save your /etc content, or not. Even more, in the short term most of packages of most of the distros won't be ready to package everyting in /usr/etc. It will take time. And as we all know, if you get rid of /etc right now most of the packages in /usr will fail, making the system unusable.

Let's put it in another way: assuming that we want this to work now (this removes the possibility of not baking /etc up), we can let the user also pick the location in /usr where to store this backup, optionally to even chose to not back up (disabled by default for now, otherwise it's a spectacular way to destroy your filesystem).

First of all, it is worth noticing that the user actually performs the backup, and decides whether /etc has to be saved, and where. We could argue that the user should pick an empty directory, or at least use `cp -aZn /etc /usr/etc` to avoid overwriting files if they are set, but this is up to the user.

Second, we can let the user pick the backup location. Something like this in /etc/create-missing-root.conf (again, happy to rename names and locations of these files :) )
```
ETC_BACKUP_LOCATION=/usr/backup
```
will ensure that /etc will be restored from /usr/backup.
If we have
```
ETC_BACKUP=/usr/etc
```
then it could have two effects: 1) initialize all /etc to system defaults once all RPMs adhere to hermetic-usr idea, providing a "template" to the admin that is allowed to modify those files (won't hurt anyways as they are exactly the same thing) or 2) simply act as another folder no different from `/usr/backup` above where the files are stored.
As last possibility,
```
ETC_BACKUP=""
# or 
# ETC_BACKUP
```
will not copy anything from /usr into /etc. Right now, at least on RHEL/Fedora, I wish you good luck booting a system in such way.

keszybz:
> I don't believe there is an actual conflict, this feature could be orthogonal to it. 

That's not what I was trying to say. Let me rephrase. Suse *already* uses `/usr/etc` with a specific purpose. This means that if you try to use `/usr/etc`, this will create a lot of confusion, and most likely make your module unusable on the suse-family of distros. The suse convention has been adopted by some other distributions and various upstream projects too. (Whether this location can be overridden is not important. The confusion would happen anyway.)

Now I'm not sure what you mean by "backup". If the goal is to save everything that was in `/etc` and then restore it after the factory reset, then it's not a factory reset ;(  And also `/usr/` is the wrong place to store site-specific things, incl. backups. Yeah, now I'm confused about the overall idea of how this is supposed to be used.

> And as we all know, if you get rid of /etc right now most of the packages in /usr will fail, making the system unusable.

That's actually … not true. There are some minor exceptions, but the big majority of software is fine with empty `/etc` and systemd in the default configuration carries some config to populate `/etc` with the basics that are required by some old software, see `/usr/share/factory/etc` and `/usr/lib/tmpfiles.d/etc.conf`.

So my question from https://github.com/dracut-ng/dracut-ng/pull/1234#issuecomment-2690336835 seems even more relevant.

keszybz:
OK, but then the wider question still stands. How does this relate to the similar functionality being implemented in systemd upstream? (This is not a rhetorical question. This is a complex topic and it's hard to know which parts of the system do what and what the long-term plans are.)

esposem:
So you are saying that systemd is implementing the same functionality? Can you point to some work happening in there? And if it's hard to know or track it, what do you suggest to do?

I thought a dracut module would be better because it uses systemd, and it shouldn't be too complex to implement the logic in hbash. If systemd is implementing this feature, then what's the point of this module?

esposem:
I need to temporarily convert this PR to draft because I figured in the last push I broke something.

keszybz:
> So you are saying that systemd is implementing the same functionality? Can you point to some work happening in there? And if it's hard to know or track it, what do you suggest to do?

Those are all good questions ;) Yes, systemd is implementing a lot of functionality to allow clean boots with empty `/etc`. But there are many aspects and degrees of support, so depending on what you want to happen, the stuff that systemd implements may or may not be enough.

If it's not clear from the documentation is systemd implements something, the best approach is to write to systemd-devel with a high-level overview of what you want to achieve. If something is already implemented, people will know that.

> I thought a dracut module would be better because it uses systemd, and it shouldn't be too complex to implement the logic in hbash. If systemd is implementing this feature, then what's the point of this module?

Well, in the past there was a lot of logic which was implemented in dracut. I (and other folks) are trying to convince people that quite often this was misguided: features can be implemented as standalone programs that do not require the special environment and functionality that dracut provides. Then those components are easier to debug and can be used in other contexts. And a lot of other features that were implemented in dracut are now provided natively by systemd. Implementation in bash can be quick and is good for prototyping, but long-term functionality is better provided by languages that are stricter about error handling, logging, etc.

esposem:
What you say makes sense. The good thing is that all the units I implemented here (except prepare-root) do not depend on dracut services or targets, so potentially this work could be reused, if nothing has already implemented in systemd. 
I'll ask systemd-devel before proceeding with this PR.

j1nx:
@esposem Are you still looking into this? I am about to look into the exact same, however rather work together if possible.

BTW; Systemd implements the /usr/etc as /usr/share/factory/etc

more info here;
https://systemd.io/BUILDING_IMAGES/

PR #1119, opened 2025-01-13
fix(systemd-cryptsetup): only filter out modules for strict hostonly

jozzsi:
## Changes

Exclude (filter out) dracut modules only when hostonly_mode is set to strict (needs opt in even in hostonly).

In sloppy mode all installable optional dependencies will be installed.

Helps with the discussion for https://github.com/dracut-ng/dracut-ng/issues/748

## Checklist
- [ ] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

Fixes #

PR #1238, opened 2025-03-01
feat: set hostonly config by default in configure

jozzsi:
## Changes

Most existing dracut packaging and most dracut documentations enable `hostonly` (Fedora, openSUSE, Gentoo, Ubuntu, Arch, etc).

The motivation for this PR is simply to make `hostonly` not only a default during dracut packaging but also here upstream.

Discussing the issues of `hostonly` config is outside of the scope of this PR. This PR simply makes an existing and already supported option the default that is the default anyways for most distributions downstream.

Follow-up to https://github.com/dracut-ng/dracut-ng/pull/1158 .

## Checklist

CC @Nowa-Ammerlaan @bdrung @LaszloGombos

### DIFF
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -71,7 +71,6 @@ done
 
 # shellcheck disable=SC2046
 dracut -f \
-    --add-confdir hostonly \
     ${noimageifnotneeded:+--noimageifnotneeded} \
     $([[ $KERNEL_INSTALL_VERBOSE == 1 ]] && echo --verbose) \
     $([[ -n $KERNEL_IMAGE ]] && echo --kernel-image "$KERNEL_IMAGE") \


bdrung:
Then https://github.com/dracut-ng/dracut-ng/pull/1158 can be reverted.

Conan-Kudo:
Let's add the revert to this PR.

dalto8:
Am I missing something?  Isn't the revert is already in this PR.

Conan-Kudo:
Sorry, I mean I want it as a separate commit. There are two changes happening in here.

ArrayBolt3:
I'm a bit confused with the hostonly situation in dracut overall. My current understanding (correct me if I'm wrong) is:

* `hostonly=no` - Make an initramfs that will theoretically work on a variety of OSes and hardware (depending on the particular dracut modules that are installed). Lacks installation-specific details like the crypttab of the host.
* `hostonly=yes, hostonly_mode=strict` - Make an initramfs that is specific to the host, both software- and hardware-wise. Includes installation-specific details like crypttab, but lacks drivers that allow hardware independence.
* `hostonly=yes, hostonly_mode=sloppy` - Make an initramfs that is specific to the host's software, but semi-independent from the hardware. Includes installation-specific details like crypttab, and has more drivers than `hostonly_mode=strict` but less than `hostonly=no` (for instance it lacks filesystem drivers).

I'm not sure what the upstream intention behind `hostonly=no` is, there aren't a lot of situations where an initramfs will be moved from one software installation to another, besides perhaps a Xen hypervisor setup where the host provides a shared kernel and initramfs for a number of different VMs.

There doesn't seem to be any mode that includes support for as much hardware and configurations as possible, but also includes critical installation-specific data like the host's crypttab. The documentation seems to indicate that `hostonly_mode=sloppy` allows the initramfs to work on different hardware as long as the hardware is *similar*, but doesn't clarify *how* similar. Can I move the installation with initramfs from one Dell Optiplex to another Optiplex of a different model? Can I move it from my laptop to a mini PC? Can I move it from my Intel-powered embedded SBC to a multi-socket AMD EPYC server with a terabyte of RAM? How similar of hardware is "similar"?

If "similar" means "same CPU architecture and close enough firmware type" (i.e., if it works on one Intel + UEFI system, it'll probably work on any Intel + UEFI system and Intel + BIOS systems if you configure it right), and moving between radically different but still compatible systems is supported, then I think this is good, `hostonly=yes` (which defaults to `hostonly_mode=sloppy`) is good enough. If a hardware change like that won't work with `hostonly=yes hostonly_mode=sloppy`, but would work if `hostonly=no` had been used, then this is a bit problematic. Would it be possible to introduce a new mode that would include enough drivers to allow moving an installation between radically different kinds of hardware, essentially combining the flexibility of `hostonly=no` with the machine-specific details of `hostonly=yes hostonly_mode=sloppy`

AlexanderKurtz:
Hi @ArrayBolt3 and everybody!

I just wanted to give a short example where `hostonly=no`  (i.e. without `fstab`/`crypttab`/etc. but with all kernel modules is **extremly** useful:

I have a small script which generates  full disc images inside QEMU. By using UEFI+`systemd-boot`, `dracut` with `hostonly=no`  and the Discoverable Partitions Specification (DPS)/`systemd-gpt-auto-generator`, these images can be extremly simple (no `fstab`, no `crypttab`, no kernel command line, etc.) and they still boot without problems on any machine (both virtual and physical) with UEFI and a matching CPU architecture.

`dracut` is currently the **only** initramfs implementation that I know of that can do this., because it supports `hostonly=no` , and actually runs `systemd` (and therefore `systemd-gpt-auto-generator`) inside the initramfs.

I have no problem with changing the default here, but please keep in mind that an initramfs without host-specific configuration files but with all kernel modules is actually very useful when building image files which are later on run on some other hardware! 

Thanks!

ArrayBolt3:
Nifty, that makes a lot of sense to me. I'm just hoping that there could be another mode that's "all kernel modules *and* system-specific data", unless `hostonly_mode=sloppy` does good enough there.

Conan-Kudo:
Dracut is fundamentally designed to generate host-specific initramfs images. We support producing "generic" (non-hostonly) images for three main reasons:

* Rescue/recovery/migration boot (like what Fedora does)
* Install media (live media, install DVDs, netboot, etc.)
* Input for generating redistributable Unified Kernel Images (UKIs) for UEFI-only environments

We do not expect generic initramfs images to be used as the default path and even "semi-generic" (`hostonly_mode=sloppy`) is uncommonly used. The difference between it and generic (`hostonly=no`) is not substantial enough for wide usage.

ArrayBolt3:
> We do not expect generic initramfs images to be used as the default path and even "semi-generic" (hostonly_mode=sloppy) is uncommonly used. The difference between it and generic (hostonly=no) is not substantial enough for wide usage.

I fundamentally disagree with this statement. There are [numerous](https://junsun.net/wordpress/2021/06/install-full-ubuntu-on-portable-usb-drive/) [guides](https://wiki.archlinux.org/title/Install_Arch_Linux_on_a_removable_medium) [online](https://www.kicksecure.com/wiki/USB_Installation#PERSISTENT_Mode) with instructions on how to install a Linux distribution to a USB hard drive, allowing the same persistent installation to be booted on multiple computers. Some Linux distributions officially support this use case scenario and consider its failure to work a bug. Moving hard drives from one computer to another is a common practice in many scenarios such as data recovery, hardware failure diagnosis, and system upgrades. Migrating a physical machine's installation to a virtual machine is a common enough use case there's an entire tool that exists just for this, [virt-p2v](https://libguestfs.org/virt-p2v.1.html). At my workplace (a Linux laptop OEM), we use a master QA disk that is specifically intended to be moved between very different machines (mini PCs, lightweight laptops, heavyweight mobile workstations, with different CPU generations and different original device manufacturers) - the installation on this disk reconfigures itself on first boot on a new platform to adapt to the platform, but it has to be able to *boot* to get that far, which requires the initramfs to be sufficiently generic. This is just the scenarios I can think of off the top of my head, there's almost certainly more.

There is a very real need for initramfs images that are uniquely bound to a particular software installation, but compatible with a very diverse set of hardware.

Conan-Kudo:
That's what the generic mode (`hostonly=no`) is for. It's supported through that flag. In fact, this is why Fedora ships the [`dracut-config-generic`](https://packages.fedoraproject.org/pkgs/dracut/dracut-config-generic/) package to automatically configure dracut this way. This is my advice for those kinds of use-cases.

Conan-Kudo:
There seems to be a misunderstanding that `hostonly=no` implies the initramfs isn't bound to the software install. There are _no_ initramfs generation modes in dracut that do not bind the generated initramfs to the software installation by default. It _is_ possible to create _install-portable_ initramfses, but those require extra care and configuration to pull off, and aren't an expected mode of operation at all.

ArrayBolt3:
> There seems to be a misunderstanding that hostonly=no implies the initramfs isn't bound to the software install. There are no initramfs generation modes in dracut that do not bind the generated initramfs to the software installation by default. It is possible to create install-portable initramfses, but those require extra care and configuration to pull off, and aren't an expected mode of operation at all.

If that's the intention, that's great, but that doesn't seem to be reality at the moment. If you build an initramfs with `hostonly=no` on a LUKS system, then make the mistake of rebooting, you'll be greeted by an unbootable system because the initramfs won't include /etc/crypttab.

If a `hostonly=no` build is still bound to the installation, then IMO it should include data required for the installation to boot like LUKS configuration. If it isn't bound to the installation, then there's a need for some mode (or combination of config options) that creates a hardware-portable initramfs that has that kind of data in it.

Conan-Kudo:
That would specifically be a bug in the dracut module for LUKS, because each module can choose to do things differently based on the modes of configuration.

Conan-Kudo:
Specifically looking at [`01systemd-cryptsetup`](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/01systemd-cryptsetup/module-setup.sh), it looks like it just bails early if it detects hostonly mode, which is probably wrong given the logic present for handling LUKS with non-hostonly images.

ArrayBolt3:
kk, I thought that was an intentional change :sweat_smile: If it's just a bug, I'd be happy to try to file a patch to fix it.

Conan-Kudo:
It's a bug, I assure you. 😄

ArrayBolt3:
> does this PR help with the issue you're discussing - https://github.com/dracut-ng/dracut-ng/pull/1119

@jozzsi I don't think so. The systemd-cryptsetup module is getting included just fine, the issue is that the installation script for it doesn't copy in the host's /etc/crypttab into the initramfs. That's ultimately what leads to an unbootable system, since now systemd-cryptsetup has no info about what encrypted drives are part of the OS installation and how to mount them after decryption.

ArrayBolt3:
@jozzsi (made a comment, then thought I had gotten confused and deleted the comment, then realized I was confused but not in the way I thought :P) I don't think so. The problem isn't missing modules (though I can see how that could be a problem too), the problem I'm running into is /etc/crypttab not being installed into the initramfs. Without that file, Dracut is then unable to find the encrypted disks that make up the OS installation and mount them properly. If you look at the `install()` function in the systemd-cryptsetup module's `module-setup.sh`, you'll see that all of the crypttab handling code is run when `hostonly` is set, which is what leads to the issue.

ArrayBolt3:
Yes, it will fix the issue... by setting the default mode to be hostonly, which is the very thing @Conan-Kudo just explained should be unnecessary, unless I'm badly misunderstanding him.

It's a bit frustrating to be hearing two different things from different developers about how hostonly settings are intended to work. Is crypttab information not being installed in `hostonly=no` mode a bug, or isn't it?

ArrayBolt3:
@jozzsi Just using `--include` isn't really enough though. The logic in `01systemd-cryptsetup` is significantly more complicated than just installing `/etc/crypttab`. There's other LUKS information that may have to come along with an installation-specific initramfs, like key files.

If it is definitely *not* a bug that these kinds of files are not included in `hostonly=no` mode, then https://github.com/dracut-ng/dracut-ng/issues/748 will need implemented for Dracut to be usable in the scenarios I listed in https://github.com/dracut-ng/dracut-ng/pull/1238#issuecomment-2816996288. I find it very likely that most distros will want to switch to this mode, since most distros like being able to boot, and all of the existing options hand you an unbootable system if you shake them just wrong. Kicksecure, Ubuntu, probably Debian (not sure), and EndeavourOS are all interested, I suspect Fedora will be interested too since @Conan-Kudo has indicated this is also the behavior he wants and expects, but misunderstood Dracut's current ability to provide this.

All that being said, I think the clarification you proposed is good clarification on the current state of things.

ArrayBolt3:
@jozzsi Two things:

* I agree, hostonly probably should be the default, and I'd be happy to test this on Debian since it sounds like `hostonly=no` will never quite do what my workplace needs it to do.
* I do want to be clear that this change will only superficially solve the Debian bug linked earlier (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1103457), and that I will still need a fix for the root issue discussed above.

As for the speculation bit, I mostly included that because of the comment by @LaszloGombos at https://github.com/dracut-ng/dracut-ng/issues/748#issuecomment-2495150772, which strongly indicated that Dracut needed to know distros would actually use these features were they to be implemented. I was trying to show evidence of that. But perhaps that should go in the discussion on the other PR.

mwilck:
Disk encryption and non-hostonly mode don't go together well, by definition. Quite obviously, adding keyfiles or anything passphrase-like in a non-hostonly initrd is out of the question.

Only a very generic crypttab can be added in a non-hostonly initrd. It basically just needs to list the devices that should be decrypted (2nd column of crypttab(5)). Even that wouldn't be truly generic.

AFAIU systemd-gpt-auto-generator supports encrypted partitions to some extent (my guess is that the encrypted partition has to have a partition UUID that s-g-a-g recognizes, haven't tried). 

Without systemd-gpt-auto-generator, the only way to handle encrypted devices in non-hostonly dracut would be to interactively ask for a passphrase for every LUKS device encountered, until a root device is found.

For special scenarios such as a standard initrd inside a specific organization, it might be possible to use FIDO2 or a standardized key file path on an USB for providing the passphrase(s). That's obviously not an option for distribution-shipped initrds though.

AlexanderKurtz:
Hi @mwilck 

> Disk encryption and non-hostonly mode don't go together well, by definition. Quite obviously, adding keyfiles or anything passphrase-like in a non-hostonly initrd is out of the question.
> 
> Only a very generic crypttab can be added in a non-hostonly initrd. It basically just needs to list the devices that should be decrypted (2nd column of crypttab(5)). Even that wouldn't be truly generic.
> 
> AFAIU systemd-gpt-auto-generator supports encrypted partitions to some extent (my guess is that the encrypted partition has to have a partition UUID that s-g-a-g recognizes, haven't tried).
> 
> Without systemd-gpt-auto-generator, the only way to handle encrypted devices in non-hostonly dracut would be to interactively ask for a passphrase for every LUKS device encountered, until a root device is found.
> 
> For special scenarios such as a standard initrd inside a specific organization, it might be possible to use FIDO2 or a standardized key file path on an USB for providing the passphrase(s). That's obviously not an option for distribution-shipped initrds though.

IMHO this isn't as clear as you think, non-hostonly initrds and disk encryption can actually work together just fine, it's just that the configuration has to be done on the kernel command line then (instead of `/etc/crypttab`). `systemd-cryptsetup-generator` supports a large amount of kernel command line parameters which make this possible and (as mentioned by you) `systemd-gpt-auto-generator` also handles encrypted disks just fine.

`hostonly=yes` may very well be the better default for most people but `hostonly=no` also has some great advantages and IMHO should be considered a first-class citizen by the dracut project. :-)

mwilck:
@AlexanderKurtz, fair enough. `luks.key=` and `luks.options=` can be used to pull in keys from USB sticks or FIDO2 devices. I suppose this will be most useful in cases where a TPM-measured static initrd is in use. Otherwise I don't quite see the benefit over just generating a host-specific initrd.

AlexanderKurtz:
@Everybody: I'd really like this PR to be merged, mostly because of the changes to `install.d/50-dracut.install` (which I consider to be a bugfix rather than a feature). Any objections?

mwilck:
> Debian is the next distribution that is forced to work around this issue while waiting for this PR to land 

Distributions have been doing this for years. That's no reason to rush.

test/container/Dockerfile-gentoo
```diff
@@ -36,9 +36,6 @@ RUN \
     echo 'sys-apps/systemd ukify qrcode cryptsetup' >> /etc/portage/package.use/systemd ;\
     # Support thin volumes and build all of LVM2 including daemons and tools like lvchange \
     echo 'sys-fs/lvm2 thin lvm' >> /etc/portage/package.use/lvm2 ;\
-    # force hostonly mode as this is the config recommended by https://wiki.gentoo.org/wiki/Dracut \
-    mkdir -p /usr/lib/dracut/dracut.conf.d/ ;\
-    echo 'hostonly="yes"' > /usr/lib/dracut/dracut.conf.d/50-hostonly.conf ;\
```
mwilck, line 41: Does it hurt to leave this code in after changing the default? Why? I think it makes sense to have it here because it makes it explicit that Gentoo wants hostonly mode.

test/TEST-20-STORAGE/test.sh
```diff
@@ -166,7 +166,7 @@ test_setup() {
         -a "lvm" \
         $(if command -v mdadm > /dev/null; then echo "-a mdraid"; fi) \
         $(if command -v cryptsetup > /dev/null; then echo "-a crypt"; fi) \
-        $(if [ "$TEST_FSTYPE" = "zfs" ]; then echo "-a zfs"; else echo "--add-drivers ${TEST_FSTYPE}"; fi) \
+        $(if [ "$TEST_FSTYPE" = "zfs" ]; then echo "-a zfs --add-drivers zfs"; else echo "--add-drivers ${TEST_FSTYPE}"; fi) \
```
mwilck, line 169: What does this change have to do with defaulting to `hostonly`?

install.d/50-dracut.install
```diff
@@ -71,7 +71,6 @@ done
 
 # shellcheck disable=SC2046
 dracut -f \
-    --add-confdir hostonly \
```
mwilck, line 74: Why remove this? IMO it makes sense to have a config directory dedicated to running dracut from `install.d`. It's debatable whether that config dir should be called `hostonly` but that's a different question.

AlexanderKurtz: :-(

IMHO this is/was the most important part of this PR. Without this, the initramfs generated by the `50-dracut.install`  `kernel-install` module will **always** be `hostonly=yes` because of https://github.com/dracut-ng/dracut-ng/blob/main/dracut.conf.d/hostonly/50-hostonly.conf even if `/etc/dracut.conf` or `/etc/dracut.conf.d/*` specify `hostonly=no`. IMHO this is a bug.

Please consider bringing this change back.

AlexanderKurtz: PS: My concrete use-case is this Debian package:

https://packages.debian.org/sid/all/dracut-config-generic/filelist

IMHO, installing it should ensure that all initramfs files generated by `dracut` are `hostonly=no` unless `dracut` was explicitly invoked by the user with `--hostonly`.

AlexanderKurtz: @jozzsi I can understand your frustration. For what it's worth, I think the review by @mwilck lead to an ultimately better version of `test/TEST-20-STORAGE/test.sh`, which is a good thing.

Regarding `install.d/50-dracut.install`: @mwilck Is there maybe a misunderstanding here? Adding the `--add-confdir hostonly` option does **not** add a "config directory dedicated to running dracut from `install.d`". It adds a config directory which (as the name suggests) contains a config file (shipped by `dracut` upstream and installed by distros) that sets `hostonly=yes`. regardless of other configuration. I don't think that's intended, is it?

mwilck: > It adds a config directory which (as the name suggests) contains a config file (shipped by dracut upstream and installed by distros) that sets hostonly=yes.

Right. I agree that this is strange, in particular if there's no way to override this behavior. 

I was indeed confused. Yes I do agree that removing this line was the right thing to do in this PR.

dalto8 (approved):

install.d/50-dracut.install
AlexanderKurtz: Thanks! :-)

AlexanderKurtz (approved):

Nowa-Ammerlaan (approved):

AlexanderKurtz:
:partying_face:

PR #1244, opened 2025-03-13
test: add container for Azure Linux

trungams:
Add Dockerfile for Azure Linux.

The Azure Linux Dockerfile is mostly based off the Fedora version without the unavailable packages.

`TEST-43-KERNEL-INSTALL` is failing with the following message, but I have not figured out why

```
Failed to chase path /usr/lib/modules/6.6.78.1-3.azl3/vmlinuz for kernel image file specified via command line, ignoring: No such file or directory
```

## Checklist

Fixes #

### DIFF
--- a/.github/workflows/container-extra.yml
+++ b/.github/workflows/container-extra.yml
@@ -46,6 +46,7 @@ jobs:
                     - {dockerfile: 'Dockerfile-gentoo', tag: 'gentoo:amd64-openrc', option: 'amd64-openrc'}
                     - {dockerfile: 'Dockerfile-fedora', tag: 'fedora:rawhide', registry: 'registry.fedoraproject.org'}
                     - {dockerfile: 'Dockerfile-fedora', tag: 'centos:stream10-development', registry: 'quay.io/centos'}
+                    - {dockerfile: 'Dockerfile-azurelinux', tag: 'azurelinux:3.0', registry: 'mcr.microsoft.com'}
                 exclude:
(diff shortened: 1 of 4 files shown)


trungams:
Hi @LaszloGombos, thank you for the quick reply. I will update my PR to add CI. 

The test container/CI would help us evaluate our distro's readiness to move to a newer version of dracut. We'll be monitoring the CI test results and assist if any problem arises. On our side, we are also planning to add testing for the initramfs generated by dracut, so the container would definitely help with that area as well.

trungams:
> My inquiry was about if you plan to do this evaluation upstream or downstream.

We are looking to do this evaluation upstream. I have updated my PR for the CI additions (ARM64 support is included as well). Thank you

trungams:
I forgot that we're not shipping some systemd packages for ARM64. Updated.

trungams:
@jozzsi Thank you for the catch. I updated to always use `--setopt=install_weak_deps=False`

trungams:
> Would it be possible to combine `btrfs..` section and `asciidoc..` sections ? I understand that `systemd-boot` needs to be behind a condition, but why the rest of the packages can not be installed with one call ?

I was following Fedora's Dockerfile and they did split the packages into 2 separate `dnf install ..` call, I guess to enable basic tests and the rest of the test suite. I updated our Dockerfile to merge both into 1 install call.

Conan-Kudo (approved):

PR #1245, opened 2025-03-14
fix: Add $dracutsysrootdir to paths where it should be present

chewi:
## Changes

As above. I noticed a "No such file or directory" error when installing the iscsi module using a sysroot.

## Checklist
- [ ] I am providing new code and test(s) for it -- I might add a sysroot test for my other changes later.


chewi:
The Fedora failure is unrelated. I've seen this outside these changes. TEST-41-FULL-SYSTEMD requests the systemd-portabled module, but Fedora doesn't install portablectl.

chewi:
I've pushed again because I noticed `inst_simple` also needed fixing, which is even more important. Without this, it was not copying `/etc/ld.so.conf.d/*` and C++ programs were subsequently failing.

chewi:
Cancel that, it broke far more than it fixed! Seems we're not consistent on the usage of `inst_simple`. I'll have a closer look.

chewi:
Looks like we may have to make it work both ways since that's what dracut-install does.

chewi:
Hmmm, I had assumed those cancelled checks were due to CI being overloaded, but on closer inspection, maybe not. I'll have a closer look.

chewi:
Got it, it needed to handle relative paths too.

chewi:
Rebased and I've added the sysroot to the `/boot` early microcode directory. That one is perhaps debatable, but we don't install anything else (like `/lib/firmware`) from outside the sysroot.

bdrung (approved):
LGTM

PR #1258, opened 2025-03-26
Don't set LD_LIBRARY_PATH when DRACUT_LDD is set

### LINKED ISSUE #1257: dracut 104+ break cross-compiling
**Describe the bug**

After dracut 103, a change was merged that breaks the cross-compiler ldd script, which 

As a result, running the cross-compiler prefixed readelf intended to run on the host crashes, even if built for the same architecture as the host.

**Distribution used**

Host distribution is Fedora 41, cross-compiler environment is current Yocto master, the upcoming Yocto 5.2 with dracut 106.

**Dracut version**

dracut 104, 105, 106.

**Init system**

systemd

**To Reproduce**

Use https://gist.github.com/zboszor/0d6febacd2f51ab0b0d23f6143daf334.git as the cross-compiler-ldd script, which in turn uses a cross-compiler variant of `readelf`.

Use the described variables in `man 8 dracut`, especially `DRACUT_LDD` coupled with the `--sysroot` option.

**Expected behavior**

It's expected that dracut-install can proper detect dependencies of binaries and the initramfs is correctly generated.

**Additional context**

I bisected it to e0b876823d9c608db7132cab9e5edd62543a27ae : 
```
commit e0b876823d9c608db7132cab9e5edd62543a27ae
Author: Antonio Alvarez Feijoo <antonio.feijoo@suse.com>
Date:   Tue Aug 6 14:58:58 2024 +0200

    fix(dracut): ldd output borked with `--sysroot`
    
    dracut-install calls ldd to resolve dependencies, but when the `--sysroot`
    option is used, ldd is not performing the search within the sysroot directory.
    To fix this issue, the `LD_LIBRARY_PATH` variable needs to be properly set to
    the directories containing shared libraries within the specified sysroot
    directory.
    
    E.g., running dracut with `--sysroot` produces an initrd without the required
    systemd shared libraries when the version between the host system and the
    sysroot directory differs:
    
    ```
    localhost:~ # ldd /.snapshots/9/snapshot/usr/lib/systemd/systemd | grep libsystemd
        libsystemd-core-256.so => not found
        libsystemd-shared-256.so => not found
    localhost:~ # export LD_LIBRARY_PATH=/.snapshots/9/snapshot/usr/lib64/systemd
    localhost:~ # ldd /.snapshots/9/snapshot/usr/lib/systemd/systemd | grep libsystemd
        libsystemd-core-256.so => /.snapshots/9/snapshot/usr/lib64/systemd/libsystemd-core-256.so (0x00007f817b600000)
        libsystemd-shared-256.so => /.snapshots/9/snapshot/usr/lib64/systemd/libsystemd-shared-256.so (0x00007f817b000000)
    ```

diff --git a/dracut-init.sh b/dracut-init.sh
index e47e034d..e085eb3d 100755
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -101,6 +101,17 @@ if ! [[ $libdirs ]]; then
     export libdirs
 fi
 
+# ldd needs LD_LIBRARY_PATH pointing to the libraries within the sysroot directory
+if [[ -n $dracutsysrootdir ]]; then
+    for lib in $libdirs; do
+        mapfile -t -d '' lib_subdirs < <(find "$lib" -type d -print0 2> /dev/null)
+        for lib_subdir in "${lib_subdirs[@]}"; do
+            LD_LIBRARY_PATH="${LD_LIBRARY_PATH:+"$LD_LIBRARY_PATH":}$dracutsysrootdir$lib_subdir"
+        done
+    done
+    export LD_LIBRARY_PATH
+fi
+
 # helper function for check() in module-setup.sh
 # to check for required installed binaries
 # issues a standardized warning message
```

Setting LD_LIBRARY_PATH is a no-no for cross-compilation.

Apparently, `dracut --sysroot` started to be used for cases where the sysroot is identical to the host environment, or close to it, and the `ldd` command is the system command, not a cross-compiler helper.

zboszor:
When cross-compiling, setting LD_LIBRARY_PATH to the sysroot library path breaks host binaries run by the ldd wrapper script.

## Checklist

Fixes #1257


chewi:
I may be about to make the headline change redundant, as I've totally reworked how Dracut finds dependencies in a cross-friendly way.

The shadow change seems useful, but I'd check `[[ $UID != 0 ]]` instead.

zboszor:
> I may be about to make the headline change redundant, as I've totally reworked how Dracut finds dependencies in a cross-friendly way.

In the meantime that commit is necessary.

If that's not a big problem, can you please create a draft MR with your changes so I can prepare my Yocto build for it?

> The shadow change seems useful, but I'd check `[[ $UID != 0 ]]` instead.

I have changed this.

chewi:
> If that's not a big problem, can you please create a draft MR with your changes so I can prepare my Yocto build for it?

Greetings, fellow cross-compiler! :smile: As requested, please see #1260. I am doing some final tests and hope to un-draft it later today. I would really appreciate your feedback.

PR #1259, opened 2025-03-26
fix(multipath): skip default multipath.conf with mpathconf

bmarzins:
Commit 1e802f15f creates a default multipath.conf file with "find_multipaths strict" when run in non-hostonly mode if there are no multipath devices and no multipath.conf. Unfortunately for systems that want to use mpathconf to create a multipath.conf file (e.g. Fedora and Centos) either through multipathd-configure.service or multipathd.sh, this default file keeps that from occurring. To make sure mpathconf is called to create the config file, do not install a default config file if mpathconf is installed.

Fixes: ("fix(multipath): include module with find_multipaths strict")

This pull request changes...

## Checklist

Fixes #509
CC: @pvalena @LaszloGombos @mwilck @jlebon

### DIFF
--- a/modules.d/90multipath/module-setup.sh
+++ b/modules.d/90multipath/module-setup.sh
@@ -122,7 +122,7 @@ install() {
         fi
     }
 
-    [[ $hostonly ]] || {
+    [[ $hostonly ]] || mpathconf_installed || {
         for_each_host_dev_and_slaves is_mpath \
             || [[ -f /etc/multipath.conf ]] || {
             cat > "${initdir}"/etc/multipath.conf << EOF


modules.d/90multipath/module-setup.sh
```diff
@@ -122,7 +122,7 @@ install() {
         fi
     }
 
-    [[ $hostonly ]] || {
+    [[ $hostonly ]] || mpathconf_installed || {
```
jlebon, line 125: I guess technically the conditional below for installing `multipathd-configure.service` is also gated on the `systemd` module being installed. Should we be extra strict here and also check for it in this conditional?

jlebon (approved):
One comment but LGTM!

modules.d/90multipath/module-setup.sh
mwilck: That would be an extra PR. It surely makes no sense to install `.service` files without systemd, but it doesn't do much harm either. As @bmarzins already said, mpathconf is used only in Red Hat distributions, and they all use systemd as well.

mwilck (approved):
LGTM

modules.d/90multipath/module-setup.sh
jlebon: Yup, WFM.

pvalena:
LGTM.

pvalena (approved):

Conan-Kudo (approved):

PR #1262, opened 2025-03-31
fix(crypt): always install s390 crypto modules

aafeijoo-suse:
Not only are the specific encryption modules not being installed on s390/s390x in non-hostonly mode, but also it's failing the detection of the kernel modules needed to decrypt PAES-encrypted volumes in hostonly mode via dmsetup. So, since the increase in size is not that much (~ 150K), always add all the drivers under s390/crypto in this architecture.

## Checklist

### DIFF
--- a/modules.d/90crypt/module-setup.sh
+++ b/modules.d/90crypt/module-setup.sh
@@ -23,7 +23,13 @@ depends() {
 
 # called by dracut
 installkernel() {
-    hostonly="" instmods drbg dm_crypt
+    local _arch=${DRACUT_ARCH:-$(uname -m)}
+    local _s390drivers=
+    if [[ $_arch == "s390" ]] || [[ $_arch == "s390x" ]]; then
+        _s390drivers="=drivers/s390/crypto"
+    fi
+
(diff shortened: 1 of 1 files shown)


hfreude:
Well ... Steffen Maier pointed me to this commit.
I am fine with adding all the kernel modules from drivers/s390/crypto but for paes to work (and that is for example a rootfs encrypted with protected key with dm-crypt) at least s390_paes (from arch/s390/crypto) is also needed.

aafeijoo-suse:
> Well ... Steffen Maier pointed me to this commit. I am fine with adding all the kernel modules from drivers/s390/crypto but for paes to work (and that is for example a rootfs encrypted with protected key with dm-crypt) at least s390_paes (from arch/s390/crypto) is also needed.

Those are already being added elsewhere:

https://github.com/dracut-ng/dracut-ng/blob/c43b79056ffdb7b410e70550a8ad8d137b4720c0/modules.d/90kernel-modules/module-setup.sh#L135-L141

Conan-Kudo (approved):

PR #1263, opened 2025-04-01
feat: add simpledrm module (as subset of drm module)

bdrung:
## Changes

Plymouth doesn't always show a splash screen if DRM drivers are installed in initrd.

Provide a `simpledrm` module that only installs the SimpleDRM module and the potentially needed privacy screen providers. This `simpledrm` module is a subset of the `drm` module. It could be used instead of `drm` to avoid pulling in drivers like amdgpu, nouveau, or nvidia-drm.

Bug-Ubuntu: https://launchpad.net/bugs/2105377

## Checklist

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -551,6 +551,10 @@ shutdown:
   - changed-files:
       - any-glob-to-any-file: 'modules.d/[0-9][0-9]shutdown/*'
 
+simpledrm:
+  - changed-files:
+      - any-glob-to-any-file: 'modules.d/[0-9][0-9]simpledrm/*'
+
 squash:
   - changed-files:
       - any-glob-to-any-file: 'modules.d/[0-9][0-9]squash/*'
--- a/modules.d/45simpledrm/module-setup.sh
+++ b/modules.d/45simpledrm/module-setup.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+# called by dracut
+check() {
+    return 255
+}
+
+# called by dracut
+installkernel() {
+    # Include simple DRM driver
+    instmods simpledrm
+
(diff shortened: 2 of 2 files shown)


bdrung:
Okay. Removed simplefb.

aleasto:
Does dracut have some magic to resolve module dependencies automatically, or do we need to track down the dependency tree for simpledrm.ko?

bdrung:
Dracut can resolve module dependencies, but I do not see that the simpledrm module on plucky would depend on anything. Or are these dependencies not recorded when built-in?

bdrung:
> We should probably add the usual metadata for a new module in `labaler.yml` - see e.g. [7cf47b2#diff-a22c263686553013feaeb0677d26eeb0b8778a756c4311c1fce13384258026aaR73-R76](https://github.com/dracut-ng/dracut-ng/commit/7cf47b26fe2e8a1a6fcac301db74ae3dbbf523b0#diff-a22c263686553013feaeb0677d26eeb0b8778a756c4311c1fce13384258026aaR73-R76)

Done.

> [bike-shedding](https://en.wikipedia.org/wiki/Law_of_triviality) - should the name of this dracut module be `simpledrm` instead of `simple-drm` for simplicity (and match the name of the kernel module) ?

Makes sense. Renamed to `simpledrm`.

> > `simple-drm` could be used instead of `drm`
> 
> This use-case is understood, but we should be clear if `simple-drm` _and_ `drm` module can be both generated into the initrd and coexists.

I updated the commit message to say that `simpledrm` is a subset of `drm`. So installing both is the same as installing only `drm`.

Conan-Kudo (approved):
LGTM. FYI @jwrdegoede

Conan-Kudo (approved):

PR #1211, opened 2025-02-09
fix(modules): all modules with 99 ordering should have a unique number

jozzsi:
## Changes

This PR does not change the ascending sort order of the modules and only changes modules in the range of `60-99`.

As discussed in the corresponding bug, this PR also tries to assign most dracut modules the sorting order `70`.

After the PR, the highest number for dracut module is `89`.

Partially resolve https://github.com/dracut-ng/dracut-ng/issues/538#issuecomment-2487284279

[script to reproduce the ordering changes to experiment with this PR downstream and/or packaging](https://github.com/user-attachments/files/19766940/rename.txt)

<details>
  <summary>
    NEW Module ordering after the PR (for the changed modules)
  </summary>

- 68cms                                                                                                                                                                                                      
- 68lvmmerge                                                                                                                                                                                                 
- 68lvmthinpool-monitor                                                                                                                                                                                      
- 68systemd-sysusers                                                                                                                                                                                         
- 69cio_ignore                                                                                                                                                                                               
- 70bluetooth                                                                                                                                                                                                
- 70btrfs                                                                                                                                                                                                    
- 70crypt                                                                                                                                                                                                    
- 70dm                                                                                                                                                                                                       
- 70dmraid                                                                                                                                                                                                   
- 70dmsquash-live                                                                                                                                                                                            
- 70dmsquash-live-autooverlay                                                                                                                                                                                
- 70dmsquash-live-ntfs                                                                                                                                                                                       
- 70kernel-modules                                                                                                                                                                                           
- 70kernel-modules-extra                                                                                                                                                                                     
- 70kernel-network-modules                                                                                                                                                                                   
- 70livenet                                                                                                                                                                                                  
- 70lvm                                                                                                                                                                                                      
- 70mdraid                                                                                                                                                                                                   
- 70multipath                                                                                                                                                                                                
- 70numlock                                                                                                                                                                                                  
- 70nvdimm                                                                                                                                                                                                   
- 70overlayfs                                                                                                                                                                                                
- 70pcmcia                                                                                                                                                                                                   
- 70ppcmac                                                                                                                                                                                                   
- 70qemu                                                                                                                                                                                                     
- 70qemu-net                                                                                                                                                                                                 
- 70test                                                                                                                                                                                                     
- 70test-makeroot                                                                                                                                                                                            
- 70test-root                                                                                                                                                                                                
- 73crypt-gpg                                                                                                                                                                                                
- 73crypt-loop                                                                                                                                                                                               
[...]

PR #1194, opened 2025-01-31
feat(initqueue): factor out initqueue into its own module

jozzsi:
This PR allows excluding the `initqueue` module, while including the `base` or the `dracut-systemd` module.                                                                                                                                                                                      
                                                                                                                                                                                                                                      
Factoring out functionality into its own module allows to better control to include or exclude the functionality. 

## Checklist
- [ ] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

CC @keszybz @dtardon @lnykryn 

Fixes #1191 #187

PR #1268, opened 2025-04-15
fix(systemd-sysusers): remove (g)shadow created by systemd-sysusers

### LINKED ISSUE #1242: cpio: etc/gshadow: Cannot open: Permission denied
**Describe the bug**
initramfs build failure

**Distribution used**
Arch Linux

**Dracut version**
Regression bisected to v106 commit f3dacc013d90bd2c

**Init system**
systemd

**To Reproduce**

~/dracut-ng/dracut.sh -l --force --verbose --no-hostonly --show-modules --kver=6.14.0-rc1-00047-g0a14566be090-dirty --kmoddir ./qbuild/mkosi.extra/lib/modules/6.14.0-rc1-00047-g0a14566be090-dirty/ --kernel-image ./vmlinux --add 'bash systemd kernel-modules fs-lib' --omit 'iscsi fcoe fcoe-uefi' --omit-drivers 'nfit libnvdimm nd_pmem' ./qbuild/mkosi.extra/boot/initramfs-6.14.0-rc1-00047-g0a14566be090-dirty.img 

**Expected behavior**

initramfs is built

jozzsi:
## Changes

The files `shadow` and `gshadow` do NOT exist when using the commit before f3dacc0.

This commit restores the earlier behavior.

## Checklist

Fixes #1242

### DIFF
--- a/modules.d/68systemd-sysusers/module-setup.sh
+++ b/modules.d/68systemd-sysusers/module-setup.sh
@@ -23,4 +23,8 @@ install() {
         set -o pipefail
         systemd-sysusers --root="$initdir" 2>&1 >&3 | grep -v "^Creating " >&2
     } 3>&1
+
+    # delete shadow files as initramfs is not designed to ask for a non-root user password
+    # interactively and their permissions crashes the build.
+    rm -f "$initdir/etc/shadow" "$initdir/etc/gshadow"
 }


marc-hb (commented):
I have no idea whether this is the right approach or not but this does fix the build for me.

modules.d/99systemd-sysusers/module-setup.sh
```diff
@@ -23,4 +23,7 @@ install() {
         set -o pipefail
         systemd-sysusers --root="$initdir" 2>&1 >&3 | grep -v "^Creating " >&2
     } 3>&1
+
+    # delete shadow files created by systemd-sysusers
+    rm -f "$initdir/etc/shadow" "$initdir/etc/gshadow"
```
marc-hb, line 28: Maybe the comment should explain why? Not just paraphrase the code.

modules.d/99systemd-sysusers/module-setup.sh
```diff
@@ -23,4 +23,7 @@ install() {
         set -o pipefail
         systemd-sysusers --root="$initdir" 2>&1 >&3 | grep -v "^Creating " >&2
     } 3>&1
+
+    # delete shadow files as initramfs is not designed to ask for a non-root user password interactively
```
marc-hb, line 27:
```suggestion
    # delete shadow files as initramfs is not designed to ask for a non-root user password
    # interactively and their permissions crashes the build.
```

marc-hb: Did you check #1242? It may be "just" a missing permission but there's no image produced. "crash" = showstopper = nothing works.

zboszor:
> @zboszor . Szia. I can see that you also worked on this bug with a different approach. Do you have any feedback/thoughts on this PR ? I am personally open to both approaches, but slightly prefer to this PR due to less complexity (less conditionals).

I think this is better than my approach, assuming the (empty?) passwords in `/etc/{g,}shadow` are not used during boot from the initramfs.

chewi:
Does this not break the root password set in 99base for the emergency console in host-only mode?

marc-hb:
> This PR has a dependency on #1211

The chain of dependency now leads to (pretty big) #1194.  I don't know anything about dracut but I find the lack of "urgency" somewhat disappointing for such a critical bug. I bisected, pinpointed and filed regression #1242 two months ago now. I mean it's disappointing that no "quick" and temporary fix (like this one) was possible in the mean time. It's common to see quick and small fixes reverted in the PR of the "real" and bigger fix - but not here? I hope this won't increase the span of buggy releases too much. I see v107 was just released - apparently with this bug? Release notes don't have any list of "known issues". 

I'm also wondering about test coverage... I didn't feel like I was doing anything extraordinary in #1242, was I? Reproduction is a single `dracut` command. How come no one else saw this?

> Temporarily closing, plan to reopen

Demoting to a GitHub "draft" works well in situations like this one. Drafts cannot be merged and can very easily be filtered out. My 2 cents; you do you.

marc-hb:
>  I think most people run Dracut as root anyway, so maybe that's why this hasn't been treated so urgently. 

There's "people" and there's CI/automation. I see a lot of checks run in every PR, which looks nice. How come none of these caught this? Do they all run as root too?

> Maybe they shouldn't, but that's another matter.

"People" is another matter. CI is not.

marc-hb (approved):

bdrung (approved):
Approved from my side under one condition: Someone verified that this change does not break root password set in base module for the emergency console in host-only mode.

marc-hb:
I stand corrected, thanks @jozzsi for your patience. I clearly spent too much time playing with scripts and not enough time looking at documentation.

I would have placed https://dracut-ng.github.io/dracut-ng/developer/hacking.html#_testsuite before commit messages, coding style and code of conduct but that's just me :-)

marc-hb:
>  Having a test/README.md sounds like a good idea. It can point to the right location or just be a symlink.

- #1454

PR #1279, opened 2025-04-21
ci: install packages needed to test upcoming dlopen features

chewi:
As requested in #1260.

systemd's fido2 feature is used for testing because nothing else is likely to pull in libfido2.so.1.


chewi:
> There is a recent commit for a new container for azurelinux - #1244

Good point, thanks. I've added systemd-devel there. Azure Linux doesn't have a libfido2 package.

PR #1281, opened 2025-04-21
man(usage): clarify hostonly

jozzsi:
## Changes

Attempt to clarify hostonly.

Original definition is from 2013 from this commit - b6c897681

## Checklist

Fixes #

### DIFF
--- a/man/dracut.usage.adoc
+++ b/man/dracut.usage.adoc
@@ -124,6 +124,21 @@ raid with encryption and LVM on top), as long as you specify the correct
 filesystem LABEL or UUID on the kernel command line for your root device, dracut
 will find it and boot from it.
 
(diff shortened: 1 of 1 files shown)


dalto8:
This definition is clear, very much needed and helpful for further discussions/commits.

However, it further highlights that there is a missing mode in dracut that many of us need.

Conan-Kudo:
We probably need a third mode that excludes configuration. `non-hostonly` does not reasonably imply "no host configuration". Then distributions can decide whether their equivalents of `dracut-config-generic` and `dracut-config-rescue` should include this or not.

Conan-Kudo:
That is not what I'm saying. I'm saying what the code does and this definition are not reasonable paths forward and we will need to fix it.

Conan-Kudo (approved):

PR #1283, opened 2025-04-22
fix: improve hostonly sloppy mode

jozzsi:
## Changes

Sloppy mode is the original hostonly mode.
    
Strict mode was newly introduced much later in 2018 ([v048](https://github.com/dracut-ng/dracut-ng/blob/main/NEWS.md#dracut-048)) as a non-default opt-in special optimization mode to allow modules to ignore more drivers or do some extra job to save memory and disk space, while making the image less portable.
    
This PR includes more dracut modules (and with that more kernel modules) in sloppy mode.

## Checklist

Fixes #

### DIFF
--- a/modules.d/99fs-lib/module-setup.sh
+++ b/modules.d/99fs-lib/module-setup.sh
@@ -36,7 +36,7 @@ include_fs_helper_modules() {
 # called by dracut
 installkernel() {
     # xfs/btrfs/ext4 need crc32c, f2fs needs crc32
-    if [[ $hostonly ]]; then
+    if [[ $hostonly_mode == "strict" ]]; then
         for_each_host_dev_fs include_fs_helper_modules
         :
     else
@@ -59,7 +59,7 @@ install() {
             xfs_db xfs_check xfs_repair xfs_metadump
             e2fsck jfs_fsck btrfsck
         )
-        if [[ $hostonly ]]; then
+        if [[ $hostonly_mode == "strict" ]]; then
             read -r -a _helpers < <(for_each_host_dev_fs echo_fs_helper)
         fi
     else
--- a/modules.d/01systemd-cryptsetup/module-setup.sh
+++ b/modules.d/01systemd-cryptsetup/module-setup.sh
@@ -6,7 +6,7 @@ check() {
     # if cryptsetup is not installed, then we cannot support encrypted devices.
     require_binaries "$systemdutildir"/systemd-cryptsetup || return 1
 
-    [[ $hostonly ]] || [[ $mount_needs ]] && {
+    [[ $hostonly_mode == "strict" ]] || [[ $mount_needs ]] && {
         for fs in "${host_fs_types[@]}"; do
             [[ $fs == "crypto_LUKS" ]] && return 0
         done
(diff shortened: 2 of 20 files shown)


modules.d/95virtfs/module-setup.sh
```diff
@@ -2,7 +2,7 @@
 
 # called by dracut
 check() {
-    [[ $hostonly ]] || [[ $mount_needs ]] && {
+    [[ $hostonly_mode == "strict" ]] || [[ $mount_needs ]] && {
         for fs in "${host_fs_types[@]}"; do
             [[ $fs == "9p" ]] && return 0
         done
```
ArrayBolt3, line 8: The whole check function in this file seems a bit weird - it defaults to not including 9pfs support, but in hostonly mode it might include it if the system uses 9pfs, but now it will only follow that in strict hostonly mode. I'm not really sure I understand the intent here, and I suspect this will regress use cases that  depend on 9pfs in the wild.

Maybe this should be changed to match the 95virtiofs module's logic, which includes the module by default and has the `is_qemu_virtualized` check in the hostonly_strict block.

modules.d/99fs-lib/module-setup.sh
```diff
@@ -59,7 +59,7 @@ install() {
             xfs_db xfs_check xfs_repair xfs_metadump
             e2fsck jfs_fsck btrfsck
         )
-        if [[ $hostonly ]]; then
+        if [[ $hostonly_mode == "strict" ]]; then
```
ArrayBolt3, line 62: This is good, I think line 39 in `installkernel` also should change to a `hostonly_mode == "strict"` check. (Wish Github would let me comment on unchanged lines...)

ArrayBolt3 (commented):
I merged https://github.com/dracut-ng/dracut-ng/pull/1238 into this branch before testing it on a Trixie system with LUKS + LVM. The system is bootable with an initrd generated with those two PRs applied, so I think this is mostly good. I have not yet created an installation which breaks without this change, to verify that this change avoids the breakage, but I can give it a shot if that would be helpful.

Conan-Kudo (approved):

bdrung:
FTR, this change increases the initrd size significantly (see https://github.com/dracut-ng/dracut-ng/pull/1351#issuecomment-2908750506). Such kind of changes need a size evaluation before approving/merging IMO.

aafeijoo-suse:
> If you need an initramfs that's small and specific to the installation, use hostonly and strict.

That would be great if it actually worked.

IMHO the problem with this type of breaking and untested changes is the way dracut releases versions: it lacks release candidates, so when people upgrade to a new version they expect at least something stable. We were using dracut in openSUSE for +10 years (Fedora even more), so we expect that things work reliably.  If we get a different result with the same configuration we've [always](https://github.com/dracutdevs/dracut/commit/bcd352dfa105544a7531d999301c84bf88e850ed) [had](https://github.com/dracutdevs/dracut/commit/94920084351349cba579ce80a89321ab6544653c), that's seen as a regression by end users. And nowadays the initrd size is a concern since we start to follow BLS and the size of legacy ESPs is quite limited, so larger initrds mean less bootloader entries (in the openSUSE case, less btrfs snapshots).

So, I understand and fully support Benjamin's use case, but breaking the old one without having a real alternative is not a good option.

Another example, #1321 is a critical issue related to this change, fixed in git main, and Fedora rawhide just released 107 a few days ago without its patch on top...

PR #1291, opened 2025-04-23
fix: load more kernel modules in sloppy hostonly mode

### LINKED ISSUE #748: Please support hostonly mode with full portability

jozzsi:
## Changes

When `hostonly_mode` is set (e.g. to either `sloppy` or `strict`) `instmods` consults the host for currently loaded kernel modules and refuses to copy the Linux kernel module into the generated initramfs unless the kernel module is already loaded on the host.

Let's set `hostonly` variable explicitly to handle `sloppy` better before calling `instmods`.

For compatibility, we should not change the default in the `instmods` function itself.

The `hostonly=$(optional_hostonly) instmods` pattern is already used in some modules - e.g. https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/95nfs/module-setup.sh#L41

See also https://github.com/dracut-ng/dracut-ng/blob/main/dracut-init.sh#L380 

CC @dalto8  @ArrayBolt3  @bdrung @LaszloGombos @Conan-Kudo 

## Checklist

Fixes #748 https://github.com/dracutdevs/dracut/issues/2687

### DIFF
--- a/modules.d/45simpledrm/module-setup.sh
+++ b/modules.d/45simpledrm/module-setup.sh
@@ -8,9 +8,9 @@ check() {
 # called by dracut
 installkernel() {
     # Include simple DRM driver
-    instmods simpledrm
+    hostonly='' instmods simpledrm
 
-    if [[ $hostonly ]]; then
+    if [[ $hostonly_mode == "strict" ]]; then
         # if there is a privacy screen then its driver must be loaded before the
         # kms driver will bind, otherwise its probe() will return -EPROBE_DEFER
         # note privacy screens always register, even with e.g. nokmsboot
@@ -18,7 +18,7 @@ installkernel() {
             [[ -L $i ]] || continue
             modlink=$(readlink "$i")
             modname=$(basename "$modlink")
-            instmods "$modname"
+            hostonly='' instmods "$modname"
         done
     else
         # include privacy screen providers (see above comment)
--- a/man/dracut.modules.7.adoc
+++ b/man/dracut.modules.7.adoc
@@ -332,6 +332,15 @@ To install a kernel module regardless of the hostonly mode use the form:
 hostonly='' instmods <kernelmodule>
 ----
 
+a common pattern to call instmods is to use the following form:
+----
+`hostonly=$(optional_hostonly) instmods <kernelmodule>`
+----
+This statement installs the kernel module in either of the following conditions:
+* in non-hostonly mode (regardless of the host state)
+* in `sloppy` hostonly mode (regardless of the host state)
+* in hostonly mode if the kernel module is already loaded in the host
+
 options:
 
 -c:: check that kernel modules exists and can be installed (i.e., not optional).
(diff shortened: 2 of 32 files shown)


modules.d/45simpledrm/module-setup.sh
```diff
@@ -8,17 +8,17 @@ check() {
 # called by dracut
 installkernel() {
     # Include simple DRM driver
-    instmods simpledrm
+    hostonly='' instmods simpledrm
 
-    if [[ $hostonly ]]; then
+    if [[ $hostonly_mode != "strict" ]]; then
```
ArrayBolt3, line 13: I think the `!=` is supposed to be an `=` here. You want to restrict the modules in strict mode, but not in sloppy mode and definitely not in `hostonly=no` mode. Right now this will run even when `hostonly=no` is set, if I'm understanding correctly.

ArrayBolt3:
This works for me in a Trixie virtual machine (I can boot an encrypted installation), but I also am now for the first time running into https://github.com/dracutdevs/dracut/issues/723 for some reason. Unsure if the two are related.

ArrayBolt3:
OK, nevermind about the bug I ran into, it must be user error because even Debian's vanilla dracut package is doing it to me now.

Conan-Kudo:
@ArrayBolt3 @bdrung could you test to see if this makes things work the way you expect it to?

ArrayBolt3:
Will give it another check probably later today.

ArrayBolt3:
@Conan-Kudo @jozzsi Works fine for me in a Debian Trixie XFCE VM.

Conan-Kudo (approved):
I don't see any reason not to land this now. This PR still makes me nervous, but let's go for it anyway.

PR #1296, opened 2025-04-25
fix(squash-erofs): fix configuration in order to match SquashFS

hsiangkao:
This pull request changes...

## Changes
 - `-C 65536` should be specified to match the default setup `-b 131072` for SquashFS;

 - `-x -1` should be specified to match the SquashFS `-no-xattrs`;

 - `-U 00000000-0000-0000-0000-000000000000` for reproducibility;

 - `-T 0` can be specified to clear all inode timestamps for reduced metadata since I'm not sure timestamps are needed for dracut.

## Checklist

Fixes #

### DIFF
--- a/modules.d/95squash-erofs/module-setup.sh
+++ b/modules.d/95squash-erofs/module-setup.sh
@@ -22,13 +22,21 @@ erofs_installpost() {
 
     # --exclude-path requires a relative path
     _erofs_args+=("--exclude-path=${squashdir#"$initdir"/}")
+    # In order to match the default SquashFS "block size"
+    _erofs_args+=("-C" "65536")
     _erofs_args+=("-E" "fragments")
+    # In order to match the SquashFS `-no-xattrs`
+    _erofs_args+=("-x" "-1")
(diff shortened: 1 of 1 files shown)


hsiangkao:
@licliu could you please check the new image sizes out of this commit?

licliu:
> @licliu could you please check the new image sizes out of this commit?

In my tests, after applying this patch, the size of erofs-root.img was reduced by about 13%. Local kdump was reduced from 32M to 28M, and nfs kdump was reduced from 43M to 37M.

hsiangkao:
> > @licliu could you please check the new image sizes out of this commit?
> 
> In my tests, after applying this patch, the size of erofs-root.img was reduced by about 13%. Local kdump was reduced from 32M to 28M, and nfs kdump was reduced from 43M to 37M.

Ok, that is `lz4hc` result. I think you could try to use `lzma` or `deflate` to see if it already meets your requirement.

Conan-Kudo (approved):
I don't necessarily agree with all the defaults here, but they _do_ pretty much match squashfs, so we can fix it later for both.

PR #1304, opened 2025-04-27
feat: Add support for removing files/directories from initramfs

### LINKED ISSUE #1297: Add support for --omit-files
**Describe the enhancement**

`--omit-files [LIST] Omit a space-separated list of files and directories from initramfs.`

jozzsi:
## Changes

Add support for removing files/directories from initramfs.

This enhancement adds a configuration option which mirrors the                                                                                                                                                                        
behavior of the `--include` command line option, but instead if adding                                                                                                                                                                  
files/directories, it removes them. 

## Checklist

Fixes #1297

CC @LaszloGombos @ericcurtin @nagmat84

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -233,6 +233,8 @@ Creates initial ramdisk images for preloading modules
                          in the final initramfs.
   -I, --install [LIST]  Install the space separated list of files into the
                          initramfs.
+  --remove      [LIST]  Remove a space-separated list of files and directories
+                        from the initramfs.
   --install-optional [LIST]
                         Install the space separated list of files into the
                          initramfs, if they exist.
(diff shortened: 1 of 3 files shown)


bdrung:
The commit message could be a little bit more verbose. The title "Remove a space-separated list of files" is confusing on the first read. You do not remove a space-separated list, you remove files/directories (and those are specified with spaces).

dracut.sh
```diff
@@ -661,6 +664,11 @@ while :; do
             PARMS_TO_STORE+=" '$2'"
             shift
             ;;
+        -R | --remove)
```
bdrung: `-R` is accepted here, but not documented anywhere. Since using this option is not recommended, I suggest to remove `-R` and only provide the long option.

jozzsi: Great catch, thanks @bdrung

bdrung (approved):

PR #1306, opened 2025-04-28
fix(multipath): disable user_friendly_names with mpathconf

bmarzins:
If dracut is creating /etc/multipath.conf by calling mpathconf in either multipathd-configure.service or multipathd.sh, there is a chance that the multipath config in the real root differs. Specifically, it might have chosen different user_friendly_names for the devices. When the systems switches to the real root, multipath may not be able to switch the devices to their configured names because those might already be in use. To avoid this, call mpathconf with "--user_friendly_names n" to create a multipath.conf with user_friendly_names disabled. If all devices use WWID names, it is always possible for multipath to rename them later.

Fixes b8a92b715 ("multipath: add automatic configuration for multipath")

This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/modules.d/90multipath/multipathd-configure.service
+++ b/modules.d/90multipath/multipathd-configure.service
@@ -15,7 +15,7 @@ Type=oneshot
 RemainAfterExit=yes
 # mpathconf requires /etc/multipath to already exist
 ExecStartPre=-/usr/bin/mkdir -p /etc/multipath
-ExecStart=/usr/sbin/mpathconf --enable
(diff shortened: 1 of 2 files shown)


mwilck:
Ack, or rather non-nack, from my side. mpathconf is RH/Fedora specific, and this PR only touches the mpathconf code path.

pvalena:
LGTM, thanks!

jlebon:
Thanks, overall ack from me, though we'll get this tested on RHCOS and FCOS to be sure.

Conan-Kudo (approved):

PR #1312, opened 2025-05-02
fix(simpledrm): add =drivers/gpu/drm/panel on ARM

bdrung:
## Changes

On Qualcomm CRD the display only turns on after the initrd. The initrd seems to be missing the panel module.

So install `=drivers/gpu/drm/panel` on ARM also in the simpledrm module (similar to the drm module).

Bug-Ubuntu: https://launchpad.net/bugs/2107477

## Checklist

### DIFF
--- a/modules.d/45simpledrm/module-setup.sh
+++ b/modules.d/45simpledrm/module-setup.sh
@@ -8,7 +8,7 @@ check() {
 # called by dracut
 installkernel() {
     # Include simple DRM driver
-    hostonly='' instmods simpledrm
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
I'm not sure this would be ARM specific? Do we know for certain the panel drivers are an ARM-only thing?

bdrung:
The DRM module add `=drivers/gpu/drm/panel` only on ARM, but I see no compelling argument why this should be ARM specific. So I changed the code to include it on all architectures.

Conan-Kudo (approved):

PR #1322, opened 2025-05-12
fix: improve hostonly sloppy mode

### LINKED ISSUE #1321: Dracut 107 does not include kernel modules for the root filesystem
**Describe the bug**
When building a hostonly initrd for a kernel which requires kernel modules to be included for the root filesystem **AND** the running kernel has the filesystem drivers built-in the modules are not included.  This leads to an unbootable system since the root filesystem cannot be loaded.

**Distribution used**
EndeavourOS

**Dracut version**
The issue is present in 107 and the current source tree but not present in 106

**Init system**
systemd

**To Reproduce**
On EndeavourOS(or Arch Linux), install the `linux` and `linux-lts` kernel packages.  Boot into the linux kernel and then generate an initrd for the lts kernel.  It will be missing the modules.

**Expected behavior**
The expected behavior is that modules would be included as they were in Dracut 106.

**Additional context**
While this may appear to be an edge case, it is causing significant issues for EndeavourOS since:
* Some Arch kernels are built with ext4 support as modular and others not
* EndeavourOS includes both types of kernels in our installer
* We generally recommend installing both kernels

jozzsi:
## Changes

Resolve regression introduced by 8519dcd.

CC @dalto8 @LaszloGombos @Conan-Kudo 

## Checklist

Fixes #1321  .

### DIFF
--- a/modules.d/90kernel-modules/module-setup.sh
+++ b/modules.d/90kernel-modules/module-setup.sh
@@ -122,11 +122,11 @@ installkernel() {
 
         # if not on strict hostonly mode, install all known filesystems,
         # if the required list is not set via the filesystems variable
-        if [[ $hostonly_mode != "strict" ]]; then
-            if [[ -z $filesystems ]]; then
-                dracut_instmods -o -P ".*/(kernel/fs/nfs|kernel/fs/nfsd|kernel/fs/lockd)/.*" '=fs'
-            fi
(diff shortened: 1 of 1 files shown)


dalto8:
Is a regression really the PR we should be closing due to the review backlog?

dalto8:
> We encourage everyone to help with the review queue

How can we do that?  You need rights to approve.

Do we have enough approvers?

PR #1330, opened 2025-05-17
refactor(dracut-install): copy_xattr: misc cleanups

marcosfrm:
- Return early when no xattrs are present.
- Use _cleanup_free_ for buffers and remove unnecessary variable initializations.
- Replace calloc() with malloc() as flistxattr() overwrites the buffer and ensures null-terminated xattr names.
- Update the loop to more idiomatic iteration over null-terminated strings.

@HuaxinLu

## Checklist

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -329,22 +329,26 @@ static inline int clone_file(int dest_fd, int src_fd)
 static int copy_xattr(int dest_fd, int src_fd)
 {
         int ret = 0;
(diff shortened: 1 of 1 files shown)


bdrung (changes requested):
LGTM except that the second commit contains two changes which made it harder to review. Can you split the second commit to have separate commits for each bullet point from the git commit message?

marcosfrm:
Done.

bdrung (approved):
Thanks.

PR #1332, opened 2025-05-18
feat: If it is a new Fedora Release regenerate rescue entry

sergiomb2:
Propose to fix https://bugzilla.redhat.com/show_bug.cgi?id=1768132#c47 

This pull request do this: 

read -r MACHINE_ID < /etc/machine-id
FILE_ENTRY=/boot/loader/entries/${MACHINE_ID}-6.14.6-200.fc41.x86_64.conf
LOADER_ENTRY=/boot/loader/entries/${MACHINE_ID}-0-rescue.conf
line1=$(head -n 1 "$FILE_ENTRY")
line2=$(head -n 1 "$LOADER_ENTRY")
line1_mod=$(echo "$line1" | sed "s/$(uname -r)/0-rescue-${MACHINE_ID}/")
if [ "$line1_mod" != "$line2" ]; then
  rm /boot/vmlinuz-0-rescue-${MACHINE_ID}
  rm /boot/initramfs-0-rescue-${MACHINE_ID}.img 
fi

## Checklist

Fixes #
https://bugzilla.redhat.com/show_bug.cgi?id=1768132


sergiomb2:
> When we regenerate the rescue image, are we effectively overwriting the old one? 

This commit check if rescue is from the same version of Fedora. If rescue is from an previous version, we just delete the rescue image.  Because without any rescue image script already creates a new rescue image

dtardon:
> Hi, if you want move rescue generation to other place fine , but my PR just intend regenerate rescue entry after upgrade Fedora version it is an old bug

Exactly. It's an _old_ problem. It's existed for 12 years without much attention from anyone, which suggests it's not that urgent. So maybe we can think a bit and _then_ try to do something reasonable rather that put in a quick hack _right now_.

sergiomb2:
The bug is "only" 5 years and 8 months old :smile: . And is not urgent, IMO, not because haven't much attention, but because we have one workaround, which is remove old rescue manually `rm /boot/*-0-rescue-*` . This PR propose the same idea, remove old rescue when detect one Fedora version upgrade

devkontrol:
It seems that this PR did not receive support from Fedora and Fedora is heavily patching rescue image generations [it seems](https://src.fedoraproject.org/rpms/dracut/blob/rawhide/f/0001-revert-fix-install.d-correctly-install-pre-genned-im.patch), so I'd like to close this PR at this time. Please reopen if you'd like to continue this work (or open a new PR).

PR #1333, opened 2025-05-18
feat: set hostonly_cmdline config to no by default

### LINKED ISSUE #1326: consider making hostonly_cmdline default to no
Perhaps the current logic should be only applied in non-strict mode - see

https://github.com/dracut-ng/dracut-ng/blob/main/dracut.sh#L1092

jozzsi:
## Changes

`hostonly_cmdline=no` is already the default on Fedora, see https://github.com/dracut-ng/dracut-ng/blob/main/dracut.conf.d/fedora.conf.example#L44

## Checklist

Fixes #1326

CC @LaszloGombos @bdrung @Nowa-Ammerlaan @AlexanderKurtz @dalto8

### DIFF
--- a/dracut.conf.d/hostonly/50-hostonly.conf
+++ b/dracut.conf.d/hostonly/50-hostonly.conf
@@ -1,2 +1,3 @@
 # optimize initrd to be as small as possible for faster boot performance, tailored to the current host
 hostonly="yes"
+hostonly_cmdline=no


dalto8:
It seems unintuitive that the default would change only when `hostonly_mode` is `strict`

Nowa-Ammerlaan (approved):

dalto8 (approved):

PR #1334, opened 2025-05-18
feat: tests are not packaged by default to simplify packaging

### LINKED ISSUE #1327: consider installing testsuite opt in instead of opt-out
discuss changing the default at https://github.com/dracut-ng/dracut-ng/blob/main/Makefile#L179

As an example openSUSE seems to exclude tests during packaging - https://github.com/openSUSE/dracut/blob/SUSE/059/suse/dracut.spec#L398

https://github.com/openSUSE/dracut-ng/blob/SUSE/107/suse/dracut.spec#L176

jozzsi:
## Changes

If it is desired to package tests, set `enable_test=yes` .

## Checklist

Fixes #1327

CC @bdrung @Nowa-Ammerlaan @LaszloGombos

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ install: all
 	for i in $(configprofile) ; do \
 		cp -arx dracut.conf.d/$$i/* $(DESTDIR)$(pkglibdir)/dracut.conf.d/ ;\
 	done
-ifneq ($(enable_test),no)
+ifeq ($(enable_test),yes)
 	cp -arx test $(DESTDIR)$(pkglibdir)
 else
 	rm -rf $(DESTDIR)$(pkglibdir)/modules.d/80test*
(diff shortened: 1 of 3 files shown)


Nowa-Ammerlaan (commented):
For readability I would flip around this if, I don't like the double negation:
``
ifneq ($(enable_test),no)
       cp -arx test $(DESTDIR)$(pkglibdir)
else
       rm -rf $(DESTDIR)$(pkglibdir)/modules.d/80test*
endif
```

ifeq ($(enable_test),yes)
       cp -arx test $(DESTDIR)$(pkglibdir)
else
       rm -rf $(DESTDIR)$(pkglibdir)/modules.d/80test*
endif
```

Conan-Kudo:
I'd rather not change this default. Exposing an option makes sense though.

bdrung (approved):
+1 from me. Debian/Ubuntu is running the upstream tests, but we do not install the tests. With this MR we can simplify our debian/rules and remove:
```
# For "make install" to not install files needed for tests
export enable_test=no
```

PR #1348, opened 2025-05-24
fix: load essential storage kernel modules in sloppy hostonly mode

### LINKED ISSUE #748: Please support hostonly mode with full portability

jozzsi:
## Changes

Essential storage kernel modules (e.g. sd_mod, nvme) should be always included unless hostonly_mode is set to strict.
    
This PR installs even more drivers in hostonly sloppy mode as a follow-up to 8519dcd .

These drivers are already always installed in `non-hostonly` (`generic`) mode.
    
## Checklist

Fixes #748

### DIFF
--- a/modules.d/70kernel-modules/module-setup.sh
+++ b/modules.d/70kernel-modules/module-setup.sh
@@ -107,9 +107,9 @@ installkernel() {
 
         awk -F: '/^\// {print $1}' "$srcmods/modules.dep" 2> /dev/null | instmods
 
-        # if not on hostonly mode, or there are hostonly block device
+        # if not on strict hostonly mode, or there are hostonly block device
         # install block drivers
-        if ! [[ ${hostonly-} ]] \
+        if [[ $hostonly_mode != "strict" ]] \
             || for_each_host_dev_and_slaves_all record_block_dev_drv; then
             hostonly='' instmods sg sr_mod sd_mod scsi_dh ata_piix
 


dalto8 (approved):

dalto8 (approved):

bdrung (commented):
Which additional modules will be installed by this change on common setups? How much impact will that have on the initrd size?

PR #1349, opened 2025-05-24
feat(rootfs-block-fallback): factor out rootfallback into its own module

jozzsi:
## Changes

This change allows to remove the initqueue dependency from rootfs-block.

initqueue will be only included into the initramfs when rootfs-block-fallback is explicitelly included.

Follow-up to 3daf67830 

## Checklist

CC @dtardon


dtardon:
FTR, I've been thinking about how to do this using systemd units and the simplest solution I've come with that may work is this:
1. Create a service (`rootfallback.service` or so) that runs the `rootfallback.sh`.
2. Create a drop-in that adds `OnFailure=rootfallback.service` for `sysroot.mount`. (Possibly do it via a generator only if there's any `rootfallback=` on the command line.)

I'm not sure it would work, but if anyone--including potential future myself--wants to play with this, this is probably the direction to follow.

PR #1350, opened 2025-05-24
test: fix the dlopen dependencies test on Gentoo

### LINKED ISSUE #1347: Test 40 seems to fail on Gentoo
`Error: libfido2.so.1 should have been included in the initramfs`

CC @chewi 

See e.g. https://github.com/dracut-ng/dracut-ng/actions/runs/15198941918

chewi:
I have done the equivalent of running the test on my local Gentoo system to verify that the lack of `USE=fido2` was the problem.

Fixes #1347


bdrung (approved):
LGTM

PR #1351, opened 2025-05-25
chore: do not maintain suse.conf upstream

jozzsi:
## Changes

There is a trend to no longer maitain distro specific configuration here upstream. Arch and Gentoo configuration has been already removed from upstream, let's also remove suse.conf.example.

It seems this file is changed downstream and the changes are not contributed back upstream - https://github.com/openSUSE/dracut-ng/commit/2eda92988d71d52d3bdb62c06819b7c2345b3c9a

## Checklist


aafeijoo-suse:
> It seems this file is changed downstream and the changes are not contributed back upstream - [openSUSE@2eda929](https://github.com/openSUSE/dracut-ng/commit/2eda92988d71d52d3bdb62c06819b7c2345b3c9a)

The description of the repository that contains this commit says: `[WIP] openSUSE's branch of dracut-ng` (WIP stands for "Work In Progress").

Regarding that change, we were forced to add `hostonly_mode="strict"` because of your "fix" 8519dcdb154b55c5feb6e4638a525c33e8fa7f66

Because we like to test things before breaking end user systems, we noticed that with that "fix" the size of the initrd exploded. E.g.:
- Separate /boot, / luks2 encrypted: ~35M -> ~50M
- RAID1 + LVM: ~31M -> ~48M

And not only the size increased, but also the system failed to boot in some cases, already reported in #1321. That, plus also other errors (already reported) made the migration to dracut-ng-107 a noop in openSUSE for now.

> There is a trend to no longer maitain distro specific configuration here upstream.

Then remove `fedora.conf.example` as well.

Conan-Kudo:
I would actually prefer to reverse this trend, especially since we're doing distro testing now in CI. It was a bad idea to drop it back when it did happen in the original repository, and it's an even worse idea now that we're doing CI on the distributions for their default configurations.

PR #1355, opened 2025-05-25
fix(modules): move more modules with unimportant ordering to 70

jozzsi:
## Changes

Dracut library modules - such as fs-lib, img-lib or uefi-lib, do not need to be ordered differently from other regular modules.

## Checklist


bdrung (approved):
uefi-lib depends on bash. So moving it to 70 to be before 85shell-interpreter is a fix.

PR #1358, opened 2025-05-26
feat(base): add support for rd.driver.pre

### LINKED ISSUE #1302: --force-drivers does not work without kernel-modules
**Describe the bug**
force-drivers dracut argument does not work unless kernel-modules dracut module is included

`dracut --force-drivers ext4 -m "rootfs-block"` is incomplete as `rd.driver.pre` requires `kernel-modules dracut module`.

A better home for `rd.driver.pre` handling is the `base` module that is guaranteed to be installed into the initrd.

This change would also make the `kernel-modules` dracut modules behave the same way as `kernel-modules-extra` and `kernel-network-modules`

jozzsi:
## Changes

Move support for rd.driver.pre from kernel-modules to base module.

One of the reason for this is as rd.driver.pre has a dependency on modprobe binary and modprobe binary is installed in the base dracut module.

Another reason is to support the use case where kernel modules are installed via dracut command line and not via kernel-modules dracut library.

## Checklist

Fixes #1302


bdrung (approved):
LGTM

Issue #1311, opened 2025-05-02
`rngd` module not included


aafeijoo-suse:
**Describe the bug**
`rngd` module not included.

**Distribution used**
openSUSE Tumbleweed

**Dracut version**
107

**Additional context**

https://github.com/dracut-ng/dracut-ng/blob/01b369a5866853c8dc5c53c7d4d8613e5721be34/modules.d/06rngd/module-setup.sh#L24

There is no `rngd.service` in openSUSE, its name is `rng-tools.service`:

```
# rpm -ql rng-tools
/usr/bin/randstat
/usr/bin/rngtest
/usr/lib/systemd/system/rng-tools.service
/usr/lib/udev/rules.d/90-hwrng.rules
/usr/sbin/rcrng-tools
/usr/sbin/rngd
/usr/share/doc/packages/rng-tools
/usr/share/doc/packages/rng-tools/NEWS
/usr/share/doc/packages/rng-tools/README
/usr/share/licenses/rng-tools
/usr/share/licenses/rng-tools/COPYING
/usr/share/man/man1/rngtest.1.gz
/usr/share/man/man8/rngd.8.gz

```

aafeijoo-suse:
> Upstream calls it `rngd.service` - https://github.com/nhorman/rng-tools/pull/216/files

Yeah, who cares about specific code for downstream distros (in this case, breaking something that used to work before):
- https://github.com/dracut-ng/dracut-ng/pull/1117#pullrequestreview-2545598682
- https://github.com/dracut-ng/dracut-ng/pull/1306#issuecomment-2838883492
- https://github.com/dracut-ng/dracut-ng/pull/950
- ...

aafeijoo-suse:
BTW, out of curiosity, does this upstream version work in the initrd without `DefaultDependencies=no`?

PR #1361, opened 2025-05-26
fix(modules): free up range 00-09 to out of tree dracut modules

### LINKED ISSUE #538: free up 00-09 dracut module namespace
**Describe the enhancement**
Consider making 00modname (or even better 00modname-09modname) namespace dedicated for out of tree dracut modules.

Justification is to allow the same functionality for out-of-tree modules than in-tree modules without interference and with a level of compatibility promise. 

See https://github.com/dracut-ng/dracut-ng/blob/main/docs/HACKING.md#writing-modules

jozzsi:
## Changes

This PR does not change the ascending sort order of the modules.

It adds 10 to all dracut modules in the range of 00-10 to move it up to the 10-20 range.                                                                                                                                                                                                                   
                                                                                                                                                                                                                                      
The motivation for this PR is to have namespace (dracut module numbers) reserved at the beginning (similar to the end) of the range.  

## Checklist

Fixes #538


Conan-Kudo:
Is there a specific out-of-tree dracut module that needs this range freed up?

aafeijoo-suse:
> As a (not very good example) - #1311 can be worked around by openSUSE easily (meaning without maintaining patch files) downstream now as a 00rngd module can be created downstream that gets executed before the upstream 16rngd gets executed.

openSUSE is not going to add workarounds like that on top of dracut-ng.

doc_site/modules/ROOT/pages/developer/modules.adoc
```diff
@@ -14,9 +14,10 @@ so you should install it on your system -- dash aims for strict POSIX
 compliance to the extent possible.
 * Hooks MUST be POSIX compliant -- they are sourced by the init script,
 and having a bashism break your user's ability to boot really sucks.
-* Generator modules should have a two digit numeric prefix -- they run in
-ascending sort order. Anything in the 90-99 range is stuff that dracut
-relies on, so try not to break those hooks.
+* Generator modules should have a two digit numeric prefix in the range
+of 50-59 -- they run in ascending sort order.
+Anything in the 80-89 range is stuff that dracut relies on,
+so try not to break those hooks.
```
bdrung: This change should be in a separate commit fixing the 90-99 to 80-89 move.

man/dracut.modules.7.adoc
```diff
@@ -14,9 +14,9 @@ DESCRIPTION
 
 dracut uses a modular system to build and extend the initramfs image. All
 modules are located in _/usr/lib/dracut/modules.d_ or in _<git-src>/modules.d_.
-The most basic dracut module is _99base_. In _99base_ the initial shell script
+The most basic dracut module is _80base_. In _80base_ the initial shell script
```
bdrung: This change should be in a separate commit fixing the 90-99 to 80-89 move.

jozzsi: Landed as 7b2296e . Thanks for the review @bdrung .

doc_site/modules/ROOT/pages/developer/modules.adoc
jozzsi: Landed as 7b2296e . Thanks for the review @bdrung .

bdrung (approved):

PR #1364, opened 2025-05-27
feat(watchdog,resume): do not depend on initqueue if systemd is used

### LINKED ISSUE #1362: `resume` module should not depend on `initqueue` module if systemd is being used
https://github.com/dracut-ng/dracut-ng/blob/e14520035286dbb4ee856cac2e05c1f21c6e148d/modules.d/74resume/module-setup.sh#L67-L75

jozzsi:
## Changes

- feat(resume): do not depend on initqueue if systemd is used
- feat(watchdog): do not depend on initqueue if systemd is used
- perf(watchdog): only install wdctl for the non-systemd case

## Checklist

Fixes #1362 

CC @dtardon


bdrung:
The last commit says:
> Move the installation of wdctl in install() to the non-systemd block.

This does not explain the "Why?".

Besides that it looks good to me.

PR #1369, opened 2025-06-05
74nvmf: Lookup required NIC kernel modules for NBFT interfaces

tbzatek:
When dracut is called with '--hostonly-mode strict' and perhaps even with --hostonly-nics '' as done by kdump/mkdumprd, all the kernel modules related to hostonly networking are stripped down. In such a case it is viable to lookup and install the required kernel modules ourselves, based on any currently present nbft interfaces.

Note: the lookup depends on the current 'nbftX' network interface naming.

Cc: @mwilck 

## Checklist
- [ ] ~I am providing new code and test(s) for it~


pvalena (approved):

tbzatek:
> > Note: the lookup depends on the current 'nbftX' network interface naming.
> 
> Perhaps we should drop this assumption rather sooner than later. As you know I like this interface naming but (assuming you do too) we're in a minority position.

Right, so this PR was just the most straightforward solution, given that the module itself creates interfaces with the `nbftX` naming. The proper solution to be fully independent would likely involve parsing the NBFT table again, extract MAC addresses for all HFIs and somehow look them up in the system.

I still wonder whether it makes sense to use some kind of network interface tagging and where/how to store such tags (udev?).

Anyway, should it block this PR or shall we postpone it as a TODO (in a separate ticket here)?

modules.d/74nvmf/module-setup.sh
```diff
@@ -72,6 +72,14 @@ installkernel() {
     hostonly=$(optional_hostonly) instmods nvme_fc nvme_tcp nvme_rdma lpfc qla2xxx
     # 802.1q VLAN may be set up in Firmware later. Include the module always.
     hostonly="" instmods 8021q
+    # lookup NIC kernel modules for active NBFT interfaces
+    if [[ $hostonly ]]; then
```
jozzsi, line 76: Should this conditional check include checking for strict ? What is the rational of doing these extra steps in sloppy mode ?

tbzatek:
> Would it be possible to fix this root cause instead ? Would this bug impact other dracut modules as well (e.g. nfs) ? What are some example reproduction steps (missing kernel modules) for this case ?

Cc: @coiby

I think it boils down to the `--hostonly-nics ''` argument that appears to be always supplied by kdump. So this PR was meant merely as a safeguard, in case some other tool attempts to strip the NIC drivers out. Networking is the crucial prerequisite for the NBFT functionality.

modules.d/74nvmf/module-setup.sh
coiby: I also think the NIC drivers needed by NBFT interfaces should be included regardless of the Host-only mode. Because the dracut nvmf module  won't work without these NIC drivers so they should be included unconditionally.

tbzatek:
Added a check that `$_driver` is not empty since the `readlink` call may return empty string in case there's no device/driver sysfs path, e.g. in case of a vlan interface that matches the `nbft*` name.

modules.d/74nvmf/module-setup.sh
mwilck: @colby: That makes no sense to me. NBFT is, by definition, a host-only feature. In non-hostonly mode the only way to ensure NBFT functionality is to include **all** network drivers that might be used for NBFT interfaces.

mwilck (approved):
LGTM

modules.d/74nvmf/module-setup.sh
tbzatek: > is to include **all** network drivers that might be used for NBFT interfaces.

Correct. Always makes me wonder that the universal non-hostonly initramfs image (e.g. an installer image) is quite small yet containing all the drivers necessary.

pvalena (approved):

PR #1370, opened 2025-06-05
fix(74nvmf): remove netroot=nbft

tbzatek:
Setting netroot causes further issues with stripped down strict hostonly initramfs (e.g. kdump initramfs) and it appears there's no strong reason to have it set.

```
[   23.599862] dracut: FATAL: No or empty root= argument
[   23.604956] dracut: Refusing to continue
```

It also causes excessive connection attempts as a result of network interface events (e.g. link up/interface ready) with NetworkManager, significantly delaying boot (observed ~12 minutes) due to connection attempts of unavailable subsystems and timing out for each such attempt.

Cc: @mwilck 

We're still not 100% certain what effects exactly does setting `netroot` have, this removal proposal is based upon lots of testing and experiments. The module appears to be working just fine without it.

## Checklist
- [ ] ~I am providing new code and test(s) for it~

### DIFF
--- a/modules.d/74nvmf/module-setup.sh
+++ b/modules.d/74nvmf/module-setup.sh
@@ -148,7 +148,6 @@ install() {
     inst_multiple ip sed
 
     inst_script "${moddir}/nvmf-autoconnect.sh" /sbin/nvmf-autoconnect.sh
-    inst_script "${moddir}/nbftroot.sh" /sbin/nbftroot
 
     inst_multiple nvme jq
     inst_hook cmdline 92 "$moddir/parse-nvmf-boot-connections.sh"
--- a/modules.d/74nvmf/nbftroot.sh
+++ b/modules.d/74nvmf/nbftroot.sh
@@ -1,5 +1,0 @@
-#! /bin/sh
-# This script is called from /sbin/netroot
-
-/sbin/nvmf-autoconnect.sh online
-exit 0
--- a/modules.d/74nvmf/parse-nvmf-boot-connections.sh
+++ b/modules.d/74nvmf/parse-nvmf-boot-connections.sh
@@ -318,9 +318,6 @@ done
 
 if [ -e /tmp/nvmf_needs_network ] || [ -e /tmp/valid_nbft_entry_found ]; then
     echo "rd.neednet=1" > /etc/cmdline.d/nvmf-neednet.conf
-    # netroot is a global variable that is present in all "sourced" scripts
-    # shellcheck disable=SC2034
-    netroot=nbft
     rm -f /tmp/nvmf_needs_network
 fi
 


tbzatek:
For the reference, this is the `rd.debug` kdump run
* with the `netroot=nbft` assignment in place: [kdump-rd.debug_fail.log](https://github.com/user-attachments/files/20613441/kdump-rd.debug_fail.log)
* and without (this patch applied): [kdump-rd.debug-no_nbftroot.log](https://github.com/user-attachments/files/20613176/kdump-rd.debug-no_nbftroot.log)

(serial console capture, slightly messy)

mwilck:
I'm afraid I disagree with this PR.

Currently we try to connect the NBFT targets in 3 places, using `nvmf-autoconnect.sh`:

1. "settled" initqueue,
2. "timeout" initqueue,
3. via /sbin/netroot, which is called from different places depending on the network management module.
   * network-legacy calls it from `/sbin/ifup`, which in turn is run from the initque where it's inserted from an udev rule (only once),
   * networkmanager calls it from nm-run.sh, which is run from the "settled" initqueue

Out of these three, the netroot hook is the "right" one, IMO. It's designed to be run when a network interface comes up (although the networkmanager module doesn't implenment it this way, see below). The other two should be seen as fallbacks for the case that the netroot hook doesn't work.

I agree that it's wrong that (for networkmanager at least) we now have two "settled" hooks that both call `nvmf-autoconnect.sh`. But I'd rather remove `nvmf-autoconnect.sh` from the "settled" hook then stop calling netroot.

```
[   23.599862] dracut: FATAL: No or empty root= argument
```

This is caused by the fact that there is no `root=` cmdline parameter in the kdump environment. I don't see it on openSUSE, because we set `root=kdump` on the kdump command line. I think this can be fixed by setting `rootok=1` and `root=nvmf` in `parse-nvmf-boot-connections.sh`, similar to what `74iscsi`, `74nfs` and other netroot modules do.

> It also causes excessive connection attempts as a result of network interface events

I need to understand this better. AFAICS `nm-run.sh` is run from the "settled" initqueue, and it iterates over all network interfaces and calls `/sbin/netroot` for every configured interface. `/sbin/netroot`, in turn, will call `nvme connect-all --nbft`, which will attempt all connections, even for those whose network connections are not up yet. Moreover, `nvmf-autoconnect.sh` will also be called from the "settled" initqueue, adding yet another `connect-all` command.

The main difference wrt the legacy module seems to be that the latter tries to run `ifup` (and thus, `netroot`) only once when an interface is first brought up.

Summarizing, I'd propose changes like this to fix the issue:

```
diff --git a/modules.d/74nvmf/parse-nvmf-boot-connections.sh b/modules.d/74nvmf/parse-nvmf-boot-connections.sh
index aad2a45..48939b7 100755
--- a/modules.d/74nvmf/parse-nvmf-boot-connections.sh
+++ b/modules.d/74nvmf/parse-nvmf-boot-connections.sh
@@ -324,5 +324,8 @@ if [ -e /tmp/nvmf_needs_network ] || [ -e /tmp/valid_nbft_entry_found ]; then
     rm -f /tmp/nvmf_needs_network
 fi
 
-/sbin/initqueue --settled --onetime --name nvmf-connect-settled /sbin/nvmf-autoconnect.sh settled
 /sbin/initqueue --timeout --onetime --name nvmf-connect-timeout /sbin/nvmf-autoconnect.sh timeout
+
+# shellcheck disable=SC2034
+rootok=1
+[ -z "$root" ] && root="nvmf"
```

In addition to that, we should discuss improving `nm-run.sh` such that it calls `/sbin/netroot` no more than once per interface, like the network-legacy module.

mwilck:
I'm working on an alternative patch set.

tbzatek:
Updated patch per the suggestion above, only left out the initqueue/settled call in place until it's properly tested with NVMe/FC.

mwilck:
@Conan-Kudo & @pvalena, I think we're good here. Please confirm.

mwilck:
> I'll soon send a PR on top of this one

... which is #1622. It is **not** meant to delay this PR.

PR #1385, opened 2025-06-19
chore: version lock commisery-action

jozzsi:
@master seems to report the folloeing error                                                                                                                                                                                           
                                                                                                                                                                                                                                      
Error: Resource not accessible by integration                                                                                                                                                                                         
                                                                                                                                                                                                                                      
Switch to @v3, as it is the last stable version.

## Checklist


bdrung (approved):
v4 was just released. We could consider switching to it.

PR #1386, opened 2025-06-19
fix(dmsquash-live): erofs collision with latest util-linux

### LINKED ISSUE #1384: erofs collision with latest util-linux
The issue has been reported originally for Fedora, see: https://bugzilla.redhat.com/show_bug.cgi?id=2367956

Attempt to upgrade to the latest util-linux upstream release v2.41 failed (see Adam's report at https://bodhi.fedoraproject.org/updates/FEDORA-2025-dd6b6950ff)
  

The problem is that recent kernels can mount EROFS directly without loop devices, and this feature is enabled in Fedora kernels (>=6.12; CONFIG_EROFS_FS_BACKED_BY_FILE=y). This feature is now supported by mount/libmount, too.
 
The difference is how mount(8) reacts to
```
   # mount foo.img /mnt
```
The old version creates a loop device and mounts it. The new mount(8) directly mounts the image.
 
The issue is possible to reproduce with the Fedora Live image:
https://adamwill.fedorapeople.org/05017930-Fedora-Workstation-Live-x86_64-FEDORA-2025-dd6b6950ff.iso
 
You can see LiveOS/squashfs.img is directly mounted, but there is also /dev/loop0 associated with the image. This is probably an unwanted situation (and potentially risky).
 ```
$ findmnt --submounts /run
TARGET                                      SOURCE                                  FSTYPE          OPTI
/run                                        tmpfs                                   tmpfs           rw,n
├─/run/initramfs/live                       /dev/sr0                                iso9660         ro,r
├─/run/rootfsbase                           /run/initramfs/live/LiveOS/squashfs.img erofs           ro,r
├─/run/credentials/systemd-journald.service tmpfs                                   tmpfs           ro,n
├─/run/credentials/systemd-resolved.service tmpfs                                   tmpfs           ro,n
└─/run/user/1000                            tmpfs                                   tmpfs           rw,n
  ├─/run/user/1000/doc                      portal                                  fuse.portal     rw,n
  └─/run/user/1000/gvfs                     gvfsd-fuse                              fuse.gvfsd-fuse rw,n
 
$ lsblk
NAME  MAJ:MIN RM  SIZE RO TYPE MOUNTPOINTS
loop0   7:0    0  2.1G  1 loop
sr0    11:0    1  2.3G  0 rom  /run/initramfs/live
zram0 251:0    0  3.8G  0 disk [SWAP]
vda   253:0    0   10G  0 disk
 
$ losetup
NAME       SIZELIMIT OFFSET AUTOCLEAR RO BACK-FILE            DIO LOG-SEC
/dev/loop0         0      0         0  1 /LiveOS/squashfs.img   0     512
```

See https://bugzilla.redhat.com/attachment.cgi?id=2093109 for complete debug log with rd.debug.

Here are the relevant lines (grep -E '(losetup|mount -[a-z])' log):
```
 losetup -r /dev/loop0 /run/initramfs/live/LiveOS/squashfs.img     
 mount -n -o ro /dev/loop0 /run/initramfs/squashfs
         
 mount -r /run/initramfs/live/LiveOS/squashfs.img /run/rootfsbase  
 umount -l /run/initramfs/squashfs
```         
You can see that `modules.d/70dmsquash-live/dmsquash-live-root.sh` tries to mount the same image twice, first by losetup + mount; second as "mount squashfs.img".
         
This worked as expected in the old version because libmount is smart and detects that the same backing file (squashfs.img) is already associated with any loop device. Then it reuses the device rather than creating a new one.
         
After updating to util-linux v2.41, it works too, but the second mount does not create (or reuse) a loop device, but mounts squashfs.img directly.

**Solution.**

I think the best approach would be to avoid the second mount altogether, independently of the util-linux version. It is more robust to use a bind mount there than to attempt to create a second instance of the same file system. Something like:    
```diff
diff --git a/modules.d/70dmsquash-live/dmsquash-live-root.sh b/modules.d/70dmsquash-live/dmsq
index 2ea393e1..68c26a3e 100755
--- a/modules.d/70dmsquash-live/dmsquash-live-root.sh              
+++ b/modules.d/70dmsquash-live/dmsquash-live-root.sh              
@@ -422,7 +422,11 @@ fi
 if [ -n "$overlayfs" ]; then                                      
     if [ -n "$FSIMG" ]; then                                      
         mkdir -m 0755 -p /run/rootfsbase                          
-        mount -r "$FSIMG" /run/rootfsbase                         
+       if [ "$FSIMG" = "$SQUASHED" ]; then                        
+           mount --bind /run/initramfs/squashfs /run/rootfsbase   
+       else
+           mount -r "$FSIMG" /run/rootfsbase                      
+       fi  
     else
         ln -sf /run/initramfs/live /run/rootfsbase                
     fi  
```         

This change should be portable and allow avoiding direct erofs image mounts. It will also avoid dependence on libmount duplicate loopdev backing file detection.

jozzsi:
## Changes

This PR fixes the issue that recent kernels can mount EROFS directly without loop devices, and this feature is enabled in Fedora kernels (>=6.12; CONFIG_EROFS_FS_BACKED_BY_FILE=y). This feature is now supported by mount/libmount, too.

I think the best approach would be to avoid the second mount altogether, independently of the util-linux version.
It is more robust to use a bind mount there than to attempt to create a second instance of the same file system.

## Checklist

Fixes #1384

CC @FGrose @karelzak

### DIFF
--- a/modules.d/70dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/70dmsquash-live/dmsquash-live-root.sh
@@ -422,7 +422,11 @@ fi
 if [ -n "$overlayfs" ]; then
     if [ -n "$FSIMG" ]; then
         mkdir -m 0755 -p /run/rootfsbase
-        mount -r "$FSIMG" /run/rootfsbase
+        if [ "$FSIMG" = "$SQUASHED" ]; then
+            mount --bind /run/initramfs/squashfs /run/rootfsbase
+        else
+            mount -r "$FSIMG" /run/rootfsbase
+        fi
     else
         ln -sf /run/initramfs/live /run/rootfsbase
     fi


modules.d/70dmsquash-live/dmsquash-live-root.sh
```diff
@@ -422,7 +422,11 @@ fi
 if [ -n "$overlayfs" ]; then
     if [ -n "$FSIMG" ]; then
         mkdir -m 0755 -p /run/rootfsbase
-        mount -r "$FSIMG" /run/rootfsbase
+        if [ "$FSIMG" = "$SQUASHED" ]; then
+            mount --bind /run/initramfs/squashfs /run/rootfsbase
```
FGrose, line 426: The `/run/initramfs/squashfs` mountpoint is no longer needed.  We could
`[ "${DRACUT_SYSTEMD-}" ] && mount --make-rprivate /run`
`mount --move /run/initramfs/squashfs /run/rootfsbase`

jozzsi: @FGrose Thanks. I tried to understand your suggestion - which sounds like a follow-up "optimization" (and not an essential part of fixing the bug that was reported). Would you agree ? Do you think it could be part of a follow-up PR ? 

Sadly this module seems to be quite convoluted, so I want to minimize the size of the commits.

Conan-Kudo (approved):

Issue #1310, opened 2025-05-02
Cannot use `$kernel` in `initrdname=`


aafeijoo-suse:
**Describe the bug**

Cannot use `$kernel` in `initrdname=`, so this configuration parameter is useless for openSUSE.

**Dracut version**
107

**To Reproduce**
```
# grep -w initrdname /etc/dracut.conf
initrdname="initrd-$kernel"
# dracut -f
...
dracut[I]: *** Creating image file '/boot/initrd-.tmp' ***
dracut[I]: *** Creating initramfs image file '/boot/initrd-.tmp' done ***
dracut[I]: *** Moving image file '/boot/initrd-.tmp' to '/boot/initrd-' ***
dracut[I]: *** Moving image file '/boot/initrd-.tmp' to '/boot/initrd-' done ***
```

**Additional context**

`$kernel` assigned after conf files are sourced.

https://github.com/dracut-ng/dracut-ng/blob/01b369a5866853c8dc5c53c7d4d8613e5721be34/dracut.sh#L967-L979

https://github.com/dracut-ng/dracut-ng/blob/01b369a5866853c8dc5c53c7d4d8613e5721be34/dracut.sh#L989-L992

https://github.com/dracut-ng/dracut-ng/blob/01b369a5866853c8dc5c53c7d4d8613e5721be34/dracut.sh#L1034-L1036

https://github.com/dracut-ng/dracut-ng/blob/01b369a5866853c8dc5c53c7d4d8613e5721be34/dracut.sh#L1204

aafeijoo-suse:
> Cannot use $kernel in initrdname=, so this configuration parameter is useless for openSUSE.

And for Debian/Ubuntu:

https://sources.debian.org/src/dracut/107-1/debian/patches/Default-initrdname-to-initrd.img-kernel.patch/

PR #1387, opened 2025-06-19
feat: allow the use of $kernel in initrdname= config

jozzsi:
## Changes

Assign `$kernel` before conf files are sourced.

## Checklist

Fixes: #1310

CC @aafeijoo-suse @bdrung

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -968,6 +968,17 @@ export DRACUT_LOG_LEVEL=warning
 # These config variables needs to be exported for dracut-install.
 export add_dlopen_features="" omit_dlopen_features=""
 
+if ! [[ $kernel ]] && [[ $regenerate_all_l != "yes" ]]; then
+    if type -P systemd-detect-virt &> /dev/null && ! systemd-detect-virt -c &> /dev/null; then
+        kernel="$(uname -r)"
+    else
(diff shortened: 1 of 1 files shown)


bdrung (dismissed):
LGTM

dracut.sh
```diff
@@ -968,6 +968,17 @@ export DRACUT_LOG_LEVEL=warning
 # These config variables needs to be exported for dracut-install.
 export add_dlopen_features="" omit_dlopen_features=""
 
+if ! [[ $kernel ]] && [[ $regenerate_all_l != "yes" ]]; then
```
bdrung, line 971: What happens for --regenerate-all? `kernel` will not be set in this case.

jozzsi: Yes, `kernel` can not be set for `--regenerate-all`. Since we move code around, we need to explicitelly check for this condition here - see https://github.com/dracut-ng/dracut-ng/blob/main/dracut.sh#L1030

Conan-Kudo (approved):

bdrung (approved):
I don't like how the sourcing of the configuration (with setting `kernel`) is entangled with `--regenerate-all`, but this is a topic for the future (and I do not see a simple solution to make the code better).

PR #1390, opened 2025-06-20
fix(systemd-repart): allow partition format

esposem:
systemd-repart is capable not only of creating a partition, but also of formatting it. According with repart.d, it is able to create the following filesystems: ext4, btrfs, xfs, vfat, erofs and squashfs.

Add support in the systemd-repart module for the underlying tools to allow systemd-repart to format the partition.

Failure to do so would make systemd-repart initramfs unit fail, if Format= option is provided in a repart.d config file.

## Checklist

### DIFF
--- a/modules.d/01systemd-repart/module-setup.sh
+++ b/modules.d/01systemd-repart/module-setup.sh
@@ -20,6 +20,16 @@ install() {
         "$systemdsystemunitdir"/initrd-root-fs.target.wants/systemd-repart.service \
         systemd-repart
 
+    # Systemd-repart is capable of also formatting the created partition.
+    # Support all filesystems that repart.d supports.
+    inst_multiple -o \
+        "mkfs.ext4" \
+        "mkfs.btrfs" \
+        "mkfs.xfs" \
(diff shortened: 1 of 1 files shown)


vittyvk:
FWIW, looks good to me too. 'inst_multiple -o' should guarantee that these new dependencies are optional. E.g. mkfs.btrfs may not be available in some distros (e.g. RHEL).

pvalena (approved):
LGTM.

bdrung (commented):
Should be look for Format= and only include the `mkfs` commands it that config is found? Or is is possible to provide systemd-repart configuration via the kernel cmdline?

vittyvk:
Not sure about kernel cmdline, probably not, but repart configuration can also come via systemd's configuration extensions (systemd-sysext dracut module supports these).

Conan-Kudo:
Please do not drop btrfs support here, most distributions have support for it, and both Fedora and SUSE family distributions default to it.

Conan-Kudo (approved):

PR #1395, opened 2025-06-23
fix(systemd-sysext): intall the required kerenel modules

vittyvk:
systemd-repart uses 'erofs' for Discoverable Disk Image (DDI) ("--make-ddi=TYPE" option), include it into initramfs by default. While on in, include 'loop' module as well as systemd-sysext can't work without it.

## Checklist

Fixes #


vittyvk:
Rebased, no changes.

bdrung (approved):
LGTM

PR #1398, opened 2025-06-27
test: setup test environment variables only when needed

bdrung:
## Changes

The function `set_test_envonment_variables` contains a spelling mistake. Rename it to `set_kversion_and_vmlinux` to make it more explicit.

The test environment variables `KVERSION` and `VMLINUZ` are only needed when calling `run-qemu` or using them directly. They are not needed for TEST-80-GETARG and TEST-81-SKIPCPIO.

So move the `set_kversion_and_vmlinux` call to the relevant `test_check` functions.

## Checklist

Fixes #


bdrung:
The alternative approach would be to split the tests into unit and integration tests.

Conan-Kudo (approved):

Issue #1389, opened 2025-06-20
crypsetup (ENC-RAID-LVM) test fails 


jozzsi:
See CI failure: 26 on fedora:rawhide

https://github.com/dracut-ng/dracut-ng/actions/workflows/integration-extra.yml

```
[    5.809019] systemd-cryptsetup invoked oom-killer: gfp_mask=0x140dca(GFP_HIGHUSER_MOVABLE|__GFP_ZERO|__GFP_COMP), order=0, oom_score_adj=500
```

Increasing the RAM available for the VM does not seem to fix the issue.

This is happening only on fedora:rawhide. Other Fedora versions and other distributions do not seem ot be impacted.

CC @AdamWill @pvalena @Conan-Kudo

mbroz:
ok, it seems that there are two cryptsetup processes spawn. I though systemd used serialization (there is option for it)...
```
[    712]     0   712   103711    67369    65152     2217         0   598016        0           500 systemd-cryptse
[    714]     0   714   105465   103700   101440     2260         0   884736        0           500 systemd-cryptse
```

Anyway,  as the test here is testing storage stack, the used LUKS PBKDF is irrelevant - I suggest to add ``--pbkdf pbkdf2`` to luksFormat commands  in this test and this issue should disappear.

PR #1404, opened 2025-06-28
test: add --pbkdf pbkdf2 to luksFormat commands

jozzsi:
## Changes

Make the test work with recent release (v2.8.0) of cryptsetup.

This PR keeps memory requirements for LUKS unlocking minimal, allowing test cases run in system with minimal memory.
The storage stack and boot process that is being tested, remains exactly the same.
Default for LUKS2 is now Argon2 (memory-hard KDF, so it means that it requires some amount of memory).

Fixes: https://github.com/dracut-ng/dracut-ng/issues/1389

## Checklist

### DIFF
--- a/test/TEST-20-STORAGE/create-root.sh
+++ b/test/TEST-20-STORAGE/create-root.sh
@@ -24,7 +24,7 @@ else
 
     if ! grep -qF 'rd.luks=0' /proc/cmdline && command -v cryptsetup > /dev/null; then
         printf test > keyfile
-        cryptsetup -q luksFormat /dev/md0 /keyfile
+        cryptsetup --pbkdf pbkdf2 -q luksFormat /dev/md0 /keyfile
         echo "The passphrase is test"
         cryptsetup luksOpen /dev/md0 dracut_crypt_test < /keyfile
         lvm pvcreate -ff -y /dev/mapper/dracut_crypt_test
--- a/test/TEST-26-ENC-RAID-LVM/create-root.sh
+++ b/test/TEST-26-ENC-RAID-LVM/create-root.sh
@@ -3,8 +3,8 @@
 trap 'poweroff -f' EXIT
 set -ex
 printf test > keyfile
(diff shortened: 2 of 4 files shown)


bdrung:
Which is the default PBKDF algorithm? Is that a bug of the default algorithm? If yes, is there an upstream bug to reference? Can you extent the commit (message) with that information?

mbroz:
> Which is the default PBKDF algorithm? Is that a bug of the default algorithm? If yes, is there an upstream bug to reference? Can you extent the commit (message) with that information?

Default for LUKS2 is Argon2 (memory-hard KDF, so it means that it requires some amount of memory. We try hard to detect amount of memory that is sane but still provide expected security margin. The problem is that the system use overallocation - it will let you allocate a large amout memory, but once you try to write to it, OOM will kill the process (this is designed to work for Java and similar systems that allocates large amount of memory but never use it...). This is a known problem, and it is not a bug in fact. Anyway, it is irrelevant for your use case. Here it is even complicated that something spawn more systemd-cryptsetup processes in parallel.

Just use PBKDF2, exactly as this PR does  (that has small constant memory footprint), it was default for LUKS1.  If this happens in real system, I will try to fiddle with cryptsetup allocation logic again, but here it really makes no sense.

mbroz:
IOW, this PR keeps memory requirements for LUKS unlocking minimal, allowing testcases run in system with minimal memory. The storage stack, that is being tested, is exactly the same.

bdrung:
Thanks @mbroz for the explanation. Do you recommend to tweak the PBKDF cost parameters or better use the default ones for pbkdf2?

@jozzsi can you include this explanation into the commit message? Can you tweak the other test cases as well (test 41 and 72 call luksFormat as well)?

bdrung:
Last small nitpick: `(ENC-RAID-LVM)` in the commit title does not match any more.

mbroz:
> Thanks @mbroz for the explanation. Do you recommend to tweak the PBKDF cost parameters or better use the default ones for pbkdf2?

You can decrease unlocking time to minimum (with ``-i 1``), but as it is up to 2 seconds anyway, I would keep the default to not complicate it, it should work fine.

bdrung (approved):

PR #1413, opened 2025-07-03
fix: improve reproducibility

### LINKED ISSUE #1409: `--reproducible` does not produce identical UKIs
**Describe the bug**
When using both `--uefi` and `--reproducible`  to generate multiple UKIs, the UKIs always differ in hash, even if no configuration or module changes are made.

In contrast, generating a regular initramfs (without `--uefi`) is reproducible as expected.

**Why it matters:**
Without reproducibility, identical input leads to different UKIs. This causes unnecessary duplication when multiple snapshots are taken, leading to inefficient use of storage over time.

**Distribution used**
LFS with Arch Linux repos, or Arch Linux

**Dracut version**
dracut 107

**Init system**
systemd 257

**To Reproduce**
```bash
sudo dracut --force --uefi --reproducible --hostonly --no-hostonly-cmdline --kver "$(uname -r)" /tmp/uki-a
sudo dracut --force --uefi --reproducible --hostonly --no-hostonly-cmdline --kver "$(uname -r)" /tmp/uki-b
sudo b2sum /tmp/uki-*
```
You will see that two UKI files are not identical.

**Expected behavior**
Two UKIs should be identical and have the same hash if inputs haven't changed.

Note: `mkinitcpio` produces reproducible UKIs out of the box, with identical hashes.

**Additional context**
https://github.com/dracut-ng/dracut-ng/pull/141
This PR doesn't work?

jozzsi:
## Changes

Why ? This PR increases reproducibility of the generated initramfs as it is immune to timestamp changes on the host.

Setting SOURCE_DATE_EPOCH is recommended (but not required) for reproducibility.

## Checklist

Fixes #1409

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2447,8 +2447,14 @@ if [[ $uefi == yes ]]; then
 fi
 
 if [[ $DRACUT_REPRODUCIBLE ]]; then
-    find "$initdir" -newer "$dracutbasedir/dracut-functions.sh" -print0 \
-        | xargs -r -0 touch -h -m -c -r "$dracutbasedir/dracut-functions.sh"
+    if [ -n "${SOURCE_DATE_EPOCH}" ]; then
+        # ensure that no timestamps are newer than $SOURCE_DATE_EPOCH
+        find "$initdir" -newermt "@${SOURCE_DATE_EPOCH}" -print0 \
+            | xargs -0r touch --no-dereference --date="@${SOURCE_DATE_EPOCH}"
+    else
+        # set all timestamps to 0
+        find "$initdir" -mindepth 1 -execdir touch -hcd '@0' '{}' +
+    fi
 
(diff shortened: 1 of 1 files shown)


bdrung:
I don't like this approach. dracut should honor SOURCE_DATE_EPOCH instead.

I suggest to do the same as initramfs-tools:

```sh
# if SOURCE_DATE_EPOCH is set, try and create a reproducible image
if [ -n "${SOURCE_DATE_EPOCH}" ]; then
	# ensure that no timestamps are newer than $SOURCE_DATE_EPOCH
	find "${DESTDIR}" -newermt "@${SOURCE_DATE_EPOCH}" -print0 | \
		xargs -0r touch --no-dereference --date="@${SOURCE_DATE_EPOCH}"
fi
```

See https://salsa.debian.org/kernel-team/initramfs-tools/-/blob/debian/latest/mkinitramfs?ref_type=heads

Issue #1422, opened 2025-07-05
network-manager does not require dbus when systemd is not used


chenxiaolong:
**Describe the bug**

The `network-manager` module currently depends on `dbus` unconditionally:

https://github.com/dracut-ng/dracut-ng/blob/79ffbd28294818a36e515a441142125d70e3acbb/modules.d/35network-manager/module-setup.sh#L13

However, NetworkManager doesn't seem to actually depend on dbus when systemd is not used in the initramfs. The `NetworkManager.conf` manpage confirms this:

> __configure-and-quit__
> ...
> When set to 'initrd', NetworkManager does not connect to D-Bus and quits after configuring the network. This is an implementation detail how the NetworkManager module of dracut can run NetworkManager. An alternative to this is having NetworkManager as a systemd service with D-Bus in initrd.

My use case is a bit esoteric because I'm using zfsbootmenu to boot Fedora 42 on a root-on-zfs setup. zfsbootmenu has its own kernel + initramfs. After selecting what to boot, it loads Fedora's kernel + initramfs and does a kexec.

* The zfsbootmenu initramfs is built with dracut, but unfortunately doesn't support systemd.
* The Fedora initramfs is built with dracut and uses systemd.

I'm trying to use the `network-manager` dracut module in both initramfs images. Out of the box, building the zfsbootmenu initramfs image does not work because they omit `systemd`, which prevents pulling in `dbus` and thus `network-manager`.

**Distribution used**

Fedora 42

**Dracut version**

`dracut-105-3.fc42.x86_64` and also the main branch (commit 79ffbd28294818a36e515a441142125d70e3acbb)

**Init system**

Both systemd and non-systemd

**To Reproduce**

```
❯ dracut -o systemd -a network-manager /tmp/test.img
...
dracut[E]: Module 'dbus-broker' depends on module 'systemd', which can't be installed
dracut[E]: Module 'dbus-daemon' depends on module 'systemd', which can't be installed
dracut[E]: Module 'dbus' depends on module 'dbus-daemon', which can't be installed
dracut[E]: Module 'network-manager' depends on module 'dbus', which can't be installed
...
```

**Expected behavior**

The dbus dependency should be required only when using systemd as it doesn't seem to be required for non-systemd setups.

**Additional context**

I can work around the problem with this little patch:

```diff
diff --git a/modules.d/35network-manager/module-setup.sh b/modules.d/35network-manager/module-setup.sh
index 09107708..03a80575 100755
[...]

PR #1423, opened 2025-07-06
fix(network-manager): depend on dbus only when using systemd

chenxiaolong:
## Changes

network-manager does not require dbus to function when using the `--configure-and-quit=initrd` option. This should be guaranteed as it is documented behavior in the [NetworkManager.conf(5) manpage](https://man.archlinux.org/man/extra/networkmanager/NetworkManager.conf.5.en) and specifically mentions dracut's use case.

With this change, initramfs images that don't or can't use systemd can use network-manager instead of the old network-legacy module.

## Checklist
  - Tested with zfsbootmenu (no systemd) and Fedora (with systemd), both built from a Fedora 42 host
  - I don't believe there are any docs to update

Fixes: #1422

### DIFF
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -10,7 +10,10 @@ check() {
 
 # called by dracut
 depends() {
-    echo dbus bash net-lib kernel-network-modules initqueue
+    echo bash net-lib kernel-network-modules initqueue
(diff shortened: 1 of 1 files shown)


classabbyamp:
seems fine

bdrung (approved):
LGTM

Conan-Kudo (approved):

PR #1425, opened 2025-07-08
fix(systemd-sysext): enable initrd-specific units

vittyvk:
Commit 68a09b43 ("fix(systemd-sysext): install new initrd-specific units") correctly handled the name change for systemd-{confext,sysext}.service units but left systemctl enablement intact. Besides the name change, the newly introduced '-initrd' units don't have an "[Install]" section so just doing 'enable' is not enough. Similar to other modules, e.g '11systemd-portabled', use 'add-wants initrd.target' instead.

## Checklist


vittyvk:
@aafeijoo-suse hm, I don't see any symlinks for sysext/confext services in systemd-257.7 in Fedora, e.g. my Fedora-42 has:
```
$ ls -1 /lib/systemd/system/initrd.target.wants/
dracut-cmdline.service
dracut-initqueue.service
dracut-mount.service
dracut-pre-mount.service
dracut-pre-pivot.service
dracut-pre-trigger.service
dracut-pre-udev.service
systemd-battery-check.service
systemd-bsod.service
systemd-pcrphase-initrd.service
```
where should they come from?

vittyvk:
> > I don't see any symlinks for sysext/confext services in systemd-257.7
> 
> This is v258 material.
> 

Ah, all right -- so for old systemds we will keep doing 'systemctl --enable' as before and for the new one we will just package the symlinks and avoid hacks.

aafeijoo-suse (approved):
Thanks for fixing this missing piece.

modules.d/11systemd-sysext/module-setup.sh
```diff
@@ -41,13 +41,18 @@ install() {
...
         systemd-confext systemd-sysext
 
-    # Enable systemd type unit(s)
+    # Enable systemd type unit(s) for systemd < v258 which doesn't ship
+    # initrd.target.wants symlinks.
     for i in \
         systemd-confext.service \
         systemd-sysext.service; do
-        $SYSTEMCTL -q --root "$initdir" enable "$i"
+        if [[ -e "$initdir$systemdsystemunitdir"/"$i" ]]; then
```
bdrung, line 59: Shouldn't that check for the absence of the `initrd.target.wants` symlinks instead?

aafeijoo-suse: These 2 services could run in the initrd and in the host before systemd-v258, before the initrd-specific units were added, so there were no `initrd.target.wants` symlinks before. With `_suffix="-initrd"`  we control that they are not present in `$initdir`, and if they are present (`_suffix=`), it means there are no initrd-specific units so these 2 services must be enabled.

bdrung: Thanks. The commit message mislead me. IMO those two changes should be split into two separate commits.

jozzsi: @vittyvk  - would you be able to split this up into two commits as suggested by @bdrung ?

vittyvk: Sure, will split.

vittyvk: Pushed. The fix for the issue (sysext/confext services not enabled in initrd) still requires both changes but the commit changelog is somewhat more fine grained. No functional change compared to the previous version. Please take a look, thanks!

PR #1429, opened 2025-07-09
fix(dracut.sh): Ensure hardlink deduplication is reproducible

cjriches:
By default, `hardlink` will only deduplicate files with identical mtimes, down to one-second granularity. If a dracut module rapidly generates multiple identical files, it is completely up to chance as to whether their mtimes cross a second boundary or not, and thus whether they get deduplicated or not. This results in non-reproducible output.

Perform timestamp clamping before invoking `hardlink` if a reproducible build is requested.

## Checklist


cjriches:
There's a few open questions / alternative approaches here:

* Doing this mtime clamping will also result in more files getting more consistently deduplicated, so it potentially has benefits for everyone - perhaps we should do it unconditionally regardless of --reproducible?
* We could also avoid touching mtimes and just run hardlink -t to deduplicate regardless of mtime match, but there's a small chance this breaks someone who was relying on identical files having different mtimes. The -t option also isn't present in all distros (e.g. it's in RHEL9 but not RHEL8).

bdrung:
`dracut.sh` calls `hardlink "$initdir"`. Looking at the man page of hardlink, I found the option `--ignore-time` which says:
> Link and compare files even if their time of modification is different. This is usually a good choice.

So I suggest to set this parameter.

cjriches:
> I found the option --ignore-time

That's what I was getting at with the above comment:

> We could also avoid touching mtimes and just run hardlink -t to deduplicate regardless of mtime match, but there's a small chance this breaks someone who was relying on identical files having different mtimes. The -t option also isn't present in all distros (e.g. it's in RHEL9 but not RHEL8).

I felt this wasn't the best option due to the above drawbacks, but can do so if you prefer?

bdrung:
Your PR triggered me to write https://github.com/dracut-ng/dracut-ng/pull/1431 - the first commit there is inspired/taken from you. May I set you as author of that "refactor(dracut): introduce clamp_mtimes helper function" commit?

cjriches:
> Move the clamping code before the hardlink code. The code after hardlink might either be moved before hardlink or need to be made sure that it does not change the mtime incorrectly.

Yeah, I was deliberately avoiding this due to concerns about subsequent steps re-introducing variance.

>  (e.g. I expect the strip command to keep the mtime untouched).

Really? I would expect the opposite, and a quick test shows that it does affect the mtime at least on my machine.

As far as I can see, the current order of operations is this:

1. hardlink
2. cleanup empty ldconfig_paths directories
3. generate early-microcode cpio image
4. package ACPI tables
5. store current command line parameters
6. inst_libdir_file workarounds
7. build_ld_cache
8. strip various things
9. module_install squash-lib
10. clamp mtimes

I think it's reasonable to assume that most of those intervening steps might alter mtimes. If we're going to try and clamp once, then I think we should do the opposite and move the hardlink much later, after all of that and the existing clamp. Presumably that should be harmless, as all it could affect is deduplicating _more_ things that were generated in the meantime.

Shall I alter the patch to take that approach?

bdrung:
Yes, please. That (moving hardlinks to a later point) sounds like the better approach.

cjriches:
> Yes, please. That (moving hardlinks to a later point) sounds like the better approach.

Done, though I haven't retested yet. Will tackle that tomorrow.

dracut.sh
```diff
@@ -2471,6 +2465,13 @@ if [[ $DRACUT_REPRODUCIBLE ]]; then
     fi
 fi
 
+# Hardlink is mtime-sensitive; do it after the above clamp.
```
cjriches, line 2472: Nope, this approach doesn't work. Running hardlink itself affects the mtimes of the directories, since we end up creating and deleting files inside those directories.

Therefore, clamping must be done after hardlink as well as before.

cjriches: The original version of this PR works. @bdrung shall we revert to that, or do you have an alternative in mind?

bdrung: suggestion: Clamp all non-directories, hardlink, clamp all directories?

cjriches: Done; testing in progress.

cjriches: Yep, this works.

bdrung (approved):
When enabling hardlinks, dracut will run clamp_mtimes on directories twice. So just assume that this makes no performance difference.

PR #1440, opened 2025-07-12
fix(fcoe-uefi): exit early on empty vlan

jozzsi:
## Changes

Exit early in case `get_fcoe_boot_vlan` exits with error or just an empty string, instead of producing invalid config entry.

Found this PR at https://github.com/dracutdevs/dracut/pull/2379 from @pvalena . 

CC @pvalena @mwilck @aafeijoo-suse @lnykryn @LaszloGombos 

## Checklist

### DIFF
--- a/modules.d/74fcoe-uefi/parse-uefifcoe.sh
+++ b/modules.d/74fcoe-uefi/parse-uefifcoe.sh
@@ -9,19 +9,21 @@ print_fcoe_uefi_conf() {
     mac=$(get_fcoe_boot_mac "$1")
     [ -z "$mac" ] && return 1
     dev=$(set_ifname fcoe "$mac")
-    vlan=$(get_fcoe_boot_vlan "$1")
-    if [ "$vlan" -ne "0" ]; then
-        case "$vlan" in
-            [0-9]*)
-                printf "%s\n" "vlan=$dev.$vlan:$dev"
-                dev="$dev.$vlan"
-                ;;
-            *)
-                printf "%s\n" "vlan=$vlan:$dev"
-                dev="$vlan"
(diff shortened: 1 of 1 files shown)


modules.d/74fcoe-uefi/parse-uefifcoe.sh
```diff
@@ -9,19 +9,22 @@ print_fcoe_uefi_conf() {
...
-                ;;
-            *)
-                printf "%s\n" "vlan=$vlan:$dev"
-                dev="$vlan"
-                ;;
-        esac
-    fi
+    vlan=$(get_fcoe_boot_vlan "$1") || return 1
+    case "$vlan" in
+        "0") ;;
```
bdrung, line 14: That looks inconsistent formatted to me: `;;` is on the same line. It is on a separate line for others.

jozzsi: This change would apparently fails the linter test

```
'shfmt -s' returned error 1 finding the following formatting issues:

----------
diff modules.d/74fcoe-uefi/parse-uefifcoe.sh.orig modules.d/74fcoe-uefi/parse-uefifcoe.sh
--- modules.d/74fcoe-uefi/parse-uefifcoe.sh.orig
+++ modules.d/74fcoe-uefi/parse-uefifcoe.sh
@@ -11,8 +11,7 @@
     dev=$(set_ifname fcoe "$mac")
     vlan=$(get_fcoe_boot_vlan "$1") || return 1
     case "$vlan" in
-        "0")
-            ;;
+        "0") ;;
 
         '')
             return 1
----------
```

PR #1442, opened 2025-07-12
fix(dracut.conf.d): reserve namespace 50 to out-of-tree configurations

jozzsi:
## Changes

50 should be reserved for out-of-tree dracut configurations, just like 50 is reserved for out-of-tree dracut modules. This policy makes dracut more consistent and easier to reason about.

Change the built-in dracut configurations from namespace 50 to namespace 10.

Document that the recommended ordering for distribution or user provided configuration files is in the range of 50-59.

CC @Nowa-Ammerlaan @LaszloGombos @bdrung @dalto8 

## Checklist


test/TEST-43-KERNEL-INSTALL/test.sh
```diff
@@ -64,7 +64,7 @@ test_setup() {
     cp /usr/lib/dracut/test/dracut.conf.d/test/test.conf /usr/lib/dracut/dracut.conf.d/
 
     # enable rescue boot config
-    cp /usr/lib/dracut/dracut.conf.d/rescue/50-rescue.conf /usr/lib/dracut/dracut.conf.d/
+    cp /usr/lib/dracut/dracut.conf.d/rescue/* /usr/lib/dracut/dracut.conf.d/
```
bdrung: Shouldn't that be `*-rescue.conf`?

jozzsi: The way these `dracut.conf.d` "profiles" are designed by convention is that each profile has its own directory and each directory has exactly one file. Here it really does not make a difference one way or another, so I went for what is shorter and more readable. I changed it to *conf, hopefully it is more clear that way.

bdrung: Thanks for explaining.

PR #1444, opened 2025-07-12
fix(man): document what to expect running dracut non-root

jozzsi:
## Changes

Running dracut non-root supports only a limited set of functionalities.

Follow-up to 97b5f91 .

CC @marc-hb @bdrung @LaszloGombos 

## Checklist


marc-hb (approved):
Thank you!

PR #1445, opened 2025-07-13
fix(systemd-sysusers): maintain users and groups

jozzsi:
## Changes

Currently dracut deletes entries for users and groups created by systemd-sysusers.
    
This PR maitaines users and groups created by systemd-sysusers and merges root password from the host with the shadow file created by systemd-sysusers.
    
Add test assertion to check if nobody user is present, which is expected on a systemd-enabled system.

Allow the owner to read and write /etc/shadow and /etc/gshadow in the initramfs.

This fixes a permission error during dracut when it's not run as root:

```
dracut[I]: *** Creating image file '.../initramfs.img' ***
cpio: etc/gshadow: Cannot open: Permission denied
cpio: etc/shadow: Cannot open: Permission denied
dracut[F]: Creation of .../initramfs.img failed
```

Indeed, the permissions were set empty:

```
$ cd $initdir
$ ls -l etc/gshadow etc/shadow
----------. 1 myuser mygroup  179 mar  26 10:29 gshadow
----------. 1 myuser mygroup   32 mar  26 10:29 shadow
```

This approach also enables systemd-sysusers credentials, see https://www.freedesktop.org/software/systemd/man/latest/systemd-sysusers.html#Credentials

@LaszloGombos @bdrung @zboszor @AdamWill @marc-hb @chewi @aafeijoo-suse  - I received some feedback on the earlier PR (that is now landed) - so I wanted to open this up for discussion agin. I am glad the previous PR landed, but perhaps there is room for improvement.

This PR is based on @zboszor earlier suggestion, but need to be tested in more scenarios. 

@AdamWill would you be open to test this PR on Fedora rawhide ? 

## Checklist


modules.d/80base/module-setup.sh
```diff
@@ -65,7 +65,18 @@ install() {
     [[ ${hostonly-} ]] && pwshadow='x'
     grep '^root:' "$initdir/etc/passwd" > /dev/null 2>&1 || echo "root:$pwshadow:0:0::/root:/bin/sh" >> "$initdir/etc/passwd"
 
-    [[ ${hostonly-} ]] && grep '^root:' "${dracutsysrootdir-}"/etc/shadow >> "$initdir/etc/shadow"
+    if [[ ${hostonly-} ]]; then
+        # check if other dracut modules already created an entry for root in /etc/shadow
+        if grep '^root:' "$initdir/etc/shadow" > /dev/null 2>&1; then
```
marc-hb:
```suggestion
        if grep -q '^root:' "$initdir/etc/shadow"; then
```
Discarding stderr is a bad idea 99.99% of the time.

When `grep` uses `stderr` (very rarely) 1. things have gone very seriously wrong and you really don't want to lose the message 2. it is never going to "spam" the logs.

modules.d/80base/module-setup.sh
```diff
@@ -65,7 +65,18 @@ install() {
...
+            # replace root password in the existing entry in etc/shadow
+            # root password from host takes precedence over root password set by systemd-sysuser in hostonly mode
+            root_password=$(grep '^root:' "${dracutsysrootdir-}"/etc/shadow 2> /dev/null | cut -d':' -f2)
```
marc-hb:
```suggestion
            root_password=$(grep '^root:' "${dracutsysrootdir-}"/etc/shadow | cut -d':' -f2)
```

If stderr is really a problem in this function, then it should be redirected only once (ideally: to some logfile) for the entire function; not here and there for various commands.

test/TEST-41-FULL-SYSTEMD/test.sh
```diff
@@ -139,6 +139,14 @@ test_setup() {
         -a "dracut-systemd $dracut_modules" \
         --add-drivers "btrfs"
 
+    # verify that systemd-coredump user exists generated by systemd-sysuser
+    if ! grep '^systemd-coredump:' "$TESTDIR"/initrd/dracut.*/initramfs/etc/passwd > /dev/null 2>&1; then
```
marc-hb:
```suggestion
    if ! grep -q '^systemd-coredump:' "$TESTDIR"/initrd/dracut.*/initramfs/etc/passwd ; then
```

test/TEST-41-FULL-SYSTEMD/test.sh
```diff
@@ -139,6 +139,14 @@ test_setup() {
...
+        # fail the test
+        echo "systemd-coredump user is not present in /etc/passwd" >&2
```
marc-hb: So it looks like stderr is NOT a problem here... then no reason to drop other useful error messages elsewhere. There may be some tools abusing stderr but they really are the exception, not the rule.

modules.d/80base/module-setup.sh
jozzsi: Thanks. Done !

modules.d/80base/module-setup.sh
jozzsi: Thanks. Done !

test/TEST-41-FULL-SYSTEMD/test.sh
jozzsi: Thanks. Done !

test/TEST-41-FULL-SYSTEMD/test.sh
jozzsi: done

PR #1446, opened 2025-07-13
test(NETWORK): coverage for bringing up network interface          

jozzsi:
## Changes

Currently, all networking tests in the dracut CI require a client and a server VM, and all the server VMs are enabled to work with network-legacy only.
    
Introduce a simpler networking test that tests the `rd.neednet` dracut boot time argument and ensures that the interface is up without needing a server VM.
    
This simple networking test could be run in most networking configurations and pass in the Void test containers as well, which previously have not had any network tests enabled.

The test as added to the ARM64 CI as well .

## Checklist


test/TEST-50-NETWORK/test.sh
```diff
@@ -0,0 +1,42 @@
...
+    declare -a disk_args=()
+    # shellcheck disable=SC2034  # disk_index used in qemu_add_drive
+    declare -i disk_index=0
+    qemu_add_drive disk_index disk_args "$TESTDIR"/marker.img marker
+    qemu_add_drive disk_index disk_args "$TESTDIR"/root.img root
+
+    test_marker_reset
+
+    "$testdir"/run-qemu \
+        -nic user,model=virtio-net-pci \
```
bdrung: Since we explicitly check for the IP address, I suggest to be explicitly here (e.g. `user,id=lan0,net=10.0.3.0/24`). We might want to set the hostname and domain and check that too.

bdrung (commented):
Having network tests that use the qemu user network stack is great.

For the other network tests I like to see them ported to use dnsmasq instead of isc-dhcp-server.

test/TEST-50-NETWORK/assertion.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Get the output of 'ip addr show eth0'
+ip_output=$(ip addr show eth0)
```
bdrung, line 4: Will the interface always be called `eth0` - even on other architectures?

In initramfs-tools we have a similar test case and there we use this snippet to name the interface `lan0`:

```sh
persist_net() {
	name="$1"
	mac="$2"

	mkdir -p "${DESTDIR}/etc/systemd/network"
	cat >"${DESTDIR}/etc/systemd/network/10-persistent-${name}.link" <<EOF
[Match]
MACAddress=${mac}

[Link]
Name=${name}
EOF
}

persist_net lan0 "52:54:00:65:43:21"
persist_net lan1 "52:54:00:12:34:56"
persist_net lan9 "52:54:00:12:33:21"
```

test/TEST-50-NETWORK/test.sh
jozzsi: > I suggest to be explicitly here

Great idea, done !

> We might want to set the hostname and domain and check that too.

I hope to expand test coverage as follow-up commits/PRs.

test/TEST-50-NETWORK/assertion.sh
jozzsi: > Will the interface always be called eth0 - even on other architectures?

Given the number of architectures (and distributions and networking options) dracut supports, I really want to start slow and expand as follow-up PRs. I think allowing predictable naming schema would make it harder to make this test pass in all architectures, but I think we should try to do that as follow-up PRs. 

eth0 is determined by kernel drivers not by user space as I understand, see the `net.ifnames=0` passed in as kernel command line option.

Issue #1424, opened 2025-07-08
failed to find llvm-objcopy


ghost:
**Describe the bug**
dracut failed to find llvm-objcopy.
`dracut[F]: Need 'objcopy' to create a UEFI executable`

The following is the only directly related snippet I found in the codebase (dracut.sh):
```
if [[ $uefi == yes ]]; then
    if ! command -v objcopy &> /dev/null; then
        dfatal "Need 'objcopy' to create a UEFI executable"
        exit 1
    fi
```
My understanding is that as long as `command -v objcopy` return a non-empty output, no error will occur.
Here is the output: 
`/usr/lib/llvm/20/bin/objcopy`

**Distribution used**
Gentoo Linux x86_64 (musl/llvm)

**Dracut version**
107

**Init system**
OpenRC 0.56

**To Reproduce**
Ensure that GNU binutils is not installed.
(Do not use `sys-kernel/gentoo-kernel-bin` because GNU binutils will be installed)
`echo "sys-kernel/installkernel dracut uki" >> /etc/portage/package.use/sys-kernel`
`echo "sys-kernel/dracut -dracut-cpio" >> /etc/portage/package.use/sys-kernel`
`emerge -av sys-kernel/gentoo-kernel`
If kernel build completed, but UKI generation failed:
`emerge --config sys-kernel/gentoo-kernel`

**Expected behavior**
llvm-objcopy should be found and UKI should be generated successfully.

Conan-Kudo:
Is LLVM `objcopy` compatible? Also, why is `/usr/lib/llvm/20/bin` in your path?

Nowa-Ammerlaan:
> Is LLVM `objcopy` compatible? 

It is (sort of). Extracting stuff from the UKI is more difficult though.

> Also, why is `/usr/lib/llvm/20/bin` in your path?

There should be a symlink at /usr/bin/objcopy. Please check if it is there, and if not please enable the native-symlinks USE flag 

Dracut should maybe use the OBJCOPY environment variable to find/select which objcopy to use, which would make it compatible with the case where native-symlinks is not enabled

Nowa-Ammerlaan:
I have a hunch that `sys-devel/binutils-config` and it's eselect utility are supposed to manage this symlink (though not 100% sure).

Anyway, please let me know if this fixes your problem: https://github.com/dracut-ng/dracut-ng/pull/1447

There is also something fishy going on with the `PATH` as you mentioned, not sure what is up with that.

[...]

PR #1447, opened 2025-07-13
fix({dracut,lsinitrd}.sh): Use obj{copy,dump} set by toolchain

Nowa-Ammerlaan:
If dracut is invoked in a context where the toolchain is configured to use a certain `objcopy` implementation, use that one. This may be the case if dracut is invoked via systemd's `kernel-install` in a Gentoo ebuild on a llvm-only system.

Closes: https://github.com/dracut-ng/dracut-ng/issues/1424

## Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1042,7 +1042,7 @@ get_dev_module() {
 pe_file_format() {
     if [[ $# -eq 1 ]]; then
         local magic
-        magic=$(objdump -p "$1" \
+        magic=$("${OBJDUMP:-objdump}" -p "$1" \
             | awk '{if ($1 == "Magic"){print $2}}')
         # 010b (PE32), 020b (PE32+)
         [[ $magic == "020b" || $magic == "010b" ]] && return 0
@@ -1055,7 +1055,7 @@ pe_get_header_data() {
     local data_header
     [[ $# -ne "2" ]] && return 1
     [[ $(pe_file_format "$1") -eq 1 ]] && return 1
-    data_header=$(objdump -p "$1" \
+    data_header=$("${OBJDUMP:-objdump}" -p "$1" \
         | awk -v data="$2" '{if ($1 == data){print $2}}')
     echo "$data_header"
 }
--- a/dracut.sh
+++ b/dracut.sh
@@ -1649,7 +1649,7 @@ if [[ ! $print_cmdline ]]; then
     fi
 
     if [[ $uefi == yes ]]; then
-        if ! command -v objcopy &> /dev/null; then
+        if ! command -v "${OBJCOPY:-objcopy}" &> /dev/null; then
             dfatal "Need 'objcopy' to create a UEFI executable"
             exit 1
         fi
@@ -2638,7 +2638,7 @@ clean_sbat_string() {
 get_sbat_string() {
     local inp=$1
     local out=$uefi_outdir/$2
-    objcopy -O binary --only-section=.sbat "$inp" "$out"
+    "${OBJCOPY:-objcopy}" -O binary --only-section=.sbat "$inp" "$out"
     clean_sbat_string "$out"
 }
 
@@ -2656,7 +2656,7 @@ if [[ $uefi == yes ]]; then
         fi
     fi
 
-    offs=$(($(objdump -h "$uefi_stub" 2> /dev/null | awk 'NF==7 {size=$3;
+    offs=$(($("${OBJDUMP:-objdump}" -h "$uefi_stub" 2> /dev/null | awk 'NF==7 {size=$3;
                 offset=$4} END {print "16#"size" + 16#"offset}')))
     if [[ $offs -eq 0 ]]; then
         dfatal "Failed to get the size of $uefi_stub to create UEFI image file"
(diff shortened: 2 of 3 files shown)


bdrung:
I saw that we have `DRACUT_PATH` in dracut.sh. So you should be able to set `DRACUT_PATH=/path/to/objcopy:$PATH` to use a specific objcopy.

Nowa-Ammerlaan:
> How would dracut be called/used after this PR?

No change.

There will only be a behavioural change if OBJCOPY or OBJDUMP is set in the environment. Which is the case if Dracut is invoked from a Gentoo ebuild environment, or some other build system environment. The change involves that it is now less broken on systems without binutils.

bdrung (commented):
This makes the code longer / more complex. IMO this use case should be covered by `DRACUT_PATH` (or make dracut honour `PATH`).

Nowa-Ammerlaan:
> This makes the code longer / more complex. IMO this use case should be covered by `DRACUT_PATH` (or make dracut honour `PATH`).

`PATH` is only part of the problem, the second problem is that the users toolchain preference is not respected. This is important because, as Bug #1424 shows, binutils is not present on every system. The `OBJCOPY` and `OBJDUMP` environment variables are the standard solution to allow users to set this preference. I do agree though that the `PATH` manipulation in Dracut is weird, I did not remove it here since I am unsure why it was introduced in the first place. Forcing users to manually set `DRACUT_PATH` just to have their kernel build+install not error out is not an acceptable workaround.

dracut.sh
```diff
@@ -23,6 +23,11 @@
 unset BASH_ENV
 unset GZIP
 
+: "${OBJCOPY:=objcopy}"
+[[ ${OBJCOPY} != /* ]] && OBJCOPY=$(command -v "${OBJCOPY}")
+: "${OBJDUMP:=objdump}"
+[[ ${OBJDUMP} != /* ]] && OBJDUMP=$(command -v "${OBJDUMP}")
```
bdrung: Suggestion: keep the code simple here. I fail to see a reason why we should make the path absolute here.

```suggestion
OBJCOPY="${OBJCOPY-objcopy}"
OBJDUMP="${OBJDUMP-objdump}"
```

dracut-functions.sh
```diff
@@ -1042,7 +1042,7 @@ get_dev_module() {
 pe_file_format() {
     if [[ $# -eq 1 ]]; then
         local magic
-        magic=$(objdump -p "$1" \
+        magic=$("${OBJDUMP}" -p "$1" \
```
bdrung: Since dracut-functions.sh can be sourced from everywhere, I suggest to keep the fallback:

```suggestion
        magic=$("${OBJDUMP-objdump}" -p "$1" \
```

dracut-functions.sh
```diff
@@ -1055,7 +1055,7 @@ pe_get_header_data() {
     local data_header
     [[ $# -ne "2" ]] && return 1
     [[ $(pe_file_format "$1") -eq 1 ]] && return 1
-    data_header=$(objdump -p "$1" \
+    data_header=$("${OBJDUMP}" -p "$1" \
```
bdrung:
```suggestion
    data_header=$("${OBJDUMP-objdump}" -p "$1" \
```

lsinitrd.sh
```diff
@@ -357,11 +360,11 @@ unpack_files() {
 
 read -r -N 2 bin < "$image"
 if [ "$bin" = "MZ" ]; then
-    command -v objcopy > /dev/null || {
+    command -v "${OBJCOPY}" > /dev/null || {
```
bdrung: It looks like objcopy is only used here. So the variable could be evaluated only here:

```suggestion
    command -v "${OBJCOPY-objcopy}" > /dev/null || {
```

or (not tested):

```suggestion
    OBJCOPY=$(command -v "${OBJCOPY-objcopy}" > /dev/null) || {
```

dracut.sh
Nowa-Ammerlaan: > I fail to see a reason why we should make the path absolute here.

This is because Dracut ignores `PATH`. The user who reported #1424 confirmed that we need to find the absolute path of the specified `objcopy` first and save it before the `PATH` is manipulated.

lsinitrd.sh
Nowa-Ammerlaan: > or (not tested):

This works, but we do have to drop the `> /dev/null`.

dracut-functions.sh
Nowa-Ammerlaan: Makes sense, fixed!

dracut.sh
bdrung: I fail to see why this is related to `PATH`. The `command` command will rely on `PATH` to determine the absolute path.

Nowa-Ammerlaan: > I fail to see why this is related to `PATH`. The `command` command will rely on `PATH` to determine the absolute path.

It does, but we set `OBJDUMP` and `OBJCOPY` (line 27 and 28) **before** `dracut.sh` changes the `PATH` (on line 1102). So the call to `command` here uses the usual system `PATH`, not the manipulated `PATH`.

bdrung: Thanks. Let's fix dracut's modification of PATH: https://github.com/dracut-ng/dracut-ng/pull/1468

Nowa-Ammerlaan:
Rebased and simplified since we dropped the `PATH` manipulation in https://github.com/dracut-ng/dracut-ng/pull/1468

bdrung (approved):
LGTM now.

PR #1448, opened 2025-07-13
fix: network-manager dracut module no longer depends on systemd

jozzsi:
## Changes

Previously dracut network-manager module only worked with systemd.

Now after 58baf86, network-manager dracut module should work in non-systemd Linux distributions as well, such as Void Linux.

This commit re-enables network-manager dracut module for all Linux distributions.

CC @classabbyamp @chenxiaolong 

## Checklist


bdrung (commented):
LGTM, but the commit message could be improved:

* It says "dracut network-manager package" or "network-manager dracut package" but don't we call them modules instead of packages?
* it says "this PR" but it should say "this commit"

PR #1452, opened 2025-07-15
test(SYSTEMD): convert test to run without initqueue

### LINKED ISSUE #1191: rfe: make initqueue optional
It'd be nice to be able to construct an initrd with dracut, but not have the dracut initqueue active in the initrd. For example, it could be moved to a new initqueue module and made optional. Dracut modules which require the initqueue would declare that they need the initqueue. Other modules would need to run their stuff via systemd units and/or udev rules.

The benefit is that we get a simpler system. Currently we have two "event loops", the one in the systemd transaction and the older-but-now-duplicate dracut job queue. Modules that only need systemd and/or udev have the additional advantage that they can run unchanged in the main system.

jozzsi:
## Changes

Systemd itself provides similar functionality to the dracut initqueue module. It has been a long-standing criticism of dracut systemd integration that initqueue should not be necessary for most system boot scenarios.

While there is still a long way to go, we can start testing some of the common systemd boot scenarios without initqueue.

See https://github.com/dracut-ng/dracut-ng/issues/1191.

Commit 2676f1a allowed to make this change, without that commit this test case would fail.

CC @dtardon @keszybz 

## Checklist

Fixes #1191

### DIFF
--- a/test/TEST-40-SYSTEMD/test.sh
+++ b/test/TEST-40-SYSTEMD/test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -e
 # shellcheck disable=SC2034
-TEST_DESCRIPTION="root filesystem on a ext4 filesystem"
+TEST_DESCRIPTION="root filesystem on a ext4 filesystem with systemd but without initqueue"
 
 test_check() {
     command -v systemctl &> /dev/null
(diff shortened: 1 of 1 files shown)


dtardon (approved):
LGTM

PR #1456, opened 2025-07-16
fix(dracut.sh): do not use uname to detect kernel version in a container

### LINKED ISSUE #1455: default to running for /usr/lib/modules/$kver in a container
This is a followup to https://github.com/dracut-ng/dracut-ng/pull/582#issuecomment-2294210362

Basically let's do

```
if systemd-detect-virt -c != none; then
  kver=$(cd /usr/lib/modules && echo *)
fi
```

jozzsi:
## Changes

A big general papercut with dracut right now in that it always assumes the kernel it's running on is the kernel to target.

This commit lets dracut to detect that it's in a container (e.g. systemd-detect-virt -c), and check for a single /usr/lib/modules/$kver directory and automatically use that kernel.

Currently this kernel version detection has three copies in the source (dracut.sh, lsinitrd.sh,test-functions). This commit keeps the logic for the three copies the same. As a follow-up commit, we should try to actually share the code for this logic instead of copying it.

CC @cgwalters 

## Checklist

Fixes #1455

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1076,9 +1076,9 @@ if ! [[ $kernel ]]; then
     if type -P systemd-detect-virt &> /dev/null && container=$(systemd-detect-virt -c) &> /dev/null; then
         dinfo "*** Detected container: $container ***"
         # shellcheck disable=SC2012
-        kernel="$(cd /lib/modules && ls -1 | tail -1)"
+        kernel="$(cd /lib/modules && ls -1v | tail -1)"
         # shellcheck disable=SC2012
-        [[ $kernel ]] || kernel="$(cd /usr/lib/modules && ls -1 | tail -1)"
+        [[ $kernel ]] || kernel="$(cd /usr/lib/modules && ls -1v | tail -1)"
     fi
     [[ $kernel ]] || kernel="$(uname -r)"
 fi
--- a/lsinitrd.sh
+++ b/lsinitrd.sh
@@ -99,9 +99,9 @@ done
 if ! [[ $KERNEL_VERSION ]]; then
     if type -P systemd-detect-virt &> /dev/null && systemd-detect-virt -c &> /dev/null; then
         # shellcheck disable=SC2012
-        KERNEL_VERSION="$(cd /lib/modules && ls -1 | tail -1)"
+        KERNEL_VERSION="$(cd /lib/modules && ls -1v | tail -1)"
         # shellcheck disable=SC2012
-        [[ $KERNEL_VERSION ]] || KERNEL_VERSION="$(cd /usr/lib/modules && ls -1 | tail -1)"
+        [[ $KERNEL_VERSION ]] || KERNEL_VERSION="$(cd /usr/lib/modules && ls -1v | tail -1)"
     fi
     [[ $KERNEL_VERSION ]] || KERNEL_VERSION="$(uname -r)"
 fi
--- a/test/test-functions
+++ b/test/test-functions
@@ -106,9 +106,9 @@ determine_kernel_version() {
     # we are running inside a container already
 
     # shellcheck disable=SC2012
-    [[ ${KVERSION-} ]] || KVERSION="$(cd /lib/modules && ls -1 | tail -1)"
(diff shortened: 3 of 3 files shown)


dracut.sh
```diff
@@ -1073,7 +1073,13 @@ if [[ $regenerate_all == "yes" ]]; then
 fi
 
 if ! [[ $kernel ]]; then
-    kernel=$(uname -r)
+    if type -P systemd-detect-virt &> /dev/null && systemd-detect-virt -c &> /dev/null; then
+        # shellcheck disable=SC2012
+        kernel="$(cd /lib/modules && ls -1 | tail -1)"
```
cgwalters: We're relying on the default sorting from `ls` here...which is probably fine, but *in theory* should maybe be doing some sort of version number parsing.

On the bootc side we strictly disallow multiple kernels, so it's not a concern either way.

dracut.sh
```diff
@@ -1073,7 +1073,13 @@ if [[ $regenerate_all == "yes" ]]; then
...
+        # shellcheck disable=SC2012
+        [[ $kernel ]] || kernel="$(cd /usr/lib/modules && ls -1 | tail -1)"
```
cgwalters: Do we really care about non-usrmerge systems i.e. where `/lib` isn't just a symlink to `/usr/lib`?

dracut.sh
```diff
@@ -1073,7 +1073,13 @@ if [[ $regenerate_all == "yes" ]]; then
 fi
 
 if ! [[ $kernel ]]; then
-    kernel=$(uname -r)
+    if type -P systemd-detect-virt &> /dev/null && systemd-detect-virt -c &> /dev/null; then
```
cgwalters: Perhaps worth logging something here, like:
```
    if type -P systemd-detect-virt &> /dev/null && container=systemd-detect-virt -c &> /dev/null; then
dracut_log "Detected container: $container"
```

or so?

dracut.sh
jozzsi: Went with ls -v here, seems a low hanging fruit and hopefully the second commit does not slow down the PR review. Thanks for the tip.

dracut.sh
jozzsi: Done.

dracut.sh
jozzsi: Fair question. Dracut CI uses Alpine to test some conditions that are relevant also for Fedora, but easier to test with Alpine - e.g. busybox compatibility (busybox is not available in CentOS).

Alpine still uses /lib/modules. I understand why this caught your eyes, but perhaps best to handle this as a separate PR. Some ideas:
 - change the order of detection as `/usr/lib/modules` is the common case
 - remove support for `/lib/modules` and add a workaround in the Alpine CI container with a sym link - this would break Alpine dracut packaging if Alpine packaging decide to run tests as part of packaging

cgwalters: From my PoV a cleanup here should probably happen after the code dedup you mention first. I think it's OK as is.

dracut.sh
```diff
@@ -1073,7 +1073,14 @@ if [[ $regenerate_all == "yes" ]]; then
 fi
 
 if ! [[ $kernel ]]; then
-    kernel=$(uname -r)
+    if type -P systemd-detect-virt &> /dev/null && container=$(systemd-detect-virt -c) &> /dev/null; then
+        printf "%s\n" "dracut[W]: Detected container: $container" >&2
```
bdrung: Why print the detected container as warning?

bdrung:
What I would like to see as follow-up PR: Move the code into a function that is identical everywhere.

dracut.sh
jozzsi: https://github.com/dracut-ng/dracut-ng/pull/1456#discussion_r2211222252

CC @cgwalters

cgwalters: I don't strongly think we need to log at all if that's the question, but I think @bdrung is more asking "why isn't this `dinfo`" right?

We already log a whole lot of stuff at `dinfo` level.

jozzsi: Oh, I see. Changed to `dinfo`.

PR #1462, opened 2025-07-19
fix: do not use the running kernel to determine kernel version

jozzsi:
## Changes

If kernel version is not passed in the command line for a dracut invocation than dracut.sh should be the only place where kernel version is determined.

Remove the heuristic to determine kernel version from Makefile and from dracut-init.sh.

Follow-up to 2b2debd .

## Checklist


bdrung (commented):
The change looks good, but the commit title "do not use a single uname -r call to determine kernel version" confuses me. With these changes only one singe `uname -r` remains, doesn't it?

bdrung (approved):

PR #1463, opened 2025-07-19
fix(dracut): consolidate reporting running in a container

jozzsi:
## Changes

`dinfo` call to report running in a container does not work as expected as `dinfo` is not yet defined.

In addition we're already reporting running in a container in some other conditions, which would lead to reporting it twice.

Consolidate reporting running in a container into one single place and make it work, and reporting is as information and not as a warning as using dracut inside container is common.
    
Reordered two code blocks to enable the usage of dinfo to simplify reporting.

Follow-up to 2b2debd .

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1073,8 +1073,7 @@ if [[ $regenerate_all == "yes" ]]; then
 fi
 
 if ! [[ $kernel ]]; then
-    if type -P systemd-detect-virt &> /dev/null && container=$(systemd-detect-virt -c) &> /dev/null; then
-        dinfo "*** Detected container: $container ***"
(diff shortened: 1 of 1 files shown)


dracut.sh
```diff
@@ -1437,6 +1429,11 @@ else
     exit 1
 fi
 
+if container=$(systemd-detect-virt -c) &> /dev/null; then
+    export DRACUT_NO_MKNOD=1
+    dinfo "Detected $container container !"
```
```suggestion
    dinfo "Detected $container container."
```

or

```suggestion
    dinfo "Detected container: $container"
```

jozzsi: Thanks for the review. Changed.

bdrung (approved):
LGTM

PR #1465, opened 2025-07-19
feat: let lsinitrd report back the Kernel version

jozzsi:
This pull request changes...

## Checklist

Fixes #

### DIFF
--- a/lsinitrd.sh
+++ b/lsinitrd.sh
@@ -25,6 +25,7 @@ usage() {
         echo "-h, --help                  print a help message and exit."
         echo "-s, --size                  sort the contents of the initramfs by size."
         echo "-m, --mod                   list modules."
+        echo "-r, --kernel-release        print the kernel release and exit."
         echo "-f, --file <filename>       print the contents of <filename>."
         echo "--unpack                    unpack the initramfs, instead of displaying the contents."
         echo "                            If optional filenames are given, will only unpack specified files,"
@@ -50,6 +51,7 @@ TEMP=$(getopt \
     --long kver: \
     --long file: \
     --long mod \
+    --long kernel-release \
     --long help \
     --long size \
     --long unpack \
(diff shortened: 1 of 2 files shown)


test/test-functions
```diff
@@ -108,7 +108,7 @@ command -v test_cleanup &> /dev/null || test_cleanup() {
 }
 
 determine_kernel_version() {
-    lsinitrd "$1" | grep modules.dep | head -1 | rev | cut -d'/' -f2 | rev
+    lsinitrd "$1" | grep ^Kernel: | cut -d' ' -f2
```
bdrung: Could be add a parameter to `lsinitrd` to only print the kernel version?

jozzsi: We need to define the return value if we can not determine the kernel version.

Here in this version pf the PR for now, we simply do not report it.

jozzsi: I think it is a bit of overkill to introduce an extra command line option for a simple one line feature, but did it anyways to see how it would look and solicited feedback.

lsinitrd.sh
```diff
@@ -533,7 +536,10 @@ else
     ((ret += $?))
     echo "Version: $version"
     echo
-    if [ "$modules" -eq 1 ]; then
+    if [ "$kernel" -eq 1 ]; then
+      kernel=$($CAT "$image" | cpio --extract --quiet --list 'usr/lib/modules/*/modules.dep' | head -1 | sed 's;.*/\([^/]\+\)/modules.dep;\1;')
```
bdrung, line 540: Can you support 3cpio as well here?

jozzsi: I tried to use `cpio_extract` function and failed. 

```
3cpio: Error: Target directory '.' is not empty. Use --force to overwrite existing files!
```

jozzsi:
I am no longer interested pursuing this PR.

PR #1466, opened 2025-07-19
test: compute VMLINUZ in run-qemu only when it is needed

jozzsi:
## Changes

Dracut can call ukify when available to generate UKI (see acfddd69f). Prefer testing this flow, over testing calling ukify directly from the test-case.
    
This change also remove the need for test-cases to have direct access to VMLINUZ variable.

Currently VMLINUZ is computed and set even when it is not needed (e.g. for test 80). Move the computation to run-qemu instead and only set it when needed.

## Checklist


bdrung (approved):
I like that approach!

PR #1469, opened 2025-07-20
fix(dracut): only call uname -r if it is safe to do

jozzsi:
## Changes

Change the priority order to determine the kernel version if it is not specified on the command line.
    
Only call uname -r if it is safe to do, meaning if it is confirmed that dracut is not called from within a container.

Eliminate redundant test code from determine_kernel_version function. Update test cases where knowledge of kernel version is required and use the generated initramfs to extract the kernel version.

Follow-up to 2b2debd .

@bdrung @LaszloGombos @cgwalters 

## Checklist

### DIFF
--- a/test/TEST-60-NFS/test.sh
+++ b/test/TEST-60-NFS/test.sh
@@ -239,7 +239,7 @@ test_setup() {
         --add-confdir test-root \
         -a "url-lib nfs" \
         -I "ip grep setsid" \
-        -f "$TESTDIR"/initramfs.root "$KVERSION" || return 1
+        -f "$TESTDIR"/initramfs.root || return 1
 
     KVERSION=$(determine_kernel_version "$TESTDIR"/initramfs.root)
     export kernel=$KVERSION
--- a/test/run-qemu
+++ b/test/run-qemu
@@ -8,33 +8,34 @@ ARCH="${ARCH-$(uname -m)}"
 QEMU_CPU="${QEMU_CPU:-max}"
 
 set_vmlinux_env() {
-    VMLINUZ=${VMLINUZ-"/lib/modules/${KVERSION}/vmlinuz"}
-    if ! [ -f "$VMLINUZ" ]; then
-        VMLINUZ="/lib/modules/${KVERSION}/vmlinux"
-    fi
-
-    if ! [ -f "$VMLINUZ" ]; then
-        [[ -f /etc/machine-id ]] && read -r MACHINE_ID < /etc/machine-id
-
-        if [[ ${MACHINE_ID-} ]] && { [[ -d /boot/${MACHINE_ID} ]] || [[ -L /boot/${MACHINE_ID} ]]; }; then
-            VMLINUZ="/boot/${MACHINE_ID}/$KVERSION/linux"
-        elif [ -f "/boot/vmlinuz-${KVERSION}" ]; then
-            VMLINUZ="/boot/vmlinuz-${KVERSION}"
-        elif [ -f "/boot/vmlinux-${KVERSION}" ]; then
-            VMLINUZ="/boot/vmlinux-${KVERSION}"
-        elif [ -f "/boot/kernel-${KVERSION}" ]; then
-            VMLINUZ="/boot/kernel-${KVERSION}"
-        elif [ -f "/boot/Image-${KVERSION}" ]; then
-            VMLINUZ="/boot/Image-${KVERSION}"
+    if [[ ${KVERSION-} ]]; then
+        VMLINUZ=${VMLINUZ-"/lib/modules/${KVERSION}/vmlinuz"}
+        if ! [ -f "$VMLINUZ" ]; then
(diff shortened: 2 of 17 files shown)


dracut.sh
```diff
@@ -1073,13 +1073,14 @@ if [[ $regenerate_all == "yes" ]]; then
 fi
 
 if ! [[ $kernel ]]; then
-    if type -P systemd-detect-virt &> /dev/null && systemd-detect-virt -c &> /dev/null; then
+    if type -P systemd-detect-virt &> /dev/null && ! systemd-detect-virt -c &> /dev/null; then
```
LaszloGombos: @jozzsi  - I noticed that https://github.com/dracut-ng/dracut-ng/issues/1455#issuecomment-3092579683 includes testing for `hostonly` as well. Is that condition missing from here ?

jozzsi: Prefer to discuss it as a follow-up PR as it is orthogonal to this PR (and requires to rearrange some code blocks). Thanks !

bdrung:
IMO we can have multiple PRs open for separate issues (unless they are entangled). Priority can be handled by labels. Having more open PRs will probably have no negative effect on my reviewing speed.

test/TEST-60-NFS/test.sh
```diff
@@ -231,11 +231,20 @@ test_run() {
...
 
     rm -rf -- "$TESTDIR"/overlay
+
+    "$DRACUT" -N --keep --tmpdir "$TESTDIR" \
+        --add-confdir test-root \
+        -a "url-lib nfs" \
+        -I "ip grep setsid" \
+        -f "$TESTDIR"/initramfs.root "$KVERSION" || return 1
+
+    KVERSION=$(determine_kernel_version "$TESTDIR"/initramfs.root)
```
bdrung: `KVERSION` is used before it is set.

bdrung (commented):
The `fix(dracut): only call uname -r if it is safe to do` commit looks good to me and could be moved in a separate merge request.

The description of the `test: eliminate redundant code by eliminating determine_kernel_version` commit does not match the commit description (any more). Is that commit atomic (i.e. would the tests be green after just picking that one)?

test/TEST-60-NFS/test.sh
jozzsi: What change do you prefer here @bdrung ? Remove the usage of KVERSION from line 242 ? Happy to do so, but that would make the PR even bigger.

bdrung: Good question. Removing KVERSION is probably the cleanest solution. I am not sure if moving `export kernel=$KVERSION` down is correct, because `kernel` is probably set for one of those `inst` calls.

jozzsi: Removing KVERSION will touch every tests, not just the test that are involved currently in this commit. IMO that should be a spearate PR, or it should be done after https://github.com/dracut-ng/dracut-ng/pull/1221 is landed to avoid rebase hell.

jozzsi: Resolved in the latest revision.

test/run-qemu
```diff
@@ -8,7 +8,15 @@ ARCH="${ARCH-$(uname -m)}"
 QEMU_CPU="${QEMU_CPU:-max}"
 
 set_vmlinux_env() {
-    VMLINUZ=${VMLINUZ-"/lib/modules/${KVERSION}/vmlinuz"}
+    if ! [[ ${KVERSION-} ]]; then
+        VMLINUZ=$(find /boot/vmlinuz-* -type f 2> /dev/null | tail -1)
+        ! [ -f "$VMLINUZ" ] && VMLINUZ=$(find /lib/modules/ -type f -name vmlinuz 2> /dev/null | tail -1)
+    fi
```
bdrung: Why do you move this part up in the function? How is that related to this merge request?

bdrung (commented):
The commits are not atomic. Each commit should result in working test cases. But in this case at least the last-1 commit has failing tests.

Probably the "test: extract KVERSION from the generated initramfs" commit should be the first one and then the remaining changes should follow (probably squashed since they go together).

test/run-qemu
jozzsi: now run-qemu gets called with empty KVERSION. Kept the ordering of the code blocks (but now I had to touch almost all lines due to nested conditions.

bdrung (approved):
Looks good now.

test/run-qemu
```diff
@@ -8,28 +8,28 @@ ARCH="${ARCH-$(uname -m)}"
 QEMU_CPU="${QEMU_CPU:-max}"
 
 set_vmlinux_env() {
-    VMLINUZ=${VMLINUZ-"/lib/modules/${KVERSION}/vmlinuz"}
```
bdrung, line 11: Since some tests are failing: How about determining KVERSION using lsinitrd in case it is not set?

PR #1471, opened 2025-07-20
fix(dracut): detect if systemd-detect-virt is available before calling it

jozzsi:
## Changes

When `systemd-detect-virt` is not available (e.g. in a non-systemd host) The following error is reported

```
/usr/sbin/dracut: line 1435: systemd-detect-virt: command not found
```

Let's detect if `systemd-detect-virt` is available before calling it.

Follow-up to 000f5db .

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1431,7 +1431,7 @@ else
     exit 1
 fi
 
-if container=$(systemd-detect-virt -c) &> /dev/null; then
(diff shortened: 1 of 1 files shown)


bdrung (approved):
LGTM. It's the only place that is not covered by `command -v` or `type -P`.

PR #1477, opened 2025-07-29
docs: switch to asciidoctor for man generation, with fallback

bdrung:
The man page warnings trigger me to push for this asciidoctor change again:

## Changes

lintian complains about the man pages in the Debian/Ubuntu package. The following warnings are only present when building the man pages with asciidoc and they will be fixed by using asciidoctor:

```
W: dracut-core: groff-message troff:<standard input>:183: warning: macro 'an-trap' not defined [usr/share/man/man8/dracut.8.gz:1]
W: dracut-core: groff-message troff:<standard input>:537: warning: macro 'an-trap' not defined [usr/share/man/man7/dracut.cmdline.7.gz:1]
W: dracut-core: groff-message troff:<standard input>:537: warning: macro 'an-trap' not defined [usr/share/man/man7/dracut.kernel.7.gz:1]
W: dracut-core: groff-message troff:<standard input>:68: warning: macro 'an-trap' not defined [usr/share/man/man7/dracut.modules.7.gz:1]
W: dracut-core: groff-message troff:<standard input>:72: warning: macro 'an-trap' not defined [usr/share/man/man5/dracut.conf.5.gz:1]
```

Other projects like git use asciidoctor.

This switches the man page generator to asciidoctor, with a fallback option to configure to use the existing asciidoc/docbook path.

The reason to do this is to consolidate the published site generation and man page generation around asciidoctor.  With that in place, we can use some asciidoctor features like extensions to better link the man pages together, or provide other tweaks to the documentation.

This switches the default generation to asciidoctor, but adds a --disable-asciidoctor switch as a temporary escape hatch for packagers.

The testing is converted to asciidoctor, with an additional test of the asciidoc toolchain just to avoid major regressions there.

## Checklist

Fixes #


Conan-Kudo (approved):
I don't love this, but...

PR #1481, opened 2025-07-29
fix(dracut-initramfs-restore): check for Debian initrd.img symlink

bdrung:
## Changes

In case no initrd was found for the kernel version, also check the the symlinks `/boot/initrd.img` (for Ubuntu) and `/initrd.img` (for Debian).

## Checklist

Fixes #


bdrung:
Maybe. `dracut-initramfs-restore.sh` has the symlink checking fallback that lsinitrd does not have.

PR #1482, opened 2025-07-29
fix(dracut): default to initrd.img-${kernel} on Debian/Ubuntu

bdrung:
## Changes

Debian/Ubuntu carries a patch to change the default `initrdname` to `initrd.img-${kernel}`. This is done to keep the name the same as initramfs-tools which has been the default for a long time. This change will be carried for quite some time to ease the switch from initramfs-tools to dracut.

So better include this default change for Debian/Ubuntu upstream.

## Checklist

Fixes #


bdrung:
IMO I still prefer to change the dracut code for changing the default. This allows to run dracut with all configurations files removed and still getting a good result.

PR #1483, opened 2025-07-29
fix(dracut): only call uname -r as last resort

bdrung:
## Changes

Running dracut in a schroot environment with `systemd-detect-virt` installed will pick `uname -r` as kernel version despite that kernel not being available. The same could happen inside containers without `systemd-detect-virt` installed.

So always check `/lib/modules` and `/usr/lib/modules` before falling back to calling `uname -r`.

## Checklist

Fixes #


bdrung:
Good point. We could change the check to check for `--container` or `--chroot`.

PR #1484, opened 2025-07-30
feat(debian.conf): default to initrd.img-${kernel} on Debian/Ubuntu

bdrung:
## Changes

Add an example configuration that Debian/Ubuntu can ship to change the default `initrdname` to `initrd.img-${kernel}`. This is done to keep the name the same as initramfs-tools which has been the default for a long time. This change will be carried for quite some time to ease the switch from initramfs-tools to dracut.

## Checklist

Fixes #

### DIFF
--- a/dracut.conf.d/debian/01-debian.conf
+++ b/dracut.conf.d/debian/01-debian.conf
@@ -1,2 +1,3 @@
 # Debian/Ubuntu specific Dracut configuration
 i18n_vars="/etc/locale.conf:LANG /etc/default/console-setup:FONT,FONT_MAP,CONSOLE_MAP-FONT_UNIMAP /etc/default/keyboard:XKBLAYOUT-KEYMAP,XKBVARIANT-EXT_KEYMAPS"
+initrdname="initrd.img-${kernel}"


bdrung:
I like to keep the kernel name close to what initramfs-tools does for the transition phase. Or at least co-ordinate the name change. That probably needs to be discussed with the kernel team and agreed on a plan for the rename.

bdrung:
Rebased this merge request on https://github.com/dracut-ng/dracut-ng/pull/1487. Please review https://github.com/dracut-ng/dracut-ng/pull/1487 first.

man/dracut.conf.5.adoc
```diff
@@ -370,7 +370,8 @@ The initrdname configuration option is required to match the
 _initr*$\{kernel\}*_ file pattern and only one file with this pattern should
 exists in the directory where initramfs is loaded from.
 +
-Defaults to _initramfs-$\{kernel\}.img_.
+Defaults to _initramfs-$\{kernel\}.img_. Debian/Ubuntu based systems normally
+ship a configuration file to change the defaults to _initrd.img-$\{kernel\}_.
```
Conan-Kudo: Please drop this change.

Conan-Kudo (changes requested):

man/dracut.conf.5.adoc
jozzsi: This seems harsh. The Linux ecosystem is using initramfs very inconsistently and we still use the name initrd in critical places, such as in systemd (e.g. `initrd.target` and `/usr/lib/initrd-release`).

When it comes to distribution specific changes, I suggest to consider the size of the commit and the size of the user base that the commit is impacting. 

That said, this commit in documentation would be quite easy to carry downstream as a patch.

bdrung: Would there be a wording that would be accepted upstream? For example:
> Distributions might set a different default via a configuration file in /usr/lib/dracut/dracut.conf.d

Conan-Kudo: Make it more generic, since it's not just distributions, anyone can make a snippet to change the name format.

AlexanderKurtz: I have been watching this PR for a while and I think that maybe @Conan-Kudo is right here after all. IMHO the manpage should document the *upstream* defaults, at least that's what I'd personally prefer as a user of dracut.

The fact that that Debian (or any other distro) ships a config file which changes some of these options doesn't need to be documented in the manpage, it can easily be discovered by looking at at the contents of `/usr/lib/dracut/dracut.conf.d/` and can be overridden by dropping a file in `/etc/dracut.conf.d`. IMHO the correct place to to mention the existence of `10-debian.conf` would be in a `README.Debian` file in `/usr/share/doc/dracut` (which AFAIK is inherently Debian-specific).

bdrung: I removed the man page change.

PR #1485, opened 2025-07-30
test: drop setting DRACUT_PATH

bdrung:
## Changes

Commit 2606f985d6f3 ("feat(dracut): drop DRACUT_PATH and rely on PATH") removed the `DRACUT_PATH` variable. So `DRACUT_PATH` does not need to be set in the tests any more.

@LaszloGombos @Conan-Kudo 

## Checklist

Fixes #


bdrung:
rebased

PR #1487, opened 2025-07-30
feat: add Debian/Ubuntu specific Dracut configuration

bdrung:
## Changes

Add the Debian/Ubuntu specific Dracut configuration, but do not install this configuration by default. It is meant to be used via `./configure --configprofile debian` during Debian/Ubuntu package build.

## Checklist

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,6 +188,7 @@ install: all
 	install -m 0755 dracut-logger.sh $(DESTDIR)$(pkglibdir)/dracut-logger.sh
 	install -m 0755 dracut-initramfs-restore.sh $(DESTDIR)$(pkglibdir)/dracut-initramfs-restore
 	cp -arx modules.d dracut.conf.d $(DESTDIR)$(pkglibdir)
+	rm -r $(DESTDIR)$(pkglibdir)/dracut.conf.d/debian/
 	for i in $(configprofile) ; do \
 		cp -arx dracut.conf.d/$$i/* $(DESTDIR)$(pkglibdir)/dracut.conf.d/ ;\
 	done
--- a/dracut.conf.d/debian/10-debian.conf
+++ b/dracut.conf.d/debian/10-debian.conf
@@ -1,0 +1,3 @@
+# Debian/Ubuntu specific Dracut configuration
+hostonly="yes"
+hostonly_cmdline="no"


Makefile
```diff
@@ -188,6 +188,7 @@ install: all
 	install -m 0755 dracut-logger.sh $(DESTDIR)$(pkglibdir)/dracut-logger.sh
 	install -m 0755 dracut-initramfs-restore.sh $(DESTDIR)$(pkglibdir)/dracut-initramfs-restore
 	cp -arx modules.d dracut.conf.d $(DESTDIR)$(pkglibdir)
+	rm -r $(DESTDIR)$(pkglibdir)/dracut.conf.d/debian/
```
jozzsi, line 191: If we're doing this step, should we also remove dracut.conf.d/*.example ? 

Not expecting it to be as part of this PR, just expecting a level of consistency eventually... 

CC @pvalena @aafeijoo-suse for visibility

bdrung: Good point. I'll do a follow-up PR to remove those.

aafeijoo-suse: That will break [this](https://github.com/openSUSE/dracut-ng/blob/eb9bf4cb62d74be7478f14631088bbae77c5d1c4/suse/dracut.spec#L163) and [this](https://src.fedoraproject.org/rpms/dracut/blob/rawhide/f/dracut.spec#_252).

bdrung: These two references install dracut.conf.d/$distro.conf.example into `%{buildroot}%{dracutlibdir}/dracut.conf.d/01-dist.conf`. The proposed removal is `%{buildroot}%{dracutlibdir}/dracut.conf.d/$distro.conf.example`. But let's discuss it in the upcoming merge request.

bdrung: Here is the merge request for it: https://github.com/dracut-ng/dracut-ng/pull/1491

aafeijoo-suse: > These two references install dracut.conf.d/$distro.conf.example into `%{buildroot}%{dracutlibdir}/dracut.conf.d/01-dist.conf`. The proposed removal is `%{buildroot}%{dracutlibdir}/dracut.conf.d/$distro.conf.example`.

My bad, you're right, thanks for clarifying it. In fact, [we are already removing those in our spec file](https://github.com/openSUSE/dracut-ng/blob/eb9bf4cb62d74be7478f14631088bbae77c5d1c4/suse/dracut.spec#L162).

PR #1494, opened 2025-08-01
fix(btrfs): write cmdline in install()

aafeijoo-suse:
This fixes the build with `--kernel-only` option.

```
$ dracut -f --kernel-only
...
dracut[I]: *** Including module: btrfs ***
/usr/lib/dracut/modules.d/90btrfs/module-setup.sh: line 35: /var/tmp/dracut.dqV4ugj/initramfs/etc/cmdline.d/00-btrfs.conf: No such file or directory
```

## Checklist

### DIFF
--- a/modules.d/70btrfs/module-setup.sh
+++ b/modules.d/70btrfs/module-setup.sh
@@ -32,7 +32,6 @@ cmdline() {
 # called by dracut
 installkernel() {
     hostonly='' instmods btrfs
-    printf "%s\n" "$(cmdline)" > "${initdir}/etc/cmdline.d/00-btrfs.conf"
 }
 
 # called by dracut
@@ -57,4 +56,8 @@ install() {
 
     inst_multiple -o btrfsck btrfs-zero-log btrfstune
     inst btrfs /sbin/btrfs
+
+    if [[ $hostonly_cmdline == "yes" ]]; then
+        printf "%s\n" "$(cmdline)" > "${initdir}/etc/cmdline.d/00-btrfs.conf"
+    fi
 }


Conan-Kudo (approved):

modules.d/70btrfs/module-setup.sh
```diff
@@ -57,4 +56,8 @@ install() {
 
     inst_multiple -o btrfsck btrfs-zero-log btrfstune
     inst btrfs /sbin/btrfs
+
+    if [[ $hostonly_cmdline == "yes" ]]; then
+        printf "%s\n" "$(cmdline)" > "${initdir}/etc/cmdline.d/00-btrfs.conf"
```
LaszloGombos, line 61: The content of this file (`00-btrfs.conf`) inside the initrafms will be always  ` rd.driver.pre=btrfs` regardless of the host configuration. This would imply that this file is generic and not hostonly, so I think we should either remove this hack (word borrowed from the comment above) or we should not add the `$hostonly_cmdline == "yes"` test as a condition.

This hack was added 5+ years ago (https://github.com/dracutdevs/dracut/issues/658) and I do not think any other initramfs generators carry this hack. Would this be a good time to discuss why would this change still needed ? 

If the hack is indeed still needed and we add the `hostonly` test, that might break generic initramfs.

aafeijoo-suse: > The content of this file (`00-btrfs.conf`) inside the initrafms will be always ` rd.driver.pre=btrfs` regardless of the host configuration. This would imply that this file is generic and not hostonly, so I think we should either remove this hack (word borrowed from the comment above) or we should not add the `$hostonly_cmdline == "yes"` test as a condition.

`hostonly` and `hostonly_cmdline` are independent options if the latter is set.

> This hack was added 5+ years ago ([dracutdevs/dracut#658](https://github.com/dracutdevs/dracut/issues/658)) and I do not think any other initramfs generators carry this hack.

Quoting a sentence that is written in that issue: "_this problem also affected debian initramfs-tools before._"

> Would this be a good time to discuss why would this change still needed ?

The comment says "Hack for slow machines". I'm not interested in discussing whether all machines are fast enough these days, only in fixing the `--kernel-only` option.

> If the hack is indeed still needed and we add the `hostonly` test, that might break generic initramfs.

We are not adding the `hostonly` test, we are adding `hostonly_cmdline`, exactly like all dracut modules are doing before extending the kernel command line via /etc. The same patch submitted in this PR is present in openSUSE Tumbleweed since October 2024, no complaints yet.

LaszloGombos: > We are not adding the hostonly test, we are adding hostonly_cmdline

Good point, indeed I misunderstood `hostonly_cmdline`. The name of this option is confusing, but that is beyond the scope of this PR.

> The same patch submitted in this PR is present in openSUSE Tumbleweed since October 2024, no complaints yet.

This is useful info for the review. openSUSE sets `hostonly_cmdline=yes`, so the `hostonly_cmdline==yes` makes no difference for openSUSE.

Most other distributions however (e.g. Fedora) set `hostonly_cmdline=no` and this would mean that `rd.driver.pre=btrfs` will no longer applied - effectively disabling this "hack" by default.

Since I was arguing for revisiting this hack, I do not mind having it disabled now by default for most distributions except openSUSE.

PR #1501, opened 2025-08-02
 test: fix Linux kernel detection for qemu runs

jozzsi:
## Changes

Determine KVERSION from the generated initramfs for qemu runs. Extend fallback detection even when KVERSION is specified.
 
Resolves CI regression (on [arm daily integration tests](https://github.com/dracut-ng/dracut-ng/actions/workflows/integration-extra-arm64.yml)) introduced by db608c3 .

## Checklist

### DIFF
--- a/test/run-qemu
+++ b/test/run-qemu
@@ -29,10 +29,10 @@ set_vmlinux_env() {
                 VMLINUZ="/boot/Image-${KVERSION}"
             fi
         fi
-    else
-        VMLINUZ=$(find /boot/vmlinuz-* -type f 2> /dev/null | tail -1)
-        ! [ -f "$VMLINUZ" ] && VMLINUZ=$(find /lib/modules/ -type f -name vmlinuz 2> /dev/null | tail -1)
     fi
+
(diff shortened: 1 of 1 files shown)


test/run-qemu
```diff
@@ -90,6 +90,11 @@ fi
 
 # only set -kernel if -initrd is specified
 if [[ $* == *-initrd* ]]; then
+    for arg in "$@"; do
+        [[ $1 == *-initrd* ]] && break
+    done
+
+    KVERSION=$(lsinitrd "$arg" | grep modules.dep | head -1 | rev | cut -d'/' -f2 | rev)
```
bdrung, line 97: Could we rely on successfully determine `KVERSION` in all cases?

jozzsi: For the purpose of the CI test, yes we can determine KVERSION (I tested and CI regressions are resolved by this change).

But in the overall general case, we no we can not. 

This is the main reason I have not (yet) proposed to add the command line option for lsinitrd to return this value in https://github.com/dracut-ng/dracut-ng/pull/1465 (it is using the exact same sequence of commands).

PR #1502, opened 2025-08-02
fix(dracut): enforce that hostonly is set for hostonly_install

jozzsi:
## Changes

Before honoring `-H` command line argument, check if `hostonly` is actually set.

[Here](https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/74nvmf/module-setup.sh#L145) is an example where `inst` is called even in generic mode which would result in picking up files from the host from the /etc directory.

```
inst_simple -H "/etc/nvme/hostnqn"
```

## Checklist

Fixes #

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -240,7 +240,7 @@ inst_dir() {
 inst() {
     local dstdir="${dstdir:-"$initdir"}"
     local _ret _hostonly_install
-    if [[ $1 == "-H" ]]; then
+    if [[ $1 == "-H" ]] && [[ $hostonly ]]; then
         _hostonly_install="-H"
         shift
     fi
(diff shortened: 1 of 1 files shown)


dracut-init.sh
```diff
@@ -245,7 +245,7 @@ inst_dir() {
 inst() {
     local dstdir="${dstdir:-"$initdir"}"
     local _ret _hostonly_install
-    if [[ $1 == "-H" ]]; then
+    if [[ $1 == "-H" ]] && [[ ${hostonly-} ]]; then
```
mwilck: What's the point of using `${hostonly-}` here instead of the more familiar `"${hostonly}"`?

jozzsi: Not much really from my perspective, just a copy paste from other parts of the code

I am in the process of trying to undo some of that - https://github.com/dracut-ng/dracut-ng/pull/1508

Thanks for pointing it out, will remove it from here.

PR #1508, opened 2025-08-02
 chore: always set hostonly and hostonly_mode variables

jozzsi:
It makes it easier to reason both for built-in and out-of-tree dracut modules with the assumption that `hostonly` and `hostonly_mode` variables are always set.
    
Document most commonly used shell variables for dracut modules and document that they are always set.

Partial revert of c85c93245bf .

Note: Run the following in the `modules.d` directory

```
sed -i 's/\${hostonly\-}/hostonly/g' 
```

## Checklist


bdrung:
I agree to make `hostonly` and `hostonly_mode` consistent.

In case you prefer to always set these variables, we should document that behavior IMO.

bdrung (approved):

PR #1511, opened 2025-08-03
test(NFS): use test dracut modules

### LINKED ISSUE #1534: TEST-72-NBD fails in test_check if KVERSION is unset
test_check in TEST-72-NBD:
```sh
    if ! modinfo -k "$KVERSION" nbd &> /dev/null; then
        echo "Kernel module nbd does not exist"
        return 1
    fi
```

A recent commit removed setting `KVERSION`. So this check does not work any more and `set -u` will cause it to fail.

@jozzsi

jozzsi:
## Changes

Make use of test dracut modules and configuration and make the test consistent with the rest of the CI.

## Checklist

Partially Fixes #1534

### DIFF
--- a/test/TEST-60-NFS/test.sh
+++ b/test/TEST-60-NFS/test.sh
@@ -231,80 +231,32 @@ test_run() {
 }
 
 test_setup() {
-    DRACUT_PATH=${DRACUT_PATH:-/sbin /bin /usr/sbin /usr/bin}
-    export no_kernel=
-    # Detect lib paths
-
     "$DRACUT" -N --keep --tmpdir "$TESTDIR" \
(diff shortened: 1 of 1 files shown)


test/TEST-60-NFS/test.sh
```diff
@@ -314,26 +266,18 @@ test_setup() {
     cp ./client-init.sh "$initdir"/sbin/init
 
     # second, install the files needed to make the root filesystem
```
bdrung, line 268: Is that comment still needed?

jozzsi: https://github.com/dracut-ng/dracut-ng/pull/1545/files

PR #1514, opened 2025-08-04
feat(Makefile): set Dracut version on install

bdrung:
## Changes

Distributions can override `DRACUT_VERSION` or `DRACUT_FULL_VERSION` to include the distribution's revision in the dracut version. This override is not reflected in the installed `dracut-version.sh` and needs to be modified by the distribution.

Persist the version in the installed `dracut-version.sh` during `make install`.

## Checklist

Fixes #


bdrung:
This one line change results in removing 5 lines in `debian/rules`.

PR #1516, opened 2025-08-05
feat(dracut.conf.d): rename hostonly config to default config

jozzsi:
## Changes

Some distributions (such as Arch) do not provide default configuration. For those distributions, it would be an increased value to provide improve default beyond just setting hostonly.

This commit introduces the notion of default config (and eliminiates the idea for the need of independent hostonly config).

Distributions maintaining independent configurations default (e.g. Fedora) can choose to not package this default configuration.

Set the default log level to warning (matching current openSUSE and Fedora) defaults.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ pkgconfigdatadir ?= $(datadir)/pkgconfig
 configs = \
 	fips/10-fips.conf \
 	generic/11-generic.conf \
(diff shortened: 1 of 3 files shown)


bdrung:
IMO the default should be the default instead having a default plus a "default" configuration. So we should make "hostonly=yes" the default and not install the hostonly config by default.

PR #1525, opened 2025-08-06
feat: change the default for hostonly_cmdline to no

### LINKED ISSUE #1536: Document `hostonly-cmdline` a bit better to prevent nasty surprises and/or warn on no root?
After upgrading from dracut 107 to 108, my setup no longer generated a bootable initramfs since the image is missing the `root=` parameter (as reported by dracut on boot).

It took a bit of digging, but eventually I narrowed it down to a default being flipped: specifically, `--hostonly-cmdline` defaulted to `yes` in 107 but has been flipped to `no` in 108, per efaee4473680db78f4f3244ba952e783ea3b5aa9.

Setup: Manjaro rolling distribution, systemd. My `local.conf` looks like this:

```
hostonly="yes"
hostonly_mode="strict"
do_strip="yes"
aggressive_strip="yes"
compress="zstd"
dracutmodules="rootfs-block kernel-modules dash"
add_drivers+=" xhci_pci_renesas xhci_pci xhci_hcd usbhid nvme "
force_drivers+=" xhci_pci_renesas xhci_pci xhci_hcd usbhid nvme "
uefi="yes"
machine_id="no"
kernel_cmdline="sysrq_always_enabled=1 amdgpu.ppfeaturemask=0xfff7ffff fbcon=nodefer mitigations=off zswap.enabled=0 nowatchdog rd.shell"
```

With dracut 107, the initramfs generated has
```
root=UUID=1917c015-dae0-4501-a53a-9c850edab31d rootfstype=ext4 rootflags=rw,relatime
```
Which is indeed the root partition; with dracut 108, this is missing and the system doesn't boot. This all comes down to:

```
/usr/lib/dracut/modules.d/[74,95]rootfs-block/module-setup.sh@67(install): [[ no == yes ]]

 # called by dracut
     66 install() {
     67     if [[ $hostonly_cmdline == "yes" ]]; then
     68         local _journaldev
     69         _journaldev=$(cmdline_journal)
```

Setting the option explicitly to `yes` in `local.conf` made dracut 108 produce bootable images again.

Thing is, the documentation for what `hostonly_cmdline` does is ever so slightly scant, so it's rather hard to anticipate not having it can render your system unbootable. The `--help` only says: "--hostonly-cmdline  Store kernel command line arguments needed", which seems like a good thing (they're needed, no?) and fails to explain why anyone would ever set it to `no`, let alone why this should be the default (nor does the help of course indicate which one is the default).

This is not necessarily to suggest the dracut behavior itself should be changed (well, except maybe that it might report having generated an image with no `root=`, since even if the user asked for this, I expect that's at least warning material); this report is mostly so any other victims can find this info and/or to solicit tips on how to do things "properly" if this configuration is unusual/broken/not recommended, so the relevant wikis can be updated with better info.

jozzsi:
## Changes

This change in theory break compatibility with earlier dracut releases.

In practice, Fedora-based and Debian-based distributions have all configured `hostonly_cmdline` to default to "no" already in their respective configuration files.

This configuration is set to "yes" in openSUSE based distributions. For those distributions that explicitly set this variable , this change in defaults will not make have an impact.

Intentionally did not modified Fedora default config, I think it is hard to get attention from the Fedora team to review.

## Checklist

Partially fixes #1536

### DIFF
--- a/dracut.conf.d/debian/10-debian.conf
+++ b/dracut.conf.d/debian/10-debian.conf
@@ -1,3 +1,2 @@
 # Debian/Ubuntu specific Dracut configuration
 hostonly="yes"
-hostonly_cmdline="no"
--- a/dracut.conf.d/hostonly/10-hostonly.conf
+++ b/dracut.conf.d/hostonly/10-hostonly.conf
@@ -1,3 +1,2 @@
 # optimize initrd to be as small as possible for faster boot performance, tailored to the current host
 hostonly="yes"
-hostonly_cmdline=no
--- a/dracut.conf.d/test-makeroot/test-makeroot.conf
+++ b/dracut.conf.d/test-makeroot/test-makeroot.conf
@@ -4,5 +4,4 @@ compress="cat"
 do_strip="no"
 do_hardlink="no"
 early_microcode="no"
-hostonly_cmdline="no"
 stdloglvl=2
--- a/dracut.conf.d/test-root/test-root.conf
+++ b/dracut.conf.d/test-root/test-root.conf
@@ -4,5 +4,4 @@ compress="cat"
 do_strip="no"
 do_hardlink="no"
 early_microcode="no"
-hostonly_cmdline="no"
 stdloglvl=2
--- a/dracut.conf.d/test/test.conf
+++ b/dracut.conf.d/test/test.conf
@@ -3,7 +3,6 @@ add_dracutmodules+=" test "
 do_strip="no"
 do_hardlink="no"
 early_microcode="no"
-hostonly_cmdline="no"
 
 # systemd on arm64 on GitHub CI needs the 2 min timeout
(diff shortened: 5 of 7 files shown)


JustTNE:
Changing defaults out from under people is scary, and there have been a fair share of GitHub issues here in this repository because of that due to unbootable systems. It's probably pretty likely other people outside of these GitHub issues were affected as well.

Garuda Linux specifically already used the `=no` version since Garuda Linux supplies these via the bootloader (grub) by default, so there would be absolutely no objections from the Garuda Linux end for changing the default.

I suppose it depends on how much breakage for custom configurations dracut-ng as a project is comfortable with, to change a default setting.

dalto8:
EndeavourOS is currently setting it to "no" via the command line.  If it was no by default, we would probably not do that anymore and if someone wanted to use it they could toggle it on in the config.

So, for us, "no" would be a better default.

That being said, it isn't a deal breaker either way.

PR #1527, opened 2025-08-06
test: drop requiring iscsiuio.socket

bdrung:
## Changes

The iscsiuio socket/service is only needed on specific hardware. It is not needed for the ISCSI and ISCSI-MULTI test cases. Without this check the tests will succeed on Debian and Ubuntu.

This needs https://github.com/dracut-ng/dracut-ng/pull/1528 to succeed.

## Checklist

Fixes #

### DIFF
--- a/.github/workflows/integration-extra.yml
+++ b/.github/workflows/integration-extra.yml
@@ -128,13 +128,6 @@ jobs:
                     - "70"
                     - "71"
                     - "72"
-                exclude:
-                    # https://github.com/dracut-ng/dracut-ng/issues/446
-                    - container: debian:sid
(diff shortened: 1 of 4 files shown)


bdrung:
I successfully tested with unstable which should be close to testing (i.e. trixie). open-iscsi on bookworm ships `/sbin/iscsi-iname`. So I do not know why it failed to find it.

bdrung:
Since the other tests switched from `ubuntu:rolling` to `ubuntu:devel`, let's use `ubuntu:devel` here as well.

Conan-Kudo (dismissed):

PR #1531, opened 2025-08-07
feat(systemd): drop unnecessary dependency on libgcrypt

pvalena:
## Changes

I am not sure whether other distro use it, but we don't need it anymore in Centos Stream (and probably Fedora).

Ref: https://issues.redhat.com/browse/RHEL-95542

## Checklist

Fixes #

### DIFF
--- a/modules.d/11systemd-journald/module-setup.sh
+++ b/modules.d/11systemd-journald/module-setup.sh
@@ -59,7 +59,6 @@ install() {
     # Install library file(s)
     if [[ ! $USE_SYSTEMD_DLOPEN_DEPS ]]; then
         inst_libdir_file \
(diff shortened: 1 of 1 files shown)


chewi:
This was discussed in #639 and partly addressed in #1260. If you have at least systemd 257, gcrypt will only be pulled in by the systemd-journald module. If you want to avoid that, you can add `omit_dlopen_features+=" libsystemd-shared-*.so:gcrypt "` to your config. Perhaps the line you've highlighted could still be dropped to avoid it in older versions though since it is a very niche feature.

devkontrol (dismissed):
Thanks @chewi for the pointers and feedback. LGTM

pvalena:
- rebased

chewi (approved):

devkontrol (approved):

PR #1539, opened 2025-08-09
feat(fstab-initramfs): add module to add a fstab to the initramfs

svvac:
Add a simple module to provide the initramfs with an `/etc/fstab`.

My personal use-case is mounting a partition early in order to provide secrets for the boot process, including mounting the sysroot, which means fstab-sys comes in too late.

## Checklist


svvac:
Sorry for wasting your time.

PR #1544, opened 2025-08-09
fix: always include modules required for networking

### LINKED ISSUE #1437: nfs module fails with strict hostonly mode
According to [this](https://github.com/dracut-ng/dracut-ng/pull/1348#issuecomment-3051962123) comment.

To make this actionable, it would be great to have reproduction steps 

CC @bdrung @aafeijoo-suse

jozzsi:
## Changes

For static IPv4 configurations, we use arping to check for duplicate IP addresses. arping requires the af_packet module to work, and if arping fails, the IP address will not be set and booting will fail. af_packet may not be loaded / required in the running system, for example if the system had been booted in an IPv6 configuration, or if it had been manually unloaded. Make sure it's included in initramfs in hostonly mode, too.

network-manager always requires nf_tables and other kernel modules in the generated initramfs to function, regardless if these modules are loaded on the host or not at the time of initramfs generation.

Remove the hostonly_mode=sloppy workaround. The NETWORK test now passes regardless of the value of `hostonly_mode`.      

Follow-up to b074216, 5e1fec168 and to de862885e .

## Checklist

Fixes #1437

### DIFF
--- a/modules.d/35network-legacy/module-setup.sh
+++ b/modules.d/35network-legacy/module-setup.sh
@@ -15,12 +15,6 @@ depends() {
 }
 
 # called by dracut
-installkernel() {
-    # arping depends on af_packet
-    hostonly='' instmods af_packet
-}
-
-# called by dracut
 install() {
     local _arch
 
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -19,7 +19,7 @@ depends() {
 
 # called by dracut
 installkernel() {
-    hostonly=$(optional_hostonly) instmods nf_tables nfnetlink nft_fwd_netdev
+    hostonly='' instmods nf_tables nfnetlink nft_fwd_netdev
 }
 
 # called by dracut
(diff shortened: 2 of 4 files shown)


modules.d/70kernel-network-modules/module-setup.sh
```diff
@@ -36,5 +36,8 @@ installkernel() {
         '=drivers/net/phy' \
         '=drivers/net/team' \
         '=drivers/net/ethernet' \
-        ecb arc4 bridge stp llc ipv6 bonding 8021q ipvlan macvlan af_packet virtio_net xennet
+        ecb arc4 bridge stp llc ipv6 bonding 8021q ipvlan macvlan virtio_net xennet
+
+    # arping depends on af_packet
+    hostonly='' instmods af_packet
```
bdrung: The `arping` call is in the net-lib module. So shouldn't the `af_packet` package installed there?

jozzsi: Great idea. Done.

bdrung (approved):

PR #1548, opened 2025-08-10
feat(configure): autoconfigure configprofile

jozzsi:
## Changes

The original motivation of this PR was to set the openSUSE config                                                                                                                                                                   
inside the openSUSE CI container.                                                                                                                                                                                                   
                                                                                                                                                                                                                                    
This commit will autoconfigure those configprofiles that are                                                                                                                                                                        
maintained upstream for fedora, debian and opensuse based distro's.                                                                                                                                                                 
                                                                                                                                                                                                                                    
One notable change from this commit is that it would now allow testing                                                                                                                                                              
hostonly_cmdline=yes configuration which is currently untested                                                                                                                                                                      
on the CI. 

## Checklist

### DIFF
--- a/configure
+++ b/configure
@@ -22,8 +22,27 @@ fi
 CC="${CC:-cc}"
 PKG_CONFIG="${PKG_CONFIG:-pkg-config}"
 
-# set hostonly by default
-configprofile="${configprofile:-hostonly}"
+# set configprofile based on ID and ID_LIKE from os-release
+if [[ -f /etc/os-release ]]; then
+    . /etc/os-release
+elif [[ -f /usr/lib/os-release ]]; then
+    . /usr/lib/os-release
+fi
+
+case " $ID $ID_LIKE " in
+    *" debian "*)
+        configprofile=debian
+        ;;
+    *" fedora "*)
+        configprofile=fedora
+        ;;
+    *" suse "*)
+        configprofile=opensuse
+        ;;
+    *)
+        configprofile=hostonly
+        ;;
+esac
 
 # Little helper function for reading args from the commandline.
 # it automatically handles -a b and -a=b variants, and returns 1 if


configure
```diff
@@ -22,8 +22,27 @@ fi
 CC="${CC:-cc}"
 PKG_CONFIG="${PKG_CONFIG:-pkg-config}"
 
-# set hostonly by default
-configprofile="${configprofile:-hostonly}"
+# set configprofile based on ID_LIKE from os-release
```
bdrung:
```suggestion
# set configprofile based on ID and ID_LIKE from os-release
```

configure
```diff
@@ -22,8 +22,27 @@ fi
...
+    debian*)
+        configprofile=debian
+        ;;
+    *suse)
+        configprofile=opensuse
+        ;;
+    *)
+        configprofile=hostonly
+        ;;
+esac
```
bdrung: `$ID$ID_LIKE` would be `ubuntudebian` on Ubuntu.

`ID_LIKE` is a space separated list. So I suggest following to search for matching IDs (plus alphabetical order):

```suggestion
case " $ID $ID_LIKE " in
    *" debian "*)
        configprofile=debian
        ;;
    *" fedora "*)
        configprofile=fedora
        ;;
    *" suse "*)
        configprofile=opensuse
        ;;
    *)
        configprofile=hostonly
        ;;
esac
```

I grabbed samples from https://github.com/chef/os_release to check that the distributions set ID or ID_LIKE to those values.

bdrung (commented):
Excellent idea!

configure
jozzsi: Thanks, done !

configure
jozzsi: Thanks, done !

bdrung (approved):

Issue #1520, opened 2025-08-05
Missing dependency on unit: initrd.target


bonaccolto90:
Describe the bug
When i do an upgrade that update: kernel, nvidia driver or various modules of the system, when it comes to updating initramfs with dracuts, it shows a voice that says "Failed to add dependency on unit: Unit initrd.target does not exist" right after the ":: Building fallback initramfs for linux-zen (6.15.9-zen1-1-zen)"

Distribution used
Garuda Linux (based on Archlinux)

Dracut version
dracut 108

Init system
systemd

To Reproduce
sudo pacman -Syu

Expected behavior
It shouldn't report any error on updateing initramfs

Additional context
When i do:

sudo pacman -S systemd

This is the result (trimmed the code to hide superfluous information)

`... (13/19) Updating initramfs with dracut 
:: Rebuilding all initramfs due to changes in dracut files. 
:: Building initramfs for linux-zen (6.15.9-zen1-1-zen) 
:: Building fallback initramfs for linux-zen (6.15.9-zen1-1-zen) 
Failed to add dependency on unit: Unit initrd.target does not exist 
Failed to add dependency on unit: Unit initrd.target does not exist 
Failed to add dependency on unit: Unit initrd.target does not exist ...
`

This started since the release of the new update.

bonaccolto90:
> This could be caused by an out-of-tree dracut module.
> 
> Thanks for reporting this issue and I hope you can help with more information. Could you please post the output of calling dracut with `-v` option ? If dracut gets called from an Arch hook, than maybe just modify the hook temporary. Thanks

`dracut -v
dracut[I]: Executing: /usr/bin/dracut -v
dracut[F]: Can't write to /boot/efi/b52c4e9baa6040bdb6ce0d9f11da78b9/6.15.9-zen1-1-zen: Directory /boot/efi/b52c4e9baa6040bdb6ce0d9f11da78b9/6.15.9-zen1-1-zen does not exist or is not accessible.
`

bonaccolto90:
I executed the `dracut --list-modules` and initrd.target shows up.

On the forum of Garuda they say "just ignore it" like it's not a problem...



> From https://forum.garudalinux.org/t/cant-update-dracut/45329/45
> 
> > The interesting thing is that the dracut contributors on GitHub blame this issue on us
> 
> This is unwelcome commentary of this project, when there was a genuine effort to try to help .

[...]

PR #1549, opened 2025-08-10
feat(rdma): dracut module for rdma

jozzsi:
## Changes

Unify and improve dracut rdma module.

An addition this version of the module code:
 - moves the dracut module ordering from distro specific packaging files to common
 - adds supports for distributions not explicitly supported by https://github.com/linux-rdma/rdma-core, such as Arch, Debian
 - fixes dracut dependency and shellcheck issues

The intention of this PR is _increase_ the adaption of rdma and dracut. It is obvious (to me) that it is not desired to have a copy of this module both in https://github.com/linux-rdma/rdma-core and in here. Let's see if we can work together to achieve this goal.

CC @jgunthorpe @rleon @bdrung @yishaih 

Ps.: The very initial trigger for this conversation is the issue with the rdma module ordering. The current '05' value will not work. If this module is continue to be maintained by https://github.com/linux-rdma/rdma-core, than the module ordering should be `50`, see https://github.com/dracut-ng/dracut-ng/issues/1520#issuecomment-3164321429

## Checklist

### DIFF
--- a/modules.d/70rdma/module-setup.sh
+++ b/modules.d/70rdma/module-setup.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 check() {
-    [ -n "$hostonly" ] && [ -c /sys/class/infiniband_verbs/uverbs0 ] && return 0
+    [ -n "$hostonly" ] && [ -e /sys/class/infiniband_verbs/uverbs0 ] && return 0
     [ -n "$hostonly" ] && return 255
     return 0
 }


bdrung (commented):
It's a good idea to move the RDMA Dracut module from per-distribution files into the Dracut project.

The commit message for the first commit could be improved. Where does this script come from? What is the purpose, etc?

Maybe merging the shellcheck fix into the first commit (or al least do the formatting in the initial commit).

jgunthorpe:
For a long time we've had issues with distros preferring to override the packaging and not work upstream, I guess this includes the dracut script too? How do you imagine this working? Won't rdma-core and all the distros still provide the script in addition to the one in dracut now? If the problem is the duplication get the distros to use the existing upstream script..

PR #1550, opened 2025-08-10
ci: remove testing with ubuntu:latest

jozzsi:
## Changes

ubuntu:latest could be as old as two years old, and, while testing with debian:stable makes sense, testing with ubuntu:latest seems redundant.

Ubuntu is now well integrated with dracut and dracut usptream main no longer needs to target (or backported) to ubuntu:latest.

## Checklist

Fixes #


bdrung (approved):
Yes. Testing only ubuntu:rolling and ubuntu:devel is totally sufficient.

PR #1556, opened 2025-08-11
fix(fedora): always include chroot binary

jozzsi:
## Changes

It seems RedHat requires chroot, which is addressed by a commit revert currently instead of a proper permanent solution, see https://github.com/redhat-plumbers/dracut-rhel10/pull/49

This commit provides a long term permanent solution that does not require a downstream revert.

CC @dtardon @pvalena @Conan-Kudo 

## Checklist

Fixes #

### DIFF
--- a/dracut.conf.d/fedora/01-dist.conf
+++ b/dracut.conf.d/fedora/01-dist.conf
@@ -8,7 +8,10 @@ i18n_install_all="yes"
 
 stdloglvl=3
 sysloglvl=5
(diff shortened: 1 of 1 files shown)


mrc0mmand:
> What the heck why is `chroot(8)` required? @pvalena?

Pavel is currently from PTO, but I can answer this too - apparently customers use it to debug (and fix) stuff when things go south in initrds (rescue initrds, that include such debug tools by default, are not present in some cloud environments).

bdrung:
Then I suggest to record that reason as comment in the configuration file. Are `vi` and others included for the same reason?

PR #1558, opened 2025-08-11
fix(dracut-install): handle multiple sonames in dlopen JSON with less noise

### LINKED ISSUE #1552: dracut-install: WARNING: could not locate dlopen dependency libqrencode
```
./test/test.sh ubuntu:rolling 41 
```

```
dracut-install: WARNING: could not locate dlopen dependency libqrencode.so.3 requested by '/usr/lib/x86_64-linux-gnu/systemd/libsystemd-shared-257.so'
```

```
./test/test.sh arch 41 
```

```
dracut-install: WARNING: could not locate dlopen dependency libqrencode.so.3 requested by '/usr/lib/systemd/libsystemd-shared-257.8-2.so'
```

CC @chewi

chewi:
## Changes

We should not try to install every library referenced in the dlopen JSON soname array, only the first one present. The array is intended to be an ordered preference list.

The dlopen dependency failure warning was particularly noisy and likely to trigger. We were already caching the processed items in `resolve_lazy`, but `resolve_deps` recurses many times, so it was necessary to move the cache down a level. I didn't reuse `items` here because it would have clashed with its usage elsewhere.

I had to think about whether the cache would function correctly with changing values of `pdeps`. If a dependency is not found on the first attempt, it does not prevent its consumer from being installed, so it does not matter that it might be found via a `RUNPATH` on a subsequent attempt.

## Checklist
- [ ] I have reviewed and updated any documentation if relevant -- N/A

Fixes #1552

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -951,9 +951,10 @@ static void resolve_deps_dlopen_parse_json(Hashmap *pdeps, Hashmap *deps, const 
         for (size_t entry_idx = 0; entry_idx < sd_json_variant_elements(dlopen_json); entry_idx++) {
                 sd_json_variant *entry = sd_json_variant_by_index(dlopen_json, entry_idx);
                 sd_json_variant *feature_json = sd_json_variant_by_key(entry, "feature");
+                const char *feature = NULL;
 
                 if (feature_json && sd_json_variant_is_string(feature_json)) {
-                        const char *feature = sd_json_variant_string(feature_json);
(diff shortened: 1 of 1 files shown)


LaszloGombos (approved):

jozzsi (approved):
checked the CI logs, and based on the logs, the issue seems to be resolved. Thanks @chewi

chewi:
Please hit the merge button then. :slightly_smiling_face:

chewi:
Ah sorry, thought you did.

PR #1563, opened 2025-08-12
docs: Consistent use of "initramfs"

MggMuggins:
It was a little bit comical to read "dracut creates _initramfs_ archives." and then "dracut creates an initrd image..." :slightly_smiling_face: 

## Checklist

### DIFF
--- a/doc_site/modules/ROOT/pages/index.adoc
+++ b/doc_site/modules/ROOT/pages/index.adoc
@@ -60,7 +60,7 @@ dracut creates _initramfs_ archives.
 
 == dracut's approach
 
-dracut creates an initrd image by copying tools and files from an installed
+dracut creates an initramfs archive by copying tools and files from an installed
 system and combining it with dracut modules, usually found on an installed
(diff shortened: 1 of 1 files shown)


dtardon (changes requested):
The other way around, please...

dtardon:
> Can you clarify what you mean? Everything I've been able to find (primarily [here](https://www.kernel.org/doc/html/latest/filesystems/ramfs-rootfs-initramfs.html#what-is-initramfs) but haven't seen anything that conflicts with that) indicates that initrd is old/deprecated and initramfs is preferred. 

Note that:
1. That document is >20 years old.
2. It speaks about specific _kernel implementations_ of the concept (one of whose is long dead now).

> Or do you mean "initramfs image" and "initrd archive"?

No, I meant "initrd image" (or just "initrd"). Because that's the term [The Boot Loader Specification](https://uapi-group.org/specifications/specs/boot_loader_specification/) uses.

PR #1564, opened 2025-08-12
feat(dracut): support 3cpio for creating initrds

bdrung:
## Changes

`3cpio` is written in Rust and faster than `cpio`. Use `3cpio` in `dracut.sh` in case it is available. `3cpio` supports `SOURCE_DATE_EPOCH` and therefore clamping the files is not needed.

### Quick benchmark on Ubuntu questing in a chroot

With only cpio installed:
```
# hyperfine "dracut -f"
Benchmark 1: dracut -f
  Time (mean ± σ):      6.700 s ±  0.052 s    [User: 8.724 s, System: 3.668 s]
  Range (min … max):    6.646 s …  6.828 s    10 runs
```

With only 3cpio installed:

```
# hyperfine "dracut -f"
Benchmark 1: dracut -f
  Time (mean ± σ):      6.606 s ±  0.033 s    [User: 8.737 s, System: 3.435 s]
  Range (min … max):    6.552 s …  6.662 s    10 runs
```

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -270,7 +270,7 @@ Creates initial ramdisk images for preloading modules
                         Specify the compressor and compressor specific options
                          used by mksquashfs if squash module is called when
                          building the initramfs.
-  --enhanced-cpio       Attempt to reflink cpio file data using dracut-cpio.
+  --enhanced-cpio       Attempt to reflink cpio file data using 3cpio/dracut-cpio.
   --list-modules        List all available dracut modules.
   -M, --show-modules    Print included module's name to standard output during
                          build.
@@ -1356,6 +1356,7 @@ hostonly_mode=${hostonly_mode-}
 
 if [[ $reproducible == yes ]] && [[ -z ${SOURCE_DATE_EPOCH-} ]]; then
     SOURCE_DATE_EPOCH=$(stat -c %Y "$dracutbasedir/dracut-functions.sh")
+    export SOURCE_DATE_EPOCH
 fi
 
 if [[ -z $DRACUT_KMODDIR_OVERRIDE && -n $drivers_dir ]]; then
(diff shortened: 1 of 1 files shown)


bdrung:
`3cpio` provides everything that `dracut-cpio` provides (as of https://github.com/bdrung/3cpio/pull/26) and more (manifest support, listing + extracting).

This merge request will use 3cpio if available. If 3cpio is not available, nothing changes.

The midterm plan:
1. Land this merge request.
2. Wait for all distributions that currently build dracut-cpio to package and ship 3cpio
3. Then drop dracut-cpio from dracut

3cpio is already available in Arch (see https://repology.org/project/3cpio/versions).

Once this merge request landed I can work on the steps to use the manifest. Which means: Modify dracut-install to write to a manifest file instead of copying files. Then probably start with the kernel modules. Then all other files.

bdrung:
3cpio is still stuck in Debian new (see rust-threecpio on https://ftp-master.debian.org/new.html), but we can add 3cpio on both Ubuntu containers.

Nowa-Ammerlaan:
> I think this step should be "replace dracut-cpio with 3cpio", so that it is not required to be a separate package (distributions prefer to package it separately can continue to do so).
> 

Please don't bundle external dependencies. Gentoo's policy is to unbundle these wherever possible. Bundeling dependencies always creates surprises and is a potential security issue.

Conan-Kudo:
Yes, I will not accept having stuff bundled in the dracut sources. When we get rid of dracut-cpio for 3cpio, we will not replace it with a bundled copy of 3cpio. It _might_ be worth it for @bdrung to move 3cpio to the dracut-ng github organization, but even that's not a requirement.

Conan-Kudo (approved):

PR #1566, opened 2025-08-12
feat(dracut): support setting compression level separately

bdrung:
## Changes

Extend the Dracut configuration to configure the default compression level for each compression format (e. g. `compress_level_zstd` for zstd). Add `--compress-level` option to `dracut` to override the compression level from the configuration.

Then users could just call `dracut --zstd --compress-level 3` instead of `dracut --compress="zstd -3 -T0 -q"`.

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -264,6 +264,8 @@ Creates initial ramdisk images for preloading modules
                          passed compression program.  Make sure your kernel
                          knows how to decompress the generated initramfs,
                          otherwise you will not be able to boot.
+  --compress-level [LEVEL]
+                        Compression level passed to compression program.
   --no-compress         Do not compress the generated initramfs. This will
                          override any other compression options.
   --squash-compressor [COMPRESSION]
@@ -419,6 +421,7 @@ rearrange_params() {
             --long sysroot: \
             --long stdlog: \
             --long compress: \
+            --long compress-level: \
             --long squash-compressor: \
             --long prefix: \
             --long rebuild: \
@@ -750,6 +753,11 @@ while :; do
             ;;
         --compress)
             compress_l="$2"
+            PARMS_TO_STORE+=" '$2'"
+            shift
+            ;;
+        --compress-level)
+            compress_level_l="$2"
             PARMS_TO_STORE+=" '$2'"
             shift
             ;;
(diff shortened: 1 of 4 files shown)


dracut.sh
```diff
@@ -2553,34 +2570,34 @@ fi
...
-        compress="$DRACUT_COMPRESS_LZOP -9"
+        compress="$DRACUT_COMPRESS_LZOP -${compress_level_lzop-9}"
         ;;
     lz4)
-        compress="$DRACUT_COMPRESS_LZ4 -l -9"
+        compress="$DRACUT_COMPRESS_LZ4 -l -${compress_level_lz4-9}"
         ;;
     zstd)
-        compress="$DRACUT_COMPRESS_ZSTD -15 -q -T0"
+        compress="$DRACUT_COMPRESS_ZSTD -${compress_level_zstd-15} -q -T0"
```
pvalena, line 2694: isn't the syntax missing `:`?
```suggestion
        compress="$DRACUT_COMPRESS_ZSTD -${compress_level_zstd:-15} -q -T0"
```
Same for all the above.

dracut.sh
```diff
@@ -1151,6 +1159,15 @@ drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
...
 [[ $INITRD_COMPRESS ]] && compress=$INITRD_COMPRESS
 [[ $compress_l ]] && compress=$compress_l
+[[ $compress_level_l ]] && {
+    compress_level_bzip2=$compress_level_l
+    compress_level_gzip=$compress_level_l
+    compress_level_lz4=$compress_level_l
+    compress_level_lzma=$compress_level_l
+    compress_level_lzop=$compress_level_l
+    compress_level_xz=$compress_level_l
+    compress_level_zstd=$compress_level_l
```
pvalena, line 1173: Why do we need separate variables for each?

pvalena (changes requested):

dracut.sh
bdrung: No. the `${foo-15}` will return `15` if `foo` is not defined. The additional `:` would set this value, but this is not needed here.

dracut.sh
bdrung: We have separate config variables for each compression (zstd and lz4 can have higher values) and at that stage we don't know which compression is picked. The `--compress-level` overwrites all those config settings.

pvalena: AFAICT you're introducing separate config variables for each compression - I simply don't see the benefit; on a code level, this can be easily handled by single variable. Before we do that, I simply want to know the reasoning behind it.

bdrung: The reason is that the different compression levels have different compression level ranges. Imagine setting the compression level to 19, because you use zstd. Then you call `dracut --xz` and the compression level 19 does not exist for xz.

There might be other use cases like: for lz4 I want to use level 1 (to get the fastest compression) but for xz I want the highest compression.

pvalena: We don't have any such checks for those variables; you can set it _wrongly_ anyway... Also, I don't see why would you set two compression levels if you only use one compression method.

pvalena: (And defaults are handled separately from the respective variables.)

pvalena: > The reason is that the different compression levels have different compression level ranges. Imagine setting the compression level to 19, because you use zstd. Then you call `dracut --xz` and the compression level 19 does not exist for xz.

Oh I see, when you set it in config, and you override it on commandline, you basically want it to reset/revert to some default (or predefined value). That's why there's discrepancy ... you'd need to always override it on commandline as well, off course, as with any config value `dracut --xz --compress-level ...`.

In any case, if switching a compression method on a command line is really that common (while having a compression level defined in a config), let's introduce additonal 7 variables :).

bdrung:
Rebased and adapted the PR for the landed 3cpio support.

dracut.sh
Conan-Kudo: It's already the case in Fedora where some variants use zstd while the rest use xz. This will change in the future, but the reality is there today.

PR #1573, opened 2025-08-14
ci(debian): install 3cpio on Ubuntu

bdrung:
## Changes

In preparation for supporting 3cpio, install it on the Ubuntu containers.

## Checklist

Fixes #


bdrung:
> On ubuntu:rolling it does not seems to install it.. from the log
> 
> ```
> #6 [3/5] RUN if [ "debian" = "ubuntu:rolling" ] || [ "debian" = "ubuntu:devel" ] ; then     DEBIAN_FRONTEND=noninteractive apt-get install -y -qq --no-install-recommends -o Dpkg::Use-Pty=0     3cpio     ; fi
> #6 DONE 0.1s
> ```
> 
> Should this be "$DISTRIBUTION" = "ubuntu:rolling" instead of "debian" ?

Congratulation, you found bug https://github.com/dracut-ng/dracut-ng/issues/1581.

PR #1575, opened 2025-08-14
feat(dracut): make hostonly the default

jozzsi:
## Changes

Hostonly is already the default in the last dracut release, but the default was set as a configuration file.

This commit changes that and sets the default inside the dracut.sh script (unless of course a distribution or user overrides the default).

The hostonly configuration file is kept intentionally, but it is now not used by the configure option by default.

## Checklist


dalto8:
Why retain the config file?

PR #1580, opened 2025-08-15
feat(Makefile): filter out s390 specific dracut modules

jozzsi:
## Changes

This commit filters out s390 specific dracut modules, so that they only get packaged for s390 specific packages.

This commit would allow sharing the list of s390 specific dracut modules between distributions upstream.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ mandir ?= ${prefix}/share/man
 CFLAGS ?= -O2 -g -Wall -std=gnu99 -D_FILE_OFFSET_BITS=64 -Wformat -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2
 bashcompletiondir ?= ${datadir}/bash-completion/completions
 pkgconfigdatadir ?= $(datadir)/pkgconfig
+ARCH ?= $(shell uname -m)
 
 configs = \
 	fips/10-fips.conf \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -184,6 +184,8 @@ dracut.pc: Makefile.inc Makefile
 	@echo "dracutmodulesdir=$(pkglibdir)/modules.d" >> dracut.pc
 	@echo "dracutconfdir=$(pkglibdir)/dracut.conf.d" >> dracut.pc
 
+ARCH := $(shell uname -m)
```
bdrung: That should be overridable.

Makefile
```diff
@@ -283,6 +285,11 @@ endif
 			rm -rf $(DESTDIR)$(pkglibdir)/modules.d/[0-9][0-9]$${i}; \
 		done \
 	fi
+ifeq ($(ARCH),s390x)
+	for f in cio_ignore cms dasd dasd_mod dcssblk zfcp zipl znet; do \
+		rm -rf $(DESTDIR)$(pkglibdir)/modules.d/[0-9][0-9]$${i}; \
```
bdrung: I would drop `-f` since we expect those directories to be present. That way we would notice if we forgot to update the list here.

Makefile
jozzsi: ok, done

Makefile
jozzsi: ok, done here and the block above for the same reason - hope the same commit is ok here.

PR #1583, opened 2025-08-15
fix(dracut): decrease logging by default

jozzsi:
## Changes

The current default set in dracut is "4". All distributions (except Debian), sets the default to "3".

Instead of relying on good defaults only within distribution configurations, let's set the dracut default to be less verbose.

Most users new to dracut perceive the verbose dracut logs as errors. This change will help prevent and eliminate user confusion and improve the user experience.

## Checklist


bdrung (commented):
+1 with my Debian/Ubuntu hat on.

There are more places (like dracut man page) that talk about the default log level.

bdrung:
And btw, the dracut man page could say that `-v` and `-q` will increase/decrease the level once and you could specify that multiple times.

PR #1589, opened 2025-08-15
ci(integration-extra): disable UEFI test on debian:sid

jozzsi:
## Changes

This test has been failing on debian:sid for a while and passing on all other test containers.

I was unable to find a root cause and what is even more interesting is that the test is passing on debian:latest (trixie).

At this point of time debian:sid and debian:latest (trixie) should be very similar, so there is a decent chance that this is a distribution issue and not a dracut issue.

This is the last remaining failure to make the integration-extra GA green.

## Checklist

### DIFF
--- a/.github/workflows/integration-extra.yml
+++ b/.github/workflows/integration-extra.yml
@@ -59,6 +59,9 @@ jobs:
                     # https://github.com/dracut-ng/dracut-ng/issues/1224
                     - container: gentoo:amd64-openrc
                       test: "43"
+                    # https://github.com/dracut-ng/dracut-ng/issues/1407
+                    - container: debian:sid
+                      test: "12"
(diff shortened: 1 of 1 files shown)


bdrung:
But the UEFI tests are green on Debian sid in the package build. Search for TEST-12-UEFI on https://ci.debian.net/data/autopkgtest/unstable/amd64/d/dracut/63220958/log.gz

bdrung:
I have two branches where I test the latest upstream with the latest packaging (build + autopkgtest) on amd64:

* https://github.com/bdrung/dracut-ng/tree/debian/autopkgtest-successful
* https://github.com/bdrung/dracut-ng/tree/ubuntu/autopkgtest-successful

I only push those branches if they succeed. So that it becomes easier for packaging new upstream versions and finding regressions. All tests are green there.

PR #1593, opened 2025-08-18
fix(dracut): use "-name" to avoid find matching temporary directory

DeliZhangX:
$initdir includes a temporary directory, and the "-path" option of find attempts to match the entire path. If the temporary directory happens to contain ".ko", e.g.
  initdir=/var/tmp/dracut.ko79pT/initramfs
this can lead to a incorrect match, finally all executable files are excluded.

This pull request changes...

## Checklist

Fixes #


DeliZhangX:
Can anyone help to review?

bdrung (approved):

PR #1596, opened 2025-08-21
fix(lsinitrd, dracut-initramfs-restore): detect initrd for BLS Type #1 entries

aafeijoo-suse:
dracut is making wrong assumptions with BLS Type #1 entries.

- The ESP is not always used to store boot loader entries, there may be also an Extended Boot Loader Partition (XBOOTLDR). The spec defines the concept of the `$BOOT` partition placeholder, as the primary place to put boot menu entry resources into.

- The `machine-id` is not always used as subdirectory, it can be the `entry-token`.

- Also, although the name "initrd" is referenced in the examples, its use is not mandatory. E.g., in openSUSE we add a suffix with its hash value.

Manually implementing this autodetection logic can be complex, but all the necessary information is provided by `bootctl`, so use it if available.

More info: https://uapi-group.org/specifications/specs/boot_loader_specification

## Checklist

### DIFF
--- a/lsinitrd.sh
+++ b/lsinitrd.sh
@@ -109,6 +109,25 @@ find_initrd_for_kernel_version() {
     local kernel_version="$1"
     local base_path files initrd machine_id
 
+    if command -v bootctl > /dev/null && command -v jq > /dev/null; then
+        # get proper path to $BOOT
+        base_path=$(bootctl -x)
+        # get initrd key of the selected bootloader entry (i.e., the one that
+        # is actually used to boot the system)
+        mapfile -t files < <(bootctl --json=pretty list 2> /dev/null | jq -r '.[] | select(.isSelected).initrd[]' 2> /dev/null)
+        if [[ ${#files[@]} -ge 1 ]] && [[ -e "${base_path}${files[0]}" ]]; then
+            echo "${base_path}${files[0]}"
+            return
+        fi
+        # if the selected bootloader entry does not have any initrd keys, check
+        # the default (maybe the current selected entry was removed)
+        mapfile -t files < <(bootctl --json=pretty list 2> /dev/null | jq -r '.[] | select(.isDefault).initrd[]' 2> /dev/null)
+        if [[ ${#files[@]} -ge 1 ]] && [[ -e "${base_path}${files[0]}" ]]; then
+            echo "${base_path}${files[0]}"
+            return
+        fi
+    fi
+
     if [[ -d /efi/Default ]] || [[ -d /boot/Default ]] || [[ -d /boot/efi/Default ]]; then
         machine_id="Default"
     elif [[ -s /etc/machine-id ]]; then
--- a/dracut-initramfs-restore.sh
+++ b/dracut-initramfs-restore.sh
@@ -21,6 +21,28 @@ find_initrd_for_kernel_version() {
     local kernel_version="$1"
     local base_path files initrd machine_id
 
+    if command -v bootctl > /dev/null && command -v jq > /dev/null; then
+        # get proper path to $BOOT
+        base_path=$(bootctl -x)
+        # get initrd key of the selected bootloader entry (i.e., the one that
+        # is actually used to boot the system)
+        mapfile -t files < <(bootctl --json=pretty list 2> /dev/null | jq -r '.[] | select(.isSelected).initrd[]' 2> /dev/null)
+        if [[ ${#files[@]} -ge 1 ]] && [[ -e "${base_path}${files[0]}" ]]; then
+            echo "${base_path}${files[0]}"
+            return
+        fi
+        # if the selected bootloader entry does not have any initrd keys, check
+        # the default (maybe the current selected entry was removed)
+        # also check at least that the default bootloader entry has the same
+        # kernel version
+        mapfile -t files < <(bootctl --json=pretty list 2> /dev/null | jq -r '.[] | select(.isDefault).initrd[]' 2> /dev/null)
+        if [[ ${#files[@]} -ge 1 ]] && [[ -e "${base_path}${files[0]}" ]] \
+            && [[ ${files[0]} == *"${kernel_version}"* ]]; then
+            echo "${base_path}${files[0]}"
+            return
+        fi
+    fi
+
     if [[ -d /efi/Default ]] || [[ -d /boot/Default ]] || [[ -d /boot/efi/Default ]]; then
         machine_id="Default"
     elif [[ -s /etc/machine-id ]]; then


lsinitrd.sh
```diff
@@ -109,6 +109,18 @@ find_initrd_for_kernel_version() {
     local kernel_version="$1"
     local base_path files initrd machine_id
 
+    if command -v bootctl > /dev/null && command -v jq > /dev/null; then
+        # get proper path to $BOOT
+        base_path=$(bootctl -x)
+        # get initrd key of the selected bootloader entry (i.e., the one that
+        # is actually used to boot the system)
+        mapfile -t files < <(bootctl --json=pretty list 2> /dev/null | jq -r '.[] | select( .isSelected ) | .initrd[]' 2> /dev/null)
```
dalto8: I did some quick testing on this and noticed that in my test machine, the `loader` type has `isSelected` as true but this does not have an `initrd`.  The `initrd` key is set on the matching entry with type `type1`.  This causes the filter to return nothing.

dalto8: Also, currently, only the nvmf module requires `jq`.  Do we want to use it here as well?

aafeijoo-suse: Thanks for trying this, I could only test it in openSUSE Tumbleweed (grub2-bls / sd-boot) and Fedora (grub2-bls / sd-boot), and it just does not work in Fedora with its grub2-bls implementation (there is no entry with the key `isSelected`). But if it works with systemd-boot on multiple operating systems, I'd say we're at least improving the current situation where `lsinitrd` and loading the exitrd are broken.

Are you using any bootloader that follows the BLS? Could you print me the output of `bootctl --json=pretty list` on your system?

aafeijoo-suse: For me is the easiest way to parse JSON via bash scripting, and as you said, it's already being used in the `nvmf` module.

dalto8: I was using systemd-boot.  I see this on both Arch and EndeavourOS.

```
[
        {
                "type" : "type1",
                "source" : "esp",
                "id" : "fc6662aae8a245a0a773eb6825b6e9d6-6.12.43-1-lts.conf",
                "path" : "/efi/loader/entries/fc6662aae8a245a0a773eb6825b6e9d6-6.12.43-1-lts.conf",
                "root" : "/efi",
                "title" : "EndeavourOS",
                "showTitle" : "EndeavourOS (6.12.43-1-lts)",
                "sortKey" : "endeavouros-6.12.43-1-lts",
                "version" : "6.12.43-1-lts",
                "machineId" : "fc6662aae8a245a0a773eb6825b6e9d6",
                "options" : "nvme_load=YES nowatchdog rw rd.luks.uuid=edd8ce92-0c1d-4eee-953e-d063cf9c0be6 root=/dev/mapper/luks-edd8ce92-0c1d-4eee-953e-d063cf9c0be6 rd.luks.uuid=5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b resume=/dev/mapper/luks-5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b systemd.ma>
                "linux" : "/fc6662aae8a245a0a773eb6825b6e9d6/6.12.43-1-lts/linux",
                "initrd" : [
                        "/fc6662aae8a245a0a773eb6825b6e9d6/6.12.43-1-lts/initrd"
                ],
                "isReported" : false,
                "isDefault" : true,
                "isSelected" : false,
                "addons" : null,
                "cmdline" : "nvme_load=YES nowatchdog rw rd.luks.uuid=edd8ce92-0c1d-4eee-953e-d063cf9c0be6 root=/dev/mapper/luks-edd8ce92-0c1d-4eee-953e-d063cf9c0be6 rd.luks.uuid=5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b resume=/dev/mapper/luks-5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b systemd.ma>
        },
        {
                "type" : "type1",
                "source" : "esp",
                "id" : "fc6662aae8a245a0a773eb6825b6e9d6-6.12.43-1-lts-fallback.conf",
                "path" : "/efi/loader/entries/fc6662aae8a245a0a773eb6825b6e9d6-6.12.43-1-lts-fallback.conf",
                "root" : "/efi",
                "title" : "EndeavourOS",
                "showTitle" : "EndeavourOS (6.12.43-1-lts-fallback)",
                "sortKey" : "endeavouros-6.12.43-1-lts-fallback",
                "version" : "6.12.43-1-lts-fallback",
                "machineId" : "fc6662aae8a245a0a773eb6825b6e9d6",
                "options" : "nvme_load=YES nowatchdog rw rd.luks.uuid=edd8ce92-0c1d-4eee-953e-d063cf9c0be6 root=/dev/mapper/luks-edd8ce92-0c1d-4eee-953e-d063cf9c0be6 rd.luks.uuid=5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b resume=/dev/mapper/luks-5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b systemd.ma>
                "linux" : "/fc6662aae8a245a0a773eb6825b6e9d6/6.12.43-1-lts/linux",
                "initrd" : [
                        "/fc6662aae8a245a0a773eb6825b6e9d6/6.12.43-1-lts/initrd-fallback"
                ],
                "isReported" : false,
                "isDefault" : false,
                "isSelected" : false,
                "addons" : null,
                "cmdline" : "nvme_load=YES nowatchdog rw rd.luks.uuid=edd8ce92-0c1d-4eee-953e-d063cf9c0be6 root=/dev/mapper/luks-edd8ce92-0c1d-4eee-953e-d063cf9c0be6 rd.luks.uuid=5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b resume=/dev/mapper/luks-5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b systemd.ma>
        },
        {
                "type" : "type1",
                "source" : "esp",
                "id" : "fc6662aae8a245a0a773eb6825b6e9d6-6.16.1-arch1-1.conf",
                "path" : "/efi/loader/entries/fc6662aae8a245a0a773eb6825b6e9d6-6.16.1-arch1-1.conf",
                "root" : "/efi",
                "title" : "EndeavourOS",
                "showTitle" : "EndeavourOS (6.16.1-arch1-1)",
                "sortKey" : "endeavouros-6.16.1-arch1-1",
                "version" : "6.16.1-arch1-1",
                "machineId" : "fc6662aae8a245a0a773eb6825b6e9d6",
                "options" : "nvme_load=YES nowatchdog rw rd.luks.uuid=edd8ce92-0c1d-4eee-953e-d063cf9c0be6 root=/dev/mapper/luks-edd8ce92-0c1d-4eee-953e-d063cf9c0be6 rd.luks.uuid=5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b resume=/dev/mapper/luks-5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b systemd.ma>
                "linux" : "/fc6662aae8a245a0a773eb6825b6e9d6/6.16.1-arch1-1/linux",
                "initrd" : [
                        "/fc6662aae8a245a0a773eb6825b6e9d6/6.16.1-arch1-1/initrd"
                ],
                "isReported" : false,
                "isDefault" : false,
                "isSelected" : false,
                "addons" : null,
                "cmdline" : "nvme_load=YES nowatchdog rw rd.luks.uuid=edd8ce92-0c1d-4eee-953e-d063cf9c0be6 root=/dev/mapper/luks-edd8ce92-0c1d-4eee-953e-d063cf9c0be6 rd.luks.uuid=5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b resume=/dev/mapper/luks-5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b systemd.ma>
        },
        {
                "type" : "type1",
                "source" : "esp",
                "id" : "fc6662aae8a245a0a773eb6825b6e9d6-6.16.1-arch1-1-fallback.conf",
                "path" : "/efi/loader/entries/fc6662aae8a245a0a773eb6825b6e9d6-6.16.1-arch1-1-fallback.conf",
                "root" : "/efi",
                "title" : "EndeavourOS",
                "showTitle" : "EndeavourOS (6.16.1-arch1-1-fallback)",
                "sortKey" : "endeavouros-6.16.1-arch1-1-fallback",
                "version" : "6.16.1-arch1-1-fallback",
                "machineId" : "fc6662aae8a245a0a773eb6825b6e9d6",
                "options" : "nvme_load=YES nowatchdog rw rd.luks.uuid=edd8ce92-0c1d-4eee-953e-d063cf9c0be6 root=/dev/mapper/luks-edd8ce92-0c1d-4eee-953e-d063cf9c0be6 rd.luks.uuid=5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b resume=/dev/mapper/luks-5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b systemd.ma>
                "linux" : "/fc6662aae8a245a0a773eb6825b6e9d6/6.16.1-arch1-1/linux",
                "initrd" : [
                        "/fc6662aae8a245a0a773eb6825b6e9d6/6.16.1-arch1-1/initrd-fallback"
                ],
                "isReported" : false,
                "isDefault" : false,
                "isSelected" : false,
                "addons" : null,
                "cmdline" : "nvme_load=YES nowatchdog rw rd.luks.uuid=edd8ce92-0c1d-4eee-953e-d063cf9c0be6 root=/dev/mapper/luks-edd8ce92-0c1d-4eee-953e-d063cf9c0be6 rd.luks.uuid=5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b resume=/dev/mapper/luks-5a20ac1e-eecd-4e8b-b7a7-e7092288bb3b systemd.ma>
        },
        {
                "type" : "loader",
                "source" : "esp",
                "id" : "fc6662aae8a245a0a773eb6825b6e9d6-6.16.0-arch2-1.conf",
                "path" : "/sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f",
                "showTitle" : "fc6662aae8a245a0a773eb6825b6e9d6-6.16.0-arch2-1.conf",
                "isReported" : true,
                "isDefault" : false,
                "isSelected" : true,
                "addons" : null
        },
        {
                "type" : "loader",
                "source" : "esp",
                "id" : "fc6662aae8a245a0a773eb6825b6e9d6-6.16.0-arch2-1-fallback.conf",
                "path" : "/sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f",
                "showTitle" : "fc6662aae8a245a0a773eb6825b6e9d6-6.16.0-arch2-1-fallback.conf",
                "isReported" : true,
                "isDefault" : false,
                "isSelected" : false,
                "addons" : null
        },
        {
                "type" : "auto",
                "source" : "esp",
                "id" : "auto-reboot-to-firmware-setup",
                "path" : "/sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f",
                "title" : "Reboot Into Firmware Interface",
                "showTitle" : "Reboot Into Firmware Interface",
                "isReported" : true,
                "isDefault" : false,
                "isSelected" : false,
                "addons" : null
        }
]
```

dalto8: At least on Arch and EndeavourOS, `jq` is not a hard dependency.  I guess this would change that.

aafeijoo-suse: Ok, now I understand. You're not booting a BLS Type #1 entry (defined `<ESP>/loader/entries/*.conf`, which is what this patch tries to detect), you're booting an entry from the LoaderEntries EFI variable. Is the initrd self-contained? is it like an UKI?

dalto8: No, it is not a UKI.


> You're not booting a BLS Type #1 entry (defined `<ESP>/loader/entries/*.conf`

It is actually.  The actual entry it is booting is `/efi/loader/entries/fc6662aae8a245a0a773eb6825b6e9d6-6.16.1-arch1-1.conf`

You can see this if you look at the entry with the matching `id`.  This is what systemd-boot/kernel-install does when it generates the entries before you boot them.  After you reboot, it will change the `isSelected` to the entry you are expecting.  But dracut can be run after the entries are generated but before the system is rebooted.  I don't think that would be uncommon.

aafeijoo-suse: Thanks for your feedback. So far I failed to create `"type" : "loader"` entries with `"id"` pointing to a .conf file... so I'll dig into the code. @aplanas do you know how to create on of these by any chance?

aafeijoo-suse: > It is actually. The actual entry it is booting is `/efi/loader/entries/fc6662aae8a245a0a773eb6825b6e9d6-6.16.1-arch1-1.conf`
> 
> You can see this if you look at the entry with the matching `id`.

hmm...

```
{
        "type" : "loader",
        ...
        "id" : "fc6662aae8a245a0a773eb6825b6e9d6-6.16.0-arch2-1.conf",
        ...
},
```

 `fc6662aae8a245a0a773eb6825b6e9d6-6.16.0-arch2-1.conf` is not `fc6662aae8a245a0a773eb6825b6e9d6-6.16.1-arch1-1.conf`, there is not any other matching id on the list.

aplanas: This is a bit weird:

```json
        {
                "type" : "loader",
                "source" : "esp",
                "id" : "fc6662aae8a245a0a773eb6825b6e9d6-6.16.0-arch2-1.conf",
                "path" : "/sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f",
                "showTitle" : "fc6662aae8a245a0a773eb6825b6e9d6-6.16.0-arch2-1.conf",
                "isReported" : true,
                "isDefault" : false,
                "isSelected" : true,
                "addons" : null
        }
```

The path is an EFI variable: `LoaderEntries` that is defined by BLI (https://systemd.io/BOOT_LOADER_INTERFACE/)

* The EFI variable `LoaderEntries may contain a series of boot loader entry identifiers, one after the other, each individually NUL terminated. This may be used to let the OS know which boot menu entries were discovered by the boot loader. A boot loader entry identifier should be a short, non-empty alphanumeric string (possibly containing -, too). The list should be in the order the entries are shown on screen during boot. See below regarding a recommended vocabulary for boot loader entry identifiers.

I am not sure why this entry was created by the boot loader, but IIUC this EFI variable will contain several `id`s, as there are multiple `"type": "loader"` entries with the same `path` to the same EFI variable.

For this case I guess we can use the ID and the EFI variable content to deduce the real ID and look back to it in the json.  But how `"fc6662aae8a245a0a773eb6825b6e9d6-6.16.0-arch2-1.conf"` is mapped to `fc6662aae8a245a0a773eb6825b6e9d6-6.16.1-arch1-1.conf` is not clear for me.

@dalto8 any chance you can dump `/sys/firmware/efi/efivars/LoaderEntries-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f`

aafeijoo-suse: > When boot loader entries are defined through the Boot Loader Specification files, the identifier should be derived directly from the file name, **but with the .conf (Type #1 snippets) or .efi (Type #2 images) suffix removed**.

aplanas: So I guess @dalto8 have a `/efi/loader/entries/fc6662aae8a245a0a773eb6825b6e9d6-6.16.1-arch2-1.conf`. Can you confirm that?

In any case the `id` is not always directly the file name of the file. The boot count (https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/) will add a '+n.conf' suffix into the file, without changing the ID, that is why makes sense to use the path

Vogtinator: IMO if a BLS entry file was used to boot and it still exists when running `bootctl list`, this command needs to work. Sounds like a bug in `bootctl`, maybe it doesn't do the matching correctly.

Vogtinator:
I was actually looking into implementing this for `lsinitrd` myself, but hit similar issues as you. I'm also not sure what `lsinitrd` is meant to do by default: Print info about the initrd that was used for booting or the initrd that will be used for booting next? With the non BLS code, both could happen, but with BLS there's an explicit choice (`select(.isSelected)` vs. `select(.isDefault)`.

lsinitrd.sh
dalto8: > So far I failed to create `"type" : "loader"` entries with `"id"` pointing to a .conf file.

I did some more testing to understand what is happening.

It is caused by calling `kernel-install remove`.  When you remove the booted .conf, it creates those loader entries.

Why would this happen?  Because when the kernel version is changed the old kernel is removed and the new one is installed.

Of note is that the IDs won't always match.  The reason they matched in my example is I removed and replaced the same kernel.

Basically, I don't think you can use `isSelected` to entry to find the initrd.

Here are the reproduction steps:
* Install the latest EndeavourOS ISO choosing systemd-boot as the bootloader.
* Run `sudo downgrade linux` and choose a different kernel version.

Vogtinator: Ok, so you hit precisely what I described in https://github.com/dracut-ng/dracut-ng/pull/1596#issuecomment-3214265309: used initrd vs. next initrd. In your case, the used initrd no longer exists, so with the current choice in the PR it will fail by design.

aafeijoo-suse: > It is caused by calling kernel-install remove. When you remove the booted .conf, it creates those loader entries.

Ahh ok.

> Basically, I don't think you can use isSelected to entry to find the initrd.

Well, I usually want to check the initrd of the "selected" entry with lsinitrd (for the exitrd, there is no doubt that we have to unpack the same initrd used to boot the system). And, we could add a fallback to check the "default" entry in case there is no "selected". Opinions?

Vogtinator: > for the exitrd, there is no doubt that we have to unpack the same initrd used to boot the system

If the initrd is regenerated with similar contents, it will work. It's also how the code has worked previously FWICT. IMO that's better than just failing completely.

Vogtinator: > And, we could add a fallback to check the "default" entry in case there is no "selected". Opinions?

I'm in favor, but it has to make sure it actually fits the running kernel. Maybe it needs to iterate entries that use the booted kernel and pick the best one...

aplanas: Another option is to drop bootctl and jq:

1. Search the EFI variable `LoaderEntrySelected`
2. If not found fallback to the current logic
3. If found read the ID from it and `grep initrd /boot/efi/loader/entries/$ID`
4. If the conf file is not present, then boot assessment is active and do `grep initrd /boot/.../entries/${ID%.conf}\+*.conf`

It will cover a lot of new cases

aafeijoo-suse: > > for the exitrd, there is no doubt that we have to unpack the same initrd used to boot the system
> 
> If the initrd is regenerated with similar contents, it will work. It's also how the code has worked previously FWICT. IMO that's better than just failing completely.

Yes, that's true, it's also how the code has worked previously.

> > And, we could add a fallback to check the "default" entry in case there is no "selected". Opinions?
> 
> I'm in favor, but it has to make sure it actually fits the running kernel. Maybe it needs to iterate entries that use the booted kernel and pick the best one...

Why should it fit the running kernel for `lsinitrd`? We simply print the contents of the initrd, often for debugging when we detect a problem with the current boot. If we've already reached the point where the system has been updated and the current initrd has been removed, the contents of the next initrd might be more interesting, regardless of whether the kernel version is different.

For the exitrd is different...

aafeijoo-suse: > Another option is to drop bootctl and jq:
> 
>     1. Search the EFI variable `LoaderEntrySelected`

Ok, but it's not always populated... e.g. Fedora:

```
root@fedora:/home/dev# ls -l /sys/firmware/efi/efivars/LoaderEntry*
-rw-r--r--. 1 root root 174 Aug 25 08:42 /sys/firmware/efi/efivars/LoaderEntryDefault-4a67b082-0a4c-41cf-b6c7-440b29bb8c4f
```

aplanas: In this case you fallback to the old logic (step 2)

aafeijoo-suse: The spec that defines the EFI variables (https://systemd.io/BOOT_LOADER_INTERFACE/) is not under any UAPI spec, it's systemd-specific. So I don't know if adding any additional logic other than just using `bootctl` will provide more value here.

aplanas: In grub2[-bls] there is a module that must be loaded (bli.mod) that includes those variables (to some extend at least)

> So I don't know if adding any additional logic other than just using bootctl will provide more value here.

`bootctl` is using those variables to mark those entries. The proposal was about replacing bootctl and jq (as they are new dependencies), to do something that is equivalent to what bootctl is doing. But if they are OK to be used, then sure the solution is simpler.

aafeijoo-suse:
PR stalled for 6 months. I'll keep a downstream patch, closing.

Issue #1509, opened 2025-08-03
fold dracut-init.sh into dracut.sh


jozzsi:
After TEST 60, 61, 62 have been converted to use the the dracut test modules (same as other tests), dracut-init.sh would be only sourced from dracut.sh, so as such could be folded into one file and open the door for additional optimizations.

Note that `dracut-functions.sh` make sense to stay independent and that is critical file for backward compatibility (dracut-init.sh is not).

fpokryvk:
@jozzsi thanks for the heads-up. So, the correct way to stay backward compatible is to do the following?
```
if [ -f $basedir/dracut-init.sh ]; then
  . $basedir/dracut-init.sh
else
  . $basedir/dracut.sh
fi
```

PR #1619, opened 2025-09-01
refactor: fold dracut-init.sh into dracut.sh

jozzsi:
dracut-init.sh is only sourced from dracut.sh, so it could be folded into one file  and open the door for additional optimizations.

## Checklist

Fixes: #1509


Conan-Kudo:
I'm not a _huge_ fan of making the shell script files bigger. I'm a little less concerned about "performance" and I would prefer to see us modularize and make individual files smaller instead.

bdrung (changes requested):
The `--local` parameter still checks for `dracut-init.sh`.

bdrung (dismissed):

bdrung:
I would call this commit a "refactor" instead of "perf".

bdrung:
Version 109 has been released. Can you rebase this PR?

bdrung (approved):
I'm not a _huge_ fan of making shell script files bigger, but in this case the merged dracut.sh is better than sourcing one extra shell script. To reduce the size, we could move the microcode handling code to a module.

LaszloGombos (approved):
I hope this PR opens the door for moving more code to dracut modules and improve shellcheck hygiene

PR #1620, opened 2025-09-04
fix(udev-rules): exclude udev rules from snapd

bdrung:
## Changes

There are no snaps in the initrd and these udev rules generated by snapd can cause warning messages. So just exclude udev rules in `/etc` that have `70-snap` in the name.

## Checklist


Conan-Kudo:
Ugh, I'm not sure I want this in here... We have no way of knowing that `70-snap` is because of snapd.

Conan-Kudo:
Yeah, I'm firmly against this change. Please come up with a better way to deal with this, preferably in snapd.

Conan-Kudo:
Yeah, I'm closing this. This change is not appropriate here.

PR #1622, opened 2025-09-05
fix(74nvmf): fix livenet + nvmf

mwilck:
This PR goes on top of #1370.

Booting hangs if the `root=live:` parameter is given on a system that has a valid NBFT table when the nvmf module is included in the initrd.

This pull request fixes that by not overwriting the `netroot=` variable if it is already set. This change can cause the NVMF connections not to be activated. Therefore I am adding another change on top that makes sure that the nvmf-autoconnect script is run at least once if a valid NBFT is detected.

Cc: @aafeijoo-suse, @tbzatek, @johnmeneghini


modules.d/74nvmf/parse-nvmf-boot-connections.sh
```diff
@@ -326,6 +326,8 @@ fi
 
 /sbin/initqueue --settled --onetime --name nvmf-connect-settled /sbin/nvmf-autoconnect.sh settled
 /sbin/initqueue --timeout --onetime --name nvmf-connect-timeout /sbin/nvmf-autoconnect.sh timeout
+# Make sure the autoconnect script is run at least once
+echo '[ -f /tmp/nvmf.done ]' > "$hookdir/initqueue/finished/95-nvmf.sh"
```
tbzatek: 95 -> 74

mwilck: I don't mind, but is that mandatory? The module also still contains `95-nvmf-initqueue.rules`...

mwilck: ... and uses `/etc/cmdline.d/95nvmf-args.conf`

mwilck: anyway, fixed here. We can fix the other 95's later if it's required.

tbzatek: ahh, right, that should've been done at the moment the module name was renamed.

PR #1626, opened 2025-09-06
fix(i18n): remove Gentoo workaround that is no longer needed

jozzsi:
## Changes

Gentoo no longer sets KEYMAP to something like "-u pl2", instead it is set to "pl12".

This piece of code is part of the very original commit of the i18n dracut module from 2010 - 87122af                                                                                                                                                                                             
                                                                                                                                                                                                                              
There is no mention of '-u' in the oldest version of the wiki from 2013 https://wiki.gentoo.org/index.php?title=Keyboard_layout_switching&diff=1320347&oldid=16937

Remove this workaround that is no longer used or tested to make it easier to reason about the code.

## Checklist

### DIFF
--- a/modules.d/20i18n/module-setup.sh
+++ b/modules.d/20i18n/module-setup.sh
@@ -243,9 +243,6 @@ install() {
         fi
         shopt -q -u nocasematch
 
-        # Gentoo user may have KEYMAP set to something like "-u pl2",
-        KEYMAP=${KEYMAP#-* }
-
         # openSUSE user may have KEYMAP set to something like ".gz"
         KEYMAP=${KEYMAP/.gz/}
 


bdrung:
Thanks for looking into it. Your finding satisfy me.

> This piece of code is part of the very original commit of the i18n dracut module form 2010 - [87122af](https://github.com/dracut-ng/dracut-ng/commit/87122afcdaad0c3648c14cfab68d1a40e090a957)
> 
> There is no mention of '-u' in the oldest version of the wiki from 2013 - https://wiki.gentoo.org/index.php?title=Keyboard_layout_switching&diff=1320347&oldid=16937

Having this in the commit message would be useful.

PR #1639, opened 2025-09-11
feat: dracut.sh: try $STRIP for $strip_cmd first

koenkooi:
## Changes
When using dracut in a cross enviroment, like OpenEmbedded, the host provided strip (or eu-strip) won't work, so try using the $STRIP variable from the shell environment first before falling back to path based lookups.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2337,7 +2337,8 @@ done
 if [[ $do_strip == yes ]]; then
     # Prefer strip from elfutils for package size
     declare strip_cmd
-    strip_cmd=$(command -v eu-strip)
+    strip_cmd="${STRIP-}"
+    [ -z "$strip_cmd" ] && strip_cmd=$(command -v eu-strip)
     [ -z "$strip_cmd" ] && strip_cmd="strip"
 
     for p in "$strip_cmd" xargs find; do


dracut.sh
```diff
@@ -2277,7 +2277,8 @@
 if [[ $do_strip == yes ]]; then
     # Prefer strip from elfutils for package size
     declare strip_cmd
-    strip_cmd=$(command -v eu-strip)
+    strip_cmd="${STRIP}"
+    [ -z "$strip_cmd" ] && $(command -v eu-strip)
```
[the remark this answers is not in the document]
koenkooi: Fixed

koenkooi:
> This would mean that it would be influenced by package build environments that define this variable, right? Is that a consequence we want to have?

That is precisely my usecase, using the environment provided STRIP, which will point to the right cross version of it.

> I would prefer this to be a variable that isn't used in build environments that people deliberately set.

What use case are we trying to protect against with that? I don't have a _strong_ opinion on whether to use `$STRIP` or name it differently, but I'd like to avoid creating extra effort for people with properly configured environments to shield people with improperly configured environments.

Either way, I have no issue using a differently named variable that is desired.

bdrung:
And please use a similar construct to what https://github.com/dracut-ng/dracut-ng/commit/8d870e632bc5f82ae40c52e515ba3c4c6cce037d did.

dracut.sh
```diff
@@ -2337,7 +2337,8 @@ done
 if [[ $do_strip == yes ]]; then
     # Prefer strip from elfutils for package size
     declare strip_cmd
-    strip_cmd=$(command -v eu-strip)
+    strip_cmd="${STRIP}"
```
```suggestion
    strip_cmd="${STRIP-}"
```

koenkooi: I've included this in the most recent rebase

bdrung (approved):
Thanks.

Conan-Kudo (approved):

PR #1640, opened 2025-09-12
feat(fedora): switch to strict hostonly mode by default

jozzsi:
For Fedora generated initramfs size is a priority over initramfs portability.

Change the Fedora configuration to reflect this Fedora policy decision.

## Checklist

Related bug - https://github.com/dracut-ng/dracut-ng/issues/748 

CC @AdamWill @pvalena @Conan-Kudo @dtardon


litian1992:
FYI, we have this patch downstream lately
https://gitlab.com/redhat/rhel/src/kernel/rhel-10/-/commit/fb08452dae1ac62cfd755da4cb2eb3c9c960fe1f
The hostonly mode brings in password at emergency shell which is undesired for us.
I haven't got the ability to test kdump with a LUKS disk so far on Azure due to missing features and so on. Maybe the others did.

cgwalters:
cc https://src.fedoraproject.org/rpms/dracut/pull-request/90#comment-305793

Two problems here:
- I don't think having distribution-specific configs upstream makes a lot of sense.
- For image-based systems we want to build generic initramfs and we have Fedora image based systems so we can't just say "if fedora then hostonly"

PR #1651, opened 2025-09-14
docs: document keep configuration option

jozzsi:
The `keep` configuration option keeps (does not delete) the temporary initramfs directory for debugging purposes.

## Checklist

Fixes #


LaszloGombos (dismissed):
LGTM. 

Prefer an additional review to confirm if we want to advertise this config option via documentation.

Conan-Kudo (changes requested):
I don't think we want to document this. This flag breaks expected behavior for dependent components. I'm not entirely happy that it's a non-descriptive option either.

jozzsi:

PR #1673, opened 2025-09-18
fix: partial revert for hostonly sloppy mode

jozzsi:
This commit is a _partial_ revert of 8519dcd.

The goal of this commit is to have the same dracut modules included in both `sloppy` and in `strict` mode, but the dracut modules themselves might install more kernel modules depending on the `hostonly_mode` value.

Changes in other areas of the code (e.g. changes inside `installkernel`) were not reverted as they require more discussions and agreement by the community.

As an example the `crypt` dracut module could potentially copy more kernel modules into the generated initramfs in `sloppy` `hostonly_mode` mode than in `strict` `hostonly` mode.

## Checklist

CC @pvalena @AdamWill @lnykryn @Conan-Kudo

Related issue https://github.com/dracut-ng/dracut-ng/issues/748

### DIFF
--- a/modules.d/74iscsi/module-setup.sh
+++ b/modules.d/74iscsi/module-setup.sh
@@ -6,9 +6,9 @@ check() {
     require_binaries iscsi-iname iscsiadm iscsid || return 1
     require_kernel_modules iscsi_tcp || return 1
 
-    # If hostonly strict was requested, fail the check if we are not actually
+    # If hostonly was requested, fail the check if we are not actually
     # booting from root.
-    [[ $hostonly_mode == "strict" ]] || [[ $mount_needs ]] && {
+    [[ $hostonly ]] || [[ $mount_needs ]] && {
         pushd . > /dev/null
         for_each_host_dev_and_slaves block_is_iscsi
         local _is_iscsi=$?
--- a/modules.d/11systemd-cryptsetup/module-setup.sh
+++ b/modules.d/11systemd-cryptsetup/module-setup.sh
@@ -6,7 +6,7 @@ check() {
     # if cryptsetup is not installed, then we cannot support encrypted devices.
     require_binaries "$systemdutildir"/systemd-cryptsetup || return 1
 
-    [[ $hostonly_mode == "strict" ]] || [[ $mount_needs ]] && {
+    [[ $hostonly ]] || [[ $mount_needs ]] && {
         for fs in "${host_fs_types[@]}"; do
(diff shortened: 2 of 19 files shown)


ArrayBolt3:
From Kicksecure's standpoint, this is mostly fine, but I wouldn't revert `70qemu-net` or `70qemu`. That could break an installation that is typically booted on a physical machine but is later booted in a VM. Otherwise I don't see any of the rest of the changes causing us problems.

ArrayBolt3:
Hmm, I originally was going to elect to recommend leaving it out, but looking a bit closer it looks like it may be impossible for some hardware to work: https://www.freedesktop.org/software/systemd/man/latest/hwdb.html How important it is depends on what exactly is in the hardware database; on my Ubuntu system it's mostly assigning names to devices (which I guess could be useful for dracut modules sometimes, but then those modules should depend on hwdb), there's like *one* Dell keyboard that has *one* button on it fixed by one of the hwdb files but that's all I've been able to find so far.

Honestly I think it's fine to leave it out. If one laptop series' mic mute button doesn't work in the initramfs, too bad. The system will still boot, at which point a full hwdb will be available, and that's all we really care about.

modules.d/74iscsi/module-setup.sh
```diff
@@ -8,7 +8,7 @@ check() {
 
     # If hostonly strict was requested, fail the check if we are not actually
```
bdrung:
```suggestion
    # If hostonly was requested, fail the check if we are not actually
```

bdrung (changes requested):
Can you document in the git commit message which changes where not revert and why we keep them?

bdrung:
I tested this change in a Ubuntu chroot with some stuff like plymouth installed (to get comparable sizes to bare metal installations):

* plucky with 106: 24 MB
* questing with 108: 35 MB
* questing with main+PR: 29 MB

modules.d/74iscsi/module-setup.sh
jozzsi: Thanks !

bdrung:
> > * modules.d/70kernel-modules/module-setup.sh: if not on strict hostonly mode, install all known filesystems -> what is the size impact here?
> 
> Here is some data form an earlier PR - [#1348 (comment)](https://github.com/dracut-ng/dracut-ng/pull/1348#issuecomment-3064188515)

Assuming the 317 kB from that earlier PR, this is small enough to keep.

Issue #1519, opened 2025-08-05
Improve release cadence


aafeijoo-suse:
The current method seems to create a new release every 3 months, without taking into account other considerations of the Linux ecosystem.

Some proposals:
- Do not have releases with a fixed frequency (at least 3 months seems very high), without being sure that all the important bugs have been fixed.
- Introduce release candidates to be able to detect errors before creating the final release. It does not need to be a fixed number, just until everything looks good.
- Introduce point releases for at least the latest version (with a dedicated branch), to simplify the downstream maintenance.
- Be in sync with systemd releases. systemd is very important when used as PID1 in the initrd, so there should not be a new dracut release with fixes pending for an upcoming version of systemd that is already planned to be released soon.

aafeijoo-suse:
> Syncing our releases to systemd is an interesting idea. Do they have their release schedule documented somewhere?

Not as far as I know, but I think they try to have 2 releases per year, in the middle and at the end of the year. But this is not fixed, since the release is delayed if there are issues. For example now systemd-v258 is in RC2 (normally there are 3 RCs and then the final release, but if something needs fixing there may be more RCs before).

But as far as dracut is concerned, most of the required changes that need to be included should have been detected already after the first release candidate of systemd, when some rolling distros like Arch start using it, so there is no need to wait more than 1 month until the final release.

bdrung:
We could increase the release schedule to do one release every two month.

bdrung:
The amount of changes will be the same, but cut into smaller chunks, won't it? If downstreams want to invest less time they can skip upstream releases IMO.

devkontrol:
> I meant 3 months seems very high. I'd say at least every 6 months.

I expect the number of security concerns to increase in the next few month and I think release every 3 months is justified for the project.

I agree with @bdrung  that if downstreams want to invest less time they can skip upstream releases IMO.

Let's close this issue.

bdrung:
There is one point mentioned that we haven't discussed: point releases for at least the latest version

PR #1676, opened 2025-09-18
ci: increase release cadence from 3 to 2 month

bdrung:
## Changes

The current release schedule of Dracut is time based every three months. The git main branch should be in a releasable state all the time. There are currently no point releases for Dracut. Distributions/users need to cherry-pick relevant fixes for their own or wait for the next release.

The release cadence for relevant other projects: The Linux kernel releases every 9-10 weeks normally. Systemd tries to have two releases per year.

Our development process allows to release early, release often. Increase the release cadence from three to two months. This way we make (rolling) distributions happy that are quick in introducing new upstream releases. Distributions that do not want to package Dracut updates every two months can simply skip some Dracut releases.

We could decide in the future to start releasing point releases with cherry-picked fixes.

@LaszloGombos and @Conan-Kudo what do you think?

## Checklist

Fixes: https://github.com/dracut-ng/dracut-ng/issues/1519

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -3,8 +3,8 @@ name: Release
 
 on:  # yamllint disable-line rule:truthy
   schedule:
-    ## Schedule the job quarterly to run on Feb-1, May-1, Aug-1, and Nov-1
-    - cron: '0 0 1 FEB,MAY,AUG,NOV *'
+    ## Schedule the job every two months to run on the first of each even month
+    - cron: '0 0 1 FEB,APR,JUN,AUG,OCT,DEC *'
   workflow_dispatch:
     inputs:
       tag:


Conan-Kudo:
We can certainly try it, but my experience with KDE Plasma is that if our release schedule isn't evenly divisible in such a way that it aligns with distros to ship and integrate, it lessens the real world testing of those releases that fall under those poor windows (like the theoretical upcoming release in October).

bdrung:
That applies to distributions with fixed release dates (like Ubuntu and Fedora), but there are rolling releases that will probably publish those dracut releases. Those releases would land in Debian unstable (unless it's in its freeze before their 2 years apart release).

Conan-Kudo:
Shortening by one month doesn't buy us a lot, as the distributions you list all have `dracut-git` options that let them try git mainline trivially. And I don't consider Debian sid a rolling distribution just like I don't consider Fedora Rawhide that either. Neither are designed to be consumed by people as a first class offering.

PR #1678, opened 2025-09-18
perf: improve logger efficiency

HastD:
## Changes

The logger would previously feed input from stdin one line at a time to the logging target. This is very slow and can have a major impact on performance when dealing with debug output that can be thousands of lines long, such as the output of `lsinitrd`.

Instead, add a `_do_dlog_batch` function that accepts input on stdin rather than as an argument, splits stdin into multiple streams, and uses `awk` to process multiline inputs efficiently for each target. When `dlog` is called without arguments, it passes stdin on to this function. This is orders of magnitude faster for large inputs.

Also make the internal logging functions return 1 if supplied an invalid log level.

I tested this locally in a Fedora Kinoite VM, and it reduced total dracut runtime by a little over 20% in the default configuration (which includes `sysloglvl=5`).

## Checklist
- [x] I have reviewed and updated any documentation if relevant (N/A)

### DIFF
--- a/dracut-logger.sh
+++ b/dracut-logger.sh
@@ -354,6 +354,95 @@ _do_dlog() {
     return 0
 }
 
+## @brief Like _do_dlog() but accepts messages on standard input and sends them
+# to the appropriate logging outputs in a batch.
+#
+# @param lvl Numeric logging level.
+# @retval 0 if level is in range, even if logging fails.
+# @retval 1 if level is out of range.
+_do_dlog_batch() {
+    # Return early if level out of range
+    _lvl2char "$1" > /dev/null || return 1
+
+    # shellcheck disable=SC1003
+    sed -e '$a\' - | tee \
+        >(_do_dlog_batch_stdlog "$1") \
+        >(_do_dlog_batch_syslog "$1") \
+        >(_do_dlog_batch_filelog "$1") \
+        >(_do_dlog_batch_kmsglog "$1") \
+        > /dev/null
+}
+
+## @brief Batch processor for logging to standard error.
+#
+# @param lvl Numeric logging level.
+# @retval 0 if level is in range, even if logging fails.
+# @retval 1 if level is out of range.
+_do_dlog_batch_stdlog() {
+    local lvl="$1"
+    local lvlc
+    if ((lvl <= stdloglvl)); then
+        lvlc=$(_lvl2char "$lvl") || return 1
+        awk -v prefix="dracut[${lvlc}]: " '{ print prefix $0 }' >&2
+    else
+        cat > /dev/null
+    fi
+}
+
+## @brief Batch processor for logging to system journal.
+#
+# @param lvl Numeric logging level.
+# @retval 0 if level is in range, even if logging fails.
+# @retval 1 if level is out of range.
+_do_dlog_batch_syslog() {
+    local lvl="$1"
+    local syslog_lvlc
+    if ((lvl <= sysloglvl)); then
+        if [[ "${_dlogfd}" ]]; then
+            syslog_lvlc="$(($(_dlvl2syslvl "$lvl") & 7))" || return 1
+            awk -v prefix="<${syslog_lvlc}>" '{ print prefix $0 }' >&"${_dlogfd}"
+        else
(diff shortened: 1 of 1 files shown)


HastD:
@bdrung Thanks for doing the more precise benchmarking—that looks consistent with my local testing. I think most of the additional time with `sysloglvl=5` is coming from logging the output of `lsinitrd` after the initramfs has been built. By the way, if you want to see an isolated demonstration of the speed difference, try running the following in a bash shell:
```sh
. dracut-logger.sh
dlog_init
time seq 1000 | dwarn
```
On my laptop, this takes about 15 seconds without my changes, and about 0.2 seconds with my changes. Additionally, most of that 0.2 seconds is constant overhead: even with 1 million lines it still takes under 1 second.

Also, I'm realizing now that what I wrote was ambiguous: by "debug output", I meant ordinary logging at the "debug" level, not the additional output produced by the `--debug` option (which doesn't go through the logging functions and is unaffected by my changes).

dracut-logger.sh
```diff
@@ -354,6 +354,95 @@ _do_dlog() {
...
+#
+# @param lvl Numeric logging level.
+# @retval 0 if level is in range, even if logging fails.
+# @retval 1 if level is out of range.
+_do_dlog_batch() {
+    # Return early if level out of range
+    _lvl2char "$1" > /dev/null || return 1
+
+    # shellcheck disable=SC1003
+    sed -e '$a\' - | tee \
```
bdrung, line 368: What is this `sed` command doing?

HastD: This adds a newline to the end of the input (from stdin) if and only if it doesn't already end with a newline. (The `$` is the address of the last line of the input, and the `a\` command appends what follows—in this case, the empty string—to a new line after the given line.)

bdrung: Okay. Found an explanation: The sed command adds a newline at the end in case there is none.

dracut-logger.sh
```diff
@@ -354,6 +354,95 @@ _do_dlog() {
...
+# @retval 0 if level is in range, even if logging fails.
+# @retval 1 if level is out of range.
+_do_dlog_batch_stdlog() {
+    local lvl="$1"
+    local lvlc
+    if ((lvl <= stdloglvl)); then
+        lvlc=$(_lvl2char "$lvl") || return 1
+        awk -v prefix="dracut[${lvlc}]: " '{ print prefix $0 }' >&2
+    else
+        cat > /dev/null
```
bdrung, line 388: What does the `cat` command do here? Isn't stdin thrown away in that case without calling `cat`?

dracut-logger.sh
```diff
@@ -344,36 +347,129 @@ _do_dlog() {
...
+    # Return early if level out of range
+    _lvl2char "$1" > /dev/null || return 1
+
+    # shellcheck disable=SC1003
+    sed -e '$a\' - | tee \
+        >(_do_dlog_batch_stdlog "$1") \
+        >(_do_dlog_batch_syslog "$1") \
+        >(_do_dlog_batch_filelog "$1") \
+        >(_do_dlog_batch_kmsglog "$1") \
+        > /dev/null
```
bdrung, line 373: The one redirect to /dev/null could be avoided by using this fd:

```suggestion
    sed -e '$a\' - | tee \
        >(_do_dlog_batch_stdlog "$1") \
        >(_do_dlog_batch_syslog "$1") \
        >(_do_dlog_batch_filelog "$1") \
        | _do_dlog_batch_kmsglog "$1"
```

dracut-logger.sh
HastD: Yes, but if we don't read from stdin at all then the pipe hangs up early and causes problems with the `tee` command's splitting of stdin. This way, all four logging targets (stdlog, syslog, filelog, and kmsglog) read from stdin unconditionally, either logging the input to the appropriate target or discarding the input, depending on the logging configuration.

dracut-logger.sh
HastD: I thought about doing that but I felt it made the code clearer to handle input to all four logging targets in the same way. Do you want me to change it anyway?

bdrung: Doing that for clarity is a valid reason.

Issue #1694, opened 2025-09-22
dracut-cpio: unit tests fail with latest version of GNU cpio


ddiss:
**Describe the bug**
dracut-cpio unit tests compare binary archive output with that of GNU cpio, for the same set of input files. A recent change to upstream GNU cpio, commit 6a94d5e ("New option --ignore-dirnlink"), causes some tests to fail.
The failure is due to GNU `cpio --reproducible` now hardcoding directory `nlink` values to 2, instead of using the `st_nlink` value reported by `stat()`.

**Distribution used**
openSUSE Tumbleweed with GNU cpio-2.15, but it should affect any version shipping recent upstream GNU cpio versions.

**Dracut version**
Any version carrying a9c67046 ("feat(cpio): add newc archive creation utility") and GNU cpio commit 6a94d5e should fail `cargo test`. Tested with a2ae12c9689885797edb307799aad423065a92cb .

**Init system**
N/A

**To Reproduce**
Any unit test which archives directories and compares with GNU cpio output fails, e.g.
`cargo test tests::test_archive_dir -- --test-threads=1 --exact`

PR #1695, opened 2025-09-22
fix(cpio): fix unit tests alongside recent GNU cpio

ddiss:
dracut-cpio unit tests compare binary archive output with that of GNU cpio, for the same set of input files. A recent change to upstream GNU cpio, commit 6a94d5e ("New option --ignore-dirnlink"), causes some tests to fail.
The failure is due to GNU cpio `--reproducible` now hardcoding directory nlink values to 2, instead of using the st_nlink value reported by stat().

Fix the unit tests by dropping the GNU cpio `--reproducible` alias parameter, and instead specify `--ignore-devno --renumber-inodes` explicitly, matching pre-6a94d5e GNU cpio `--reproducible` behaviour.

Fixes: https://github.com/dracut-ng/dracut-ng/issues/1694

## Checklist

Fixes #

### DIFF
--- a/src/dracut-cpio/src/main.rs
+++ b/src/dracut-cpio/src/main.rs
@@ -813,8 +813,6 @@ fn main() -> std::io::Result<()> {
     Ok(())
 }
 
-// tests change working directory, so need to be run with:
-// cargo test -- --test-threads=1 --nocapture
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -823,7 +821,12 @@ mod tests {
     use std::path::PathBuf;
     use std::process::{Command, Stdio};
 
-    struct TempWorkDir {
+    pub static TEST_LOCK: std::sync::Mutex<u32> = std::sync::Mutex::new(0);
+
+    struct TempWorkDir<'a> {
+        // Hold a mutex alongside directory change, to avoid failures due to
+        // multi-threaded "cargo test".
+        cwd_lock: std::sync::MutexGuard<'a, u32>,
         prev_dir: PathBuf,
         parent_tmp_dir: PathBuf,
         cleanup_files: Vec<PathBuf>,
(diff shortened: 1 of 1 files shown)


ddiss:
v3:
- add extra fix for https://github.com/dracut-ng/dracut-ng/issues/1702

src/dracut-cpio/src/main.rs
```diff
@@ -1810,4 +1797,40 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_archive_all_single_threaded() {
+        // Each test_archive_ function changes its CWD and cargo test spawns
+        // multiple threads for concurrent testing by default, unless given
+        // --test-threads=1. Therefore, use a single wrapper test so that people
+        // don't get confused by multi-threaded "cargo test" failures.
+        // If someone finds a way make single-threaded the default for
+        // "cargo test" then we can drop this wrapper in future.
```
bdrung: I am not sure if that solution is the best one. I ran into a similar issue with 3cpio and the solution there is quite simple. I created:

```
    // Lock for tests that rely on / change the current directory
    pub static TEST_LOCK: std::sync::Mutex<u32> = std::sync::Mutex::new(0);
```

Then each test that relies on cwd gets this call as first step:

```
        let _lock = TEST_LOCK.lock().unwrap();
```

ddiss: I did play around with mutexes for this bug initially, but wasn't happy with the complexity (was using Arc, etc.).
IMO it should be possible to disable multi-threading with a `#[test-single-threaded]` parameter or something.
Anyhow, your suggestion above seems reasonable, even with a lifetime parameter for plumbing into TempWorkDir. I'll resubmit with a change based on it.

Oh BTW, I didn't mean to step on your 3cpio changes with this PR; I still think moving Dracut towards archive-in-place-without-staging is a worthy goal. I've been reusing dracut-cpio for some other work (https://github.com/rapido-linux/rapido) and didn't want to these fixes to get lost.

bdrung (approved):
Looks good now. I probably should implement that release-lock-from-failed-tests in 3cpio as well.

PR #1698, opened 2025-09-22
fix(systemd-portabled): do not call portablectl

jozzsi:
Drop the portablectl call altogether and stop adding {/etc,/var,usr}/lib/portables/*.raw to the initramfs automatically.                                                                                                                               
                                                                                                                                                                                                  
These .raw files are full OS trees and are rather huge, adding a specific .raw one manually seems like a legit use-case. 

## Checklist

Fixes #


chewi:
I don't know why you need more feedback. Looks like you have plenty. I roughly know what portable services are, but I've never used them, and I also don't know why anyone would want this in the initrd. I'm fine with the proposed changes or dropping the module entirely.

PR #1710, opened 2025-09-26
chore(dracut): enforce that /etc is only used in hostonly mode

jozzsi:
Only set global variables using /etc in hostonly mode, ot make sure errors are exposed if /etc is used in non-hostonly mode.

The goal if this PR is to ensure reproducible non-hostonly images.

## Checklist

Fixes #


chewi:
Was this change supposed to leak into /etc/conf.d/systemd.conf?

```diff
@@ -1,4 +1,4 @@
 systemdutildir="/usr/lib/systemd"
 systemdsystemunitdir="/usr/lib/systemd/system"
-systemdsystemconfdir="/etc/systemd/system"
-systemdnetworkconfdir="/etc/systemd/network"
+systemdsystemconfdir=""
+systemdnetworkconfdir=""
```

Maybe so, but modules.d/11systemd-networkd/networkd-config.sh has this line:

```bash
rm -f "$systemdnetworkconfdir"/zzzz-dracut-default.network
```

Isn't that a problem?

PR #1737, opened 2025-09-29
feat(plymouth): use simpledrm if Plymouth is configured to it

bdrung:
## Changes

Plymouth has a configuration setting `UseSimpledrm` for using `simpledrm`. Check this configuration setting and only pull in `simpledrm` instead of `drm` in case it is set.

Ubuntu is using `simpledrm` by default.

This merge requests depends on https://github.com/dracut-ng/dracut-ng/pull/1735 which should be reviewed and merged before.

## Checklist

Fixes #


modules.d/45plymouth/module-setup.sh
```diff
@@ -12,6 +12,26 @@ pkglib_dir() {
...
 
+read_setting_from_config() {
+    local setting="$1"
+    local config_file="$2"
+    sed -n "s/^${setting}= *\([^ ]\+\) */\1/p" "$config_file"
+}
+
+use_simpledrm() {
+    local config use_simpledrm
+    for config in "${dracutsysrootdir-}/etc/plymouth/plymouthd.conf" "${dracutsysrootdir-}/usr/share/plymouth/plymouthd.defaults"; do
```
jozzsi: I think checking `/etc` this should be behind a `hostonly` flag

bdrung: Good catch. Fixed.

Conan-Kudo (dismissed):

PR #1746, opened 2025-10-06
fix(network-manager): hostname not set with old systemd service units

### LINKED ISSUE #1743: hostname is no longer set from kernel arguments
**Describe the bug**
We see a regression in NetworkManager dracut testsuite - when we set hostname, it is not honored and transient hostname `fedora` is used

**Distribution used**
Fedora-Rawhide

**Dracut version**
dracut-108-2.fc44.x86_64

**Init system**
systemd

**To Reproduce**
boot with hostname set, e.g. add this kernel option in static IP setting `ip=192.168.50.201::192.168.50.1:255.255.255.0:nfs-cl:52-54-00-12-34-00:none`

**Expected behavior**
Hostname `nfs-cl` should be set, not `fedora`

**Additional context**
It seems this commit might be the cause:
https://github.com/dracut-ng/dracut-ng/commit/83dffc58f606ad7ad47a32716ce240831d7f018f

aafeijoo-suse:
Regression introduced in 83dffc58f606ad7ad47a32716ce240831d7f018f

## Checklist

Fixes #1743

### DIFF
--- a/modules.d/35network-manager/nm-run.sh
+++ b/modules.d/35network-manager/nm-run.sh
@@ -13,10 +13,11 @@ if [ -z "${DRACUT_SYSTEMD-}" ]; then
             break
         done
     fi
+fi
 
-    if [ -s /run/NetworkManager/initrd/hostname ]; then
-        cat /run/NetworkManager/initrd/hostname > /proc/sys/kernel/hostname
-    fi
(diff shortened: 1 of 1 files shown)


bdrung:
Can you explain this change a little bit?

PR #1747, opened 2025-10-06
feat(IMA): Ability to halt system if IMA policy fails to load

doublez13:
Add a new boolean cmdline option that, when set, will halt the system if the IMA policy file is not present or fails to load

When both `ima_policy=secure_boot` and `rd.ima.require_policy_file` are set, the system will halt if the required policy file is not also signed.

This option provides higher assurance that the system can boot into a trusted IMA state.

This pull request changes...

## Checklist

Fixes #


doublez13:
I will close this and submit a new PR with proper commit message formatting

PR #1748, opened 2025-10-06
feat(IMA): ability to halt system if IMA policy fails to load

doublez13:
Add a new boolean cmdline option that, when set, will halt the system if the IMA policy file is not present or fails to load

When both `ima_policy=secure_boot` and `rd.ima.require_policy_file` are set, the system will halt if the required policy file is not also signed.

This option provides higher assurance that the system can boot into a trusted IMA state

## Checklist

### DIFF
--- a/modules.d/77integrity/ima-policy-load.sh
+++ b/modules.d/77integrity/ima-policy-load.sh
@@ -28,10 +28,14 @@ load_ima_policy() {
     IMAPOLICYPATH="${NEWROOT}${IMAPOLICY}"
 
     # check the existence of the IMA policy file
+    # shellcheck disable=SC2015
     [ -f "${IMAPOLICYPATH}" ] && {
         info "Loading the provided IMA custom policy"
         printf '%s' "${IMAPOLICYPATH}" > "${IMASECDIR}"/policy \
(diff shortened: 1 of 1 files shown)


doublez13:
For documentation, should this option be added to the Misc section?

modules.d/77integrity/ima-policy-load.sh
```diff
@@ -28,10 +28,14 @@ load_ima_policy() {
     IMAPOLICYPATH="${NEWROOT}${IMAPOLICY}"
 
     # check the existence of the IMA policy file
+    # shellcheck disable=SC2015
```
bdrung: Instead of disabling SC2015, I would prefer using if-then-else statements that are easier to read than these multi-line statements.

doublez13: Sounds good! done

doublez13:
> This should be `feat(integrity)` not `feat(IMA)`.

Should the boolean be changed to `rd.integrity.require_policy_file` or is there something else you'd prefer?

devkontrol:
> This option provides higher assurance that the system can boot into a trusted IMA state

Does it make sense to have it as an option and allow to opt out of it ?

Issue #1749, opened 2025-10-07
dracut-cpio: incorrect dev_t minor field setting


ddiss:
**Describe the bug**
A typo that I made in the initial dracut-cpio implementation sees the stat()-derived major number used for both major and minor numbers in the cpio archive, instead of the unique major/minor tuple.
This bug shouldn't cause any real-world issues, as archive-renumbered inode numbers are already unique and hardlink tracking checks against the proper dev_t tuple.

**Distribution used**
N/A

**Dracut version**
Current HEAD (82cd3d3726db25a39b26b5039fcb76cc414e6b89). Introduced in a9c67046 ("feat(cpio): add newc archive creation utility")

**Init system**
N/A

**To Reproduce**
Use dracut-cpio to archive two input files located on differing filesystems / devices. Both files will carry the same major/minor tuple in the archive.

**Expected behavior**
The archived major/minor tuple should be unique to reflect that the input files are located on differing devices.

ddiss:
copying what I said in the PR:

FWIW, I had a look at GNU cpio behaviour, checking how it archived major / minor numbers when archive source files cross multiple devices. The --reproducible (--ignore-devno) parameter actually zeros the fields regardless, as the renumbered inode numbers ( --renumber-inodes) ensure (major,minor,inode) tuple uniqueness for non-hardlinks.

I've proposed additional changes to match GNU cpio `--ignore-devno` with major/minor zeroing.

PR #1750, opened 2025-10-07
fix(cpio): archive correct minor numbers

ddiss:
A typo that I made in the initial dracut-cpio implementation sees the stat()-dev_t-derived major number used for both major and minor number in the cpio archive.

This bug shouldn't cause any real-world issues, as archive-renumbered inode numbers are already unique and hardlink tracking checks against the proper dev_t tuple.

Fixes: https://github.com/dracut-ng/dracut-ng/issues/1749

## Checklist

### DIFF
--- a/src/dracut-cpio/src/main.rs
+++ b/src/dracut-cpio/src/main.rs
@@ -1797,4 +1797,46 @@ mod tests {
             assert!(gmd.rdev() == dmd.rdev());
         }
     }
+
+    // Inode numbers are unique (for non-hardlinks) within the archive, so
+    // device ID mapping is unnecessary. Confirm that dracut-cpio behaves like
+    // GNU cpio --ignore-devno. Check this by archiving the /tmp directory
+    // alongside a working-directory nested file; despite differing source
+    // device IDs, the archived major/minor numbers should be zero.
+    #[test]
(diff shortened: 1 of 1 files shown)


ddiss:
FWIW, I had a look at GNU cpio behaviour, checking how it archived major / minor numbers when archive source files cross multiple devices. The `--reproducible` (`--ignore-devno`) parameter actually zeros the fields regardless, as the renumbered inode numbers (` --renumber-inodes`) ensure (major,minor,inode) tuple uniqueness for non-hardlinks.
We could do the same, but the hardlink tracking code would need a slight rework.

ddiss:
> FWIW, I had a look at GNU cpio behaviour, checking how it archived major / minor numbers when archive source files cross multiple devices. The `--reproducible` (`--ignore-devno`) parameter actually zeros the fields regardless, as the renumbered inode numbers (` --renumber-inodes`) ensure (major,minor,inode) tuple uniqueness for non-hardlinks. We could do the same, but the hardlink tracking code would need a slight rework.

I've ripped out the unnecessary major / minor device mapping, so that we match GNU cpio `--ignore-devno`. I'll push the changes (with a test) on top of the initial typo fix here.

Conan-Kudo (approved):

PR #1752, opened 2025-10-07
Support additional hooks locations in /etc and /lib 

### LINKED ISSUE #1618: Make dracut hooks compatible with systemd-sysext/systemd-confext
Currently, dracut hooks (e.g. cmdline|pre-udev|pre-trigger|...) live in /var/lib/dracut/hooks. Previously, these hooks used to be in {/usr,}/lib/dracut/hooks but this doesn't work well when /usr is mounted read-only. Dracut need a writable location. See
```
commit a45048b80c27ee5a45a380052a6d29ab1925f7f9
Author: Laszlo Gombos <laszlo.gombos@gmail.com>
Date:   Fri Apr 5 15:50:20 2024 +0200

    fix(dracut): move hooks directory from /usr/lib to /var/lib
```
for additional details.

The scheme works well, however, it is incompatible with systemd-sysext/systemd-confext (see e.g. https://www.freedesktop.org/software/systemd/man/latest/systemd-sysext.html) way of extending initramfs. Systemd's sysext can only extend /usr or /opt and confext can extend /etc. (and the /lib/dracut/hooks symlink does not really help in any way).

The easiest way to make dracut systemd-sysext/confext compatible seems to be to add a *second* location for hooks. E.g. we can add /etc/dracut.hooks.d/{cmdline|pre-udev|pre-trigger,...} and then execute hooks from both directories if present. (In theory, it is also possible to add support for 'overriding' certain hooks, e.g. if both /etc/dracut.hooks.d/pre-pivot/01-myhook.sh and /var/lib/dracut/hooks/pre-pivot/01-myhook.sh exist, only the one in /etc gets to run. I'm not sure if there's a use-case for that though.)

I'm happy to work on an implementation if the whole idea seems acceptable.

vittyvk:
## Problem

Dracut init supports running custom hooks which are placed to `/var/lib/dracut/hooks/`. The location was previously changed to /var as the place needs to be writeable as some hooks are created and removed in runtime. The current location, however, may come inconvenient in some scenarios when the user wants to extend initramfs with custom scripts. In particular, systemd allows extending initrd with 'sysext' and 'confext' mechanism. This comes handy for extending e.g. UKI's initramfs without the need to rebuild (and re-sign) the UKI. The problem is that 'sysext' can only be used to extend /usr and /opt and 'confext/ can only extend /etc. Both services make the location read-only and thus can't be used for the main dracut hooks location even if we move it somewhere.

## Changes

Add additional locations where users can put dracut hooks:
- `/lib/dracut/hooks` -- this location is supposed to be used for distro-specific static hooks.
- `/etc/dracut.hooks.d` -- this location can be used by users for locally  created hooks.

`/var/lib/dracut/hooks` stays the default location which is supposed to be  used by dracut modules. This location is always writeable so modules can place and remove hooks from there in runtime. 

## Checklist

Fixes #1618


bdrung:
The current config directory is `/etc/dracut.conf.d`. When we add more stuff to Dracut we could decide to use one common `/etc/dracut` directory. Using `/etc/dracut.conf.d/hooks.d` sounds wrong to me. Since configs and hooks are not the same.

For consistency for the hooks directory, I would go with `/usr/lib/dracut/hooks`, `/etc/dracut/hooks`, and `/var/lib/dracut/hooks`.

vittyvk:
I've updated the location to `/etc/dracut/hooks` in v2.

vittyvk:
> (Unfortunately) the new test needs to be added explicitly to `.github/workflows/integration.yml` (and .github/workflows/integration-extra.yml), otherwise it would not run.

Missed that, added (and hope I did it correctly)! Thanks!

bdrung:
@vittyvk Can you rebase your branch?

vittyvk:
@bdrung sure, done!

LaszloGombos (approved):
Let's go! 

I large and uncontroversial PR should not stay in the review queue long to encourage active community participation.

vittyvk:
It seems we're still stuck here as we need more ACKs: @bdrung could you take a look please? Thanks a lot!

pvalena:
@LaszloGombos are we waiting for some additonal review or testing?

LaszloGombos:
> @LaszloGombos are we waiting for some additonal review or testing?

Approved on Dec-3, so not sure why I was pinged on this PR. This PR now has a conflict, that needs rebasing.

LaszloGombos (dismissed):

test/TEST-14-HOOKS/test.sh
```diff
@@ -0,0 +1,110 @@
...
+        done
+        _i=$((_i + 1))
+    done
+
+    echo "Testing hooks with the same name in all hooks location: /var gets priority"
+    # Same name -- only the hook in the highest priority dir gets to run.
+    for _h in $hookdirs; do
+        # Only the hook in /var/lib/dracut/hooks must execute
+        testlog_check "@0$_i-/var/lib/dracut/hooks-$_h-samename3@"
+        testlog_check "@0$_i-/etc/dracut/hooks-$_h-samename3@" || exit 1
```
bdrung: Since run with `set -eu`, `|| exit 1` should not be needed.

test/TEST-14-HOOKS/test.sh
```diff
@@ -0,0 +1,110 @@
...
+        done
+    done
+
+    # create initramfs
+    test_dracut --keep \
+        -I "dd sync" \
+        -i ./lasthook.sh /var/lib/dracut/hooks/cleanup/99-lasthook.sh
+
+    # create a separate volume for test results
+    dd if=/dev/zero of="$TESTDIR"/testlog.img bs=1MiB count=1 status=none && sync "$TESTDIR"/testlog.img
```
bdrung: We do not need `sync` on the host side, do we?

doc_site/modules/ROOT/pages/developer/modules.adoc
```diff
@@ -123,26 +123,33 @@ You are encouraged to provide a `README` that describes what the module is for.
...
-
-`/lib/dracut/hooks/cmdline/*.sh`::
+Dracut init looks for custom hook scripts in the following locations
+(`HOOKDIR` below): `/var/lib/dracut/hooks`, `/etc/dracut/hooks`, and
+`/lib/dracut/hooks`. The intended use of these locations is: standard,
+distribution shipped scripts are put to `/lib/dracut/hooks`;
+`/etc/dracut/hooks` is used as a local override; dracut modules create
+(and remove) scripts in runtime in `/var/lib/dracut/hooks` (and `$hookdir`
+variable is provided). The following hook points to inject scripts are
+currently supported:
```
bdrung, line 133: Can we have this information in one of the man pages as well?

That reminds me that systemd tools often list the path they are looking at in a nice format.

test/TEST-14-HOOKS/test.sh
vittyvk: It was supposed to be "&&" and because I made this mistake, I did not notice that hooks from all locations (and not the most privileged one, as documented) are actually executed. Fixed now, thanks!

test/TEST-14-HOOKS/test.sh
vittyvk: Indeed, not sure where I copied it from, removed now.

doc_site/modules/ROOT/pages/developer/modules.adoc
vittyvk: Added to dracut.modules.7 as well.

dracut.sh
```diff
@@ -2731,20 +2733,18 @@ mkdir -p "${initdir}"/lib/dracut
...
     ((${#install_items[@]} > 0)) && inst_multiple ${install_items[@]}
     # shellcheck disable=SC2068
     ((${#install_optional_items[@]} > 0)) && inst_multiple -o ${install_optional_items[@]}
 
-    # symlink to old hooks location for compatibility
-    ln_r /var/lib/dracut/hooks /lib/dracut/hooks
-
-    for _d in $hookdirs; do
-        # shellcheck disable=SC2174
-        mkdir -m 0755 -p "${initdir}/lib/dracut/hooks/$_d"
```
bdrung: The first commit should change the path of this line.

modules.d/80base/dracut-lib.sh
```diff
@@ -356,35 +356,35 @@ setdebug() {
...
-    for f in "/$_dir"/*.sh; do
-        if [ -e "$f" ]; then
-            # shellcheck disable=SC1090
-            # shellcheck disable=SC2240
-            . "$f" "$@"
-        fi
+    _pattern=$2
+    [ -z "$_pattern" ] && _pattern="*.sh"
+    for hook in "$hookdir/$_dir/"$_pattern; do
+	    [ "$hook" = "$hookdir/$_dir/$_pattern" ] && break
```
bdrung: This check doesn't seem to be needed. `[ "$hook" = "$hookdir/$_dir/$_pattern" ]` does the same as the following `[ -f "$hook" ]`

modules.d/80base/dracut-lib.sh
```diff
@@ -356,35 +356,35 @@ setdebug() {
...
 setdebug
 
-source_all() {
-    local f
+hookdir=/var/lib/dracut/hooks
+export hookdir
+
+list_hooks() {
     local _dir
+    local _pattern
```
bdrung: Common pattern would be:
```
local _dir="$1"
local _pattern="$2"
```

modules.d/80base/dracut-lib.sh
```diff
@@ -1063,7 +1074,8 @@ check_meminfo() {
 }
 
 remove_hostonly_files() {
-    rm -fr /etc/cmdline /etc/cmdline.d/*.conf "$hookdir"/initqueue/finished/*.sh
```
bdrung, line 1074: You use `rm_hooks` only once. So I see no point in replacing one line by 12 lines.

modules.d/86shutdown/module-setup.sh
```diff
@@ -9,15 +9,14 @@ depends() {
...
     inst_multiple -o kexec
     inst "$moddir/shutdown.sh" "$prefix/shutdown"
-    mkdir -m 0755 -p "${initdir}"/var/lib/dracut/hooks
 
-    # symlink to old hooks location for compatibility
-    ln_r /var/lib/dracut/hooks /lib/dracut/hooks
-
-    for _d in $hookdirs shutdown shutdown-emergency; do
-        mkdir -m 0755 -p "${initdir}"/var/lib/dracut/hooks/"$_d"
+    for _h in "/var/lib/dracut/hooks" "/etc/dracut/hooks" "/lib/dracut/hooks"; do
```
bdrung: Why should we create these directories in /lib and /etc? Scripts that places theses scripts can create these directories.

modules.d/80base/dracut-lib.sh
```diff
@@ -362,12 +362,26 @@ export hookdir
...
+        # which has higher priority. '/var/lib/dracut/hooks' gets top priority,
+        # '/etc/dracut/hooks' comes after and /lib/dracut/hooks is the least
+        # priviliged location.
+        if [ -f "/var/lib/dracut/hooks/$_dir/$_hook" ]; then
+            echo "/var/lib/dracut/hooks/$_dir/$_hook"
+        elif [ -f "/etc/dracut/hooks/$_dir/$_hook" ]; then
+            echo "/etc/dracut/hooks/$_dir/$_hook"
+        else
+            echo "/lib/dracut/hooks/$_dir/$_hook"
+        fi
```
bdrung: I didn't like the need to pull in `sort` and thought about an alternative.

```
    for hook in "/var/lib/dracut/hooks/$_dir/"$_pattern; do
        [ -f "$hook" ] || continue
        echo "$hook"
    done
    for hook in "/etc/dracut/hooks/$_dir/"$_pattern; do
        [ -f "$hook" ] || continue
        [ ! -f "/var/lib/dracut/hooks/$_dir/${hook##*/}" ] || continue
        echo "$hook"
    done
    for hook in "/lib/dracut/hooks/$_dir/"$_pattern; do
        [ -f "$hook" ] || continue
        [ ! -f "/var/lib/dracut/hooks/$_dir/${hook##*/}" ] || continue
        [ ! -f "/etc/dracut/hooks/$_dir/${hook##*/}" ] || continue
        echo "$hook"
    done
```

or in the short format:

```
    for hook in "/var/lib/dracut/hooks/$_dir/"$_pattern; do
        [ -f "$hook" ] && echo "$hook"
    done
    for hook in "/etc/dracut/hooks/$_dir/"$_pattern; do
        [ -f "$hook" ] && [ ! -f "/var/lib/dracut/hooks/$_dir/${hook##*/}" ] && echo "$hook"
    done
    for hook in "/lib/dracut/hooks/$_dir/"$_pattern; do
        [ -f "$hook" ] && [ ! -f "/var/lib/dracut/hooks/$_dir/${hook##*/}" ] \
            [ ! -f "/etc/dracut/hooks/$_dir/${hook##*/}" ] && echo "$hook"
    done
```
What do you think?

bdrung:
Here is the proposal for the test case: https://github.com/dracut-ng/dracut-ng/commit/e4d9e5033354dec027dc1c6e6f2283bf89673368

modules.d/80base/dracut-lib.sh
vittyvk: Hey @bdrung sorry I missed your comment while at FOSDEM -- it would be wonderful to meet! I'm back now and I will try to address your feedback quickly (thanks for the review btw!).

dracut.sh
vittyvk: Did in https://github.com/dracut-ng/dracut-ng/pull/2181

vittyvk:
> 
> To speed up landing this change, can you create another PR with only the first commit?
> 

Done: https://github.com/dracut-ng/dracut-ng/pull/2181

modules.d/80base/dracut-lib.sh
vittyvk: Thank you for the suggestion, it's certainly better without the sort.

modules.d/86shutdown/module-setup.sh
vittyvk: Ok, sure, we don't have to create them. Removed.

modules.d/80base/dracut-lib.sh
vittyvk: Fixed!

modules.d/80base/dracut-lib.sh
vittyvk: Removed.

test/TEST-14-HOOKS/test.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+set -eu
+# shellcheck disable=SC2034
+TEST_DESCRIPTION="dracut hooks in various locations"
+
```
bdrung, line 5:
```suggestion
# Uncomment this to debug failures
#DEBUGFAIL="rd.debug rd.shell"

```

vittyvk: Added!

modules.d/80base/dracut-lib.sh
```diff
@@ -364,35 +364,45 @@ setdebug() {
...
-    done
-}
-
 hookdir=/var/lib/dracut/hooks
 export hookdir
 
+list_hooks() {
+    local _dir="$1"
+    local _pattern="$2"
+    [ -z "$_pattern" ] && _pattern="*.sh"
```
bdrung:
```suggestion
    [ -z "$_pattern" ] && _pattern="*.sh"
    local hook
```

modules.d/80base/dracut-lib.sh
```diff
@@ -364,35 +364,45 @@ setdebug() {
...
+
+    # Hooks are executed in the alphabetical order. It is also allowed to
```
bdrung: > Hooks are executed in the alphabetical order.

This is not the case any more and we haven't made the promise anywhere.

modules.d/80base/dracut-lib.sh
```diff
@@ -364,35 +364,45 @@ setdebug() {
...
-        fi
-    done
-}
-
 hookdir=/var/lib/dracut/hooks
 export hookdir
 
+list_hooks() {
+    local _dir="$1"
+    local _pattern="$2"
```
bdrung: Since these variables are declared as private, the underscore is not needed.

modules.d/80base/dracut-lib.sh
vittyvk: Fixed.

modules.d/80base/dracut-lib.sh
vittyvk: Dropped.

modules.d/80base/dracut-lib.sh
vittyvk: Removed.

Conan-Kudo:
One last question: why are we adding paths in `/lib` when we've been trying to push things into `/usr`?

bdrung:
Good point. We should better use `/usr/lib` (even for not-/usr-merged systemd).

vittyvk:
@bdrung @Conan-Kudo do you want to me update /lib/dracut/hooks -> /usr/lib/dracut/hooks here or can we make it a follow-up change? I can see that a number of tests hardcode "/lib/dracut/hooks/" path as well.

PR #1755, opened 2025-10-08
fix(dracut-rescue): add support for /run/initramfs/dracut.conf.d                                                                                                                                  

jozzsi:
Allow dracut read configuration files from /run as well, in addition to /etc/ and /usr including the dracut-rescue kernel-install script.                                                                                                                                                                            
                                                                                                                                                                                                  
Modify test case to demonstrate use-case and provide test coverage.                                                                                                                                                                                         
                                                                                                                                                                                                  
Follow-up to 82cd3d3 . 

## Checklist

Fixes #

### DIFF
--- a/install.d/51-dracut-rescue.install
+++ b/install.d/51-dracut-rescue.install
@@ -110,7 +110,7 @@ case "$COMMAND" in
         fi
 
         # source our config dir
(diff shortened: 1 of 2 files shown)


bdrung:
The change looks good, but I fail to understand what you want to say in your git commit message.

PR #1763, opened 2025-10-10
feat(kernel-modules-export): make kernel modules from initramfs available to rootfs

### LINKED ISSUE #1762: make kernel modules from initramfs available to rootfs
When booting with an external-to-rootfs kernel you need to ensure that /lib/modules contains any necessary modules not already loaded.

This feature arranges for the kernel modules in the initramfs to be placed into /lib/modules.

Although the kernel module for the filesystem is in the initramfs, it might not loaded in the initramfs, because it is not needed to mount the root filesystem. 

An implementation of this idea from another initramfs generator - https://git.launchpad.net/cloud-initramfs-tools/tree/copymods/scripts/init-bottom/copymods

CC @bdrung

jozzsi:
## Changes

Although the kernel module for the filesystem is in the initramfs, it might not loaded in the initramfs, because it is not needed to mount the root filesystem. 

This feature arranges for the kernel modules in the initramfs to be placed into /lib/modules.

When booting with an external-to-rootfs kernel you need to ensure that /lib/modules contains any necessary modules not already loaded.

## Checklist

Fixes #1762

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -121,6 +121,7 @@ jobs:
                     - "41"
                     - "42"
                     - "43"
+                    - "44"
                 exclude:
                     # https://github.com/dracut-ng/dracut-ng/issues/1677
                     - container: arch:latest
--- a/doc_site/modules/ROOT/pages/modules/core.adoc
+++ b/doc_site/modules/ROOT/pages/modules/core.adoc
@@ -205,6 +205,10 @@ code, there are no specific types or categories for dracut modules.
 | mount block fallback device as rootfs
 | device
 
+| kernel-modules-export
+| make kernel modules from initramfs available to rootfs
+| kernel
+
 | securityfs
 | mount securityfs early
 | filesystem
--- a/man/dracut.cmdline.7.adoc
+++ b/man/dracut.cmdline.7.adoc
@@ -149,6 +149,10 @@ Misc
     force loading kernel module <drivername> after all automatic loading modules
     have been loaded. This parameter can be specified multiple times.
 
+**rd.driver.export**::
+    making all drivers from initramfs available to rootfs after switch_root.
+    This parameter can be set to 'force' for exporting over existing modules.
+
 **rd.retry=**__<seconds>__::
     specify how long dracut should retry the initqueue to configure devices.
     The default is 180 seconds. After 2/3 of the time, degraded raids are force
--- a/modules.d/70kernel-modules-export/modules-export.sh
+++ b/modules.d/70kernel-modules-export/modules-export.sh
@@ -1,0 +1,43 @@
+#!/bin/sh
+
+command -v getarg > /dev/null || . /lib/dracut-lib.sh
+
+if ! getargbool 0 rd.driver.export; then
+    exit 0
+fi
+
+driver_export=$(getarg rd.driver.export=)
+
+KVERSION="$(uname -r)"
+
+if [ ! -d "/lib/modules/$KVERSION" ]; then
+    warn "Something odd, no /lib/modules/$KVERSION in initramfs."
+    exit 0
+fi
+
+[ -d "$NEWROOT/lib/modules" ] || mkdir -p "$NEWROOT/lib/modules" \
+    || {
+        warn "No /lib/modules in target. cannot help."
+        exit 0
+    }
+
+if [ -d "$NEWROOT/lib/modules/$KVERSION" ]; then
+    if [ "${driver_export#*force}" = "$driver_export" ]; then
+        warn "/lib/modules/$KVERSION exists. To export modules set rd.driver.export=force!"
+        exit 0
+    else
+        info "Due to rd.driver.export=force exporting over existing modules!"
+    fi
+fi
+
+mount -t tmpfs driver_export "$NEWROOT/lib/modules" \
+    || {
+        warn "Failed mount of tmpfs."
+        exit 0
+    }
+
+cp -a "/lib/modules/$KVERSION" "$NEWROOT/lib/modules" \
+    || {
+        warn "Failed to export modules to target root."
+        exit 0
+    }
--- a/test/TEST-44-DRIVERS/test.sh
+++ b/test/TEST-44-DRIVERS/test.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+set -eu
+# shellcheck disable=SC2034
+TEST_DESCRIPTION="no (xfs) driver on root filesystem"
+
+test_check() {
+    if ! type -p mkfs.xfs &> /dev/null; then
+        echo "Test needs mkfs.xfs.. Skipping"
+        return 1
+    fi
+
+    if ! command -v systemctl > /dev/null; then
+        echo "This test needs systemd to run."
+        return 1
+    fi
+}
+
+# Uncomment this to debug failures
+#DEBUGFAIL="rd.shell=1 rd.break=pre-mount"
+test_run() {
+    declare -a disk_args=()
+    declare -i disk_index=0
+    qemu_add_drive disk_index disk_args "$TESTDIR"/marker.img marker
+    qemu_add_drive disk_index disk_args "$TESTDIR"/root.img root
+    qemu_add_drive disk_index disk_args "$TESTDIR"/mnt.img mnt
+
+    test_marker_reset
+
+    # This test should fail if rd.driver.export is not passed at kernel command-line
+    "$testdir"/run-qemu \
+        "${disk_args[@]}" \
+        -append "$TEST_KERNEL_CMDLINE rd.driver.export" \
+        -initrd "$TESTDIR"/initramfs.testing
+
+    test_marker_check
+}
+
+test_setup() {
+    # Create what will eventually be our root filesystem onto an overlay
+    call_dracut --tmpdir "$TESTDIR" \
+        --no-kernel \
+        --add "systemd-udevd systemd-journald systemd-tmpfiles systemd-ldconfig systemd-ask-password shutdown" \
+        --mount "/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_mnt /mnt xfs rw" \
+        --add-confdir test-root \
+        -f "$TESTDIR"/initramfs.root
+    mkdir -p "$TESTDIR"/overlay/source
+    mv "$TESTDIR"/dracut.*/initramfs/* "$TESTDIR"/overlay/source
+    rm -rf "$TESTDIR"/dracut.*
+
+    # make sure no linux kernel driver is included in the rootfs
+    rm -rf "$TESTDIR"/overlay/source/lib/modules/*
+
+    # make sure /lib/modules directory exists inside the rootfs
+    mkdir -p "$TESTDIR"/overlay/source/lib/modules "$TESTDIR"/overlay/source/mnt
+
+    # create an initramfs that will create the target root filesystem.
+    # We do it this way so that we do not risk trashing the host mdraid
+    # devices, volume groups, encrypted partitions, etc.
+    call_dracut -i "$TESTDIR"/overlay / \
+        --add-confdir "test-makeroot" \
+        -I "mkfs.xfs" \
+        -i ./create-root.sh /lib/dracut/hooks/initqueue/01-create-root.sh \
+        -f "$TESTDIR"/initramfs.makeroot
+
+    declare -a disk_args=()
+    # shellcheck disable=SC2034  # disk_index used in qemu_add_drive
+    declare -i disk_index=0
+    qemu_add_drive disk_index disk_args "$TESTDIR"/marker.img marker 1
+    qemu_add_drive disk_index disk_args "$TESTDIR"/root.img root 1
+    qemu_add_drive disk_index disk_args "$TESTDIR"/mnt.img mnt 1
+
+    # Invoke KVM and/or QEMU to actually create the target filesystem.
+    "$testdir"/run-qemu \
+        "${disk_args[@]}" \
+        -append "root=/dev/fakeroot quiet console=ttyS0,115200n81" \
+        -initrd "$TESTDIR"/initramfs.makeroot
+    test_marker_check dracut-root-block-created
+    rm -- "$TESTDIR"/marker.img
+
+    test_dracut \
+        --add-drivers xfs \
+        --add kernel-modules-export
+}
+
+# shellcheck disable=SC1090
+. "$testdir"/test-functions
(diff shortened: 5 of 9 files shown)


modules.d/70kernel-modules-apply/copymods.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+
+command -v getarg > /dev/null || . /lib/dracut-lib.sh
+
+copymods=$(getarg rd.driver.copy=)
+
+if [ -n "$copymods" ]; then
```
bdrung:
```suggestion
if [ -z "$copymods" ]; then
    exit 0
fi
```

Then everything can be one level higher.

doc_site/modules/ROOT/pages/modules/core.adoc
```diff
@@ -205,6 +205,10 @@ code, there are no specific types or categories for dracut modules.
 | mount block fallback device as rootfs
 | device
 
+| kernel-modules-apply
+| make kernel modules from initramfs available to rootfs
+| kernel
```
bdrung, line 210: It should say that `mount -t tmpfs copymods "$NEWROOT/lib/modules"` is done.

man/dracut.cmdline.7.adoc
```diff
@@ -149,6 +149,9 @@ Misc
     force loading kernel module <drivername> after all automatic loading modules
     have been loaded. This parameter can be specified multiple times.
 
+**rd.driver.copy=**::
+    make all drivers from initramfs available to rootfs after switch_root.
```
bdrung: It should say that `mount -t tmpfs copymods "$NEWROOT/lib/modules"` is done and which values it can take (force, yes, no?)

test/TEST-44-DRIVERS/test.sh
```diff
@@ -0,0 +1,84 @@
...
+
+    test_marker_reset
+
+    # This test should fail if rd.driver.copy is not passed at kernel command-line
+    "$testdir"/run-qemu \
+        "${disk_args[@]}" \
+        -append "$TEST_KERNEL_CMDLINE rd.driver.copy=yes" \
+        -initrd "$TESTDIR"/initramfs.testing
+
+    test_marker_check
```
bdrung, line 35: Can you add a test that the xfs partition was actually mounted?

.github/workflows/integration.yml
```diff
@@ -121,6 +121,7 @@ jobs:
                     - "41"
                     - "42"
                     - "43"
+                    - "44"
```
bdrung, line 124: Can you install xfsprogs (for /usr/sbin/mkfs.xfs) in the debian container to get the test also run on Ubuntu?

jozzsi: https://github.com/dracut-ng/dracut-ng/pull/1793

modules.d/70kernel-modules-apply/copymods.sh
jozzsi: good idea, made the changes.

doc_site/modules/ROOT/pages/modules/core.adoc
jozzsi: This table currently just have a one line overview of what problem the module is solving (and not how it is solving it).

man/dracut.cmdline.7.adoc
jozzsi: Should we really lock ourselves for using tmpfs by describing implementation deatil in the interface documentation? 

Could at a later implementation bind mount be considered here ?

test/TEST-44-DRIVERS/test.sh
jozzsi: Would this actually increase the test coverage ? If the xfs partition was not mounted the test case already fails. What is the goal of checking this twice during the test run ?

jozzsi: To add, this is where the current code indirectly forced to check that xfs mount is actually worked

```
--mount "/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_mnt /mnt xfs rw" \
```

test/TEST-44-DRIVERS/test.sh
```diff
@@ -0,0 +1,84 @@
...
+
+test_setup() {
+    # Create what will eventually be our root filesystem onto an overlay
+    "$DRACUT" --tmpdir "$TESTDIR" \
+        --no-kernel \
+        --add "systemd-udevd systemd-journald systemd-tmpfiles systemd-ldconfig systemd-ask-password shutdown" \
+        --mount "/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_mnt /mnt xfs rw" \
+        --add-confdir test-root \
+        -f "$TESTDIR"/initramfs.root
+    mkdir -p "$TESTDIR"/overlay/source && mv "$TESTDIR"/dracut.*/initramfs/* "$TESTDIR"/overlay/source && rm -rf "$TESTDIR"/dracut.*
```
bdrung: Please avoid `&&` here.

man/dracut.cmdline.7.adoc
```diff
@@ -149,6 +149,10 @@ Misc
     force loading kernel module <drivername> after all automatic loading modules
     have been loaded. This parameter can be specified multiple times.
 
+**rd.driver.copy=**::
```
bdrung: The man page says:
> "rd.*" parameters mentioned without "=" are boolean parameters. They can be turned on/off by setting them to {0|1}.

So we should probably make that consistent. Take 0, 1, or force

doc_site/modules/ROOT/pages/modules/core.adoc
```diff
@@ -205,6 +205,10 @@ code, there are no specific types or categories for dracut modules.
 | mount block fallback device as rootfs
 | device
 
+| kernel-modules-apply
```
bdrung: Why do you name the module `kernel-modules-apply`? `apply` is not used anywhere else in this diff. How about making it more align with the parameter: `kernel-modules-copy`?

man/dracut.cmdline.7.adoc
bdrung: Good point. Then let's not mention tmpfs.

test/TEST-44-DRIVERS/test.sh
bdrung: Okay. So the test will fail with:
```
made it to the test rootfs!
● mnt.mount loaded failed failed /mnt
Powering down.
```

doc_site/modules/ROOT/pages/modules/core.adoc
jozzsi: I do not find 'apply' or 'copy' a particularly good name. We should discuss if there is a better name. For now I consistently renamed it to 'export' as it exports kernel modules from initramfs to rootfs. I am open for better name suggestions as long as we do not corner ourselves to a particular implementation and instead focusing on wgat this module does (and not how it does it).

man/dracut.cmdline.7.adoc
jozzsi: ok, I can changed to `copymods.sh ` script accordingly.

test/TEST-44-DRIVERS/test.sh
jozzsi: ok. done.

doc_site/modules/ROOT/pages/modules/core.adoc
bdrung: I agree that 'export' is a better name than 'apply'. IMO 'export' is a good enough name to go with it (i.e. change this module name and `rd.driver.export`). I wanted us to think hard enough about the name since the cmdline name will stick.

man/dracut.cmdline.7.adoc
bdrung:
```suggestion
**rd.driver.export**::
```
Looking at the other boolean values, `=` is not specified there.

man/dracut.cmdline.7.adoc
```diff
@@ -149,6 +149,10 @@ Misc
...
+    boolean (yes or no) or 'force' for making making all drivers from initramfs
```
bdrung: The other boolean options just define the True case. So we should probably do the same here. Something like:
> making all drivers from initramfs available to rootfs after switch_root. This parameter can be set to 'force' for ...

modules.d/70kernel-modules-export/copymods.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+command -v getarg > /dev/null || . /lib/dracut-lib.sh
+
+copymods=$(getarg rd.driver.export=)
```
bdrung: Since we settled on the name, we can drop the `copy` everywhere. I would rename `copymods` to `driver_export` and `copymods.sh` to `modules-export.sh`.

test/TEST-44-DRIVERS/test.sh
```diff
@@ -0,0 +1,86 @@
...
+    qemu_add_drive disk_index disk_args "$TESTDIR"/marker.img marker
+    qemu_add_drive disk_index disk_args "$TESTDIR"/root.img root
+    qemu_add_drive disk_index disk_args "$TESTDIR"/mnt.img mnt
+
+    test_marker_reset
+
+    # This test should fail if rd.driver.export is not passed at kernel command-line
+    "$testdir"/run-qemu \
+        "${disk_args[@]}" \
+        -append "$TEST_KERNEL_CMDLINE rd.driver.export=1" \
```
bdrung: Just setting `rd.driver.export` should enable it as well and we should test this default way.

man/dracut.cmdline.7.adoc
bdrung: Should be mention that this option is only available if the kernel-module-export module is included?

modules.d/70kernel-modules-export/modules-export.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/sh
+
+command -v getarg > /dev/null || . /lib/dracut-lib.sh
+
+copymods=$(getarg rd.driver.export=)
+
+if ! getargbool 0 rd.driver.export; then
```
bdrung, line 5: Slight performance improvement: this can be moved above the getarg call.

man/dracut.cmdline.7.adoc
jozzsi: done.

modules.d/70kernel-modules-export/copymods.sh
jozzsi: ok. done.

man/dracut.cmdline.7.adoc
jozzsi: This is a larger issue. Most options are only available if certain dracut modules are included. I think this should be handled as a follow-up where we discuss what is the best way to document this for all modules.

modules.d/70kernel-modules-export/modules-export.sh
jozzsi: nice. done.

test/TEST-44-DRIVERS/test.sh
jozzsi: ok. done.

man/dracut.cmdline.7.adoc
jozzsi: ok. done.

test/TEST-44-DRIVERS/test.sh
```diff
@@ -0,0 +1,86 @@
...
+    # make sure no linux kernel driver is included in the rootfs
+    rm -rf "$TESTDIR"/overlay/source/lib/modules/*
+
+    # make sure /lib/modules directory exists inside the rootfs
+    mkdir -p "$TESTDIR"/overlay/source/lib/modules "$TESTDIR"/overlay/source/mnt
+
+    # create an initramfs that will create the target root filesystem.
+    # We do it this way so that we do not risk trashing the host mdraid
+    # devices, volume groups, encrypted partitions, etc.
+    "$DRACUT" -i "$TESTDIR"/overlay / \
```
bdrung: Can you rebase and pick up the changes for https://github.com/dracut-ng/dracut-ng/pull/1807?

man/dracut.cmdline.7.adoc
```diff
@@ -149,6 +149,11 @@ Misc
     force loading kernel module <drivername> after all automatic loading modules
     have been loaded. This parameter can be specified multiple times.
 
+**rd.driver.export**::
+    making all drivers from initramfs available to rootfs after switch_root.
+    This parameter can be set to 'force' for making making all drivers from
+    initramfs available to rootfs after switch_root.
```
bdrung: `--force` is for exporting over existing modules.

modules.d/70kernel-modules-export/modules-export.sh
```diff
@@ -0,0 +1,42 @@
...
+fi
+
+[ -d "$NEWROOT/lib/modules" ] || mkdir -p "$NEWROOT/lib/modules" \
+    || {
+        warn "No /lib/modules in target. cannot help."
+        exit 0
+    }
+
+if [ -d "$NEWROOT/lib/modules/$KVERSION" ]; then
+    if [ "${driver_export#*force}" = "$driver_export" ]; then
```
bdrung, line 25: Should we emit a warning in this case?

modules.d/70kernel-modules-export/modules-export.sh
```diff
@@ -0,0 +1,42 @@
...
+        exit 0
+    else
+        warn "Due to rd.driver.export=force exporting over existing modules!"
```
bdrung: Should we lower that from a warning to an info level? The user requested that.

test/TEST-44-DRIVERS/test.sh
jozzsi: done

man/dracut.cmdline.7.adoc
jozzsi: done

modules.d/70kernel-modules-export/modules-export.sh
jozzsi: added a warning

modules.d/70kernel-modules-export/modules-export.sh
jozzsi: set it to info

modules.d/70kernel-modules-export/modules-export.sh
```diff
@@ -0,0 +1,43 @@
...
+
+[ -d "$NEWROOT/lib/modules" ] || mkdir -p "$NEWROOT/lib/modules" \
+    || {
+        warn "No /lib/modules in target. cannot help."
+        exit 0
+    }
+
+if [ -d "$NEWROOT/lib/modules/$KVERSION" ]; then
+    if [ "${driver_export#*force}" = "$driver_export" ]; then
+        warn "To export modules set rd.driver.export=force!"
```
bdrung:
```suggestion
        warn "/lib/modules/$KVERSION exists. To export modules set rd.driver.export=force!"
```

Issue #1677, opened 2025-09-18
adapt to systemd 258


jozzsi:
Arch upgraded to systemd 258, which resulted in the following symptom in TEST 41

```
 *** systemd units included in initrd from mkinitcpio but not from dracut:breakpoint-pre-basic.service
breakpoint-pre-mount.service
breakpoint-pre-switch-root.service
breakpoint-pre-udev.service
```

PR #1765, opened 2025-10-11
fix(FULL-SYSTEMD): omit plymouth dracut module         

jozzsi:
## Changes

This test case with the new version of systemd (258) fails because of plymouth-start.service.
    
It requires either including the plymouth dracut module (which can not yet be done in all CI containers) or it requires explicitly removing plymouth-start.service.

See https://github.com/dracut-ng/dracut-ng/issues/1677 .

## Checklist


jozzsi:
> We should start including the plymouth module everywhere, even if it's not being activated.

Sure, but that fails with systemd v258. Let's than continue not even running this test with systemd v258. Closing.

bdrung:
Has someone checked why it fails? Debian unstable has systemd v258 as well.

PR #1773, opened 2025-10-15
ci(alpine): add seabios package

jozzsi:
## Changes

seabios no longer gets installed in the alpine container which leads to the following error.

```
run-qemu: /usr/bin/qemu-system-x86_64 ...

qemu: could not load PC BIOS 'bios-256k.bin'
```

To resolve the error, make sure seabios is installed.

## Checklist

Fixes #


jozzsi:
no longer needed

PR #1774, opened 2025-10-16
fix(dracut): use grep -q/-s to silence output/error

coiby:
## Changes

There is no need to redirect output/error to /dev/null. Simply use -q/-s option provided by grep.

Note we should use avoid using "set -o pipefail lsinitrd PAHT_PATH_INITRAMFS | grep -q PATTERN". Because with set -o pipefail, a pipeline’s exit status becomes non-zero if any element of the pipeline fails and grep -q exits as soon as it sees a match.

## Checklist

Fixes #

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -475,15 +475,15 @@ inst_rule_group_owner() {
 
     # shellcheck disable=SC2013
     for i in $(sed -nr 's/.*OWNER=?"([^ "]+).*/\1/p' "$1"); do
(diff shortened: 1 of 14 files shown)


Conan-Kudo (approved):

Conan-Kudo (changes requested):
Please use `fix` instead of `chore`.

coiby:
> Please use `fix` instead of `chore`.

Addressed in new version. Thanks!

PR #1776, opened 2025-10-20
feat(dmsquash-live-autooverlay): support readlink from uutils

bdrung:
## Changes

The readlink command from GNU coreutils and the one from uutil coreutils behave differently in case the path contains symlinks and `..`. GNU readlink traverses symlinks first, but uutils resolves `..` first.

Example from Ubuntu 25.10 (questing):

```sh
$ gnureadlink -f /sys/class/block/nvme0n1p1/..
/sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/nvme/nvme0/nvme0n1
$ /usr/lib/cargo/bin/coreutils/readlink -f /sys/class/block/nvme0n1p1/..
/sys/class/block
```

Enforce the correct order by resolving the symlinks before `..`. That way the result does not rely on implementation specifics of different readlink implementations.

## Checklist

Fixes #

### DIFF
--- a/modules.d/70dmsquash-live-autooverlay/create-overlay.sh
+++ b/modules.d/70dmsquash-live-autooverlay/create-overlay.sh
@@ -51,7 +51,8 @@ gatherData() {
         info "Skipping overlay creation: unpartitioned or read-only media detected"
         exit 0
     fi
-    fullDriveSysfsPath=$(readlink -f "${rootDeviceSysfsPath}/..")
+    fullSysfsPath=$(readlink -f "${rootDeviceSysfsPath}")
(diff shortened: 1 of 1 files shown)


bdrung:
> How would the busybox readlink implementation compare to GNU coreutils ?

busybox behaves the same as GNU coreutils:

```
$ busybox readlink -f /sys/class/block/nvme0n1p1/..
/sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/nvme/nvme0/nvme0n1
```

ruct coreutils should be fixed (see https://github.com/uutils/coreutils/issues/9057) but I think this change makes the code more readable.

PR #1779, opened 2025-10-22
fix(base): set udev log level properly

turret-dev:
since udev logging functionality was added in af67d62151 to dracut, in commit 25e773eeb4 in systemd, udev has since migrated from its own logging system to the logging system found in the rest of systemd, which uses a different environment variable for setting the log level.  this will allow for systems to truely be quiet during boot (i.e., before showing a splash or before the init system begins being loud)

hilariously, udev did the migration the day after dracut added support! if systems were kept up to date with live branches, that means the original change would have only worked for one day.

## Checklist

Fixes af67d62151

### DIFF
--- a/modules.d/80base/init.sh
+++ b/modules.d/80base/init.sh
@@ -146,8 +146,11 @@ getargbool 0 rd.udev.log_level=info -d rd.udev.log-priority=info -d rd.udev.info
 getargbool 0 rd.udev.log_level=debug -d rd.udev.log-priority=debug -d rd.udev.debug -d -y rdudevdebug \
     && UDEV_LOG=debug
 
+mkdir -p /etc/udev
+echo "udev_log=${UDEV_LOG}" >> /etc/udev/udev.conf
+
 # start up udev and trigger cold plugs
-UDEV_LOG=$UDEV_LOG "$systemdutildir"/systemd-udevd --daemon --resolve-names=never
+"$systemdutildir"/systemd-udevd --daemon --resolve-names=never
 
 UDEV_QUEUE_EMPTY="udevadm settle --timeout=0"
 


modules.d/80base/init.sh
```diff
@@ -147,7 +147,7 @@ getargbool 0 rd.udev.log_level=debug -d rd.udev.log-priority=debug -d rd.udev.de
     && UDEV_LOG=debug
 
 # start up udev and trigger cold plugs
-UDEV_LOG=$UDEV_LOG "$systemdutildir"/systemd-udevd --daemon --resolve-names=never
+SYSTEMD_LOG_LEVEL=$UDEV_LOG "$systemdutildir"/systemd-udevd --daemon --resolve-names=never
```
Conan-Kudo: Do we need both variables for things like eudev?

Conan-Kudo (changes requested):

modules.d/80base/init.sh
turret-dev: No, eudev doesn't check any environment variables for log levels.  The old variant does not work at all for the current version of eudev, released in 2023.

If we wanted to support eudev, then we would have to set log level in udev.conf instead, which would actually set it for both implementations

turret-dev: Also, the command explicitly starts systemd-udevd.  I'm not sure if that is whatever version of udevd is in the host is copied or if it is only systemd's udevd.

aafeijoo-suse: > Also, the command explicitly starts systemd-udevd. I'm not sure if that is whatever version of udevd is in the host is copied or if it is only systemd's udevd.

Any udevd, see:

https://github.com/dracut-ng/dracut-ng/blob/d361c3f0cfca6ce65e648934447e0ae355b6e06b/modules.d/74udev-rules/module-setup.sh#L12-L25

jozzsi: > If we wanted to support eudev, then we would have to set log level in udev.conf instead, which would actually set it for both implementations

So maybe `/run/udev/udev.conf`, since eudev is also supported (and tested e.g. with Void and Alpine).

turret-dev: what is /run/udev/udev.conf? eudev uses /etc/udev/udev.conf by default and doesn't check anything else .. just checked (using latest commit) and /etc/udev/udev.conf works for systemd udev as well

jozzsi: Should this chunk of code be conditional - only write `udev.conf` if `UDEV_LOG` is set ?

turret-dev: > Should this chunk of code be conditional - only write `udev.conf` if `UDEV_LOG` is set ? 
> 

`UDEV_LOG` is always set. it is by default set to `err` which eliminates the systemd-udevd starting message

Conan-Kudo (approved):

PR #1781, opened 2025-10-27
fix(dmsquash-live): install required stat binary

### LINKED ISSUE #1780: command stat is missing in dependency
When adding rd.live.ram=1 If get this during the boot process

dmsquash-live-root: 327: stat: not found

Looking at the source, in line 327 the stat command is called, but It's missing in the dependencies of the dracut module and the stat executable is only`added by 86shutdown/module-setup.sh.

aafeijoo-suse:
Follow-up for 52351cfa049a9594d539e6a5337d591e8039ab80

## Checklist

Fixes #1780

### DIFF
--- a/modules.d/70dmsquash-live/module-setup.sh
+++ b/modules.d/70dmsquash-live/module-setup.sh
@@ -22,7 +22,7 @@ installkernel() {
 
 # called by dracut
 install() {
-    inst_multiple umount dmsetup blkid dd losetup blockdev find rmdir grep
+    inst_multiple umount dmsetup blkid dd losetup blockdev find rmdir grep stat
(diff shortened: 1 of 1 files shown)


aafeijoo-suse:
> > > ... and link the bug report?
> > 
> > 
> > Fixes #1780
> 
> Oh, I missed that.
> 
> I suggest to be a little bit more verbose. Suggestion:
> 
> > dmsquash-live-root.sh calls `stat` since [52351cf](https://github.com/dracut-ng/dracut-ng/commit/52351cfa049a9594d539e6a5337d591e8039ab80) ("feat(livenet): add memory size check depending on live image size")

ok, no problem.

PR #1783, opened 2025-10-28
test: support systemd in target rootfs

bdrung:
## Changes

The test-root module installs `test-init.sh` in `/sbin/test-init` and adds `testsuite.service` if systemd is used. When systemd is used in the target client/server rootfs, the systemd in the initrd hands over the control to the systemd in the rootfs. Then `testsuite.service` executes `test-init.sh` instead of the wanted client/server init scripts.

Add the helper function `inst_init` to override the init script from the test-root dracut module.

This is one step in the direction to make these tests work on Debian/Ubuntu.

## Checklist

Fixes #


Conan-Kudo:
This needs rebasing.

bdrung:
Rebased.

PR #1792, opened 2025-10-29
feat(plymouth): drop unused plymouth-populate-initrd

bdrung:
## Changes

The plymouth `check` function calls `pkglib_dir` which searches for an executable `plymouth-populate-initrd` script. Otherwise it will exit with a failure and plymouth will not be included. Therefore `plymouth-populate-initrd` will always be present when `install` is called. Thus the code path for calling `plymouth-populate-initrd.sh` will never taken.

So drop `plymouth-populate-initrd.sh` and its unused code path.

## Checklist

Fixes #


jozzsi:
CC @zboszor

PR #1798, opened 2025-10-30
test: determine rootfs size dynamically

bdrung:
## Changes

Do not hard-code the size of the rootfs, but determine the needed size dynamically. This will speed up the creation time for smaller content and prevent failures in case of bigger content.

## Checklist

Fixes #

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -179,7 +179,12 @@ build_ext4_image() {
     local image="$2"
     local label="$3"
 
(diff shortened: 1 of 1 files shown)


jozzsi:
This approach does not seem to work, seems CI failure is legit. I have not looked into why.

> This will speed up the creation time for smaller content 

This goal is maybe not worthwhile to pursue

> prevent failures in case of bigger content.

This on the other hand I think is a reoccurring problem

Conan-Kudo (approved):

PR #1804, opened 2025-10-31
feat(pcsc): add Debian/Ubuntu support

### LINKED ISSUE #1794: Dracut compatibility with Ubuntu and Debian
**Describe the bug**

Using a YubiKey (PKCS#11 via OpenSC/pcscd) to unlock an encrypted root at boot works on Fedora/RHEL, but fails on Debian/Ubuntu when initramfs is built with dracut(-ng). The 73pcsc module appears incomplete/out-of-sync across distros:

It references opensc.module instead of the actual `opensc-pkcs11.module` provided by Debian/Ubuntu (p11-kit).

It does not install all required OpenSC-related libraries (e.g., libeac.so*), which leads to missing symbols at boot.

It only considers `/etc/opensc.conf` (Fedora/RedHat layout) and not `/etc/opensc/opensc.conf` (Debian/Ubuntu layout).

**Distribution used**

Reproduced on (same machine and token; Fedora works, Debian/Ubuntu fail):

- Ubuntu 25.10 LTS (questing)

Works with:

Fedora42 same OpenSC/YubiKey setup succeeds

Dracut version 108

Init system
systemd (pcscd as systemd service/socket in the initramfs)

To Reproduce

On Debian/Ubuntu:

Install OpenSC, pcsc-lite, p11-kit modules; enroll a LUKS key with PKCS#11 (YubiKey).

Build initramfs with dracut(-ng) including PKCS#11 support and the 73pcsc module (and crypt/systemd modules as usual).

Ensure the token is present at boot.

Boot the system.

Observe:

The YubiKey is visible under `/dev/bus/usb/...`, but systemd-cryptsetup cannot use PKCS#11.

pcsc/opensc in the initramfs cannot load all dependencies (missing config and/or libs).

Fallback to the passphrase unlock works.

Expected behavior
The PKCS#11 unlock flow should work on Debian/Ubuntu the same way it works on Fedora/RHEL: 
pcscd starts in the initramfs, OpenSC/p11-kit are correctly configured, required libs are present, and systemd-cryptsetup can use the token to unlock without needing the password fallback.

Additional context
We compared Fedora/RHEL (working) vs Debian/Ubuntu (failing) and found concrete issues in module `73pcsc` (`/usr/lib/dracut/modules.d/73pcsc/module-setup.sh`). 
Highlights from our notes (line numbers from our local copy):

L33: p11-kit module filename mismatch on Debian/Ubuntu
Current code expects:

`/usr/share/p11-kit/modules/opensc.module`

On Debian/Ubuntu, the file is:

`/usr/share/p11-kit/modules/opensc-pkcs11.module`

Proposal: use a wildcard to be robust across distros.

L56/L57: Missing libraries for OpenSC (e.g., libeac.so*)

We had to add:
```
{"tls/$_arch/",tls/,"$_arch/",}"libeac.so*"
{"tls/$_arch/",tls/,"$_arch/",}"libcrypto.so*" (Not sur it's really required)
```

Maybe: dracut-install (used by inst_*) could pull ldd dependencies recursively. If a dependency is weak/optional, it may still be skipped; in practice, **explicitly including libeac.so* fixed our boot**. 

Guidance on the preferred upstream approach (explicit vs relying on recursive dep resolution) would be appreciated.

L62: OpenSC configuration path differs across distros
Fedora/RHEL use:
`/etc/opensc.conf`

Debian/Ubuntu use:
`/etc/opensc/opensc.conf`

Proposal: try both paths (install whichever exists on the host).

L12: return 255 in check() we don't know if 0 or 255 needed ?

Below is a minimal illustration of the changes we used to get Debian/Ubuntu working (for discussion, not a final patch):

```
diff --git a/modules.d/73pcsc/module-setup.sh b/modules.d/73pcsc/module-setup.sh
--- a/modules.d/73pcsc/module-setup.sh
+++ b/modules.d/73pcsc/module-setup.sh
@@ -30,7 +30,7 @@ install() {
     inst_multiple -o \
         pcscd \
-        /usr/share/p11-kit/modules/opensc.module
+        /usr/share/p11-kit/modules/opensc*.module

     # Enable systemd type unit(s)
     for i in \
@@ -53,8 +53,10 @@ install() {
         {"tls/$_arch/",tls/,"$_arch/",}"libpcsclite.so.*" \
         {"tls/$_arch/",tls/,"$_arch/",}"libpcsclite_real.so.*" \
+        {"tls/$_arch/",tls/,"$_arch/",}"libeac.so*" \
+        {"tls/$_arch/",tls/,"$_arch/",}"libcrypto.so*"
 
     if [[ ${hostonly-} ]]; then
         inst_multiple -H -o \
-            /etc/opensc.conf \
+            /etc/opensc.conf \
+            /etc/opensc/opensc.conf \
             "/etc/reader.conf.d/*"
     fi
```

Thanks! We’re happy to iterate based on your guidance.

jozzsi:
## Changes

On Debian/Ubuntu, the filename is `opensc-pkcs11.module` instead of `opensc.module`.

On Debian/Ubuntu, missing `libeac.so` libraries for OpenSC.

On Debian/Ubuntu the OpenSC configuration path is `/etc/opensc/opensc.conf` .

## Checklist

Fixes #1794

### DIFF
--- a/modules.d/73pcsc/module-setup.sh
+++ b/modules.d/73pcsc/module-setup.sh
@@ -30,7 +30,8 @@ install() {
 
     inst_multiple -o \
         pcscd \
-        /usr/share/p11-kit/modules/opensc.module
+        /usr/share/p11-kit/modules/opensc.module \
+        /usr/share/p11-kit/modules/opensc-pkcs11.module
 
     # Enable systemd type unit(s)
     for i in \
@@ -51,6 +52,7 @@ install() {
         {"tls/$_arch/",tls/,"$_arch/",}"pcsc/drivers/ifd-ccid.bundle/Contents/Info.plist" \
         {"tls/$_arch/",tls/,"$_arch/",}"pcsc/drivers/ifd-ccid.bundle/Contents/Linux/libccid.so" \
         {"tls/$_arch/",tls/,"$_arch/",}"pcsc/drivers/serial/libccidtwin.so" \
+        {"tls/$_arch/",tls/,"$_arch/",}"libeac.so*" \
         {"tls/$_arch/",tls/,"$_arch/",}"libpcsclite.so.*" \
         {"tls/$_arch/",tls/,"$_arch/",}"libpcsclite_real.so.*"
 
@@ -58,6 +60,7 @@ install() {
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
             /etc/opensc.conf \
+            /etc/opensc/opensc.conf \
             "/etc/reader.conf.d/*"
     fi
 


bdrung (approved):
LGTM. There is an additional space after `/etc/opensc/opensc.conf` in the git description.

Checked on Ubuntu:

* The opensc package depends on libeac3. So including this library makes sense.
* The opensc package ships /etc/opensc/opensc.conf
* The opensc package depends on opensc-pkcs11. So including /usr/share/p11-kit/modules/opensc-pkcs11.module makes sense.

Conan-Kudo (approved):

PR #1809, opened 2025-10-31
fix: line-buffer log output to /dev/kmsg

### LINKED ISSUE #1803: awk: cmd. line:1: (FILENAME=- FNR=1) fatal: not enough arguments to satisfy format string
**Describe the bug**

```
dracut[E]: No '/dev/log' or 'logger' included for syslog logging
awk: cmd. line:1: (FILENAME=- FNR=1) fatal: not enough arguments to satisfy format string
	`<%s>15'
	  ^ ran out for this one
awk: cmd. line:1: (FILENAME=- FNR=1) fatal: not enough arguments to satisfy format string
	`<%s>15'
	  ^ ran out for this one
Not removing /var/tmp/dracut-test.AU94u3/dracut.dVzL7KC.
```

@HastD That looks like a regression from https://github.com/dracut-ng/dracut-ng/pull/1678.

**Distribution used**
Fedora

**Dracut version**
main

**Additional context**
Found in https://github.com/dracut-ng/dracut-ng/actions/runs/18969852664/job/54175109178?pr=1763

HastD:
/dev/kmsg is only designed to accept single lines at a time, and therefore has a relatively small message size limit. This means we need to line-buffer the output of awk (and prepend the message level prefix to each line rather than just the first).

(This was missed in local testing because it only becomes an issue for sufficiently large inputs. I tested this with larger inputs and it works properly with the line-buffering.)

Fixes #1803 properly—the previous PR fixed the immediate issue but still ran into problems because of the lack of line-buffering.

### DIFF
--- a/dracut-logger.sh
+++ b/dracut-logger.sh
@@ -436,8 +436,9 @@ _do_dlog_batch_kmsglog() {
     local kmsg_lvlc
     if ((lvl <= kmsgloglvl)); then
         kmsg_lvlc="$(_dlvl2syslvl "$lvl")" || return 1
-        awk -v lvlc="${kmsg_lvlc}" -v prefix="dracut[$$] " \
-            '{ if (NR==1) printf "<%s>", lvlc; print prefix $0 }' > /dev/kmsg
+        # /dev/kmsg is designed to only accept single lines at a time, so the
+        # output of awk has to be line-buffered.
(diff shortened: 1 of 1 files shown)


HastD:
Note, the issue was visible in the output of some integration tests, such as test 43 on fedora:latest, for #1806, but it didn't cause a CI failure because logging failures intentionally aren't fatal. We should double-check the output of some tests to make sure there's no "Invalid argument" messages in the logs, which is how overly long inputs to /dev/kmsg get rejected.

Conan-Kudo (approved):

PR #1812, opened 2025-10-31
feat(install.d): skip if initrd_generator is unset

bdrung:
## Changes

The Dracut 12-uefi and 43-kernel-install test cases fail on Debian/Ubuntu. dracut-core ships
`/usr/lib/kernel/install.d/50-dracut.install` which generates an initrd in `KERNEL_INSTALL_STAGING_AREA`.
`/usr/lib/kernel/install.d/55-initrd.install` (from the systemd package) then links to the initrd in /boot. This results in two initrd being included in the UKI and the content of later one overwrites the former one.

`/usr/lib/kernel/install.d/55-initrd.install` should only act if there is no initrd generator configured, but it needs to know when dracut is the initrd generator.

Be more explicit by changing the kernel-install hooks to only act if the initrd generator is set to dracut. But also ship a configuration snippet that sets the initrd generator to dracut.

Bug-Debian: https://bugs.debian.org/1117563

## Checklist
- [x] I have tested it locally in combination with https://salsa.debian.org/systemd-team/systemd/-/merge_requests/295

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -269,6 +269,8 @@ endif
 ifeq ($(enable_dracut_cpio),yes)
 	install -m 0755 dracut-cpio $(DESTDIR)$(pkglibdir)/dracut-cpio
 endif
+	mkdir -p $(DESTDIR)${prefix}/lib/kernel/install.conf.d
+	install -m 0644 install.conf.d/50-dracut.conf $(DESTDIR)${prefix}/lib/kernel/install.conf.d/50-dracut.conf
 	mkdir -p $(DESTDIR)${prefix}/lib/kernel/install.d
 	install -m 0755 install.d/50-dracut.install $(DESTDIR)${prefix}/lib/kernel/install.d/50-dracut.install
 	install -m 0755 install.d/51-dracut-rescue.install $(DESTDIR)${prefix}/lib/kernel/install.d/51-dracut-rescue.install
--- a/install.conf.d/50-dracut.conf
+++ b/install.conf.d/50-dracut.conf
@@ -1,0 +1,1 @@
+initrd_generator=dracut
--- a/install.d/50-dracut.install
+++ b/install.d/50-dracut.install
@@ -16,7 +16,7 @@ if [[ "$KERNEL_INSTALL_IMAGE_TYPE" = "uki" ]]; then
     exit 0
 fi
 
-if [[ "${KERNEL_INSTALL_INITRD_GENERATOR:-dracut}" = "dracut" ]]; then
+if [[ "${KERNEL_INSTALL_INITRD_GENERATOR-}" = "dracut" ]]; then
     # We are the initrd generator
     IMAGE="initrd"
     UEFI_OPTS="--no-uefi"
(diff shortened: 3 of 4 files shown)


bdrung:
And I found this in https://github.com/systemd/mkosi/blob/main/mkosi.conf:
```
RemoveFiles=
[...]
        # The dracut install plugin doesn't honor KERNEL_INSTALL_INITRD_GENERATOR.
        /usr/lib/kernel/install.d/50-dracut.install
```

Conan-Kudo:
So honoring the flag seems to be the correct fix, not making it so the flag is mandatory.

Conan-Kudo:
> > Where does this file come from? It doesn't exist in other distros' systemd/systemd-udev packages. And I didn't see it in the upstream sources. If this is a Debian-specific file, then it needs to be fixed.
> 
> Yes. It is Debian-specific: https://salsa.debian.org/systemd-team/systemd/-/blob/debian/master/debian/extra/kernel-install.d/55-initrd.install
> 
> This hook is silenced when `KERNEL_INSTALL_INITRD_GENERATOR` is set. How else should it behave in case we do not use this PR?

Your hook is wrong. It should not fire unless something configures to enable it. You've basically made it do the wrong thing by default.

Nowa-Ammerlaan:
In Gentoo I have worked around this by doing two things:
- I added a hook that enforces layout, initrd_generator, and uki_generator to be set, else it exits and complains to the user.
- I added "none" as a recognised setting for folks who explicitly do not want an initrd/uki generator.

It is completely unclear what should happen if the setting is empty, as is also demonstrated by the discussion here. Hence it is undefined behaviour. I do not like having our users systems rely on undefined behaviour because it may randomly change and break, hence I opted to explicitly forbid this. The default for the initrd/uki generator setting is thus "none" instead of empty. I made that change a long time ago after I found that many users were confused when they found that dracut was always being run when it was installed even if it was not actually enabled via install.conf.

dalto8:
It doesn't seem like sensible behavior for dracut by default to require setting a variable to use dracut.  In addition to not liking it from a design perspective, I don't like the fact that we are potentially creating a situation where no initrd at all would be created.

Having multiple initrd generators installed is an edge case to begin with.  That being said, if a user does want this for some reason, systemd/kernel-install already provides a mechanism to inhibit the running of an install script via a drop-in.

dalto8:
> Can you point to documentation for that?

It is detailed in the man page for kernel-install

> kernel-install will run the executable files ("plugins") located in the directory /usr/lib/kernel/install.d/ and the local administration directory /etc/kernel/install.d/. All files are collectively sorted and executed in lexical order, regardless of the directory in which they live. However, files with identical filenames replace each other. Files in /etc/kernel/install.d/ take precedence over files with the same name in /usr/lib/kernel/install.d/. This can be used to override a system-supplied executables with a local file if needed; a symbolic link in /etc/kernel/install.d/ with the same name as an executable in /usr/lib/kernel/install.d/, pointing to /dev/null, disables the executable entirely. Executables must have the extension ".install"; other extensions are ignored.

Nowa-Ammerlaan:
> > ```
> > * I added "none" as a recognised setting for folks who explicitly do not want an initrd/uki generator.
> > ```
> 
> Can we have this upstreamed?

It's something I do in the ebuild, not sure if/how this could be upstreamed. We would probably have to patch kernel-install itself.

-> https://github.com/projg2/installkernel-gentoo/blob/d446bcb43c6ed1bab39a8d2f7ad752f65be4e4fe/installkernel-9999.ebuild#L151

PR #1817, opened 2025-11-01
test: do not use dsync oflag for dd calls

jozzsi:
## Changes

`dsync` `oflag` is not supported by all dd implementation. As an example `busybox` does not support `dsync`.

## Checklist

Fixes #

### DIFF
--- a/test/TEST-11-USR-MOUNT/create-root.sh
+++ b/test/TEST-11-USR-MOUNT/create-root.sh
@@ -15,5 +15,6 @@ btrfs filesystem sync /root
 # read-only snapshot of /root
 btrfs subvolume snapshot -r /root /root/snapshot-root
 umount /root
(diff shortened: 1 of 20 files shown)


Conan-Kudo:
Is this tracked by busybox as intentional or a bug?

jozzsi:
> Is this tracked by busybox as intentional or a bug?

Well, we aim to support current busybox (even if busybox will support this eventually). As far as I can tell, dsync is only supported by the coreutils implementation, so at this point I would call it a coreutils specific extension.

This is a test-suite only change - small price to pay given the benefit

Conan-Kudo (approved):

bdrung (approved):
LGTM. No `dsync` afterwards and there are `sync` calls for all.

This PR is to support busybox. The rust coretutils dd supports dsync.

PR #1820, opened 2025-11-01
fix(multipath): always forcefully load multipath modules

jozzsi:
## Changes

cmdline() function gets called both in hostonly and non-hostonly mode. Make sure that non-hostonly mode is handled properly, so that cmdline does not get added.

## Checklist

CC @mwilck

### DIFF
--- a/modules.d/70multipath/module-setup.sh
+++ b/modules.d/70multipath/module-setup.sh
@@ -42,9 +42,7 @@ depends() {
 # called by dracut
 cmdline() {
     for m in scsi_dh_alua scsi_dh_emc scsi_dh_rdac dm_multipath; do
-        if grep -m 1 -q "$m" /proc/modules; then
-            printf 'rd.driver.pre=%s ' "$m"
-        fi
+        printf 'rd.driver.pre=%s ' "$m"
     done
 }
 


Conan-Kudo (dismissed):

modules.d/70multipath/module-setup.sh
```diff
@@ -41,11 +41,13 @@ depends() {
 
 # called by dracut
 cmdline() {
-    for m in scsi_dh_alua scsi_dh_emc scsi_dh_rdac dm_multipath; do
-        if grep -m 1 -q "$m" /proc/modules; then
-            printf 'rd.driver.pre=%s ' "$m"
-        fi
-    done
+    if [[ $hostonly ]]; then
```
bdrung: I'd rather see the hostonly check done on the caller side.

Maybe even dracut unset hostonly_cmdline if not running in hostonly mode?

bdrung (dismissed):

jozzsi:
> If dm_multipath is not loaded, multipathd won't start. 

For now I removed the /proc/modules checking and always forcefully load. This might have it's own issues, but I hope it helps progressing the discussion.

> Upstream multipathd.service contains

I was under the impression that multipath dracut module works without systemd as well. Are we changing this direction  - if so than perhaps multipath dracut module should now depend on systemd dracut module.

jozzsi:
> After all, people who create non-hostonly initrds can omit the multipath module if they'd rather not have it.

We should aim higher. Defaults are important. For a built-in dracut module that is installed by default (return 0 on check()), it should either explicitly omit itself in non-hostonly or it should handle non-hostonly properly.

If I have a non-hostonly config and just install multipath distro package and my boot start failing because I did not know that after installing multipath now I need to manually disable multipath dracut module is very error prone.

That said, would you be able to provide feedback directly on the last uploaded version of the PR - as that might handle non-hostonly better already.

modules.d/70multipath/module-setup.sh
```diff
@@ -42,9 +42,7 @@ depends() {
 # called by dracut
 cmdline() {
     for m in scsi_dh_alua scsi_dh_emc scsi_dh_rdac dm_multipath; do
-        if grep -m 1 -q "$m" /proc/modules; then
-            printf 'rd.driver.pre=%s ' "$m"
-        fi
+        printf 'rd.driver.pre=%s ' "$m"
```
mwilck, line 45: I wonder if this will work on distros which don't provide these modules. But I guess such distros won't use the multipath module anyway.

mwilck:
> @mwilck for a non-hostonly system with no /etc/multipath.conf, won't dracut generate a default multipath.conf set to strict (at least on distros that don't include mpathconf)?

Right. I forgot my own patch :-(

I also forgot what kind of configuration `rd.multipath=default` would generate.

Looking at 1e802f1 again, on systems without `mpathconf`, the multipath module in a non-hostonly initrd will be a noop most of the time. The commit message talks about `/etc/multipath/wwids` but that file doesn't belong in non-hosthonly initrd.

I know I'm contradicting my own previous statement here. But Given that multipath configurations are relatively rare, and will typically neeed hostonly initrds, I wonder how much sense it makes to include the multipath module in the non-hostonly initrd by default.

bmarzins:
> I know I'm contradicting my own previous statement here. But Given that multipath configurations are relatively rare, and will typically neeed hostonly initrds, I wonder how much sense it makes to include the multipath module in the non-hostonly initrd by default.

I actually agree with your reasoning in https://github.com/dracut-ng/dracut-ng/commit/1e802f15fee3d6402d83e0efefef8bb88c5a33e3 . I think it makes sense to include these binaries in non-hostonly configs, since IIUC they are supposed to include what people could reasonably need. As an aside, I have a patch that's been NAK'ed for upstream multipath to make it so that users can add mpath.wwid=<WWID> to the kernel commandline, and those WWIDs will get added to /etc/multipath/wwids before multipathd starts up. I believe your objection was that multipath shouldn't be parsing /proc/cmdline, which is fair. But, assuming you don't have an objection to allowing users to specify mulitpath WWIDs on the kernel commandline in principle, that ability (coded differently) would let users to boot from multipath using a non-hostonly initramfs with a `find_multipaths strict` default config. They would just need to know the device wwid(s). Not totally sure how generally useful that is. But it's kinda neat to be able to set up multipathed boot while using a stock initramfs.

jozzsi:
There is a lot of good discussion here, that we need to break out to separate issues or PR - so for that reason I wanted to scope my PR and explain what I was hoping to achieve..

non-hostonly (generic) initramfs suppose to be independent of changes under /etc and /proc - meaning non-hostonly initramfs should be reproducible in a sense that given a /usr tree it should always result in the same initramfs. Right now it is not, because -e.g. cmdline() reads /proc even in non-hostonly mode.

How to make non-hostonly function properly is not in scope for this PR. I simply wanted to make at least non-hostonly multipath dracut module reproducible.

modules.d/70multipath/module-setup.sh
jozzsi: I run the following experiment

Modified BASIC test
```
--- a/test/TEST-10-BASIC/test.sh
+++ b/test/TEST-10-BASIC/test.sh
@@ -14,7 +14,7 @@ test_run() {
 
     "$testdir"/run-qemu -nic none \
         "${disk_args[@]}" \
-        -append "$TEST_KERNEL_CMDLINE" \
+        -append "$TEST_KERNEL_CMDLINE rd.driver.pre=askfjdh" \
         -initrd "$TESTDIR"/initramfs.testing
 
```

Test still run - and the nonexistent 'askfjdh' seems to be just ignored. This suggest that even if the kernel module does not exists (or built into the kernel), it would not introduce additional errors and the additional safety checks are not needed.

That said, it would be really useful for someone else to test as well, especially as we do not have test for multipath in the CI - we're only testing that multipath dracut module can be built and does not interfere with other dracut modules.

modules.d/70multipath/module-setup.sh
jozzsi: hostonly check (and other checks) are now removed. Doing this in hostonly mode only would break non-hostonly use-case.

bmarzins:
> Another, possibly even more useful, option would be to use find_multipath smart in non-hostonly mode.

Where Red Hat needed the ability to force immediate multipathing via the kernel commandline was most often on install, to keep anaconda from auto-assembling a LVM/MD storage stack that was already configured on the devices, which was making it impossible to set up a multipath device when the second path appeared. We've since switched to solving this by using `find_multipaths smart`, which is a much better solution for that problem. But potentially wasting 10 seconds on install seems like a much smaller deal than potentially wasting 10 seconds on every boot with a non-hostonly initramfs. If non-hostonly intramfs's have the multipath module by default (which I do think makes sense), but most of them are used in non-multipath environments, I can see a 10 second delay on every boot with them being annoying to most people.  You're right that it means that non-hostonly initramfs's won't automatically work correctly for people with multipathed systems. But if they could fix that by just adding a kernel commandline parameter, that seems like a reasonable trade off to avoid the delay of `find_multipaths smart`.

mwilck:
> Re. find_multipath smart, I'm wary of anything that hinges on some duration for devices to appear before taking action. I wouldn't want to bake that into any of our non-hostonly initrds.

Unfortunately it's the only way in which multipathd can behave ina reasonable way in a non-hostonly initrd. The alternatives are to either to grab nothing (which makes the module useless) or everything (which would probably annoy many users).

Arguably most users of non-hostonly initrds will not be multipath users. Requiring them all to either wait 10s or pass `multipath=off` on the kernel command line would be another annoyance.

What about introducing a new parameter `rd.multipath` and disabling multipath in the (non-hostonoly) initrd unless this parameter is set? This way not activating multipath would be the default, but people who want to use it can activate it easily.

mwilck:
Note also that `find_multipaths smart` does **not** wait 10s on typical SATA disks in users' PCs. It only waits for devices that match SCSI vendor/model combinations that show up in multipath's internal hardware table, i.e. storage arrays that are designed for multipath support. For all others, the wait time is limited to 1s, which should be hardly noticeable.

jlebon:
> What about introducing a new parameter `rd.multipath` and disabling multipath in the (non-hostonoly) initrd unless this parameter is set? This way not activating multipath would be the default, but people who want to use it can activate it easily.

This is already the case in our initramfs at least. Since `multipathd.service` has `ConditionPathExists=/etc/multipath.conf` and we don't have an `/etc/multipath.conf` in our non-hostonly initramfs, it stays off by default. If users want it, they would do `rd.multipath=default`. We also have initramfs code gated on that karg that explicitly waits until the boot disk is multipathed to be sure that multipathd took ownership before continuing the boot.

This is why `smart` doesn't buy us much. The user already has to explicitly opt into multipath, and if they did so, we're fine waiting however long for device assembly.

mwilck (approved):
ok

LaszloGombos (approved):

Conan-Kudo (approved):

PR #1821, opened 2025-11-01
fix(dracut-init): do not set DRACUT_KERNEL_MODALIASES for --no-kernel

### LINKED ISSUE #691: Better handling for monolithic kernels
dracut can generate an initrd for a monolithic kernel (`CONFIG_MODULES=n`) but not a well-optimized one.  Problems I have noticed so far include:

* --host-only mode cannot be used; it sees that /proc/modules does not exist, throws up its hands, and refuses to continue
* the dracut modules dracut-systemd, systemd-initrd, systemd-udevd, and systemd-sysctl, all of which *should* work just fine, cannot be used because they have a transitive hard dependency on systemd-modules-load (which should be unnecessary with a monolithic kernel)
* the initrd includes kmod and all its aliases even though these are never useful with a monolithic kernel (the only reason they are even available to be copied into the initrd is that depmod *is* still useful).

I expect I will find more problems as I continue testing; I will follow up to this issue as necessary.

Ideally, when dracut is invoked on a system with a monolithic kernel, the only consequence would be that everything related to kernel modules is left out of the initrd; all the other components should work just the same.

jozzsi:
## Changes

When dracut is invoked with `--no-kernel`, do not set `DRACUT_KERNEL_MODALIASES` to avoid using kmod.

## Checklist

Fixes #691

### DIFF
--- a/dracut-init.sh
+++ b/dracut-init.sh
@@ -217,7 +217,7 @@ dracut_module_path() {
     return 1
 }
 
-if [[ ${hostonly-} == "-h" ]]; then
+if [[ ${hostonly-} == "-h" ]] && [[ $no_kernel != yes ]]; then
     if ! [[ $DRACUT_KERNEL_MODALIASES ]] || ! [[ -f $DRACUT_KERNEL_MODALIASES ]]; then
         export DRACUT_KERNEL_MODALIASES="${DRACUT_TMPDIR}/modaliases"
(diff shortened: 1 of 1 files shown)


mutedbytes:
Hi @jozzsi yes this seems to work as well. If --no-kernel is set with hostonly, it builds as expected. If --no-kernel is removed, then I get the reversion to non-hostonly.
Thanks.

bdrung (approved):

LaszloGombos (approved):

PR #1822, opened 2025-11-01
feat: rename rd.live.overlay.overlayfs to rd.overlayfs

jozzsi:
## Changes

The overlayfs dracut module can be used without the dmsquash-live dracut module. This change improves on naming of command line options with better namespace scoping.
    
Move overlayfs documentation out of the live images section.

Follow-up to 40dd5c90e.

## Checklist

### DIFF
--- a/man/dracut.cmdline.7.adoc
+++ b/man/dracut.cmdline.7.adoc
@@ -1086,6 +1086,29 @@ rd.znet=qeth,0.0.0600,0.0.0601,0.0.0602,layer2=1,portname=foo
 rd.znet=ctc,0.0.0600,0.0.0601,protocol=bar
 --
 
+OverlayFS
+~~~~~~~~~
+Requires the dracut 'overlayfs' module.
+
+**rd.overlayfs=**1::
+Enables the use of the *OverlayFS* kernel module, if available, to provide a
+copy-on-write union directory for the root filesystem.  OverlayFS overlays are
+directories of the files that have changed on the read-only base (lower)
+filesystem.  The root filesystem is provided through a special overlay type
+mount that merges at least two directories, designated the lower and the upper.
+If an OverlayFS upper directory is not present on the boot device, a tmpfs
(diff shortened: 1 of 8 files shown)


man/dracut.cmdline.7.adoc
```diff
@@ -1086,6 +1086,29 @@ rd.znet=qeth,0.0.0600,0.0.0601,0.0.0602,layer2=1,portname=foo
 rd.znet=ctc,0.0.0600,0.0.0601,protocol=bar
 --
 
+OverlayFS
+~~~~~~~~~
+Requires the dracut 'overlayfs' module.
+
+**rd.live.overlay.overlayfs=**1::
```
Conan-Kudo: We probably would want something that isn't using `rd.live` namespace for non-livefs cases.

Conan-Kudo (changes requested):

man/dracut.cmdline.7.adoc
jozzsi: We could keep `live` for backward compat and go with `rd.overlay.overlayfs` - this would need code changes as well, not just documentation changes. 

I find `live` very non-descriptive (as oppose to being dead ? ) so while 'live` does not add much, I also do not find it particularly confusing.

One of the use case here is overlay with read only nfs mount root - is that "non-live" ?

Conan-Kudo: Yeah, that's the non-live case. the `rd.live` thing is intended to be a standardized prefix for liveOS environments. Adding `rd.overlayfs` prefix for the overlayfs stuff independent of liveOS stuff makes sense.

jozzsi: ok, made the changes and kept backwards compatibility with the old naming.

jozzsi:
CC @FGrose

Conan-Kudo:
Ah nevermind, I misread the diff. This is fine.

Conan-Kudo (approved):

LaszloGombos (approved):

PR #1823, opened 2025-11-02
feat(dracut): set no_kernel to "yes" for CONFIG_MODULES=n

jozzsi:
## Changes

For a monolithic kernel (CONFIG_MODULES=n), when no_kernel is not set, initialize no_kernel to "yes".

This change further improve usability for monolithic kernel users.

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1419,6 +1419,12 @@ esac
 hostonly=${hostonly-}
 hostonly_mode=${hostonly_mode-}
 
+# when no_kernel is not specified in hostonly mode and /proc/modules does not exists,
+# set no_kernel to yes
+if [[ $no_kernel ]] && [[ $hostonly ]] && ! [[ -e /proc/modules ]]; then
+    no_kernel="yes"
+fi
+
 if [[ $reproducible == yes ]] && [[ -z ${SOURCE_DATE_EPOCH-} ]]; then
     SOURCE_DATE_EPOCH=$(stat -c %Y "$dracutbasedir/dracut-functions.sh")
     export SOURCE_DATE_EPOCH


dracut.sh
```diff
@@ -1375,6 +1375,10 @@ esac
 hostonly=${hostonly-}
 hostonly_mode=${hostonly_mode-}
 
+# when no_kernel is not specified in hostonly mode and /proc/modules does not exists,
+# set no_kernel to yes
+[[ $no_kernel ]] && [[ $hostonly ]] && ! [[ -e /proc/modules ]] && no_kernel="yes"
```
bdrung: suggestion: I prefer to use if-conditions for such complex cases to make the code easier to read.

jozzsi: ok

Hawk777:
I don’t think I can properly review this on its own in detail, while avoiding being confounded by how my distro embeds it in `installkernel` and the changes from 106 (which is the current stable version) and `main`. I will note I was already using `no_kernel`. I will also note that, even with this patch, this text still appears in an invocation:

```
realpath: /lib/modules/6.12.41-gentoo: No such file or directory
libkmod: ERROR: kmod_module_new_from_loaded: could not open /proc/modules: No such file or directory
dracut-install: Could not get list of loaded modules: Unknown error -2. Switching to non-hostonly mode.
```

So it does not seem to have fixed that problem.

mutedbytes:
@Hawk777 were you taking into account #1821 that I think this is in combination with/in support of?

Hawk777:
Ah, I didn’t see that patch. That makes more sense.

devkontrol:
Let's try to reopen this discussion on this small PR that is still relevant. 

@jozzsi please us know if you able to support this work.

@mutedbytes @Hawk777  - I assume you still think this PR looks good to land, right ?

devkontrol (approved):
Not sure if hostonly mode check is needed here - perhaps that can be removed later as a follow-up PR after careful consideration.

bdrung:
Shouldn't we check for `CONFIG_MODULES=n`?

devkontrol:
> Shouldn't we check for `CONFIG_MODULES=n`?

@jozzsi WDYT ? 

CC @Nowa-Ammerlaan @chewi for familiarity with Gentoo cases.

Nowa-Ammerlaan:
/proc/modules not existing indicates that the current kernel is module less, it says nothing about the target kernel that an initramfs is being generated for. 

This will make it harder to generate an initramfs for a kernel with modules when the current one is built without.

chewi:
> /proc/modules not existing indicates that the current kernel is module less, it says nothing about the target kernel that an initramfs is being generated for.
> 
> This will make it harder to generate an initramfs for a kernel with modules when the current one is built without.

Agreed. Also consider cross.

bdrung:
And `/proc` might not be mounted.

PR #1824, opened 2025-11-05
feat(pcsc): add openSUSE support

aafeijoo-suse:
The openSUSE version of the opensc package installs the module configuration in `/etc/pkcs11/modules/opensc.module`

## Checklist


Conan-Kudo (approved):

bdrung (approved):
The line between configuration and system gets blurry. I'll approve under the assumption that `/etc/pkcs11/modules/opensc.module` is also needed when building a generic initrd (and thus file not being host-only specific).

PR #1828, opened 2025-11-05
fix(plymouth): avoid warning if /etc/plymouth/plymouthd.conf is not present

aafeijoo-suse:
openSUSE does not install `/etc/plymouth/plymouthd.conf`, so plymouthd falls back to `/usr/share/plymouth/plymouthd.defaults`.

```
$ dracut -f
sed: can't read /etc/plymouth/plymouthd.conf: No such file or directory
```

Follow-up for 7f41458b9d8350d08b0057e050d98b7a25c3e640

## Checklist

### DIFF
--- a/modules.d/45plymouth/module-setup.sh
+++ b/modules.d/45plymouth/module-setup.sh
@@ -12,7 +12,9 @@ pkglib_dir() {
 read_setting_from_config() {
     local setting="$1"
     local config_file="$2"
-    sed -n "s/^${setting}= *\([^ ]\+\) */\1/p" "$config_file"
+    if [[ -f $config_file ]]; then
+        sed -n "s/^${setting}= *\([^ ]\+\) */\1/p" "$config_file"
+    fi
 }
 
 use_simpledrm() {


modules.d/45plymouth/module-setup.sh
```diff
@@ -12,7 +12,8 @@ pkglib_dir() {
 read_setting_from_config() {
     local setting="$1"
     local config_file="$2"
-    sed -n "s/^${setting}= *\([^ ]\+\) */\1/p" "$config_file"
+    [[ -f $config_file ]] \
+        && sed -n "s/^${setting}= *\([^ ]\+\) */\1/p" "$config_file"
```
bdrung: For better readability I suggest to use if-then-else:

```suggestion
    if [[ -f $config_file ]]; then
        sed -n "s/^${setting}= *\([^ ]\+\) */\1/p" "$config_file"
    fi
```

aafeijoo-suse: ok

bdrung (approved):

LaszloGombos (approved):

PR #1836, opened 2025-11-07
feat(systemd): install new dlopened libraries

aafeijoo-suse:
Preparation for systemd-v259. In chronological order:

- libaudit: https://github.com/systemd/systemd/commit/4d8c5c657ae0829f93944a00302e7ce700913e54
- ~~libpam: https://github.com/systemd/systemd/commit/882c9ce0402ec6e37201628a9a361500ff39b1ed~~
- libacl: https://github.com/systemd/systemd/commit/7c3a7f925f83bd05a49e8b1f09726cccc26977f7
- libblkid: https://github.com/systemd/systemd/commit/c349edfe49dc2c4b8a79e5d08ecf7c8e93c4c909
- libseccomp: https://github.com/systemd/systemd/commit/aaca6bd5d97258f29c1b8c991e8617c7272308e0
- libmount: https://github.com/systemd/systemd/commit/b3243f4beead231e27a4f017f53288a303177cb2
- libselinux: https://github.com/systemd/systemd/commit/83b6ef9b62765b11bc602eae906ff13a5464a638

## Checklist

### DIFF
--- a/modules.d/10systemd/module-setup.sh
+++ b/modules.d/10systemd/module-setup.sh
@@ -16,7 +16,7 @@ check() {
 
 # Config adjustments before installing anything.
 config() {
-    add_dlopen_features+=" libsystemd-shared-*.so:kmod "
+    add_dlopen_features+=" libsystemd-shared-*.so:audit,kmod,mount,seccomp,selinux "
 }
 
 installkernel() {
--- a/modules.d/11systemd-tmpfiles/module-setup.sh
+++ b/modules.d/11systemd-tmpfiles/module-setup.sh
@@ -11,6 +11,11 @@ check() {
     return 255
 }
 
+# Config adjustments before installing anything.
+config() {
+    add_dlopen_features+=" libsystemd-shared-*.so:acl "
+}
+
 # Install the required file(s) and directories for the module in the initramfs.
 install() {
     # Excluding "$tmpfilesdir/home.conf", sets up /home /srv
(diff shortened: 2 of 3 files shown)


aafeijoo-suse:
Actually, `libpam` is the only one that also raises doubts for me, but since it was always present in the initrd, I set it here.

I'd say that the others are needed:

- `libaudit` is in the core to create audit events.
- `libacl` is called by tmpfiles with ACLs defined.
- `libblkid` is called in the udev builtin in udev rules.
- Without `libseccomp` I was getting a warning:

```
Nov 06 07:06:58 tw systemd[1]: /usr/lib/systemd/system/systemd-udevd.service:56: System call bpf cannot be resolved as libseccomp is not available, ignoring: Operation not supported
```
- `libmount` seems clear.
- `libselinux` if the system enforces SELinux.

BTW systemd-v259 RC1 is scheduled for next Monday, so there should be some feedback soon from distros living in the edge such as arch or rawhide.

chewi:
Thanks for detailing that. Looking at my own initrd, libpam is only used by systemd-executor. Possibly something to do with the `PAMName` setting? I don't think we need that, so I'm happy to approve if that gets dropped. We can always add it back later if needs be.

chewi (approved):

LaszloGombos (approved):

Conan-Kudo (approved):

Issue #1360, opened 2025-05-26
sysusers credentials ignored


aafeijoo-suse:
**Describe the bug**
sysusers credentials ignored in the initrd since f3dacc013d90bd2c0bbfa04f5f9b167b65298440

**Distribution used**
openSUSE Tumbleweed

**Dracut version**
107

**Init system**
systemd

**Expected behavior**
Calling `systemd-sysusers --root="$initdir"` as part of the build process should be an opt-in, not mandatory. In some use cases `systemd-sysusers.service` needs to run in the initrd. See the old proposal: https://github.com/dracutdevs/dracut/pull/2067

**Additional context**
- https://www.freedesktop.org/software/systemd/man/latest/systemd-sysusers.html#Credentials
- https://www.freedesktop.org/software/systemd/man/latest/systemd-stub.html#/.extra/credentials/*.cred
- https://www.freedesktop.org/software/systemd/man/latest/systemd-stub.html#/.extra/global_credentials/*.cred

PR #1837, opened 2025-11-07
feat(systemd-sysusers): remove sysusers.d configuration from the initrd

aafeijoo-suse:
Commit f3dacc013d90bd2c0bbfa04f5f9b167b65298440 removed the systemd-sysusers machinery from the initrd, so these configuration files now are useless at boot.

## Checklist

### DIFF
--- a/modules.d/78systemd-sysusers/module-setup.sh
+++ b/modules.d/78systemd-sysusers/module-setup.sh
@@ -24,6 +24,10 @@ install() {
         systemd-sysusers --root="$initdir" 2>&1 >&3 | grep -v "^Creating " >&2
     } 3>&1
 
+    # remove sysusers.d configuration from the initrd
+    rm -rf "${initdir}${sysusers}"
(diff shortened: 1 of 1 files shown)


LaszloGombos:
Is there a relationship between this PR and the issue reported earlier - https://github.com/dracut-ng/dracut-ng/issues/1360 ?

aafeijoo-suse:
> Is there a relationship between this PR and the issue reported earlier - #1360 ?

I don't see any connection. If after switching to dracut-ng in openSUSE I get any complaints about #1360, I'll simply redirect them to that upstream issue, or, in the worst case scenario I'll revert the whole thing, like Fedora did.

devkontrol:
Can the goal of this PR be achieved via a configuration file using `--remove/remove_items` for those users or distributions where this is desired?

aafeijoo-suse:
Closing PRs > 4 months old.

PR #1839, opened 2025-11-10
perf(resume): three minor performance patches

aafeijoo-suse:
## Checklist


modules.d/74resume/module-setup.sh
bdrung: Shouldn't `[[ -x "${dracutsysrootdir-}$systemdutildir/systemd-hibernate-resume" ]]` be dropped then?

aafeijoo-suse: Indeed

LaszloGombos (approved):

bdrung (approved):

PR #1842, opened 2025-11-10
fix(dracut-systemd): use expected PS1 in the emergency shell

aafeijoo-suse:
Before starting `dracut-emergency.service`, the function `_emergency_shell()` writes `PS1` to `/etc/profile`. Then `/etc/profile` is sourced in `dracut-emergency.sh`, but `PS1` is not exported, so the current hook name is never displayed in the emergency shell. Also, the default name "dracut" is not printed either in the fallback `PS1`, because the variable `_name` is not set; it should be `_rdshell_name`.

## Checklist


bdrung (approved):
LGTM. I would have preferred to have to commits (one for the export, one for fixing the variable name).

The variable name has been wrong from the starting commit dbfaae0e34507d2d1f3c186ffe26af3e8028b9f8 in 2013.

LaszloGombos (approved):

PR #1844, opened 2025-11-11
fix(base): don't wait infinitely for a device with non-persistent name

coiby:
Currently, dracut will wait finitely for a block device e.g. /dev/sda2
when users use "dracut --add-device /dev/sda2". However, /dev/sda2 is
not a persistant name and it can show as a different name and kdump can
fail [1] as dracut will wait forever for /dev/sda2.

Commit c79fc8fd0 ("fix(dracut): rework timeout for devices added
via --mount and --add-device") already sets infinite timeout for
the underlying persistent device. There is no need to also set infinite
timeout for non-persistent device name. Removing the redundant wait can
automatically resolve the case where a device name changes.

[1] https://github.com/rhkdump/kdump-utils/pull/121

Reported-by: Alex Burmashev <alexander.burmashev@oracle.com>
Reported-by: Harshvardhan Jha <harshvardhan.j.jha@oracle.com>
Fixes: c79fc8fd0 ("fix(dracut): rework timeout for devices added via --mount and --add-device")

## Checklist
- [x] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #1845, opened 2025-11-11
refactor: change TimeoutSec=0 to TimeoutSec=infinity

aafeijoo-suse:
The officially documented way to turn off the timeouts in systemd is setting them to infinity, 0 is kept as historic compat support.

See https://github.com/systemd/systemd/commit/a9b837aa34a2d0bff1687427c66bed3b74cf0fed

Spotted after reviewing #1844, this PR should be rebased after that is merged.

## Checklist

### DIFF
--- a/modules.d/70dmsquash-live/checkisomd5@.service
+++ b/modules.d/70dmsquash-live/checkisomd5@.service
@@ -10,5 +10,5 @@ ExecStart=/usr/bin/checkisomd5 --verbose %f
 StandardInput=tty-force
 StandardOutput=inherit
 StandardError=inherit
-TimeoutSec=0
+TimeoutSec=infinity
 SuccessExitStatus=2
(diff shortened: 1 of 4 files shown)


bdrung:
LGTM, but this change is not a refactor (it changes the resulting initrd). So please use a different prefix.

aafeijoo-suse:
> LGTM, but this change is not a refactor (it changes the resulting initrd). So please use a different prefix.

I'd say that it's a refactor, it just replaces the legacy name of an option with the current one.

https://github.com/dracut-ng/dracut-ng/blob/4cb4917e1222a430047de20eb49c8a66a0416860/doc_site/modules/ROOT/pages/developer/hacking.adoc?plain=1#L77

bdrung (approved):
Since you say that it is a legacy name, since which systemd version `infinity` is supported?

LaszloGombos (approved):

PR #1850, opened 2025-11-12
feat(systemd-coredump): save coredumps to journal

aafeijoo-suse:
This is the current process to get the core dump file of a process that crashed in the initrd:

- Set `rd.break` in the kernel cmdline to enter the emergency shell.
- The core dump file should be located in /var/lib/systemd/coredump/, but /var is not persistent in the initrd, and /sysroot is mounted read-only. So, the user has to manually `mount -o remount,rw /sysroot` and then `cp /var/lib/systemd/coredump/* /sysroot/var/lib/systemd/coredump`.

To make this process much easier, store in the journal coredumps generated in the initrd.

## Checklist

### DIFF
--- a/modules.d/11systemd-coredump/initrd.conf
+++ b/modules.d/11systemd-coredump/initrd.conf
@@ -1,0 +1,8 @@
+# This file is part of dracut systemd-coredump module.
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
+# The default "Storage=external" (/var/lib/systemd/coredump/) is usually lost
+# after switching root, because /var is usually not mounted in the initrd.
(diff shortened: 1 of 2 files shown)


aafeijoo-suse:
The systemd-coredump module is not added with either hostonly or non-hostonly mode; it must be explicitly specified with `-a`/`--add`/`add_dracutmodules+=`. When I see a process segfaulting on a bug report, I just request to add it so I can get the core dump file to analyze it. Otherwise, it should not flood the journal.

Other option would be #663, but it's not as clean as this one.

Conan-Kudo (approved):

PR #1858, opened 2025-11-18
fix(systemd-networkd): install and enable systemd-networkd-resolve-hook.socket

bluca:
New socket needed for v259~rc1, networkd fails if this is not installed and enabled

## Checklist


bdrung (approved):
Tested and verified that it makes the Debian autopkgtest work again.

LaszloGombos (approved):

PR #1859, opened 2025-11-19
test: exit with code 77 if test is skipped

bdrung:
## Changes

Exiting tests that are skipped with exit code 0 will hide that the test environment probably miss some tools. That can reduce the test coverage without anybody noting it.

Therefore exit tests that are skipped with exit code 77. In case tests are intentionally expected to be skipped, an environment variable could be introduce to exit with code 0 again.

## Checklist

Fixes #


bdrung:
The failing test 70 on opensuse seems unrelated.

LaszloGombos (approved):

PR #1876, opened 2025-11-26
fix(dracut): ignore shellcheck SC2329 in addition to SC2317

bdrung:
## Changes

Shellcheck <= 0.10 complains about SC2317 (info): Command appears to be unreachable. Check usage (or ignore if invoked indirectly). Shellcheck 0.11 changes this complaint to SC2329 (info): This function is never invoked. Check usage (or ignored if invoked indirectly).

So ignore shellcheck SC2329 in addition to SC2317 to make both shellcheck versions happy.

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1,7 +1,7 @@
 #!/bin/bash -p
 # temporarly disable some shellcheck warning to be able to preserve commit history
 # before additional commits.
-# shellcheck disable=SC2317
+# shellcheck disable=SC2317,SC2329
 #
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

LaszloGombos:
As part of this PR, should we also remove https://github.com/dracut-ng/dracut-ng/blob/main/dracut.sh#L4 ?

jozzsi (approved):

LaszloGombos (approved):
ok

PR #1881, opened 2025-11-26
fix(man): avoid nesting italic in bold text

bdrung:
## Changes

asciidoctor (on Ubuntu) does not format nested italic in bold text correctly. The following text will be in italic.

Since man pages do not use italic bold text, remove the nesting and pick either italic or bold.

## Checklist

Fixes #


LaszloGombos (approved):

Conan-Kudo (approved):
Meh. We can fix it if it looks bad somewhere later.

PR #1887, opened 2025-11-27
ci: create a separate job for systemd tests

### LINKED ISSUE #1886: skipped tests are now reported as failures
See https://github.com/dracut-ng/dracut-ng/actions/workflows/integration-extra.yml

40, 41, 42, 43 tests on 
 - alpine
 - gentoo openrc
 - void

This is likely a CI regression caused by https://github.com/dracut-ng/dracut-ng/pull/1859 .

We should try to find another solution as it is hard to spot regressions if the CI is not all green. Currently this results in 28 daily CI failures.

CC @bdrung @LaszloGombos

jozzsi:
## Changes

Do not attempt to run systemd tests on non-systemd enabled environment.

The motivation of this commit is to not run tests on the CI that would be skipped. This saves some CI resources and enables to get back to restore green CI, now that skipped tests are reported as failures after f120814.
    
This PR does not change code coverage. This is simply a different and more efficient way to skip tests on the CI.

## Checklist

Fixes #1886

### DIFF
--- a/.github/workflows/integration-extra.yml
+++ b/.github/workflows/integration-extra.yml
@@ -59,6 +59,25 @@ jobs:
                     # https://github.com/dracut-ng/dracut-ng/issues/1590
                     - container: ubuntu:rolling
                       test: "30"
+                    # intentionally skipped
+                    - container: centos:stream10-development
+                      test: "11"
+                    - container: alpine:latest
+                      test: "12"
+                    - container: alpine:edge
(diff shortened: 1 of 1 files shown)


jozzsi:
> The second commit lacks details and resoning IMO.

Added more reasoning to the second commit.

LaszloGombos (approved):
Thanks !

PR #1889, opened 2025-11-27
ci(gentoo): minimize packages between OpenRC and systemd

jozzsi:
## Changes

The goal is to increase test coverage and while keeping the CI maintenance low.

Instead of having two lists of packages, lets keep the systemd-only package list to the minimum.

## Checklist


LaszloGombos (approved):
ok, let's try to see how stable this configuration is for OpenRC

PR #1895, opened 2025-11-28
ci(gentoo): use emerge-webrsync to fetch the initial snapshot

### LINKED ISSUE #1891: gentoo CI container on arm64
The CI reports the following

```
Base image platform does not match expected target platform: test/container/Dockerfile-gentoo#L18
InvalidBaseImagePlatform: Base image docker.io/gentoo/portage:latest was pulled with platform "linux/amd64", expected "linux/arm64" for current build
```

Dockerfile-gentoo#L18 refers to the following line

```
COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo 
```

It does indeed look like that `docker.io/gentoo/portage:latest` does not support `platform: 'linux/arm64`

@Nowa-Ammerlaan do you have a suggestion what to do next ? 

Should we attempt to fix `linux/arm64` container or we should disable building it ? 

Now the container seems to be built, but not used.

Here is the container build log: https://github.com/dracut-ng/dracut-ng/actions/runs/19735043867

jozzsi:
## Changes

Change the approach for computing the Gentoo container to support arm64 as well.

## Checklist

Fixes #1891

### DIFF
--- a/test/container/Dockerfile-gentoo
+++ b/test/container/Dockerfile-gentoo
@@ -15,10 +15,7 @@
 
 ARG OPTION=systemd
 
-FROM docker.io/gentoo/portage:latest AS portage
-
 FROM docker.io/gentoo/stage3:${OPTION}
-COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo
 
 # export ARG
 ARG OPTION
@@ -37,6 +34,7 @@ RUN \
     # Support thin volumes and build all of LVM2 including daemons and tools like lvchange \
     echo 'sys-fs/lvm2 thin lvm' >> /etc/portage/package.use/lvm2 ;\
     # Ensure everything is up to date before we start \
+    emerge-webrsync ;\
     emerge --quiet --update --deep --newuse --autounmask-continue=y --with-bdeps=y @world
 
 # Dependencies to pass basic test


LaszloGombos (dismissed):

test/container/Dockerfile-gentoo
```diff
@@ -15,17 +15,15 @@
...
 FROM docker.io/gentoo/stage3:${OPTION}
-COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo
 
 # export ARG
 ARG OPTION
 
 RUN \
     # Speed-up using binpkgs \
     echo "MAKEOPTS=\"-j$(nproc) -l$(nproc)\"" >> /etc/portage/make.conf ;\
+    wget --progress=dot:mega -O - https://github.com/gentoo-mirror/gentoo/archive/master.tar.gz | tar -xz && mv gentoo-master /var/db/repos/gentoo ;\
```
bdrung: Why not use `emerge-webrsync` instead?

jozzsi: @bdrung . Thanks for the tip, I made the change !

test/container/Dockerfile-gentoo
```diff
@@ -15,17 +15,15 @@
...
 FROM docker.io/gentoo/stage3:${OPTION}
-COPY --from=portage /var/db/repos/gentoo /var/db/repos/gentoo
 
 # export ARG
 ARG OPTION
 
 RUN \
     # Speed-up using binpkgs \
     echo "MAKEOPTS=\"-j$(nproc) -l$(nproc)\"" >> /etc/portage/make.conf ;\
+    emerge-webrsync ;\
```
bdrung: I suggest to move this call down after the setup (so before the first emerge call).

jozzsi: Done. Thanks !

LaszloGombos (approved):

PR #1918, opened 2025-12-05
dracut_functions.sh: make sure device UUIDs are correct before creating systemd device units

Cyanoxygen:
Sometimes the information read from udev may be outdated, for example a disk has just got repartitioned. We don't know why this happens, but it is better to verify the unique identifiers before returning from get_persistent_dev().

We compare the ID provided by get_persistent_dev() and the probed result of blkid. If they don't match, we use the latter one.

This inconsistency is observed during initramfs generation stage of the OS installation, and can not be reproduced reliably, and may be platform dependent. It results in an unbootable OS, stuck at initrd trying to wait a nonexistant UUID.

This pull request adds a verification to confirm the device UUID we are going to wait for during the initramfs actually is correct. @MingcongBai can provide some details on the issue this PR is going to fix.

## Changes

- `dracut-functions.sh`: verify device UUID before creating device units for the initramfs

## Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -285,9 +285,56 @@ get_devpath_block() {
     return 1
 }
 
+# The disk paths fetched from udev can be outdated. We don't know why this
+# happens, but here the best way is to make sure the identifier is correct.
+verify_uuid() {
+    local input="$1" disk="$2" _tag _value _match _path _process=0
+
+    case "$input" in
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Is there a bug filed with systemd about this?

Conan-Kudo:
Marking as draft until we get feedback from systemd folks.

PR #1920, opened 2025-12-10
test(run-qemu): support RISC-V

bdrung:
## Changes

Use the `virt` machine on RISC-V. This machine does not have a PCI bus. So use `virtio-rng-device` as RNG device there.

This is only the first step to not make the test fail on the first qemu call, but there is probably more needed to let the tests succeed.

## Checklist

Fixes #


Conan-Kudo (approved):

LaszloGombos (approved):
great !

PR #1921, opened 2025-12-10
feat: allow booting from compressed ostree filesystems with dmsquash-live

bcl:
When combined with the ostree module it is possible to boot from a compressed ostree filesystem, but the root of this has /ostree not /usr so it fails.

Related: HMS-8933

## Changes
Adds a test for /ostree next to the check for /usr in dmsquash-live-root

## Checklist

Fixes #

### DIFF
--- a/modules.d/70dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/70dmsquash-live/dmsquash-live-root.sh
@@ -344,7 +344,7 @@ if [ -e "$SQUASHED" ]; then
         elif [ -f /run/initramfs/squashfs/LiveOS/ext3fs.img ]; then
             FSIMG="/run/initramfs/squashfs/LiveOS/ext3fs.img"
         fi
-    elif [ -d /run/initramfs/squashfs/usr ]; then
+    elif [ -d /run/initramfs/squashfs/usr ] || [ -d /run/initramfs/squashfs/ostree ]; then
         FSIMG=$SQUASHED
         if [ -z "$overlayfs" ] && [ -n "${DRACUT_SYSTEMD-}" ]; then
             reloadsysrootmountunit=":>/xor_overlayfs;"


Conan-Kudo (approved):

Conan-Kudo (changes requested):
Please use `feat(dmsquash-live):` as the prefix in your commit message

LaszloGombos:
Should we simply just test for `/run/initramfs/squashfs/` instead, instead of keep adding possible additional subdirectories ?

bcl:
> Should we simply just test for `/run/initramfs/squashfs/` instead, instead of keep adding possible additional subdirectories

I think it's useful to have some checking, and I don't really see this list expanding quickly :)

LaszloGombos (approved):

Conan-Kudo (approved):

modules.d/70dmsquash-live/dmsquash-live-root.sh
```diff
@@ -344,7 +344,7 @@ if [ -e "$SQUASHED" ]; then
         elif [ -f /run/initramfs/squashfs/LiveOS/ext3fs.img ]; then
             FSIMG="/run/initramfs/squashfs/LiveOS/ext3fs.img"
         fi
-    elif [ -d /run/initramfs/squashfs/usr ]; then
+    elif [ -d /run/initramfs/squashfs/usr ] || [ -d /run/initramfs/squashfs/ostree ]; then
```
cgwalters, line 347: In the future we're going to want to generalize this to support e.g. `/composefs`.

I think the most general way to do this would probably be if the OS image can have a drop-in config that gets sourced here or so that specifies to skip the `/usr` check?

That said ultimately what I want to do is https://github.com/bootc-dev/bootc/issues/898 which would obviate this.

PR #1929, opened 2025-12-12
feat(overlayfs): add persistent device overlay support

nadzyah:
This pull request adds persistent device overlay support to the overlayfs module, enabling compatibility with `cloud-initramfs-tools` overlayroot syntax for users migrating from this tool to dracut.

## Changes

- Add `overlayroot=` parameter supporting LABEL=, UUID=, PARTLABEL=, PARTUUID=, device paths, and `device:dev=` syntax
- Device-backed overlays mounted at `/run/overlayfs-backing` with symlinks for transparency
- Device resolution validation with graceful fallback to tmpfs on failure
- Comprehensive test suite covering tmpfs, LABEL, UUID, and fallback scenarios
- Documentation added to dracut.cmdline man page

## Checklist

This is my first contribution to dracut-ng, so I would appreciate any feedback and suggestions :)


modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -5,17 +5,89 @@ command -v getarg > /dev/null || . /lib/dracut-lib.sh
...
+        fi
+    fi
+
+    # Tmpfs mode (default or fallback)
+    if [ "$overlay_mode" = "tmpfs" ]; then
+        info "Using tmpfs overlay (changes will not persist across reboots)"
+
+        mkdir -m 0755 -p /run/overlayfs
+        mkdir -m 0755 -p /run/ovlwork
+        if [ -n "$reset_overlay" ]; then
```
FGrose: The  'tmpfs' mode needs no reset code.

nadzyah: This is actually the original code, I haven't modified the logic here. However, I can remove it if you think this is not needed :)

FGrose: The current code has the  `&& [ -h /run/overlayfs ]` condition to skip resetting a virtual overlay.

My [development version](https://github.com/FGrose/dracut-ng/blob/AO/modules.d/70overlayfs/parse-overlayfs.sh) has this logic:
```shell
[ -h /run/overlayfs ] || {
    # For temporary overlays:
    mkdir -m 0755 -p /run/overlayfs
    mkdir -m 0755 -p /run/ovlwork
}

if getargbool 0 rd.overlayfs.reset && [ -h /run/overlayfs ]; then
    ovlfsdir=$(readlink /run/overlayfs)
    info "Resetting the OverlayFS overlay directory."
    rm -r -- "${ovlfsdir:?}" > /dev/kmsg 2>&1
    mkdir -p "$ovlfsdir"
fi
```

Conan-Kudo:
Can we merge this into `dmsquash-live-autooverlay`?

modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -5,17 +5,94 @@ command -v getarg > /dev/null || . /lib/dracut-lib.sh
...
+if [ -n "$overlayroot" ]; then
+    overlayfs="yes"
+
+    case "$overlayroot" in
+        tmpfs*)
+            overlay_mode="tmpfs"
+            ;;
+        /dev/* | LABEL=* | UUID=* | PARTLABEL=* | PARTUUID=*)
+            overlay_mode="device"
+            overlay_device=$(label_uuid_to_dev "$overlayroot")
```
FGrose: This function will echo the input (it doesn't return zero length) or return a path whether or not the partition exists.
You probably want to test `[ -b "$overlay_device ] || warn ...` before proceeding...

Same issue at line 39.

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -0,0 +1,86 @@
...
+    build_ext4_image "$TESTDIR"/dracut.*/initramfs/ "$TESTDIR"/root.img dracut || return 1
+
+    dd if=/dev/zero of="$TESTDIR"/overlay.img bs=1M count=100
+    mkfs.ext4 -L OVERLAY "$TESTDIR"/overlay.img
+    # shellcheck disable=SC2034
+    OVERLAY_UUID=$(blkid -s UUID -o value "$TESTDIR"/overlay.img)
+
+    call_dracut --tmpdir "$TESTDIR" \
+        --add overlayfs \
+        --no-hostonly --no-hostonly-cmdline \
```
jozzsi: I do not have a strong opinion on forcing no-hostonly for this test., but is that actually required ?  Could you try to remove it just to see if any of the tests fails on the CI ?

nadzyah: Done, and looks like all the tests still pass

modules.d/70overlayfs/prepare-overlayfs.sh
nadzyah: I've updated it, thanks for noticing the issue!

jozzsi:
> 1. Should `overlayroot=LABEL=persist` work alongside `rd.live.overlay=LABEL=persist` for live boot, essentially treating them as equivalent or complementary?

I think they should be equivalent if possible. Do you have any concern by making them equivalent ? 

> 2. Would the auto-partitioning/formatting logic from `dmsquash-live-autooverlay` be triggered by either parameter, or should they remain separate mechanisms?

triggered by either parameter

> 3. Is the goal to eventually phase out one of these parameters in favour of the other, or to maintain both with shared underlying logic?

I am personally ok for maintaining both with shared underlying logic for a while (e.g. 3-4 releases). The project has other aliases for command line options, but would be nice to eventually deprecate one of the syntaxes. 

See https://github.com/dracut-ng/dracut-ng/blob/main/man/dracut.cmdline.7.adoc?plain=1#L1456 for deprecated aliases

jozzsi (approved):
non-owner LGTM.

Thanks @nadzyah for working with us on this project, this is an important PR.

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env bash
+set -eu
+# shellcheck disable=SC2034
+TEST_DESCRIPTION="Test overlayfs module with persistent device overlay"
+
+test_check() {
+    command -v qemu-kvm &> /dev/null || command -v qemu-system-"$(uname -m)" &> /dev/null || return 1
```
bdrung: qemu is a required test dependency.

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -0,0 +1,85 @@
...
+    qemu_add_drive disk_index disk_args "$TESTDIR"/root.img root
+    qemu_add_drive disk_index disk_args "$TESTDIR"/overlay.img overlay
+
+    test_marker_reset
+
+    echo "TEST: tmpfs overlay"
+    "$testdir"/run-qemu -nic none \
+        "${disk_args[@]}" \
+        -append "$TEST_KERNEL_CMDLINE rd.overlayfs=1" \
+        -initrd "$TESTDIR"/initramfs.testing || return 1
```
bdrung: The tests run with `set -eu`. So all the `|| return 1` are not needed any more.

man/dracut.cmdline.7.adoc
```diff
@@ -1109,6 +1109,44 @@ The **rd.overlayfs.readonly** option, which allows a persistent overlayfs to
 be mounted read-only through a higher level transient overlay directory, has
 been implemented through the multiple lower layers feature of OverlayFS.
 
+**overlayroot=**__{tmpfs|LABEL=<label>|UUID=<uuid>|PARTUUID=<uuid>|PARTLABEL=<label>|/dev/<device>|device:dev=<device>|disabled}__::
```
bdrung: So with overlayroot we make `rd.live.overlay` available for the non-live case, don't we?

How about renaming `rd.live.overlay` to `rd.overlay` (similar to commit b77ae7eb8523da696fcb0f48a747ac22bdc9eb31)? Then `overlayroot` can be an alias for `rd.overlay`.

@jozzsi what do you think?

jozzsi: This makes sense to me as well.

Conan-Kudo: Please namespace this commandline option with `rd.overlayfs`. We really want to avoid adding more non-namespaced stuff.

Conan-Kudo (changes requested):

test/TEST-30-DMSQUASH/test.sh
```diff
@@ -77,6 +77,25 @@ test_run() {
...
+
+    client_run "persistent autooverlay" "init=/sbin/init-persist rd.live.image rd.overlay=LABEL=persist rd.live.dir=LiveOS overlayroot=LABEL=persist"
+
+    rootPartitions=$(sfdisk -d "$TESTDIR"/root.img | grep -c 'root\.img[0-9]')
+    [ "$rootPartitions" -eq 2 ]
+
+    (
+        set +o pipefail
+        dd if="$TESTDIR"/root.img bs=1MiB status=none \
+            | grep -U --binary-files=binary -F -m 1 -q dracut-autooverlay-success
```
jozzsi: The source where this line was copied from changed recently, please rebase with the recent change.

test/TEST-30-DMSQUASH/test.sh
```diff
@@ -77,6 +77,25 @@ test_run() {
...
 
+    # Test overlayroot= parameter equivalence with rd.overlay=
+    test_marker_reset
+    sfdisk --delete "$TESTDIR"/root.img 2
+    # Wipe the area where partition 2 will be recreated to avoid mkfs prompt
+    dd if=/dev/zero of="$TESTDIR"/root.img bs=1M seek=320 count=50 conv=notrunc status=none
+    rootPartitions=$(sfdisk -d "$TESTDIR"/root.img | grep -c 'root\.img[0-9]')
+    [ "$rootPartitions" -eq 1 ]
+
+    client_run "persistent autooverlay" "init=/sbin/init-persist rd.live.image rd.overlay=LABEL=persist rd.live.dir=LiveOS overlayroot=LABEL=persist"
```
jozzsi: Did you mean to remove `rd.overlay=LABEL=persist` from here ?

nadzyah: I kept both parameters in the test as you suggested here: https://github.com/dracut-ng/dracut-ng/pull/1929#issuecomment-3678085943

Did you mean something different, or should I remove one of the parameters?

jozzsi: ok. thanks

man/dracut.cmdline.7.adoc
bdrung: To clarify my idea: Move `rd.live.overlay` out of the `rd.live` namespace and pick something in the `rd.` namespace (`rd.overlay` or `rd.overlayfs.root` or ???). For compatibility with `cloud-initramfs-tools` `overlayroot` can be added as alias to that variable in the `rd.` namespace.

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -0,0 +1,76 @@
...
+    test_marker_reset
+
+    echo "TEST: persistent device overlay (LABEL=OVERLAY)"
+    "$testdir"/run-qemu -nic none \
+        "${disk_args[@]}" \
+        -append "$TEST_KERNEL_CMDLINE root=LABEL=dracut rd.overlayfs.overlayroot=LABEL=OVERLAY" \
+        -initrd "$TESTDIR"/initramfs.testing
+
+    test_marker_check
+    test_marker_reset
```
bdrung: In addition to the test marker check, can you check that the overlay was mounted correctly? E.g. write something on the overlay and afterwards check the overlay.img?

modules.d/70dmsquash-live-autooverlay/create-overlay.sh
```diff
@@ -9,13 +9,13 @@ if getargbool 0 rd.live.debug; then
 fi
 
 gatherData() {
-    overlay=$(getarg rd.live.overlay)
+    overlay=$(getarg rd.overlay -d rd.live.overlay)
     if [ -z "$overlay" ]; then
-        info "Skipping overlay creation: kernel command line parameter 'rd.live.overlay' is not set"
+        info "Skipping overlay creation: kernel command line parameter 'rd.overlay' is not set"
```
bdrung, line 14: Since this merge request is on the bigger size, I recommend to put the `rd.live.overlay` rename into a separate commit and merge request.

modules.d/70overlayfs/create-overlay-genrules.sh
```diff
@@ -0,0 +1,28 @@
...
+if [ -n "$overlay_param" ]; then
+    case "$overlay_param" in
+        LABEL=*)
+            root=$(getarg root=)
+            if [ "${root%%:*}" = "live" ] || getargbool 0 rd.live.image; then
+                overlayLabel=${overlay_param#LABEL=}
+                if [ ! -b "/dev/disk/by-label/${overlayLabel}" ]; then
+                    case "$root" in
+                        live:/dev/*)
+                            printf 'SYMLINK=="%s", RUN+="/sbin/initqueue --settled --onetime --unique /sbin/create-overlay %s"\n' \
```
bdrung: This is quite nested. Suggestion: Put the code in a function and use the return early pattern.

modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -5,17 +5,98 @@ command -v getarg > /dev/null || . /lib/dracut-lib.sh
...
+            overlay_mode="device"
+            overlay_device="${overlayroot#device:}"
+            overlay_device="${overlay_device#dev=}"
+            overlay_device=$(label_uuid_to_dev "$overlay_device")
+            if [ ! -b "$overlay_device" ]; then
+                warn "Failed to resolve device from '$overlayroot', falling back to tmpfs"
+                overlay_mode="tmpfs"
+            fi
+            ;;
+        disabled)
```
bdrung: Can you also support disabling by setting `=0` like the other boolean parameters do. See https://github.com/dracut-ng/dracut-ng/blob/ddbc224b71023685c59bcb40985e83a16ae3a626/man/dracut.cmdline.7.adoc?plain=1#L27

modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -5,17 +5,98 @@ command -v getarg > /dev/null || . /lib/dracut-lib.sh
...
+fi
+
+overlay_mode="tmpfs"
+overlay_device=""
+
+if [ -n "$overlayroot" ]; then
+    overlayfs="yes"
+
+    case "$overlayroot" in
+        tmpfs*)
```
bdrung: Can you also support enabling by setting `=1` (or to an empty string) like the other boolean parameters do. See https://github.com/dracut-ng/dracut-ng/blob/ddbc224b71023685c59bcb40985e83a16ae3a626/man/dracut.cmdline.7.adoc?plain=1#L27

test/TEST-21-OVERLAYFS/test-overlayfs.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/sh
```
bdrung: This new test 21 should be added to the CI in .github/workflows/.

test/TEST-21-OVERLAYFS/test.sh
bdrung: Oh, there is `test-overlayfs.sh`. So ignore my comment.

man/dracut.cmdline.7.adoc
```diff
@@ -1109,6 +1109,55 @@ The **rd.overlayfs.readonly** option, which allows a persistent overlayfs to
...
+---
++
+[listing]
+.Examples
+---
+overlayroot=tmpfs
+overlayroot=LABEL=OVERLAY
+overlayroot=device:dev=UUID=12345678-1234-1234-1234-123456789abc
+overlayroot=disabled
+---
```
Conan-Kudo: Do we need this in dracut? Wouldn't it make sense for this to live with the tool that uses this syntax, since it's a legacy compat syntax that shouldn't exist forever?

nadzyah: I've removed the commit that implements cloud-initramfs-tools compatibility

.github/workflows/integration.yml
```diff
@@ -91,6 +91,7 @@ jobs:
                     - "12"
                     - "13"
                     - "20"
+                    - "21"
```
bdrung, line 94: Can you also add this test to daily-basic (basic+extended), daily-hostonlystrict, and daily-omitsystemd?

man/dracut.cmdline.7.adoc
```diff
@@ -1109,6 +1109,31 @@ The **rd.overlay.readonly** option, which allows a persistent overlayfs to
...
+**rd.overlay=**__{LABEL=<label>|UUID=<uuid>|PARTUUID=<uuid>|PARTLABEL=<label>|/dev/<device>}__::
+Specifies the storage device for persistent overlay (for both live images and
+regular OverlayFS).  When a block device is specified, it will be mounted and
+used for persistent overlay storage.  For live images with LABEL= specification,
+if the labeled partition does not exist, it will be automatically created using
+available free space on the device.
++
+Supported device specifications:
++
+---
```
bdrung: This is formatted strangely (see `make doc && man -l man/dracut.cmdline.7`).

modules.d/70overlayfs/module-setup.sh
```diff
@@ -15,6 +15,7 @@ installkernel() {
 
 install() {
     inst_hook pre-mount 01 "$moddir/prepare-overlayfs.sh"
-    inst_hook mount 01 "$moddir/mount-overlayfs.sh"     # overlay on top of block device
-    inst_hook pre-pivot 10 "$moddir/mount-overlayfs.sh" # overlay on top of network device (e.g. nfs)
+    inst_hook mount 01 "$moddir/mount-overlayfs.sh"
+    inst_hook pre-pivot 00 "$moddir/prepare-overlayfs.sh"
+    inst_hook pre-pivot 01 "$moddir/mount-overlayfs.sh"
```
bdrung: Why do you change the numbering and removed the comments?

modules.d/70overlayfs/mount-overlayfs.sh
```diff
@@ -6,6 +6,9 @@ getargbool 0 rd.overlayfs -d rd.live.overlay.overlayfs && overlayfs="yes"
 getargbool 0 rd.overlay.readonly -d rd.live.overlayfs.readonly && readonly_overlay="--readonly" || readonly_overlay=""
 
 if [ -n "$overlayfs" ]; then
+    # Only proceed if prepare-overlayfs.sh has run
+    [ -e /run/rootfsbase ] || return 0
```
bdrung: When would you hit this code path?

test/TEST-21-OVERLAYFS/assertion.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+if ! grep -q " overlay " /proc/mounts; then
+    echo "overlay filesystem not found in /proc/mounts" >> /run/failed
+    cat /proc/mounts >> /run/failed
+fi
+
+if ! dd if=/dev/zero of=/test-overlay-write bs=1 count=1 2> /dev/null; then
```
bdrung: Why do you use dd instead of just an shell builtin like echo?
```suggestion
if ! echo > /test-overlay-write; then
```

test/TEST-21-OVERLAYFS/assertion.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+
+if ! grep -q " overlay " /proc/mounts; then
+    echo "overlay filesystem not found in /proc/mounts" >> /run/failed
+    cat /proc/mounts >> /run/failed
```
bdrung: putting `/proc/mounts` into the /run/failed makes the failure log hard to read:

```
overlay filesystem not found in /proc/mounts
tmpfs /run tmpfs rw,nosuid,nodev,size=195204k,nr_inodes=819200,mode=755,inode64 0 0
/dev/sda / ext4 ro,relatime 0 0
devtmpfs /dev devtmpfs rw,nosuid,size=461504k,nr_inodes=115376,mode=755,inode64 0 0
tmpfs /dev/shm tmpfs rw,nosuid,nodev,inode64,usrquota 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=600,ptmxmode=000 0 0
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
securityfs /sys/kernel/security securityfs rw,nosuid,nodev,noexec,relatime 0 0
cgroup2 /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate,memory_recursiveprot,memory_hugetlb_accounting 0 0
none /sys/fs/pstore pstore rw,nosuid,nodev,noexec,relatime 0 0
efivarfs /sys/firmware/efi/efivars efivarfs rw,nosuid,nodev,noexec,relatime 0 0
bpf /sys/fs/bpf bpf rw,nosuid,nodev,noexec,relatime,mode=700 0 0
configfs /sys/kernel/config configfs rw,nosuid,nodev,noexec,relatime 0 0
proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /etc/machine-id tmpfs ro,size=195204k,nr_inodes=819200,mode=755,inode64 0 0
none /run/credentials/systemd-journald.service tmpfs ro,nosuid,nodev,noexec,relatime,nosymfollow,size=1024k,nr_inodes=1024,mode=700,inode64,noswap 0 0
overlay is not writable
persistent overlay device not mounted at /run/overlayfs-backing
tmpfs /run tmpfs rw,nosuid,nodev,size=195204k,nr_inodes=819200,mode=755,inode64 0 0
/dev/sda / ext4 ro,relatime 0 0
devtmpfs /dev devtmpfs rw,nosuid,size=461504k,nr_inodes=115376,mode=755,inode64 0 0
tmpfs /dev/shm tmpfs rw,nosuid,nodev,inode64,usrquota 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=600,ptmxmode=000 0 0
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
securityfs /sys/kernel/security securityfs rw,nosuid,nodev,noexec,relatime 0 0
cgroup2 /sys/fs/cgroup cgroup2 rw,nosuid,nodev,noexec,relatime,nsdelegate,memory_recursiveprot,memory_hugetlb_accounting 0 0
none /sys/fs/pstore pstore rw,nosuid,nodev,noexec,relatime 0 0
efivarfs /sys/firmware/efi/efivars efivarfs rw,nosuid,nodev,noexec,relatime 0 0
bpf /sys/fs/bpf bpf rw,nosuid,nodev,noexec,relatime,mode=700 0 0
configfs /sys/kernel/config configfs rw,nosuid,nodev,noexec,relatime 0 0
proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /etc/machine-id tmpfs ro,size=195204k,nr_inodes=819200,mode=755,inode64 0 0
none /run/credentials/systemd-journald.service tmpfs ro,nosuid,nodev,noexec,relatime,nosymfollow,size=1024k,nr_inodes=1024,mode=700,inode64,noswap 0 0
```

Idea, log it in case of a failure and the end of assertion.sh:
```sh
if [ -s /run/failed ]; then
    echo ">>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> /proc/mounts >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>"
    cat /proc/mounts
    echo "<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< /proc/mounts <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<"
fi
```

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -0,0 +1,50 @@
...
+        --add-confdir test-root \
+        -i ./assertion.sh /assertion.sh \
+        -f "$TESTDIR"/initramfs.root
+
+    build_ext4_image "$TESTDIR"/dracut.*/initramfs/ "$TESTDIR"/root.img dracut
+
+    dd if=/dev/zero of="$TESTDIR"/overlay.img bs=1M count=100
+    mkfs.ext4 -L OVERLAY "$TESTDIR"/overlay.img
+    # shellcheck disable=SC2034
+    OVERLAY_UUID=$(blkid -s UUID -o value "$TESTDIR"/overlay.img)
```
bdrung: I suggest to move that variable to test_run and make it a local variable. then shellcheck will be happy as well.

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -0,0 +1,50 @@
...
+
+test_setup() {
+    call_dracut --tmpdir "$TESTDIR" \
+        --add-confdir test-root \
+        -i ./assertion.sh /assertion.sh \
+        -f "$TESTDIR"/initramfs.root
+
+    build_ext4_image "$TESTDIR"/dracut.*/initramfs/ "$TESTDIR"/root.img dracut
+
+    dd if=/dev/zero of="$TESTDIR"/overlay.img bs=1M count=100
```
bdrung: truncate is faster (and we use that in build_ext4_image as well):
```suggestion
    rm -f "$TESTDIR/overlay.img"
    truncate -s 100M "$TESTDIR/overlay.img"
```

would a smaller overlay.img be sufficient?

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -0,0 +1,50 @@
...
+    mkfs.ext4 -L OVERLAY "$TESTDIR"/overlay.img
```
bdrung:
```suggestion
    mkfs.ext4 -q -L OVERLAY "$TESTDIR"/overlay.img
```

modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -5,17 +5,77 @@ command -v getarg > /dev/null || . /lib/dracut-lib.sh
...
+                warn "Failed to resolve device from '$overlay', falling back to tmpfs"
+                overlay_mode="tmpfs"
+            fi
+            ;;
+        *)
+            # For dmsquash-live compatibility, any other format uses tmpfs
+            overlay_mode="tmpfs"
+            ;;
+    esac
+fi
```
bdrung, line 32: Please also check the else case for the opposite.

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -0,0 +1,50 @@
...
+    check_qemu_log
+
+    client_test_end
+}
+
+test_run() {
+    client_run "tmpfs overlay" "rd.overlayfs=1"
+    client_run "persistent device overlay (LABEL)" "rd.overlay=LABEL=OVERLAY"
+    client_run "persistent device overlay (UUID)" "rd.overlay=UUID=$OVERLAY_UUID"
+    client_run "fallback to tmpfs (non-existent LABEL)" "rd.overlay=LABEL=NONEXISTENT"
```
bdrung, line 35: You could also test the device path overlay:
```
    client_run "persistent device overlay (device path)" \
        "rd.overlay=/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_overlay"
```

bdrung (changes requested):
Thanks. Overall this change looks good to me. I have some smaller change requests and a few open questions.

.github/workflows/integration.yml
nadzyah: done

modules.d/70overlayfs/mount-overlayfs.sh
nadzyah: I've updated the code to add an explanation

modules.d/70overlayfs/prepare-overlayfs.sh
bdrung: Sorry, I made a mistake. That comment should have been attached to the assertion.sh file instead.

modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -5,17 +5,77 @@ command -v getarg > /dev/null || . /lib/dracut-lib.sh
 getargbool 0 rd.overlayfs -d rd.live.overlay.overlayfs && overlayfs="yes"
 getargbool 0 rd.overlay.reset -d rd.live.overlay.reset && reset_overlay="yes"
 
+overlay=$(getarg rd.overlay -d rd.live.overlay)
+
+overlay_mode="tmpfs"
+overlay_device=""
+
+if [ -n "$overlay" ]; then
```
bdrung: If `overlay` is not set, you could exit early. Then you do not have to set `overlayfs="yes"`.

modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -5,17 +5,77 @@ command -v getarg > /dev/null || . /lib/dracut-lib.sh
...
+            ln -sf /run/initramfs/overlay/ovlwork /run/ovlwork
+        }
+    fi
+
     if [ -n "$reset_overlay" ] && [ -h /run/overlayfs ]; then
         ovlfsdir=$(readlink /run/overlayfs)
         info "Resetting the OverlayFS overlay directory."
-        rm -r -- "${ovlfsdir:?}"/* "${ovlfsdir:?}"/.* > /dev/null 2>&1
+        rm -r -- "${ovlfsdir:?}" > /dev/kmsg 2>&1
+        mkdir -p "$ovlfsdir"
```
bdrung: Is there a reason for this change? I do not see how this is related.

bdrung (dismissed):
Looks good to me now. Thanks for contributing and writing tests. 🥳

nadzyah:
@bdrung I've rebased the changes since the #2153 PR got merged

bdrung (approved):

Conan-Kudo (approved):

LaszloGombos (approved):

PR #1931, opened 2025-12-13
chore: upgrade shellcheck

jozzsi:
Switch CI to a newer version of Ubuntu to get a newer version of shellcheck.

## Checklist

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -54,12 +54,14 @@ jobs:
         needs: basic
         name: syncheck
         runs-on: ubuntu-24.04
+        container:
+            image: ubuntu:devel
         steps:
             - name: "Checkout Repository"
               uses: actions/checkout@v6
             - run: |
-                  sudo apt-get update
-                  sudo apt-get -y install shellcheck shfmt
+                  apt-get update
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -54,12 +54,14 @@ jobs:
         needs: basic
         name: syncheck
         runs-on: ubuntu-24.04
+        container:
+            image: ghcr.io/dracut-ng/ubuntu:devel-amd
```
bdrung: Since our ubuntu devel image does not have shellcheck and we install shellcheck afterwards, we better just use the `ubuntu:devel` image.

jozzsi: Sadly this is more involved.

`ubuntu:devel` results in

```
3: make: not found
```

bdrung: Right. We call `make` which calls shellcheck. make is small enough to install it in addition to the other tools.

jozzsi: ok, done. Seems the only remaining package needed was indeed make.

bdrung (approved):

LaszloGombos (approved):

PR #1935, opened 2025-12-14
chore: remove default argument to cp

jozzsi:
Adding argument processing for arguments that are the default anyways just adds to cognitive load of reading the code and a little bit of performance overhead.

`--sparse=auto` is the default. There is no need to re-specify it.

## Checklist


Conan-Kudo (approved):
Ehh, whatever, I guess?

LaszloGombos (approved):

PR #1959, opened 2025-12-19
feat: remove rev_lib_symlinks

jozzsi:
## Changes

rev_lib_symlinks is not used any more by dracut and it is not a dependency for any of the out-of-tree dracut modules.

`rev_lib_symlinks` function was introduced by a10a141  to be used by `inst_library`, however `inst_library` no longer uses it (neither any other function).

It is hard to maintain code that is not documented, not used and has no test coverage.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1684,37 +1684,6 @@ mark_hostonly() {
     done
 }
 
-# find symlinks linked to given library file
-# $1 = library file
-# Function searches for symlinks by stripping version numbers appended to
-# library filename, checks if it points to the same target and finally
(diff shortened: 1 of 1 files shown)


bdrung (changes requested):
Is that a typo in the git commit? "is not used **by** any more by [...]"

Besides that: LGTM. https://codesearch.debian.net shows only one user: `plymouth-populate-initrd.in` which uses it's own copy of that function.

bdrung (approved):

LaszloGombos (approved):

PR #1975, opened 2025-12-21
chore: remove default --reflink=auto argument from cp calls

jozzsi:
## Changes

Adding argument processing for arguments that are the default anyways just adds to cognitive load of reading the code and a little bit of performance overhead.

`--reflink=auto` is the default. There is no need to re-specify it.

Also, busybox cp does not support `--reflink=auto`, so this change makes dracut more portable.

## Checklist


jozzsi:
> ...and you're willing to support it, I'm fine with making adjustments to handle it.

Appreciate the support and reopening this ticket @bdrung @LaszloGombos @Conan-Kudo  - but I am not able to support this work any more. I hope someone else can pick it. Hope you understand.

PR #1978, opened 2025-12-21
ci(alpine): fix regression regarding make is not installed

jozzsi:
## Changes

It turns out make it not a dependency for dracut-tests downstream. Add make dependency to continue upstream development.

Follow-up to 52606ca .

## Checklist


LaszloGombos (approved):
ok, np

PR #1982, opened 2025-12-22
feat(systemd): pull 'drm' module unless excluded

pvalena:
This is a clone of PR https://github.com/dracutdevs/dracut/pull/2719 by @rmetrich 

_____

## Changes

Some graphic cards require the 'drm' module to be present to be able to setup the console font (e.g. 'i915' requires it).

## Checklist

> ## Manual testing only on a QEMU/KVM
> 1. Remove plymouth package from the system
> 2. Set the console font
>    ```
>    # grep FONT /etc/vconsole.conf
>    FONT="latarcyrheb-sun32"
>    ```
> 3. Rebuild the initramfs and verify there is no "drm" module (original state, without this patch)
>    ```
>    # dracut -f
>    # lsinitrd --mod | grep drm
>    --> nothing
>    ```
> 4. Reboot and add `rd.break` to drop to emergency
> 5. At emergency prompt, confirm the font was not applied (the font is very pixelized)
> 6. Apply the patch
> 7. Rebuild the initramfs and verify there is "drm" module
>    ```
>    # dracut -f
>    # lsinitrd --mod | grep drm
>    drm
>    ```
> 8. Reboot and add `rd.break` to drop to emergency
> 9. At emergency prompt, confirm the font was applied (the font is very pixelized)


Conan-Kudo (dismissed):

pvalena:
Sure, makes sense, as console fonts are set/loaded via i18n, lets move it there. I will refresh this PR. The failures / timeouts for ubuntu:devel seem to be however somewhat unrelated :) (or in the very least it's strange that this change would introduce a breakage, as drm is just loading some kernel modules).

pvalena:
@LaszloGombos @Conan-Kudo PTAL.

jozzsi:
> Same question as Laszlo: How does this relates to the work on the simpledrm dracut module?

@pvalena please see https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/45plymouth/module-setup.sh#L48 for reference

jozzsi:
```
depends() {                                                                                                                                                                                                                           
    local deps                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                                                                                                                                                                  
    # install optional dependencies unless they are omitted                                                                                                                                                                           
    for module in simpledrm drm; do                                                                                                                                                                             
        module_check $module > /dev/null 2>&1                                                                                                                                                                                         
        if [[ $? == 255 ]] && ! [[ " $omit_dracutmodules " == *\ $module\ * ]]; then                                                                                                                                                  
            deps+=" $module"
            break                                                                                                                                                                                                          
        fi                                                                                                                                                                                                                            
    done                                                                                                                                                                                                                              
                                                                                                                                                                                                                                      
    echo "$deps"                                                                                                                                                                                                                      
    return 0                                                                                                                                                                                                                          
} 
```

This is an established patter - see e.g. https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/11systemd-udevd/module-setup.sh#L27

LaszloGombos (dismissed):

pvalena:
@LaszloGombos PTAL.

modules.d/46i18n/module-setup.sh
```diff
@@ -9,6 +9,32 @@ check() {
     return 0
 }
```
bdrung, line 11: Can you add a comment documenting the ordering (see other "ordered later" comments)?

modules.d/46i18n/module-setup.sh
```diff
@@ -9,6 +9,32 @@ check() {
...
 
+# called by dracut
+depends() {
+    # Include "drm" / "simpledrm" to be able to set the console font properly
+    local _module _drm
+    local -a _modules=(drm simpledrm)
```
bdrung, line 18: What module should be pulled in preferred to be pulled in?

modules.d/46i18n/module-setup.sh
```diff
@@ -9,6 +9,32 @@ check() {
...
+
+    if [[ -z $_drm ]]; then
+        for _module in "${_modules[@]}"; do
+            module_check "$_module" > /dev/null 2>&1
+            if [[ $? == 255 ]] && ! [[ " $omit_dracutmodules " == *\ $_module\ * ]]; then
+                _drm="$_module"
+                break
+            fi
+        done
+    fi
```
bdrung, line 35: The whole logic looks identical to other modules like 40network. Maybe we should add a helper function to dracut-functions.sh for that.

bdrung (changes requested):

modules.d/46i18n/module-setup.sh
pvalena: `drm`, as I'm not even sure simpledrm has this required functionality (can someone test?).

modules.d/46i18n/module-setup.sh
pvalena: Can we do refactoring in other PR?

jozzsi:
> drm, as I'm not even sure simpledrm has this required functionality (can someone test?).

First hit when I do a search for "simpledrm" and "Fedora"

https://fedoraproject.org/wiki/Changes/PlymouthUseSimpledrm

I would think Fedora should care about simpledrm just as much as any other major distro.

>This avoids the textmode fallback if the GPU driver takes a long time to initialize.
> This shows the splash earlier instead of letting the user look at a static screen

Similar sentiment as I pointed out at https://github.com/dracut-ng/dracut-ng/pull/1982#issuecomment-3724906536

CC @jwrdegoede

modules.d/46i18n/module-setup.sh
pvalena: Please note this ordering is only a fallback in case there was no preference detected for configuration (therefore reodered). So it really shouldn't matter, I think.

modules.d/46i18n/module-setup.sh
pvalena: done

jwrdegoede:
A couple of remarks here:

1. IMHO we should try to move to never having the drm modules in the initrd not do so more often: https://hansdegoede.dreamwidth.org/28291.html

2. I think the font-setting problem here needs to be root-caused a lot better before accepting this PR. All drm/kms drivers use fbcon to do the text rendering. So there is no reason why font-setting should not work with simpledrm. 

I can see a couple of reasons why font-setting would not work without the i915 driver loaded:

1. This is on BIOS booting systems which lack a firmware-framebuffer (efifb), so no simpledrm and before the i915 driver loads the good old vga-console is used instead of fbcon

I expect this to be the most likely root-cause. The upstream kernel has recently grown support for running simpledrm over vesa for this setup and IMHO the correct fix would be to move towards using that, which means running the latest kernel with the right Kconfig. If someone wants to look into enabling this for their distro I suggest reaching out to Thomas Zimmermann <tzimmermann@suse.de> who did all the kernel work for this. I believe this is likely all in place in 6.18.

2. deferred-fbcon-takeover is in play and the delay caused by loading i915 allows some message to get printed and fbcon to takeover: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=83d83bebf40132e2d55ec58af666713cc76f9764
Before fbcon takeover the dummy console driver is used which I presume does not support font-loading and even if it does, any loaded font will likely be forgotten when fbcon takes over as the console driver.

Note deferred-fbcon-takeover takes over the console from dummycon as soon as the first character is printed to the console. So unless you're booting with "quiet" and your userspace is also fully quiet before the i18n module does the console setup this will not be the problem.

pvalena:
> A couple of remarks here:
> 
> 1. IMHO we should try to move to never having the drm modules in the initrd not do so more often: https://hansdegoede.dreamwidth.org/28291.html
> 2. I think the font-setting problem here needs to be root-caused a lot better before accepting this PR. All drm/kms drivers use fbcon to do the text rendering. So there is no reason why font-setting should not work with simpledrm.

Great! Thanks for the confirmation it works as expected. Like I said in my earlier comment, the order doesn't really matter, so it can be reversed if that is benefitial.
To clarify, there's no problem, just that there's just no drm at all:) in initrd; no drm or simpledrm, as it didn't get pulled in (plymouth is not installed).

> 
> I can see a couple of reasons why font-setting would not work without the i915 driver loaded:

I don't understand how this is relevant -- i915 is nowhere mentioned? (i18n is something completely different)

(Skipping the rest, as I need this answered first.)

jwrdegoede:
Let me rephrase things a bit. If you run lsinitrd on the generated initrd both before and after the suggested changes like this:

```
lsinitrd > lsinitrd.before
lsinitrd > lsinitrd.after
diff -u lsinitrd.before lsinitrd.after
```

What changes does the diff command show, or in other words which extra files get added to the initrd?

In your original description you sorta claim that only the `drm` module gets added. But I think that is only because of your filtering of the lsinitrd output. I presume that a whole bunch of other drm modules are also getting added ?

pvalena:
> In your original description you sorta claim that only the `drm` module gets added. But I think that is only because of your filtering of the lsinitrd output. I presume that a whole bunch of other drm modules are also getting added ?

Firstly, I'm sorry, but the description was copied from the original 1-line PR. I have made some changes since, but the idea is the same; added dependency in between dracut modules.

Also, I'm not aware that it actually says it adds only the drm module exclusively, but, more importantly: here module = dracut module; which also pulls dependencies (modules, binaries, and various files it might as well; recursively; if needed).

devkontrol:
+ @Vogtinator  - based on valuable feedback on other PRs - for visibility and potentially for more input on this. Thanks

Vogtinator:
AFAICT this would pull in the drm and simpledrm modules even if plymouth is not installed?

IMO that's a major regression in initrd size, as it will pull in all kinds of DRM modules and GPU firmware by default that way.

Rolv-Apneseth:
> IMO that's a major regression in initrd size, as it will pull in all kinds of DRM modules and GPU firmware by default that way.

Potentially relevant: https://github.com/dracut-ng/dracut-ng/issues/2362#issuecomment-4243571437

jwrdegoede:
> Diff: https://gist.github.com/pvalena/5ae5dca31b5410d89a377febfc612937

Thank you for providing this and sorry for being slow to respond. So what is new here is the addition of the bochs.ko kernel module.

I was confused before because when you mentioned adding the drm module I though you mentioned adding the drm.ko kernel module (which appears to be builtin on your kernel), but I now see you meant the drm dracut module. In cases where there can be confusion like this please fully spell out "dracut module" when talking about a dracut module.

So it seems that loading the bochs drm kernel module which will use fbcon as console, likely replacing vgacon, makes things work. This suggests that you are using a font which is not compatible with vgacon. So a less invasive fix would be to select a font which is compatible with vgacon.

An alternative fix would be to use the new vesa drivers/gpu/drm/sysfb/vesadrm.c driver which is a drm driver for vesa framebuffers and the bochs compatible VGA card in your vm should support a vesa framebuffer. This will give you fbcon over the vesadrm fb which should be just as compatible as doing this through the bochs drm driver without requiring the addition of drm kernel modules to the initramfs.

IMHO adding drm kernel modules to the base initramfs without plymouth as this PR does is not acceptable since for non-hostonly / generic initramfs builds this will make the initramfs absolutely huge since the amdgpu and especially the nova drm drivers require huge firmwares to get added to the initramfs.

Also all in all I still miss the following here:

1. A proper use-case description / description of the problem this tries to fix. You mention that rd.break shows the font not being loaded, but why is this actually a problem ? We don't put translations in the initramfs, so the default font should work fine, at least until the switch to the real rootfs happens ?   
2. A proper root-cause analysis of why adding the drm modules helps. As mentioned I suspect the problem is that setfont fails because the vgacon console driver is in use at this point and the selected font is not vgacon compatible, but why not select another vgacon compatible font ?

PR #1987, opened 2025-12-28
feat(dracut): disable xattr when creating cpios

### LINKED ISSUE #1561: Don't try to set security.selinux xattrs ever
Today dracut forks off `cp` to copy content from the build/host system into the initramfs. For systems using SELinux, this includes the `security.selinux` xattr.

For use cases where dracut is invoked with full root privileges on a client system, this works just fine.

However for cases where dracut is running with lower privileges such as in a container as part of a server-side build (e.g. podman/docker build), it may not be allowed to set the `security.selinux` xattr. 

One can set `DRACUT_NO_XATTR` to suppress this, but that's a large hammer as it disables *all* xattrs including e.g. `security.capability`.

Bigger picture though, I am not aware of any use case for setting `security.selinux` at all in the initramfs; systemd only loads policy from the real root when transitioning.
(If we were to handle labels there it'd be probably be easier to have it just all be mounted as `initramfs_t` or so)

---

Now there's a much bigger way to fix all this which is to switch to constructing the cpio archive directly in memory. Then if we wanted to we could write whatever labels are desired at that time.

jozzsi:
## Changes

Disable xattr when creating cpio, but do not change the default for squashfs/erofs.

Move setting the default for `DRACUT_CP` later in the code to account for the newly set defaults.

## Checklist

Fixes #1561

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1474,12 +1474,6 @@ fi
 
 export LC_MESSAGES=C kernel
 
-if [[ $EUID == "0" ]] && ! [[ ${DRACUT_NO_XATTR-} ]]; then
-    export DRACUT_CP="cp --reflink=auto --preserve=mode,timestamps,xattr,links -dfr"
-else
-    export DRACUT_CP="cp --reflink=auto --preserve=mode,timestamps,links -dfr"
-fi
-
 if ! [[ ${dracutbasedir-} ]]; then
     dracutbasedir=${BASH_SOURCE[0]%/*}
     [[ $dracutbasedir == dracut-functions* ]] && dracutbasedir="."
@@ -2783,6 +2777,17 @@ if [[ $kernel_only != yes ]]; then
 fi
 
 dracut_module_included "squash-lib" && mkdir -p "$squashdir"
+
+# disable xattr when creating cpio, but do not change the default for squashfs/erofs
+if ! dracut_module_included "squash-lib"; then
+    export DRACUT_NO_XATTR="${DRACUT_NO_XATTR:=1}"
+fi
+
+if [[ $EUID == "0" ]] && ! [[ ${DRACUT_NO_XATTR-} ]]; then
+    export DRACUT_CP="cp --reflink=auto --preserve=mode,timestamps,xattr,links -dfr"
+else
+    export DRACUT_CP="cp --reflink=auto --preserve=mode,timestamps,links -dfr"
+fi
 
 _isize=0 #initramfs size
 modules_loaded=" "


jozzsi:
> SELinux that depends on it

Is it ? From the referenced bug quoting @cgwalters 

>  am not aware of any use case for setting `security.selinux` at all in the initramfs; systemd only loads policy from the real root when transitioning.

Defaults are important and powerful. Distributions and users downstream started to flip this default already (like Fedora-bootc). What is the upside/benefit of having different Fedora based distributions running with different dracut defaults ?

From a quick look of it, it seems even cp does not preserve `xattr` by default. What makes `dracut` choosing a different default than `cp`? Is it really what most users want by default ?

dracut.conf.d/fedora/01-dist.conf
```diff
@@ -16,3 +16,6 @@ hostonly="yes"
 hostonly_cmdline="no"
 early_microcode="yes"
 reproducible="yes"
+
+# enable xattr
+export DRACUT_NO_XATTR=0
```
cgwalters: I think it'd be cleaner ot leave this option unset in the default config, and if it's not set, we conditionalize on the output format: if it's cpio, then disable xattrs.

cgwalters:
> SELinux is not what I'm worried about. I'm worried about everything else. [Since Fedora 15](https://fedoraproject.org/wiki/Features/RemoveSETUID), generally new stuff in Fedora uses file capabilities to get privileges as applications, and those are done as extended attributes on the filesystem too.

I think what needs to be emphasized heavily here is that the cpio format (and the Linux kernel's parser) *do not support xattrs* - they never have. There *are* however out of tree patches to do so, see https://github.com/dracut-ng/dracut-ng/issues/1561#issuecomment-3179438340 and https://lwn.net/Articles/747473/ etc.

dracut is just pointlessly trying to copy the xattrs to the staging directory, only to ignore them when generating the cpio.

dracut.conf.d/fedora/01-dist.conf
jozzsi: @cgwalters - this is also what I think. 

That is what I had as the first version of this PR, which received some concerns specifically for Fedora based distributions, so I decided to punt the decision for Fedora for a follow-up PR.

I am glad to received contrasting viewpoints from Fedora, but this should not hold up the decision for non-Fedora based distribution.

cgwalters: As is the PR title and commit message also need to be changed from "feat(dracut): disable xattr by default" to something more like "feat(dracut): disable xattrs for Fedora by default"

But how about instead adding `DRACUT_NO_XATTR=if-cpio` or so? Semantics are that xattrs are ignored only if the output format is cpio, and then make that the default? People who want to use those out-of-tree patches for cpio+xattr could set `DRACUT_NO_XATTR=0` or so.

Then we'd continue to support `DRACUT_NO_XATTR=1` even for squashfs (if someone wanted that).

I'm just lukewarm at best on this upstream-patch-for-single-distro; we can just as trivially patch it in the Fedora dracut package too...

jozzsi: Thanks @cgwalters . Please take a second look, as the PR intent to do the opposite - change the default for all distributions except Fedora (as the Fedora maintainer seems to be concerned about making this change for Fedora). 

See the commit message

> Do not change the default for Fedora-based distributions and keep xattr enabled by default.

jozzsi: Removed the Fedora specific changes based on latest [feedback](https://github.com/dracut-ng/dracut-ng/pull/1987#issuecomment-3712322724) from @Conan-Kudo .

jozzsi:
> so let's just disable it in code for cpio regardless of generator.

Ok, thanks @Conan-Kudo . Updated the PR and removed the Fedora specific changes.

LaszloGombos (approved):

Conan-Kudo (approved):

PR #1989, opened 2026-01-02
fix(dracut): avoid calling dwarning before dracut-logger is sourced

aafeijoo-suse:
Follow-up for 8d9887bd773936cf3e42798dd26d70edd4213d4d

## Checklist

Fixes: https://bugs.debian.org/1124479

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1362,7 +1362,7 @@ fi
 if [[ ${hostonly-} ]]; then
     for i in /sys /proc /run /dev; do
         if ! findmnt --target "$i" &> /dev/null; then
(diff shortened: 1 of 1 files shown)


bdrung:
I had a deeper look and this place would not the only place where `printf` is used directly. But since logging can be configured, it is better to use the logging functions. Moving sourcing `dracut-functions.sh` earlier is possible. So I created https://github.com/dracut-ng/dracut-ng/pull/1993 instead.

I am closing this PR in the assumption that the other PR will be accepted. Otherwise we can reopen it.

PR #1990, opened 2026-01-04
ci(alpine): install openRC and build-base

jozzsi:
alpine-base package is required to make the container bootable and have the openRC services available.

The alpine-base package is not installed by default in the alpine container (only the alpine-baselayout package is installed).

Instead of explicitly installing make, install build-base meta-package for future proofing.

## Checklist

Fixes #


LaszloGombos (approved):
LGTM

PR #1991, opened 2026-01-05
fix(dracut): avoid calling dwarning before dracut-logger is sourced

jozzsi:
Follow-up for 8d9887b

## Checklist

Fixes https://bugs.debian.org/1124479


bdrung:
How about moving sourcing `dracut-functions.sh` earlier: https://github.com/dracut-ng/dracut-ng/pull/1993?

PR #2003, opened 2026-01-07
feat: always preserve file ownership

jozzsi:
## Changes

Currently file ownership is almost always preserved, except when DRACUT_CP variable is used.

Make all file copying consistent and always preserve file ownership by passing `-p` to `cp`.

Follow-up to 9ef73b6ad .

## Checklist

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -437,10 +437,13 @@ static int cp(const char *src, const char *dst)
 
 normal_copy:
         pid = fork();
-        const char *preservation = (geteuid() == 0
-                                    && no_xattr == false) ? "--preserve=mode,xattr,timestamps,ownership" : "--preserve=mode,timestamps,ownership";
+        bool preservation = geteuid() == 0 && no_xattr == false;
+
         if (pid == 0) {
-                execlp("cp", "cp", "--reflink=auto", preservation, "-fL", src, dst, NULL);
+                if (preservation)
+                        execlp("cp", "cp", "--reflink=auto", "--preserve=xattr", "-fLp", src, dst, NULL);
+                else
+                        execlp("cp", "cp", "--reflink=auto", "-fLp", src, dst, NULL);
                 _exit(errno == ENOENT ? 127 : 126);
         }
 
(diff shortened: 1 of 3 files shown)


LaszloGombos (dismissed):

src/install/dracut-install.c
```diff
@@ -438,9 +438,9 @@ static int cp(const char *src, const char *dst)
 normal_copy:
         pid = fork();
         const char *preservation = (geteuid() == 0
-                                    && no_xattr == false) ? "--preserve=mode,xattr,timestamps,ownership" : "--preserve=mode,timestamps,ownership";
+                                    && no_xattr == false) ? "--preserve=xattr" : "-p";
```
bdrung: This is inconsistent: in the second case you set `-p` which is set later anyway.

jozzsi: Thanks for pointing it out @bdrung . I am trying to minimize the changes to help with the review. I removed the duplicate `-p`.

src/install/dracut-install.c
```diff
@@ -438,7 +438,7 @@ static int cp(const char *src, const char *dst)
 normal_copy:
         pid = fork();
         const char *preservation = (geteuid() == 0
-                                    && no_xattr == false) ? "--preserve=mode,xattr,timestamps,ownership" : "--preserve=mode,timestamps,ownership";
+                                    && no_xattr == false) ? "-p --preserve=xattr" : "-p";
```
bdrung: `-p --preserve=xattr` is passed as one parameter instead of two separate ones.

jozzsi: Ok, introduced an if/else case, hope this will work. Thanks @bdrung

bdrung: In this case I would get rid of the preservation variable completely.

bdrung: or make the preservation variable a boolean.

bdrung: Oh, preservation is used twice. Then only my second idea makes sense: make preservation a boolean and use it only for the if-condition check.

jozzsi: > make the preservation variable a boolean.

Yes, this change made the code more readable. Done. Thanks @bdrung

bdrung (dismissed):

LaszloGombos (approved):

PR #2004, opened 2026-01-07
fix(dracut): ensure that inst is called after DRACUT_NO_XATTR is computed

jozzsi:
## Changes

`inst` calls dracut-install which requires DRACUT_NO_XATTR.

Move code block around to make sure DRACUT_NO_XATTR is set before `inst` is called.

Follow-up to 941b18d .

## Checklist


bdrung (approved):
LGTM. I saw no problematic code between the old and new location of the moved code.

LaszloGombos (approved):

PR #2005, opened 2026-01-07
test: use syntax compatible with busybox implementation

jozzsi:
## Changes

`busybox` `du` only support `-s` for summary and `-b` for block-size.

Follow-up to d4af288f0 .

## Checklist

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -183,9 +183,9 @@ build_ext4_image() {
     local label="$3"
 
     local size
-    size="$(du --summarize --block-size 1 "${source_dir}" | cut -f 1)"
-    # Add 8 MB for ext4 journal
-    size="$((size + 8000000))"
+    size="$(du -s -b "${source_dir}" | cut -f 1)"
+    # Add 10 MB for ext4 journal
+    size="$((size + 10000000))"
 
     rm -f "$image"
     truncate -s "$size" "$image"


LaszloGombos (approved):
ok

test/test-functions
```diff
@@ -183,9 +183,9 @@ build_ext4_image() {
     local label="$3"
 
     local size
-    size="$(du --summarize --block-size 1 "${source_dir}" | cut -f 1)"
-    # Add 8 MB for ext4 journal
-    size="$((size + 8000000))"
+    size="$(du -s -b "${source_dir}" | cut -f 1)"
+    # Add 10 MB for ext4 journal
```
bdrung, line 187: Why do you increase the additional size from 8 to 10 MB? The commit does not give a reason.

jozzsi: `-b` seems to report a lower number than `--block-size 1`. The increase was necessary to pass the test.

bdrung: Ah, that makes sense now. `-b` is equivalent to `--apparent-size --block-size=1` and `--apparent-size` says:
> print apparent sizes, rather than disk usage although the apparent size is usually smaller, it may be larger due to holes in ('sparse') files, internal fragmentation, indirect blocks, and the like

PR #2007, opened 2026-01-07
ci: add a test run for basic tests using busybox

jozzsi:
## Changes

This test prevents regressions when running basic tests on Alpine with `busybox`.

## Checklist


bdrung:
Have you considered building a Alpine container that only contains busybox?

LaszloGombos (approved):

PR #2009, opened 2026-01-08
ci(alpine-busybox): alpine container with busybox 

jozzsi:
## Changes

The `coreutils` binaries are not installed and only `busybox` binaries are available.

This container is meant to prevent regressions when running basic test with `busybox`.

## Checklist

### DIFF
--- a/.github/workflows/container.yml
+++ b/.github/workflows/container.yml
@@ -42,6 +42,7 @@ jobs:
                     - {dockerfile: 'Dockerfile-debian', tag: 'ubuntu:devel'}
                     - {dockerfile: 'Dockerfile-debian', tag: 'ubuntu:rolling'}
                     - {dockerfile: 'Dockerfile-alpine', tag: 'alpine:edge'}
+                    - {dockerfile: 'Dockerfile-alpine-busybox', tag: 'alpine-busybox:edge'}
                     - {dockerfile: 'Dockerfile-void', tag: 'void:latest'}
                 exclude:
                     - config: {tag: 'arch:latest'}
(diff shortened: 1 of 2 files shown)


LaszloGombos (dismissed):

.github/workflows/container.yml
```diff
@@ -42,6 +42,7 @@ jobs:
                     - {dockerfile: 'Dockerfile-debian', tag: 'ubuntu:devel'}
                     - {dockerfile: 'Dockerfile-debian', tag: 'ubuntu:rolling'}
                     - {dockerfile: 'Dockerfile-alpine', tag: 'alpine:edge'}
+                    - {dockerfile: 'Dockerfile-tiny', tag: 'tiny:latest'}
```
bdrung: Why not name it:

```suggestion
                    - {dockerfile: 'Dockerfile-alpine-busybox', tag: 'alpine-busybox:edge'}
```

jozzsi: ok, made the change. Thanks @bdrung

test/container/Dockerfile-alpine-busybox
```diff
@@ -0,0 +1,17 @@
+# Tiny container with busybox
```
bdrung, line 1: Maybe add that GNU coreutils should not be installed?

jozzsi: ok, done !

Conan-Kudo (approved):
Looks reasonable to me, it's good to have a test against busybox, at the very least it'll help us file bug reports with them if the behavior they have should be improved. 😄

LaszloGombos (approved):

PR #2012, opened 2026-01-08
fix(systemd-networkd): ensure zzzz-dracut-default.network cleanup

jozzsi:
## Changes

fix(systemd-networkd): ensure zzzz-dracut-default.network cleanup                                                                                                                                                                     
                                                                                                                                                                                                                                      
zzzz-dracut-default.network should be only installed when systemd-network-generator does not generate any files in                                                                                                                                                                              
files in /run/systemd/network .                                                                                                                                                                                                        
                                                                                                                                                                                                                                      
Place dracut-default.network in /usr/lib/dracut/ and do not change /usr as part of managing .network files in initramfs.   
                                        
Follow-up to 523c9b0 .

## Checklist


chewi (changes requested):
`systemdnetwork` isn't currently written to /etc/conf.d/systemd.conf.

jozzsi:
@chewi WDYT ?

chewi (approved):
/lib/99-default.network is a bit weird, but a clever solution nonetheless!

I must admit that I don't know if Flatcar would even hit this path, and I'm not familiar with systemd-networkd. I only spotted the issue while comparing the initrd contents against an earlier Dracut version.

modules.d/11systemd-networkd/networkd-config.sh
```diff
@@ -19,10 +19,17 @@ for f in /run/systemd/network/*.network; do
         echo "RequestOptions=59 60"
     } >> "$f"
 
-    # Remove the default network if at least one was generated
-    rm -f "$systemdnetworkconfdir"/zzzz-dracut-default.network
+    # at least one was generated
+    : > /run/systemd/network/generated
```
bdrung: Is this file only used for this script? Why not use a variable instead then?

jozzsi: Yes, of course. Done !

modules.d/11systemd-networkd/networkd-config.sh
bdrung, line 10: Just to be safe, I would initialize `generated`.

modules.d/11systemd-networkd/module-setup.sh
```diff
@@ -63,8 +63,7 @@ install() {
     inst_simple "$moddir"/99-wait-online-dracut.conf \
         "$systemdsystemunitdir"/systemd-networkd-wait-online.service.d/99-dracut.conf
 
-    inst_simple "$moddir"/99-default.network \
-        "$systemdnetwork"/zzzz-dracut-default.network
+    inst_simple "$moddir"/99-default.network /usr/lib/99-default.network
```
bdrung: Would it make sense to install this file in a dracut namespace? Do we other places where we do a similar thing?

jozzsi: Change the filename (in the src tree and in the generated /usr) to `dracut-default.network` from `99-default.network`

modules.d/11systemd-networkd/networkd-config.sh
jozzsi: Ok, initialized to 0.

modules.d/11systemd-networkd/module-setup.sh
bdrung: Good change, but my question aimed at using `/usr/lib` in the initrd. Why `/usr/lib` and not something else?

jozzsi: Good point. 

`/usr/lib` --> `/usr/lib/dracut`

LaszloGombos (approved):

bdrung (approved):

PR #2014, opened 2026-01-09
feat(dracut): allow hostonly mode if /sys /proc or /run is missing

jozzsi:
## Changes

Missing /sys, /proc, or /run no longer forces non-hostonly mode. Hostonly initramfs generation still requires /dev to be mounted and  continues to work in chroot environments where these filesystems may be absent.                                                                                                                                                                                                                            
                                                                                                                                                                                                                                      
If /sys is missing, dracut emits a warning but does not fall back to non-hostonly mode. Hostonly generation should not depend on /sys, /proc, or /run being mounted.

## Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1362,12 +1362,12 @@ else
 fi
 
 if [[ ${hostonly-} ]]; then
-    for i in /sys /proc /run /dev; do
-        if ! findmnt --target "$i" &> /dev/null; then
-            dwarning "Turning off host-only mode: '$i' is not mounted!"
-            unset hostonly
-        fi
-    done
+    if ! findmnt --target "/dev" &> /dev/null; then
+        dwarning "Turning off host-only mode: /dev is not mounted!"
+        unset hostonly
+    elif ! findmnt --target "/sys" &> /dev/null; then
+        dwarning "Hostonly mode with no /sys mounted!"
+    fi
 fi
 
 case $hostonly_mode in


Conan-Kudo:
This is a pretty significant change, what's the rationale for this?

jozzsi:
> Yes, what some of our modules install in hostonly mode depends on info gathered from `/sys`, so a half-baked hostonly initrd created on a system without `/sys` available does not make any sense.

I do not know what is "does not make sense", but it is indeed correct that info gathered from `/sys` by dracut. My expectation here is that if `dracut` does not handle non-existent /sys during generation (not during runtime), than that should be treated as a bug.

Let's look at an example

https://github.com/dracut-ng/dracut-ng/blob/d3c3d9b073bb9a6870570da791e335566321b07f/modules.d/70bluetooth/module-setup.sh#L16

Here dracut is correctly checking the existence of `/sys` and if `/sys` does not exists, than dracut makes a decision not to warn user about the bluetooth module to in the generated initramfs - as expected. I would think that we would want to allow this usage of dracut.

If of course some Linux distributions makes strong guarantees that /sys is never empty, that is great, but that is not the place for dracut to decide. For those Linux distributions, this PR is a no-op already.

I am willing to add back /sys to the PR, so let's not turn this into a convoluted discussion. I feel more strongly about dropping /proc check from the codebase than about /sys.

bdrung:
I am fine with only checking for `/dev` under following conditions:
* the PR commit message is updated with the reasons given in this discussion (dracut hostonly should work without /sys mounted, otherwise it's a bug; /sys and /proc is optional; not being to opinionated)
* check or test the other places where `/sys` is queried. In case you see a bug, an issue or PR is filed

dracut.sh
```diff
@@ -1362,12 +1362,12 @@ else
...
-        if ! findmnt --target "$i" &> /dev/null; then
-            dwarning "Turning off host-only mode: '$i' is not mounted!"
-            unset hostonly
-        fi
-    done
+    if ! findmnt --target "/dev" &> /dev/null; then
+        dwarning "Turning off host-only mode: /dev is not mounted!"
+        unset hostonly
+    elif ! findmnt --target "/sys" &> /dev/null; then
+        dwarning "Hostonly mode with no /sys mounted!"
```
bdrung, line 1369: Printing a warning is a good idea.

bdrung (commented):
I did a quick check for `/run` and I found nothing problematic there.

```
grep -w /run $(find -name module-setup.sh)
```

I haven't checked /proc (five matches) or /sys (a long list) yet:

```
grep -w /proc $(find -name module-setup.sh)
grep -w /sys $(find -name module-setup.sh)
```

bdrung (approved):

LaszloGombos (approved):

PR #2016, opened 2026-01-09
ci: modularize GitHub Actions for daily tests

jozzsi:
## Changes

Move each job into its own GA yaml file so that each job can be individually enabled/disabled via the GitHub web interface.

This change prevents triggering all test runs every time the monolithic yaml file is modified.

## Checklist


LaszloGombos (approved):
Ok, I think this PR will make it easier to see the failing tests.

Perhaps we can combine the same x64 and arm64 tests into the same GA as a follow-up.

LaszloGombos (approved):

PR #2017, opened 2026-01-09
refactor: move defining inst_sysusers to dracut-functions.sh

jozzsi:
## Changes

`inst_sysusers` is a public interface to dracut modules, move it to `dracut-functions.sh`.

Document `inst_sysusers` in the man page as from now on it is a public interface.

## Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1304,6 +1304,15 @@ inst_rules() {
     done
 }
 
+# install sysusers files
+inst_sysusers() {
+    inst_multiple -o "$sysusers/$*" "$sysusers/acct-*-$*"
+
+    if [[ ${hostonly-} ]]; then
+        inst_multiple -H -o "$sysusersconfdir/$*" "$sysusers/acct-*-$*"
+    fi
+}
+
 # inst_libdir_dir <dir> [<dir>...]
 # Install a <dir> located on a lib directory to the initramfs image
 inst_libdir_dir() {
--- a/dracut.sh
+++ b/dracut.sh
@@ -1633,16 +1633,6 @@ build_ld_cache() {
     fi
 }
 
-# install sysusers files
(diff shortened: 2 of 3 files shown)


LaszloGombos (dismissed):

dracut-functions.sh
```diff
@@ -1304,6 +1304,16 @@ inst_rules() {
     done
 }
 
+# install sysusers files
+# shellcheck disable=SC2317,SC2329
```
bdrung: The shellcheck override is not needed here, is it?

dracut-functions.sh
```diff
@@ -1304,6 +1304,16 @@ inst_rules() {
...
+inst_sysusers() {
```
bdrung, line 1308: Didn't you want to add the functions in the order they are documented in the man page?

jozzsi: Changed the order of the documentation instead - to keep it ordered both alphabetically and somewhat chronologically . Thanks for noticing !

dracut-functions.sh
jozzsi: nice. Removed !

LaszloGombos (dismissed):

dracut-functions.sh
```diff
@@ -1304,6 +1304,15 @@ inst_rules() {
     done
 }
 
+# install sysusers files
+inst_sysusers() {
+    inst_multiple -o "$sysusers/$*" "$sysusers/acct-*-$*"
```
bdrung, line 1309: Looking at that code using `$*`: That does not support specifying multiple files. Can you update the documentation?

jozzsi: Thanks. Done !

bdrung (approved):

LaszloGombos (approved):

PR #2041, opened 2026-01-13
feat: rename rd.live.overlay to rd.overlay

nadzyah:
## Changes

The overlay functionality can be used with both dmsquash-live and overlayfs modules, and is not limited to live images. This change improves naming of command line options with better namespace scoping.

Renames:
- `rd.live.overlay` -> `rd.overlay`
- `rd.live.overlay.reset` -> `rd.overlay.reset`

Backward compatibility is maintained through deprecation flags.

Related to #1929 

## Checklist

### DIFF
--- a/test/TEST-30-DMSQUASH/test.sh
+++ b/test/TEST-30-DMSQUASH/test.sh
@@ -60,6 +60,8 @@ check_autooverlay_marker() {
 # Reset root.img to single partition state for autooverlay test
 reset_overlay_partition() {
     test_marker_reset
+    sfdisk --delete "$TESTDIR"/root.img 2 2> /dev/null || true
+    dd if=/dev/zero of="$TESTDIR"/root.img bs=1M seek=320 count=50 conv=notrunc status=none
     local rootPartitions
     rootPartitions=$(sfdisk -d "$TESTDIR"/root.img | grep -c 'root\.img[0-9]')
     [ "$rootPartitions" -eq 1 ]
--- a/test/TEST-30-DMSQUASH/test.sh
+++ b/test/TEST-30-DMSQUASH/test.sh
@@ -43,6 +43,28 @@ client_run() {
     client_test_end
 }
 
+# Verify autooverlay created a second partition and wrote the marker to it.
+# Extract only the second partition to avoid false positives from the test
+# script in the first partition.
+check_autooverlay_marker() {
+    local rootPartitions part2_info part2_start part2_size
+    rootPartitions=$(sfdisk -d "$TESTDIR"/root.img | grep -c 'root\.img[0-9]')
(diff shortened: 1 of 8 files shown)


jozzsi (approved):
non-owner LGTM

LaszloGombos (dismissed):

bdrung (commented):
I thought about the naming (`rd.overlayfs` vs `rd.overlay`) and came to the conclusion that `rd.overlay` is the better name.

There is one case of `rd.live.overlay` in `test/TEST-30-DMSQUASH/test-init.sh` which needs adjustment. This hasn't been caught by the test case due to bug
https://github.com/dracut-ng/dracut-ng/issues/2054

There are more setting below `rd.live.overlay`:

* rd.live.overlay.nouserconfirmprompt
* rd.live.overlay.cowfs
* rd.live.overlay.size
* rd.live.overlay.thin

Shouldn't we rename those as well?

bdrung:
One more: IMO we should rename `rd.overlayfs.readonly` to `rd.overlay.readonly` and merge `rd.overlayfs` into `rd.overlay` in a follow-up PR.

bdrung:
> so maybe the deprecated stuff should exist in a backward compat module that's always dragged in by the things that need to support them?

It's just deprecated names. How can that be ripped out and put into a separate module? Can you show how that should look like? Creating a backward compat module could be done in a follow-up PR.

test/TEST-30-DMSQUASH/test.sh
```diff
@@ -43,6 +43,30 @@ client_run() {
...
+    part2_size=$(echo "$part2_info" | sed -n 's/.*size= *\([0-9]*\).*/\1/p')
+    dd if="$TESTDIR"/root.img of="$TESTDIR"/overlay-part.img bs=512 skip="$part2_start" count="$part2_size" status=none
+    test_marker_check dracut-autooverlay-success overlay-part.img
+}
+
+# Reset root.img to single partition state for autooverlay test
+reset_overlay_partition() {
+    test_marker_reset
+    sfdisk --delete "$TESTDIR"/root.img 2 2> /dev/null || true
+    dd if=/dev/zero of="$TESTDIR"/root.img bs=1M seek=320 count=50 conv=notrunc status=none
```
bdrung, line 64: These two lines are behaviour change and not a refactor. Please put this change in a separate commit.

bdrung (changes requested):
LGMT, but there are two places that still refer to rd.live.overlay:

* test/TEST-60-NFS/client-init.sh
* man/dracut.usage.adoc

LaszloGombos (dismissed):

bdrung (approved):

LaszloGombos (approved):

PR #2050, opened 2026-01-14
chore(dracut): remove support for 255 error for is_qemu_virtualized   

jozzsi:
## Changes

Treat the return value as a boolean and not an integer, to simplify the API before potentially exposing this API as a public API.   

## Checklist

Fixes #

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2059,16 +2059,15 @@ fi
 
 # shellcheck disable=SC2317,SC2329
 is_qemu_virtualized() {
-    # 0 if a virt environment was detected
-    # 1 if a virt environment could not be detected
-    # 255 if any error was encountered
+    # 0 if a QEMU/KVM like VM virtualization environment was detected
+    # 1 if a QEMU/KVM like VM virtualization environment could not be detected (including if an error was encountered)
 
     # do not consult /sys and do not detect virt environment in non-hostonly mode
     ! [[ ${hostonly-} ]] && return 1
 
     if type -P systemd-detect-virt > /dev/null 2>&1; then
         if ! vm=$(systemd-detect-virt --vm 2> /dev/null); then
-            return 255
+            return 1
         fi
         [[ $vm == "qemu" ]] && return 0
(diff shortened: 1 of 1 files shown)


dracut.sh
```diff
@@ -2059,16 +2059,15 @@ fi
 
 # shellcheck disable=SC2317,SC2329
 is_qemu_virtualized() {
-    # 0 if a virt environment was detected
-    # 1 if a virt environment could not be detected
-    # 255 if any error was encountered
+    # 0 if a vm virt environment was detected
+    # 1 if a vm virt environment could not be detected (including if an error was encountered)
```
```suggestion
    # 0 if a VM virtualization environment was detected
    # 1 if a VM virtualization environment could not be detected (including if an error was encountered)
```

suggestion: Maybe make "including if an error was encountered" more specific that it returns 1 if calling systemd-detect-virt fails.

bdrung: Looking again: this function does not detect all kinds of VM virtualization. It looks only for QEMU based ones.

jozzsi: ok, fixed.

dracut.sh
```diff
@@ -2059,16 +2059,15 @@ fi
 
 # shellcheck disable=SC2317,SC2329
 is_qemu_virtualized() {
-    # 0 if a virt environment was detected
-    # 1 if a virt environment could not be detected
-    # 255 if any error was encountered
+    # 0 if a qemu/kvm like vm virt environment was detected
+    # 1 if a emu/kvm like vm virt environment could not be detected (including if an error was encountered)
```
bdrung: Since QEMU, KVM, etc are proper names, I suggest to use their capitalization in the comments:
```suggestion
    # 0 if a QEMU/KVM like VM virtualization environment was detected
    # 1 if a QEMU/KVM like VM virtualization environment could not be detected (including if an error was encountered)
```

jozzsi: ok, thanks

bdrung (approved):

Conan-Kudo (approved):

PR #2056, opened 2026-01-15
feat(xen): introduce a new dracut module for xen kernel drivers

### LINKED ISSUE #2051: Add support for xen drivers
Currently the following drives are only added to the generated initramfs if user or system adds them explicitly via additional configurations.

The issue is to see if dracut can autodetect and auto-add these drivers.

- xen-blkfront 
- xen-netfront

jozzsi:
## Changes

This PR introduces a dedicated dracut module for adding xen kernel drivers to the generated initramfs.

## Checklist

Fixes #2051


LaszloGombos:
Found this earlier commit - 028a79b . Perhaps we can get more information from the Xen project on this.

PR #2058, opened 2026-01-15
feat(qemu-net): auto-add qemu-net if net-lib and qemu included

jozzsi:
## Changes

In the interest of improving defaults, if network and qemu dracut modules are included than automatically include dracut-net dracut modules as well.

I tested with the following dracut invocation

`dracut -a "network qemu"`

## Checklist

### DIFF
--- a/modules.d/70qemu-net/module-setup.sh
+++ b/modules.d/70qemu-net/module-setup.sh
@@ -2,6 +2,11 @@
 
 # called by dracut
 check() {
+    if [[ $hostonly_mode != "strict" ]] && dracut_module_included "net-lib" \
+        && dracut_module_included "qemu"; then
+        return 0
+    fi
+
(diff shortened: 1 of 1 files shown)


modules.d/70qemu-net/module-setup.sh
```diff
@@ -2,6 +2,11 @@
 
 # called by dracut
 check() {
+    if dracut_module_included "network" \
```
bdrung: Should we also consider the other network modules?

jozzsi: Changed it to `net-lib` instead of `network`.

jozzsi:
> We shouldn't include qemu in case the initrd is built with hostonly-mode strict.

Good point. Added an additional check.

PR #2062, opened 2026-01-16
test(DRACUT-CPIO): correct test_check condition

jozzsi:
## Changes

Partial revert to correct test_check condition.

The current code checks if dracut-cpio is found in PATH and skips  the test in case dracut-cpio is found. 

Follow-up to b687756 .

## Checklist


bdrung:
The commit message should explain why the revert was necessary.

LaszloGombos (approved):

Conan-Kudo (approved):

PR #2063, opened 2026-01-16
feat(dracut): expose determine_kernel_image as a public function

jozzsi:
## Changes

Expose determine_kernel_image as a public and documented helper function.

Instead of repeating and reimplementing the function to find kernel image, call the public function.   

Follow-up to df93355 .

## Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1485,6 +1485,27 @@ require_kernel_modules() {
     return "$_ret"
 }
 
+determine_kernel_image() {
+    local kversion="$1"
+    local paths=(
+        "/lib/modules/${kversion}/vmlinuz"
+        "/lib/modules/${kversion}/vmlinux"
+        "/lib/modules/${kversion}/Image"
+        "/boot/vmlinuz-${kversion}"
+        "/boot/vmlinux-${kversion}"
+    )
+
+    for path in "${paths[@]}"; do
+        if [ -f "$path" ]; then
+            echo "$path"
+            return 0
+        fi
+    done
+
+    echo "Could not find a Linux kernel image for version '$kversion'!" >&2
+    return 1
+}
+
 if ! is_func dinfo > /dev/null 2>&1; then
     # shellcheck source=./dracut-logger.sh
     . "${BASH_SOURCE[0]%/*}/dracut-logger.sh"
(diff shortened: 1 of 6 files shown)


Conan-Kudo (dismissed):

dracut-functions.sh
```diff
@@ -1485,6 +1485,27 @@ require_kernel_modules() {
...
+    local paths=(
+        "${dracutsysrootdir-}/lib/modules/${kversion}/vmlinuz"
+        "${dracutsysrootdir-}/lib/modules/${kversion}/vmlinux"
+        "${dracutsysrootdir-}/lib/modules/${kversion}/Image"
+        "${dracutsysrootdir-}/boot/vmlinuz-${kversion}"
+        "${dracutsysrootdir-}/boot/vmlinux-${kversion}"
+    )
+
+    for path in "${paths[@]}"; do
+        if [ -f "$path" ]; then
```
bdrung: The `test -s` (from dracut.sh) is better than `test -f`.

dracut-functions.sh
```diff
@@ -1485,6 +1485,27 @@ require_kernel_modules() {
...
+            echo "$path"
+            return 0
+        fi
+    done
+
+    echo "Could not find a Linux kernel image for version '$kversion'!" >&2
```
bdrung, line 1505: Should be use `dfatal` here instead?

jozzsi: I feel this should be a decision of the caller - for now. I hope to iterate over the details of this function. At the dracut.sh call side, the caller turns this into a dfatal.

This is function is quite delicate, let's iterate over this as a follow-up.

dracut-functions.sh
jozzsi: ok, thanks, changed it !

test/test-functions
```diff
@@ -141,27 +141,6 @@ determine_kernel_version() {
...
-
-    for path in "${paths[@]}"; do
-        if [ -f "$path" ]; then
-            echo "$path"
-            return 0
-        fi
-    done
-
-    echo "Could not find a Linux kernel image for version '$kversion'!" >&2
-    exit 1
```
bdrung, line 162: Since it's a helper function, shouldn't we use return here instead?

jozzsi: yes we should. Fixed !

bdrung (approved):

LaszloGombos (approved):

PR #2068, opened 2026-01-17
ci: test UEFI generation on Alpine

jozzsi:
## Changes

Strengthen determine_kernel_version (this function only used during testing) to look for files ending with `modules.dep`                                                              
to filter out entries like e.g. `modules.dep.bin`. 

## Checklist

### DIFF
--- a/.github/workflows/daily-busybox-x64.yml
+++ b/.github/workflows/daily-busybox-x64.yml
@@ -28,6 +28,7 @@ jobs:
                 config:
                     - {test: '10', deps: ''}
                     - {test: '11', deps: 'btrfs-progs'}
+                    - {test: '12', deps: 'efistub ovmf squashfs-tools'}
                     - {test: '13', deps: ''}
                     - {test: '20', deps: 'lvm2 device-mapper'}
                     - {test: '26', deps: 'cryptsetup mdadm device-mapper lvm2 partx sed'}
--- a/.github/workflows/daily-kernel-install-x64.yml
+++ b/.github/workflows/daily-kernel-install-x64.yml
@@ -24,6 +24,7 @@ jobs:
             fail-fast: false
             matrix:
                 container:
+                    - alpine:edge
                     - arch:latest
                     - debian:latest
                     - debian:sid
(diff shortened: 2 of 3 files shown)


LaszloGombos (dismissed):

dracut-functions.sh
```diff
@@ -1493,6 +1493,8 @@ determine_kernel_image() {
         "${dracutsysrootdir-}/lib/modules/${kversion}/Image"
         "${dracutsysrootdir-}/boot/vmlinuz-${kversion}"
         "${dracutsysrootdir-}/boot/vmlinux-${kversion}"
+        "${dracutsysrootdir-}/boot/vmlinuz-"*
+        "${dracutsysrootdir-}/boot/vmlinux-"*
```
Conan-Kudo: I'm concerned that this might result in some weird outcomes. This is fairly greedy, and we deliberately don't do this for any of the other paths.

bdrung: Same concern here. I would expert the version to be part of the path.

jozzsi: The lookup is in priority order, so this change should not impact other Linux distributions. 

Also, I am not aware of "other parts", this is now the only place that needs to change (as demonstrated by the test case passing).

Unfortunately dracut does not have a command line argument for the kernel location, that would have been a workaround.

jozzsi: I will try to dig more on this, but here is the list of files for a kernel package - https://pkgs.alpinelinux.org/contents?name=linux-lts&repo=main&branch=edge&arch=x86_64

Conan-Kudo:
I suspect a versioned `vmlinuz`/`vmlinux` file in `/boot` would be a problem, because of space constraints. But most distributions contain a copy in the versioned kernel directory path too, so that there's at least a backup that can be re-copied into `/boot`. We already know how to pick that one up.

Conan-Kudo:
Discussed it with the Alpine folks, they're not opposed to fixing this. They asked me to file an issue about it, [so I have](https://gitlab.alpinelinux.org/alpine/aports/-/issues/17907).

dracut-functions.sh
jozzsi: Removed this change from this PR after our discussion.

LaszloGombos (approved):

PR #2072, opened 2026-01-17
fix: Include dkms path in dracut-install search path

morisja:
This pull request attempts to resolve #2071  by including the dkms module path in initramfs generation

## Checklist
- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [x] I am providing new code and test(s) for it
I have included code but have not included tests.

Fixes #2071

PR #2075, opened 2026-01-18
chore: remove Changes header from the tempate

jozzsi:
Remove Changes header from the tempate to make it easier to create PRs.

## Checklist

Fixes #

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,7 +1,5 @@
 This pull request changes...
 
-## Changes
-
 ## Checklist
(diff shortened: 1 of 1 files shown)


jozzsi:
IMHO this header does not add any value, and it can be confusing - see e.g. https://github.com/dracut-ng/dracut-ng/pull/2072

bdrung (approved):
I agree. I would even go one step further afterwards and move "Fixes" up and reduce the level for "## checklist"

LaszloGombos (approved):

PR #2077, opened 2026-01-19
ci: explicitly install python3 in containers

bdrung:
## Changes

Explicitly install python3 in the containers to allow testing the `livenet` module using Python 3 `http.server` as HTTP server. Many of the containers already have Python installed as dependency.

## Checklist

Fixes #

### DIFF
--- a/test/container/Dockerfile-arch
+++ b/test/container/Dockerfile-arch
@@ -52,6 +52,7 @@ RUN pacman --noconfirm -Syu \
     pciutils \
     pkgconf \
     plymouth \
+    python \
     qemu \
     qrencode \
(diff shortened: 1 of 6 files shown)


LaszloGombos (approved):

jozzsi:
This PR seems to regress debian:sid container - https://github.com/dracut-ng/dracut-ng/actions/runs/21154452879

bdrung:
Correlation is not causation. The failure was caused https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1125944 which has been fixed. I just retrigger generating the container image to pick up the fix.

PR #2088, opened 2026-01-20
feat(livenet): support root=http:* as alias for root=live:http:*

bdrung:
## Changes

For compatibility with `cloud-initramfs-rooturl` support `root=http:*` as alias for `root=live:http:*` and `root=https:*` as alias for `root=live:https:*`.

## Checklist

Fixes #


LaszloGombos (approved):

nadzyah (approved):
I don't have write access here, so my approval doesn't count, but I don't see any problem with this change. I've also checked that `cloud-initramfs-tools` have such an option
So, +1

Conan-Kudo (approved):

PR #2090, opened 2026-01-21
feat(tpm2-tss): improve autodetection

jozzsi:
## Changes

Include the `tpm2-tss` dracut module if TPM2 HW is available in  sloppy hostonly mode. 

## Checklist

### DIFF
--- a/modules.d/73tpm2-tss/module-setup.sh
+++ b/modules.d/73tpm2-tss/module-setup.sh
@@ -4,13 +4,15 @@
 
 # Prerequisite check(s) for module.
 check() {
-
     # If the binary(s) requirements are not fulfilled the module can't be installed.
     require_binaries tpm2 || return 1
 
-    # Return 255 to only include the module, if another module requires it.
-    return 255
+    # Include the tpm2-tss dracut module if TPM2 HW is available in sloppy hostonly mode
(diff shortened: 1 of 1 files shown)


modules.d/73tpm2-tss/module-setup.sh
```diff
@@ -4,13 +4,15 @@
...
 # Prerequisite check(s) for module.
 check() {
-
     # If the binary(s) requirements are not fulfilled the module can't be installed.
     require_binaries tpm2 || return 1
 
-    # Return 255 to only include the module, if another module requires it.
-    return 255
+    # Include the tpm2-tss dracut module if TPM2 HW is available in sloppy hostonly mode
+    if [[ $hostonly_mode == "sloppy" ]] && [ -d "/sys/class/tpmrm" ] && [[ "$(ls -A /sys/class/tpmrm/)" ]]; then
```
aafeijoo-suse, line 11: Having a TPM doesn't mean that it's being used to decrypt the rootfs, or IOW, this patch will only add this module (and **~5MB** more in zstd compressed initrds) in cases where is not necessary.

jozzsi: > this patch will only add this module (and ~5MB more in zstd compressed initrds) in cases where is not necessary.

Thanks for sharing this data, very useful. The 5MB number is not the same in all Linux distributions. 

There are several ways for a distribution or a user to opt out from this default 
1. set `hostonly_mode` to `strict`.
2. package tpm2-tss dracut module as a separate package and only install it when initramfs needs it

jozzsi:

PR #2095, opened 2026-01-22
fix(systemd-pcrphase): do not print an error if an optional binary is not found

### LINKED ISSUE #2094: systemd-pcrphase: Incorrect "command not found" warning when binary exists in alternative path
**Describe the bug**

modules.d/11systemd-pcrphase/module-setup.sh has a check for binaries:

https://github.com/dracut-ng/dracut-ng/blob/5dd694a4834ec7643b02d04d73f87d5c760b7ede/modules.d/11systemd-pcrphase/module-setup.sh#L9

Which checks both, but you get an incorrect warning in a log:

```
dracut[D]: Module 'systemd-pcrphase' will not be installed, because command '/usr/lib/systemd/systemd-pcrphase' could not be found!
```
although it gets included afterwards:

```
dracut[I]: *** Including module: systemd-pcrphase ***
```

**Distribution used**
Fedora rawhide

**Dracut version**
109

**Init system**
systemd

**To Reproduce**
`dracut -f -vvv`
log: https://gist.github.com/pvalena/eeb2f83264f03019eb98d5771ffa76fc

**Expected behavior**
no error

**Additional context**
-

aafeijoo-suse:
## Checklist

Fixes #2094

### DIFF
--- a/modules.d/11systemd-pcrphase/module-setup.sh
+++ b/modules.d/11systemd-pcrphase/module-setup.sh
@@ -6,8 +6,9 @@
 check() {
     # If the binary(s) requirements are not fulfilled the module can't be installed.
     # systemd-255 renamed the binary, check for old and new location.
-    if ! require_binaries "$systemdutildir"/systemd-pcrphase \
-        && ! require_binaries "$systemdutildir"/systemd-pcrextend; then
+    if ! require_any_binary \
+        "$systemdutildir"/systemd-pcrextend \
+        "$systemdutildir"/systemd-pcrphase; then
         return 1
     fi
 


modules.d/11systemd-pcrphase/module-setup.sh
```diff
@@ -6,8 +6,9 @@
 check() {
     # If the binary(s) requirements are not fulfilled the module can't be installed.
     # systemd-255 renamed the binary, check for old and new location.
-    if ! require_binaries "$systemdutildir"/systemd-pcrphase \
-        && ! require_binaries "$systemdutildir"/systemd-pcrextend; then
+    if ! require_any_binary \
+        "$systemdutildir"/systemd-pcrphase \
```
bdrung: Can we look for `systemd-pcrextend` first since that is the newer name?

aafeijoo-suse: Yes we can.

bdrung (approved):

LaszloGombos (approved):

PR #2064, opened 2026-01-16
test(DMSQUASH): fix autooverlay test false positive and setup

nadzyah:
## Changes

The autooverlay test was always passing due to multiple issues:

* The test searched for the marker string in the entire root.img, but the string also existed literally in `test-init.sh` embedded in the first partition, causing a false positive. Fix by extracting and searching only the second (overlay) partition
* `test-init.sh` tried to read `/proc/cmdline` but `/proc` was not mounted and the directory did not exist. Fix by creating and mounting `/proc` temporarily
* The test disk image contained raw rootfs instead of the live structure (LiveOS/rootfs.img), so FSIMG was empty and `do_live_overlay()` was never called. Fix by using the live structure on disk
* Add `sync` after writing the marker to ensure it is flushed to the persist partition before shutdown

## Checklist
- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [x] I am providing new code and test(s) for it

Fixes #2054

PR #2096, opened 2026-01-22
fix(dmsquash-live-autooverlay): re-read partition table after creation

nadzyah:
This PR is a follow-up on https://github.com/dracut-ng/dracut-ng/pull/2064#issuecomment-3782037835

## Changes

After parted creates the overlay partition, the kernel's in-memory partition table is not automatically updated. This causes mkfs to fail with "The file /dev/sdX does not exist" on some systems.

Add `blockdev --rereadpt` to force the kernel to re-read the partition table before attempting to create the filesystem.

## Checklist

### DIFF
--- a/modules.d/70dmsquash-live-autooverlay/create-overlay.sh
+++ b/modules.d/70dmsquash-live-autooverlay/create-overlay.sh
@@ -78,6 +78,7 @@ gatherData() {
 
 createPartition() {
     parted --script --align optimal "${blockDevice}" mkpart primary ${partitionStart}% 100%
+    blockdev --rereadpt "${blockDevice}"
 }
 
 createFilesystem() {


modules.d/70dmsquash-live-autooverlay/create-overlay.sh
```diff
@@ -97,6 +97,7 @@ createFilesystem() {
 main() {
     gatherData "$1"
     createPartition
+    blockdev --rereadpt "${blockDevice}"
```
bdrung: The main function is a high level function. Can you move that into the createPartition function?

bdrung: Do we include the `blockdev` command in the initrd? Maybe we could use partprobe (shipped by parted) instead?

nadzyah: It's installed by the `dmsquash-live` module (which `dmsquash-live-autooverlay` depends on)
https://github.com/dracut-ng/dracut-ng/blob/5dd694a4834ec7643b02d04d73f87d5c760b7ede/modules.d/70dmsquash-live/module-setup.sh#L25

bdrung: Oh, i missed that. Thanks.

modules.d/70dmsquash-live-autooverlay/create-overlay.sh
```diff
@@ -97,7 +99,6 @@ createFilesystem() {
 main() {
     gatherData "$1"
     createPartition
-    udevsettle
     createFilesystem
     udevsettle
```
bdrung, line 103: There is a second udevsettle. Let's keep moving udevsettle for a future PR.

bdrung (approved):

LaszloGombos (approved):

PR #2105, opened 2026-01-23
ci(LIVENET): enable livenet testing on Alpine

jozzsi:
## Changes

Test 31 (livenet) requires curl to be available in the test CI container.

## Checklist

Fixes #

### DIFF
--- a/test/container/Dockerfile-alpine
+++ b/test/container/Dockerfile-alpine
@@ -26,6 +26,7 @@ RUN apk add --no-cache \
     alpine-base \
     build-base \
     cargo \
+    curl \
     dnsmasq \
     dracut-tests \
     efistub \


.github/workflows/daily-busybox-x64.yml
```diff
@@ -32,6 +32,7 @@ jobs:
                     - {test: '13', deps: ''}
                     - {test: '20', deps: 'lvm2 device-mapper'}
                     - {test: '26', deps: 'cryptsetup mdadm device-mapper lvm2 partx sed'}
+                    - {test: '31', deps: 'blkid curl device-mapper grep networkmanager networkmanager-initrd-generator python3 squashfs-tools'}
```
bdrung: Why not build a bigger alpine image that contains all those packages (similar to the other ci containers)?

jozzsi: I consider having a separate Dockerfile per busybox test would be quite an overhead and I do not really see any advantage of that approach over the current approach. Also apk is incredible fast when installing packages.

As you can see GNU sed is only needed for test 26, and GNU grep is only needed for test 31. If we have one CI container that would serve both test 26 and test 31, than both GNU grep and GNU sed would have to be installed for both test, which would provide much less test coverage than the current approach. The main point here is that installing all the packages all the time is actually reducing test coverage (and also not something that is mimicking end-user environments). 

If the project (ever) gets to the point that none of the GNU tools are required to run none-of the test, than installing packages like lvm2 or networkmanager into the busybox container became a feasible approach. 

..and as I am sure you know we already have a "big" Alpine image that cotains all the dependent packages, but that Alpine image also include coreutils.

jozzsi: Thinking about this more, we could add those Alpine packages to the busybox CI image that are not overlapping with busybox binaries. Came up with a better approach here - https://github.com/dracut-ng/dracut-ng/pull/2120 (which now is a dependency for this PR).

LaszloGombos (approved):

PR #2107, opened 2026-01-23
test: let run-qemu create a log file

bdrung:
## Changes

Move `check_log` to `test/test-functions`, name it `check_qemu_log`, and do not require to specify the success message (which will be the same for all tests). Let `run-qemu` create a log file if `QEMU_LOGFILE` is set and the QEMU is not run in `-daemonize` mode.

This is a preparation to use `check_qemu_log` in other tests as well.

## Checklist

Fixes #

### DIFF
--- a/test/run-qemu
+++ b/test/run-qemu
@@ -198,6 +198,10 @@ if ! [[ $* == *-daemonize* ]] && timeout --help 2> /dev/null | grep -q -- --fore
     # Run QEMU with a timeout (defaut: 10min) unless daemonized
     ARGS=(--foreground "${QEMU_TIMEOUT-600}" "$BIN" "${ARGS[@]}")
     BIN=timeout
+
+    if [[ ${QEMU_LOGFILE-} ]]; then
+        exec > >(tee "$QEMU_LOGFILE")
+    fi
 fi
 
 echo "${0##*/}: $BIN $(quote_args "${ARGS[@]}")"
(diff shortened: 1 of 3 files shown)


LaszloGombos (approved):
great

test/run-qemu
```diff
@@ -198,6 +198,10 @@ if ! [[ $* == *-daemonize* ]] && timeout --help 2> /dev/null | grep -q -- --fore
     # Run QEMU with a timeout (defaut: 10min) unless daemonized
     ARGS=(--foreground "${QEMU_TIMEOUT-600}" "$BIN" "${ARGS[@]}")
     BIN=timeout
+
+    if [[ ${QEMU_LOGFILE-} ]]; then
+        exec > >(tee "$QEMU_LOGFILE")
```
jozzsi, line 203: This does not seem to work with busybox `tee` it seems, and requires `coreutils`.

This PR seems to have a hard conflict with https://github.com/dracut-ng/dracut-ng/pull/2105 (which is now failing after rebase).

jozzsi: https://github.com/dracut-ng/dracut-ng/pull/2115

bdrung: You came to the wrong conclusion which lead me in the wrong direction as well. I figured out the correct reason and here is the fix: https://github.com/dracut-ng/dracut-ng/pull/2116

Issue #2066, opened 2026-01-16
Release 110


bdrung:
Release 110 at the beginning of February 2026. Mention those changes in NEWS.md:

* The separate `dracut-version.sh` has been removed. For packagers: To include the packaging revision into dracut, set `DRACUT_FULL_VERSION` when calling `make`. To change the main version (e.g. when packaging a git snapshot), set `DRACUT_MAIN_VERSION` in addition.
* feat(systemd-pcrphase): rename to systemd-pcrextend

PR #2112, opened 2026-01-23
feat(systemd-pcrphase): rename to systemd-pcrextend

bdrung:
## Changes

systemd-255 renamed systemd-pcrphase to systemd-pcrextend. Do the same for the Dracut module to match that name.

## Checklist

Fixes #


Conan-Kudo (approved):

LaszloGombos:
I think this PR is breaking our backward compatibility promise, we should at least highlight it in the release NEWS.

LaszloGombos (approved):

bdrung:
> I think this PR is breaking our backward compatibility promise, we should at least highlight it in the release NEWS.

Added to https://github.com/dracut-ng/dracut-ng/issues/2066 so that we won't forget to highlight this change.

PR #2113, opened 2026-01-23
docs: document dracut_rescue_image

### LINKED ISSUE #2101: Add a configuration option to prevent fallback images from being created
Add a configuration option to prevent fallback images from being created.

jozzsi:
## Changes

Add documentation for the dracut_rescue_image in the dracut.conf man page.

Do not document default, as different distributions set defaults differently during packaging (sadly).

## Checklist

Fixes #2101

### DIFF
--- a/man/dracut.conf.5.adoc
+++ b/man/dracut.conf.5.adoc
@@ -396,6 +396,10 @@ exists in the directory where initramfs is loaded from.
 +
 Defaults to _initramfs-$\{kernel\}.img_.
 
+*dracut_rescue_image=*"__{yes|no}__"::
+Let kernel-install create a rescue image (default=no).
+This configuration option applies only to kernel-install.
+
 == Files
 _/etc/dracut.conf_::
 Old configuration file. It is recommended to use individual files in


man/dracut.conf.5.adoc
```diff
@@ -395,6 +395,9 @@ exists in the directory where initramfs is loaded from.
 +
 Defaults to _initramfs-$\{kernel\}.img_.
 
+*dracut_rescue_image=*"__{yes|no}__"::
```
bdrung, line 399: This variable is only used for kernel-install

man/dracut.conf.5.adoc
```diff
@@ -395,6 +395,9 @@ exists in the directory where initramfs is loaded from.
...
+If set to _yes_, it prevents fallback images from being created.
```
bdrung: Isn't it the opposite?

man/dracut.conf.5.adoc
jozzsi: Added.

man/dracut.conf.5.adoc
jozzsi: facepalm. Fixed.

man/dracut.conf.5.adoc
```diff
@@ -396,6 +396,10 @@ exists in the directory where initramfs is loaded from.
 +
 Defaults to _initramfs-$\{kernel\}.img_.
 
+*dracut_rescue_image=*"__{yes|no}__"::
+If set to _no_, it prevents fallback images from being created.
+This variable is only used for kernel-install (default=yes).
```
bdrung: Isn't the default `no`? It needs to be set explicitly to `yes`.

I do not like the phrase `If set to _no_`. This is a different phrasing, but I still not like it:

```suggestion
Let kernel-install create a rescue image (default=no).
Configuration only for kernel-install.
```

jozzsi: This is hard to document, but we should try.

If https://github.com/dracut-ng/dracut-ng/blob/main/install.d/51-dracut-rescue.install file is installed on your host system, than the default is yes - meaning rescue image will be created. Fedora likes this default, but on other distro's people are interested to not have fallback images.

bdrung: The code uses `if [[ $dracut_rescue_image != "yes" ]]; then` to skip. So in case dracut_rescue_image is not set, it will skip, won't it? Does Fedora include `dracut.conf.d/rescue/10-rescue.conf` by default?

jozzsi: Fedora uses package manager to enable rescue images - https://src.fedoraproject.org/rpms/dracut/blob/rawhide/f/dracut.spec#_485 

On Arch (just to have another example) usr/lib/kernel/install.d/51-dracut-rescue.install is installed every time dracut is installed - https://archlinux.org/packages/extra/x86_64/dracut/

jozzsi: > So in case dracut_rescue_image is not set, it will skip, won't it?

You're right, technically the default is no. I updated the PR.

man/dracut.conf.5.adoc
```diff
@@ -396,6 +396,10 @@ exists in the directory where initramfs is loaded from.
 +
 Defaults to _initramfs-$\{kernel\}.img_.
 
+*dracut_rescue_image=*"__{yes|no}__"::
+Let kernel-install create a rescue image (default=no).
+Configuration only for kernel-install.
```
bdrung: I played a bit with AI to come up with different wording.
```suggestion
This configuration option applies only to kernel-install.
```

jozzsi: ok

PR #2141, opened 2026-01-27
test: exclude systemd from server rootfs

bdrung:
After commit d36012143c7c ("test: use systemd inside client test rootfs if available") and commit 4aaf8f24110f ("test: include systemd-journald in client test rootfs") the test 60 on opensuse:latest fails due to not enough memory on the server rootfs image.

So revert those two changes partially: create the server rootfs images without systemd. Also create the client test rootfs without systemd.

### Checklist

### DIFF
--- a/test/TEST-60-NFS/test.sh
+++ b/test/TEST-60-NFS/test.sh
@@ -221,7 +221,7 @@ test_run() {
 
 test_setup() {
     call_dracut --tmpdir "$TESTDIR" \
-        --add-confdir test-root \
+        --add-confdir test-root --omit systemd \
         -a "$USE_NETWORK url-lib nfs" \
         -I "ip grep setsid" \
         -f "$TESTDIR"/initramfs.root || return 1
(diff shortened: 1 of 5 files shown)


test/TEST-60-NFS/test.sh
```diff
@@ -221,7 +221,7 @@ test_run() {
 
 test_setup() {
     call_dracut --tmpdir "$TESTDIR" \
-        --add-confdir test-root \
+        --add-confdir test-root --omit "systemd systemd-journald" \
```
LaszloGombos: I find it easier to read if you just omit systemd. Dependency management should also omit systemd-journald.

LaszloGombos: We could also consider `remove_items` `systemd`, so that `test-root` module does not need a change.

`remove_items` currently does not have test coverage, so maybe this is a good opportunity to add

bdrung: Good point. changed.

bdrung: Using `remove_items` would require specifying a lot of files to remove all the files added by systemd again.

LaszloGombos (approved):

PR #2142, opened 2026-01-27
test: build the test rootfs without calling dracut

bdrung:
Build the test rootfs by using the functions from `dracut-functions.sh` instead of calling dracut. Only include the bare minimum for the client rootfs.

Tested the test execution time on a Raspberry Pi 5 running Ubuntu 26.04:

```
$ hyperfine -L commit 9d5f58b4,build-rootfs -p "git checkout {commit}" -w 1 "test/test.sh ubuntu:devel 10"
Benchmark 1: test/test.sh ubuntu:devel 10 (commit = 9d5f58b4)
  Time (mean ± σ):     49.570 s ±  2.847 s    [User: 0.185 s, System: 0.239 s]
  Range (min … max):   46.697 s … 56.466 s    10 runs

Benchmark 2: test/test.sh ubuntu:devel 10 (commit = build-rootfs)
  Time (mean ± σ):     46.627 s ±  0.625 s    [User: 0.174 s, System: 0.220 s]
  Range (min … max):   45.322 s … 47.210 s    10 runs

Summary
  test/test.sh ubuntu:devel 10 (commit = build-rootfs) ran
    1.06 ± 0.06 times faster than test/test.sh ubuntu:devel 10 (commit = 9d5f58b4)
```

### Checklist

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -75,6 +75,97 @@ quote_args() {
     echo "${args[*]}"
 }
 
+# Create the base for a trivial root file system
+build_rootfs_base() {
+    local rootdir="$1"
+
+    # Create basic directory structure (with merged /usr)
+    test ! -e "${rootdir}"
+    mkdir -p "${rootdir}"
+    for subdir in etc dev proc run sys tmp usr usr/bin usr/lib usr/sbin var; do
+        mkdir "${rootdir}/${subdir}"
+    done
+    for subdir in bin lib sbin; do
+        ln -s "usr/$subdir" "${rootdir}/${subdir}"
+    done
+
+    # Include os-release to make systemd's initrd-switch-root.service happy
+    cp /usr/lib/os-release "$rootdir/usr/lib/os-release"
+}
+
+# Expects dracut-functions.sh being sourced and configured
+install_minimal_systemd() {
+    local systemdutildir=/usr/lib/systemd
+    inst_multiple \
+        "$systemdutildir"/system/basic.target \
+        "$systemdutildir"/system/ctrl-alt-del.target \
+        "$systemdutildir"/system/final.target \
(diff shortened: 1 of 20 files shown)


test/test-functions
```diff
@@ -75,6 +75,94 @@ quote_args() {
...
+    build_rootfs_base "$rootdir"
+
+    if [[ "$(type -t inst_multiple 2> /dev/null)" != "function" ]]; then
+        # shellcheck source=./dracut-functions.sh
+        . "$PKGLIBDIR/dracut-functions.sh"
+    fi
+    # variables for inst* functions
+    export DRACUT_NO_XATTR=1
+    # shellcheck disable=SC2034
+    DRACUT_RESOLVE_DEPS=1
```
LaszloGombos: Are these suppose to be initialized in `dracut-functions.sh` (so that they wouldn't need to be here) ?

bdrung: The initialization for `DRACUT_NO_XATTR`:
```sh
# disable xattr when creating cpio, but do not change the default for squashfs/erofs
if ! dracut_module_included "squash-lib"; then
    export DRACUT_NO_XATTR="${DRACUT_NO_XATTR:=1}"
fi
```
That code cannot be included in `dracut-functions.sh`.

Regarding `DRACUT_RESOLVE_DEPS`: dracut either uses `DRACUT_RESOLVE_LAZY` or `DRACUT_RESOLVE_DEPS`. I don't see how that can be put into `dracut-functions.sh`. The only idea is to change the default.

bdrung: We can get rid of setting `DRACUT_RESOLVE_DEPS`: https://github.com/dracut-ng/dracut-ng/pull/2157

LaszloGombos (approved):
thanks @bdrung

PR #2144, opened 2026-01-27
test(NFS): use build_ext4_image to build server rootfs

bdrung:
After commit d36012143c7c ("test: use systemd inside client test rootfs if available") and commit 4aaf8f24110f ("test: include systemd-journald in client test rootfs") the test 60 on opensuse:latest fails due to not enough memory for starting the QEMU process to create the server rootfs image.

The solution in commit 30894ad4855018f6ec07664ce06bf5557d65e324 did not work, because the network tests need systemd when they use systemd-networkd for networking. The tests will fail with:

```
dracut[E]: Module 'systemd-networkd' depends on module 'systemd', which can't be installed
```

Use `build_ext4_image` to build the server rootfs to completely avoid needing to start a QEMU instance.

### Checklist


LaszloGombos (approved):
ok, thanks @bdrung

PR #2151, opened 2026-01-28
fix(dracut): only take --hostonly-cmdline into account in host-only mode

bdrung:
When building an initrd with `--no-hostonly`, the parameter `--hostonly-cmdline` is still taken into account. That is
counter-intiutive.

So only take `--hostonly-cmdline`/`hostonly_cmdline` into account in host-only mode.

The default stays the same: the cmdline is included in `hostonly=yes` mode and not included in `hostonly=no` mode.

### Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -192,7 +192,7 @@ Creates initial ramdisk images for preloading modules
                          DO NOT use "strict" mode unless you know what you
                          are doing.
   --hostonly-cmdline    Store kernel command line arguments needed
-                         in the initramfs.
+                         in the initramfs if in host-only mode.
   --no-hostonly-cmdline Do not store kernel command line arguments needed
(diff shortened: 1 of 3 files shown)


LaszloGombos:
I prefer either @jozzsi or @aafeijoo-suse give some initial feedback on this PR 

[earlier feedback](https://github.com/dracut-ng/dracut-ng/pull/1494/changes/fd9147c8681d8aa44e54a720eb8a9eb258e3c0f1#r2250543526)

> hostonly and hostonly_cmdline are independent options if the latter is set.

devkontrol (approved):
Thanks for the review @aafeijoo-suse and @jozzsi .

LGTM

devkontrol:
has conflicts, marking draft

PR #2156, opened 2026-01-29
test: execute assertion.sh instead of sourcing it

bdrung:
The `assertion.sh` files do not need to access from `test-init.sh`. So execute `assertion.sh` instead of sourcing it.

### Checklist

### DIFF
--- a/test/modules.d/70test-root/test-init.sh
+++ b/test/modules.d/70test-root/test-init.sh
@@ -43,7 +43,7 @@ fi
 
 # run the test case specific test assertion if exists
 if [ -x "/assertion.sh" ]; then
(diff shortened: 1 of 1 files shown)


bdrung:
That failure looks unrelated to this PR. Why do we see
```
made it to the test rootfs!
All OK
Powering down.
```
after that failure and why does `systemctl --failed --no-legend --no-pager` do not complain about it? Can you track it down to when it first appeared?

bdrung:
Maybe that behaviour started with using systemd in the test rootfs. At least this test 11 will be fixed by https://github.com/dracut-ng/dracut-ng/pull/2142

PR #2157, opened 2026-01-29
feat: replace DRACUT_RESOLVE_DEPS by !DRACUT_RESOLVE_LAZY

bdrung:
The two variables `DRACUT_RESOLVE_DEPS` and `DRACUT_RESOLVE_LAZY` are mutual exclusive (one must be set, the other must not).

Make the code easier by dropping `DRACUT_RESOLVE_DEPS` and replacing those parts by checking if `DRACUT_RESOLVE_LAZY` is not set. This sets the default to resolving dependencies in the `inst*` functions.

### Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1102,39 +1102,42 @@ inst_dir() {
 
 inst() {
     local dstdir="${dstdir:-"$initdir"}"
-    local _ret _hostonly_install
+    local _ret _hostonly_install _resolve_deps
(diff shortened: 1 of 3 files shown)


LaszloGombos (approved):
Thanks @bdrung for being proactive on this (as it came up on another PR).

I think this is great progress, but now we should try to capture and document the meaning of `DRACUT_RESOLVE_LAZY` (when a user or distro should set it and why).

PR #2161, opened 2026-01-29
feat(overlayfs): add encrypted persistent overlay support

nadzyah:
Add support for LUKS-encrypted persistent overlay devices via the new `rd.overlay.crypt` parameter. This is similar to the `overlayroot=crypt` functionality in cloud-initramfs-tools.

### Checklist

### DIFF
--- a/man/dracut.cmdline.7.adoc
+++ b/man/dracut.cmdline.7.adoc
@@ -1157,6 +1157,38 @@ rd.overlay=UUID=12345678-1234-1234-1234-123456789abc
 rd.overlay=/dev/sdb1
 ----
 
+**rd.overlay.crypt=**__<device>__[**,**__<option>__]...::
+Requires the dracut 'overlayfs-crypt' module.
++
+Specifies an encrypted (LUKS) device for persistent overlay storage.  If the
+device already contains a valid LUKS volume, the user will be prompted for the
+passphrase interactively (via Plymouth if available, otherwise via TTY).
+Otherwise, the device will be wiped, formatted with LUKS using a randomly
+generated passphrase, and a new filesystem will be created on it.  The random
+passphrase is stored in `/run/initramfs/overlayfs.passwd` for the duration of
+the boot.
++
+The device is specified as the first argument, followed by optional
+comma-separated options:
++
+--
+* _<device>_ - (Required) Device specification: LABEL=, UUID=, PARTLABEL=,
+  PARTUUID=, or /dev/<device>.
+* _mapname=<name>_ - Device mapper name (default: overlay-crypt).
+* _fstype=<type>_ - Filesystem type: ext2, ext3, or ext4 (default: ext4).
+* _mkfs=<0|1>_ - Filesystem creation behavior: 0=never create (require existing
+  LUKS), 1=create if needed (default).
+* _timeout=<seconds>_ - Time to wait for the device to appear.  By default,
+  the global `rd.timeout` value is used.  Set to 0 to skip waiting entirely.
+--
++
+[listing]
+.Examples
+----
+rd.overlay.crypt=LABEL=CRYPTOVL
+rd.overlay.crypt=UUID=12345678-1234-1234-1234-123456789abc,fstype=ext3
+----
+
 Booting live images
 ~~~~~~~~~~~~~~~~~~~
 Requires the dracut 'dmsquash-live' module.
--- a/modules.d/70crypt/crypt-lib.sh
+++ b/modules.d/70crypt/crypt-lib.sh
@@ -135,6 +135,28 @@ ask_for_password() {
     return $ret
 }
 
+# luks_open_interactive dev mapname [prompt] [cryptsetupopts]
+#
+# Prompt for a LUKS passphrase and open the device.  Uses Plymouth if
+# available, otherwise falls back to TTY.  Reads rd.luks.timeout for
+# the passphrase input timeout.
+luks_open_interactive() {
+    local dev="$1" mapname="$2"
+    local prompt="${3:-Password ($dev)}"
+    local cryptsetupopts="${4:-}"
+    local luks_open _timeout
+
+    _timeout=$(getarg rd.luks.timeout)
+    _timeout="${_timeout:-0}"
+    luks_open="$(command -v cryptsetup) $cryptsetupopts luksOpen"
+    ask_for_password \
+        --ply-tries 5 \
+        --ply-cmd "$luks_open -T1 $dev $mapname" \
+        --ply-prompt "$prompt" \
+        --tty-tries 1 \
+        --tty-cmd "$luks_open -T5 -t $_timeout $dev $mapname"
+}
+
 # Try to mount specified device (by path, by UUID or by label) and check
 # the path with 'test'.
 #
--- a/modules.d/70crypt/cryptroot-ask.sh
+++ b/modules.d/70crypt/cryptroot-ask.sh
@@ -78,7 +78,7 @@ asked_file=/tmp/cryptroot-asked-$luksname
 # load dm_crypt if it is not already loaded
 [ -d /sys/module/dm_crypt ] || modprobe dm_crypt
 
-command -v ask_for_password > /dev/null || . /lib/dracut-crypt-lib.sh
+command -v luks_open_interactive > /dev/null || . /lib/dracut-crypt-lib.sh
 
 #
 # Open LUKS device
@@ -175,16 +175,7 @@ else
 fi
 
 if [ $ask_passphrase -ne 0 ]; then
-    luks_open="$(command -v cryptsetup) $cryptsetupopts luksOpen"
-    _timeout=$(getarg "rd.luks.timeout")
-    _timeout=${_timeout:-0}
-    ask_for_password --ply-tries 5 \
-        --ply-cmd "$luks_open -T1 $device $luksname" \
-        --ply-prompt "Password ($device)" \
-        --tty-tries 1 \
-        --tty-cmd "$luks_open -T5 -t $_timeout $device $luksname"
-    unset luks_open
-    unset _timeout
+    luks_open_interactive "$device" "$luksname" "Password ($device)" "$cryptsetupopts"
 fi
 
 if [ "$is_keysource" -ne 0 ] && [ "${luksname##luks-}" != "$luksname" ]; then
--- a/modules.d/71overlayfs-crypt/module-setup.sh
+++ b/modules.d/71overlayfs-crypt/module-setup.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+check() {
+    require_any_binary cryptsetup || return 1
+    return 255
+}
+
+depends() {
+    echo overlayfs
+}
+
+installkernel() {
+    hostonly="" instmods dm_mod dm_crypt
+}
+
+install() {
+    inst_hook pre-mount 02 "$moddir/prepare-overlayfs-crypt.sh"
+    inst_hook pre-pivot 01 "$moddir/prepare-overlayfs-crypt.sh"
+
+    inst_simple "$moddir/overlayfs-crypt-lib.sh" "/lib/overlayfs-crypt-lib.sh"
+    inst_simple "${moddir%/*}/70crypt/crypt-lib.sh" "/lib/dracut-crypt-lib.sh"
+    inst_multiple cryptsetup wipefs mkfs.ext4 mkfs.ext3 mkfs.ext2 sha512sum mktemp chmod readlink
+    inst_multiple -o stty
+}
--- a/modules.d/71overlayfs-crypt/overlayfs-crypt-lib.sh
+++ b/modules.d/71overlayfs-crypt/overlayfs-crypt-lib.sh
@@ -1,0 +1,220 @@
+#!/bin/sh
+
+command -v getarg > /dev/null || . /lib/dracut-lib.sh
+
+parse_overlay_opts() {
+    local input="$1" ns="${2:-_RET_}"
+    local _oldifs="$IFS" tok key val
+
+    _RET=""
+    set -f
+    IFS=","
+    # shellcheck disable=SC2086
+    set -- $input
+    IFS="$_oldifs"
+    set +f
+
+    for tok in "$@"; do
+        key="${tok%%=*}"
+        val="${tok#*=}"
+        [ "$key" = "$tok" ] && val=""
+        case "$key" in
+            mapname | fstype | mkfs | timeout)
+                eval "${ns}${key}='${val}'"
+                _RET="${_RET:+$_RET }${ns}${key}"
+                ;;
+            *)
+                warn "Unknown option '$key'"
+                ;;
+        esac
+    done
+    return 0
+}
+
+generate_random_password() {
+    local tmpf entropy_sources pass
+    local persist_dir="/run/initramfs"
+
+    [ -d "$persist_dir" ] || mkdir -p "$persist_dir"
+
+    tmpf=$(mktemp "${persist_dir}/overlayfs-crypt.XXXXXX") || {
+        warn "Failed to create temp file for password"
+        return 1
+    }
+
+    entropy_sources="/proc/sys/kernel/random/boot_id /proc/sys/kernel/random/uuid /dev/urandom"
+
+    stat -L /dev/* /proc/* /sys/* > "$tmpf" 2>&1 || true
+
+    for src in $entropy_sources; do
+        [ -e "$src" ] && head -c 4096 "$src" >> "$tmpf" 2> /dev/null
+    done
+
+    pass=$(sha512sum "$tmpf" 2> /dev/null) || {
+        rm -f "$tmpf"
+        warn "Failed to generate password"
+        return 1
+    }
+    pass="${pass%% *}"
+
+    printf "%s" "$pass" > "${persist_dir}/overlayfs.passwd"
+    chmod 400 "${persist_dir}/overlayfs.passwd"
+
+    rm -f "$tmpf"
+    _RET="$pass"
+    return 0
+}
+
+_overlayfs_crypt_fallback_tmpfs() {
+    [ -d /run/overlayfs ] || {
+        mkdir -m 0755 -p /run/initramfs/overlay/overlayfs
+        mkdir -m 0755 -p /run/initramfs/overlay/ovlwork
+        ln -sf /run/initramfs/overlay/overlayfs /run/overlayfs
+        ln -sf /run/initramfs/overlay/ovlwork /run/ovlwork
+    }
+    # Signal mount-overlayfs.sh to proceed with the overlay mount
+    : > /run/overlayfs-crypt-ready
+}
+
+overlayfs_crypt_setup() {
+    local options="$1"
+    local dev="" mapname="overlay-crypt" fstype="ext4" mkfs="1" timeout=""
+    local crypt_dev
+
+    # Device is the first positional argument
+    local device="${options%%,*}"
+    case "$device" in
+        LABEL=* | UUID=* | PARTLABEL=* | PARTUUID=* | /dev/*)
+            dev="$device"
+            options="${options#"$device"}"
+            options="${options#,}"
+            ;;
+    esac
+
+    parse_overlay_opts "$options" "_ovl_" || return 1
+
+    mapname="${_ovl_mapname:-$mapname}"
+    fstype="${_ovl_fstype:-$fstype}"
+    mkfs="${_ovl_mkfs:-$mkfs}"
+    timeout="${_ovl_timeout:-$timeout}"
+
+    [ -n "$dev" ] || {
+        warn "Device parameter is required"
+        return 1
+    }
+
+    case "$dev" in
+        LABEL=* | UUID=* | PARTLABEL=* | PARTUUID=*)
+            dev=$(label_uuid_to_dev "$dev")
+            ;;
+        /dev/*) ;;
+        *)
+            dev="/dev/$dev"
+            ;;
+    esac
+
+    if [ "$timeout" = "0" ]; then
+        : # Don't wait for the device
+    elif [ -n "$timeout" ]; then
+        wait_for_dev -n "$dev" "$timeout" || {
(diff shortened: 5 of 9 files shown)


nadzyah:
@LaszloGombos @bdrung thanks for the review! I've updated the code and created the `overlayfs-crypt` module. I've considered using the existing crypt module, but it looks like the `crypt` module is designed to open existing LUKS devices, and we need to create and manage LUKS devices. However, I can extend the `crypt` module with this functionality instead of creating a new one. What do you prefer?

P.S. The `20 fedora:latest` test seems to [fail in main](https://github.com/dracut-ng/dracut-ng/actions/runs/22120496096/job/63939108072) as well, so I'm not sure it's related to my changes.

man/dracut.cmdline.7.adoc
```diff
@@ -1134,6 +1134,36 @@ rd.overlay=UUID=12345678-1234-1234-1234-123456789abc
...
+--
+* _dev=<device>_ - (Required) Device specification: LABEL=, UUID=, PARTLABEL=,
+  PARTUUID=, /dev/<device>, or short device name.
+* _pass=<password>_ - Encryption password.  If omitted, a random password is
+  generated.
+* _mapname=<name>_ - Device mapper name (default: overlay-crypt).
+* _fstype=<type>_ - Filesystem type: ext2, ext3, or ext4 (default: ext4).
+* _mkfs=<0|1>_ - Filesystem creation behavior: 0=never create (require existing
+  LUKS), 1=create if needed (default).
+* _timeout=<seconds>_ - Time to wait for device to appear (default: 0).
```
bdrung: What does `0` mean in that context? infinity?

man/dracut.cmdline.7.adoc
```diff
@@ -1134,6 +1134,36 @@ rd.overlay=UUID=12345678-1234-1234-1234-123456789abc
...
++
+Specifies an encrypted (LUKS) device for persistent overlay storage.  The device
+will be formatted with LUKS encryption and a filesystem will be created on it.
+If a password is not provided, a random one will be generated and stored in
+`/run/initramfs/overlayfs.passwd`.
++
+Supported options (comma-separated):
++
+--
+* _dev=<device>_ - (Required) Device specification: LABEL=, UUID=, PARTLABEL=,
```
bdrung: Have you consider to just have the device name as first option (since it is required)?
```
rd.overlay.crypt=<device>[,<option>]
```

man/dracut.cmdline.7.adoc
```diff
@@ -1134,6 +1134,36 @@ rd.overlay=UUID=12345678-1234-1234-1234-123456789abc
 rd.overlay=/dev/sdb1
 ----
 
+**rd.overlay.crypt=**__<options>__::
+Requires the dracut 'overlayfs-crypt' module.
++
+Specifies an encrypted (LUKS) device for persistent overlay storage.  The device
+will be formatted with LUKS encryption and a filesystem will be created on it.
```
bdrung: What will happen in case the partition is already present?

modules.d/71overlayfs-crypt/overlayfs-crypt-lib.sh
```diff
@@ -0,0 +1,209 @@
+#!/bin/sh
+
+command -v getarg > /dev/null || . /lib/dracut-lib.sh
+
+parse_overlay_opts() {
+    local input="$1" ns="${2:-_RET_}"
+    local oifs="$IFS" tok key val
```
bdrung: `oifs` is not used anywhere. `oldifs` is the common name.

modules.d/71overlayfs-crypt/overlayfs-crypt-lib.sh
```diff
@@ -0,0 +1,209 @@
...
+
+    modprobe -q dm_mod 2> /dev/null || true
+    modprobe -q dm_crypt 2> /dev/null || true
+
+    crypt_dev="/dev/mapper/$mapname"
+
+    if [ -n "$pass" ]; then
+        if printf "%s" "$pass" | cryptsetup luksOpen "$dev" "$mapname" --key-file - 2> /dev/null; then
+            info "overlayfs-crypt: opened existing LUKS device at $dev"
+            wait_for_dev -n "$crypt_dev" 20 || {
```
bdrung: Why waiting for 20 seconds? Could we make that configurable via rd.timeout?

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -48,7 +62,10 @@ test_setup() {
     truncate -s 32M "$TESTDIR"/overlay.img
     mkfs.ext4 -q -L OVERLAY "$TESTDIR"/overlay.img
 
-    test_dracut --add overlayfs
+    dd if=/dev/zero of="$TESTDIR"/crypt.img bs=1M count=100 status=none
```
bdrung: Why not use `truncate`?

modules.d/71overlayfs-crypt/module-setup.sh
```diff
@@ -0,0 +1,22 @@
...
+installkernel() {
+    hostonly="" instmods dm_mod dm_crypt
+}
+
+install() {
+    inst_hook pre-mount 02 "$moddir/prepare-overlayfs-crypt.sh"
+    inst_hook pre-pivot 01 "$moddir/prepare-overlayfs-crypt.sh"
+
+    inst_simple "$moddir/overlayfs-crypt-lib.sh" "/lib/overlayfs-crypt-lib.sh"
+    inst_multiple cryptsetup wipefs mkfs.ext4 mkfs.ext3 mkfs.ext2 sha512sum mktemp chmod readlink timeout
```
bdrung: Where is the timeout command used?

modules.d/71overlayfs-crypt/prepare-overlayfs-crypt.sh
```diff
@@ -0,0 +1,47 @@
...
+
+        mkdir -m 0755 -p /run/overlayfs-backing/overlay
+        mkdir -m 0755 -p /run/overlayfs-backing/ovlwork
+
+        ln -sf /run/overlayfs-backing/overlay /run/overlayfs
+        ln -sf /run/overlayfs-backing/ovlwork /run/ovlwork
+
+        # Signal mount-overlayfs.sh to proceed with the overlay mount
+        : > /run/overlayfs-crypt-ready
+    else
```
bdrung:
```sh
if cmd1; then
   foo
   if cmd2; then
       bar
   else
      _overlayfs_crypt_fallback_tmpfs
   fi
else
   _overlayfs_crypt_fallback_tmpfs
fi
```

could use exit-early:

```sh
if ! cmd1; then
    _overlayfs_crypt_fallback_tmpfs
    return 0
fi

foo

if ! cmd2; then
    _overlayfs_crypt_fallback_tmpfs
    return 0
fi

bar
```

man/dracut.cmdline.7.adoc
nadzyah: timeout=0 means no wait, which is also defined like that in cloud-initramfs-tools:

(`overlayroot/scripts/init-bottom/overlayroot` lines 148-157)
```sh
wait_for_dev() {
	local dev="$1" timeout="${2:-0}"
	[ -b "$dev" ] && return 0
	[ "$timeout" = "0" ] && return 1
	# wait-for-root writes fstype to stdout, redirect to null
	wait-for-root "$dev" "$timeout" >/dev/null
}
crypto_setup() {
	local fstype="ext4" pass="" mapname="secure" mkfs="1" dev=""
	local timeout=0
```

man/dracut.cmdline.7.adoc
nadzyah: If the device already has LUKS and a password is provided, it gets opened as-is. Otherwise, it gets wiped and formatted. I've updated the docs to mention this

nadzyah:
> I am confused who this functionality is even for. In order to use it, you have to disclose the key/passphrase as a boot parameter, so it doesn't help anything?

The primary use case is the random password mode, where no passphrase is disclosed on the command line. A random password is generated at boot, used to format a fresh LUKS volume, and kept only in RAM. This protects the overlay data at rest. If a VM disk gets snapshotted or cloned, the overlay partition is unreadable without the in-memory key. I think this is the main value here

The `pass=` option is there for persistent overlays (same password reopens the existing LUKS volume across reboots). I agree that putting a passphrase on the kernel cmdline isn't ideal, but AFAIK in cloud environments it's typically set by the metadata service (e.g. cloud-init) rather than typed by a user. This follows the same approach as `overlayroot=crypt` in `cloud-initramfs-tools`, which is used in Ubuntu cloud images for the same purpose

man/dracut.cmdline.7.adoc
```diff
@@ -1157,6 +1157,38 @@ rd.overlay=UUID=12345678-1234-1234-1234-123456789abc
...
++
+--
+* _<device>_ - (Required) Device specification: LABEL=, UUID=, PARTLABEL=,
+  PARTUUID=, or /dev/<device>.
+* _mapname=<name>_ - Device mapper name (default: overlay-crypt).
+* _fstype=<type>_ - Filesystem type: ext2, ext3, or ext4 (default: ext4).
+* _mkfs=<0|1>_ - Filesystem creation behavior: 0=never create (require existing
+  LUKS), 1=create if needed (default).
+* _timeout=<seconds>_ - Time to wait for the device to appear (default: 0,
+  meaning don't wait).
```
bdrung: Don't wait for the device as default? What happens in case the device is detected too late?

modules.d/71overlayfs-crypt/prepare-overlayfs-crypt.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/sh
```
bdrung, line 1: This file should be executable.

modules.d/71overlayfs-crypt/module-setup.sh
```diff
@@ -0,0 +1,23 @@
...
+    echo overlayfs
+}
+
+installkernel() {
+    hostonly="" instmods dm_mod dm_crypt
+}
+
+install() {
+    inst_hook pre-mount 02 "$moddir/prepare-overlayfs-crypt.sh"
+    inst_hook pre-pivot 01 "$moddir/prepare-overlayfs-crypt.sh"
```
bdrung, line 18: Why do you use the same script in pre-mount and pre-pivot. Shouldn't you have two separate scripts?

modules.d/71overlayfs-crypt/overlayfs-crypt-lib.sh
```diff
@@ -0,0 +1,232 @@
...
+    # Signal mount-overlayfs.sh to proceed with the overlay mount
+    : > /run/overlayfs-crypt-ready
+}
+
+_overlayfs_crypt_open_interactive() {
+    local dev="$1" mapname="$2"
+
+    # Use ask_for_password if dracut-crypt-lib.sh is available, otherwise fall
+    # back to a plain cryptsetup TTY prompt.
+    if [ -f /lib/dracut-crypt-lib.sh ]; then
```
bdrung: I don't like that we have two code paths here. It would be better to always include dracut-crypt-lib.sh in the initrd and use it. That would probably require to move that file into a separate crypto-lib module (or pulling in the crypt module).

modules.d/71overlayfs-crypt/module-setup.sh
nadzyah: This follows the same pattern as 70overlayfs, which also registers `prepare-overlayfs.sh` in both pre-mount (for live images where rootfsbase is set up before mount) and pre-pivot (for regular boots where NEWROOT is mounted). The same script works in both cases because of the ismounted/rootfsbase guard. I've added an explicit early-return guard on `/run/overlayfs-crypt-ready` to prevent double execution when both hooks would otherwise proceed

modules.d/71overlayfs-crypt/prepare-overlayfs-crypt.sh
nadzyah: fixed

modules.d/71overlayfs-crypt/overlayfs-crypt-lib.sh
```diff
@@ -0,0 +1,232 @@
...
+        mkdir -m 0755 -p /run/initramfs/overlay/overlayfs
+        mkdir -m 0755 -p /run/initramfs/overlay/ovlwork
+        ln -sf /run/initramfs/overlay/overlayfs /run/overlayfs
+        ln -sf /run/initramfs/overlay/ovlwork /run/ovlwork
+    }
+    # Signal mount-overlayfs.sh to proceed with the overlay mount
+    : > /run/overlayfs-crypt-ready
+}
+
+_overlayfs_crypt_open_interactive() {
```
bdrung: There is very similar code in `cryptroot-ask.sh` that could be moved into a function in `dracut-crypt-lib.sh` and re-used instead of adding `_overlayfs_crypt_open_interactive`.

test/TEST-21-OVERLAYFS/assertion.sh
```diff
@@ -21,6 +21,13 @@ if grep -q 'test.expect=device' /proc/cmdline; then
     if ! grep -q "/run/overlayfs-backing" /proc/mounts; then
         echo "persistent overlay device not mounted at /run/overlayfs-backing" >> /run/failed
     fi
+elif grep -q 'test.expect=crypt' /proc/cmdline; then
```
bdrung, line 47: Couldn't we check that the mount is encrypted intstead of just checking for specific paths to be present?

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -40,6 +46,10 @@ test_run() {
     client_run "fallback to tmpfs (non-existent LABEL)" "rd.overlay=LABEL=NONEXISTENT test.expect=tmpfs"
     client_run "tmpfs overlay with size (rd.overlay=tmpfs:size=32M,nr_inodes=100000)" \
         "rd.overlay=tmpfs:size=32M,nr_inodes=100000 test.expect=tmpfs-sized"
+
+    reset_crypt_disk
+    client_run "encrypted overlay (new device, random password)" \
```
bdrung, line 52: Should we also check that the password file is present?

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -50,7 +60,10 @@ test_setup() {
     truncate -s 32M "$TESTDIR"/overlay.img
     mkfs.ext4 -q -L OVERLAY "$TESTDIR"/overlay.img
 
-    test_dracut --add overlayfs
+    truncate -s 100M "$TESTDIR"/crypt.img
+    mkfs.ext4 -q -L CRYPT "$TESTDIR"/crypt.img
```
bdrung: Can you combine this code with `reset_crypt_disk` into one function? `rm -f` + `truncate` should be faster than `dd`.

test/TEST-21-OVERLAYFS/assertion.sh
```diff
@@ -21,6 +21,25 @@ if grep -q 'test.expect=device' /proc/cmdline; then
...
+            grep -q "^CRYPT-" "$_dm/dm/uuid" 2>/dev/null && _dm_uuid_ok=1
+            break
+        fi
+    done
+    [ "$_dm_uuid_ok" -eq 1 ] || echo "overlay-crypt is not a dm-crypt device" >> /run/failed
+    unset _dm _dm_uuid_ok
+    # Check that the random passphrase was persisted
+    if [ ! -f /run/initramfs/overlayfs.passwd ]; then
+        echo "password file /run/initramfs/overlayfs.passwd not found" >> /run/failed
+    fi
```
bdrung: You could put those stuff in functions (using the current comments as function names). Then you could use `local _dm_uuid_ok` to avoid unsetting them.

test/TEST-21-OVERLAYFS/assertion.sh
```diff
@@ -21,6 +21,33 @@ if grep -q 'test.expect=device' /proc/cmdline; then
...
+            fi
+        done
+        [ "$_dm_uuid_ok" -eq 1 ] || echo "overlay-crypt is not a dm-crypt device" >> /run/failed
+    }
+
+    check_crypt_passphrase() {
+        if [ ! -f /run/initramfs/overlayfs.passwd ]; then
+            echo "password file /run/initramfs/overlayfs.passwd not found" >> /run/failed
+        fi
+    }
```
bdrung: Why do you define this functions inside this `elif` clause instead of placing the functions outside of it?

modules.d/70crypt/crypt-lib.sh
```diff
@@ -135,6 +135,24 @@ ask_for_password() {
     return $ret
 }
 
+# luks_open_interactive dev mapname [prompt]
+#
+# Prompt for a LUKS passphrase and open the device.  Uses Plymouth if
+# available, otherwise falls back to TTY.
+luks_open_interactive() {
```
bdrung, line 143: Can you put this function into a separate commit that does the refactoring of the code in `modules.d/70crypt/cryptroot-ask.sh` to use this new function?

modules.d/71overlayfs-crypt/overlayfs-crypt-lib.sh
```diff
@@ -0,0 +1,233 @@
...
```
bdrung: _overlayfs_crypt_open_interactive is not used any more.

test/TEST-21-OVERLAYFS/assertion.sh
```diff
@@ -3,6 +3,29 @@ set -eu
...
+    if ! grep -q "^/dev/mapper/overlay-crypt /run/overlayfs-backing " /proc/mounts; then
+        echo "encrypted overlay not mounted at /run/overlayfs-backing" >> /run/failed
+    fi
+}
+
+check_crypt_device() {
+    local _dm _dm_uuid_ok=0
+    for _dm in /sys/class/block/dm-*; do
+        if [ "$(cat "$_dm/dm/name" 2> /dev/null)" = "overlay-crypt" ]; then
+            grep -q "^CRYPT-" "$_dm/dm/uuid" 2> /dev/null && _dm_uuid_ok=1
```
bdrung: `_dm_uuid_ok` does not need to be needed here. You could just return the function as success.

modules.d/70crypt/crypt-lib.sh
nadzyah: done in b67bb0275c28cfc22926feec04c5eacec741ee52

modules.d/70crypt/cryptroot-ask.sh
```diff
@@ -175,15 +175,9 @@ else
 fi
 
 if [ $ask_passphrase -ne 0 ]; then
-    luks_open="$(command -v cryptsetup) $cryptsetupopts luksOpen"
     _timeout=$(getarg "rd.luks.timeout")
```
bdrung: Why not moving `rd.luks.timeout` to luks_open_interactive as well? This timeout setting is useful for the overlayfs as well, isn't it?

modules.d/70crypt/cryptroot-ask.sh
bdrung: ask_for_password is not called directly any more. So you should check for luks_open_interactive instead.

bdrung (approved):

modules.d/71overlayfs-crypt/overlayfs-crypt-lib.sh
```diff
@@ -0,0 +1,220 @@
...
+    info "Setting up encrypted overlay on $dev"
+
+    modprobe -q dm_mod 2> /dev/null || true
+    modprobe -q dm_crypt 2> /dev/null || true
+
+    crypt_dev="/dev/mapper/$mapname"
+
+    if cryptsetup isLuks "$dev" 2> /dev/null; then
+        info "Existing LUKS device found at $dev, prompting for password"
+        command -v luks_open_interactive > /dev/null || . /lib/dracut-crypt-lib.sh
```
jozzsi, line 152: dracut-crypt-lib.sh file is only available if `crypt` dracut module is included in the generated initramfs.

What happens if I invoke dracut with e.g.
`dracut -m "overlayfs-crypt"` ?

bdrung: We have the a bit hacky inclusion of this file in all cases:
```
inst_simple "${moddir%/*}/70crypt/crypt-lib.sh" "/lib/dracut-crypt-lib.sh"
```

Conan-Kudo: Shouldn't we just forcibly depend on it for this module?

jozzsi: We probably should to split out `crypt-lib` dracut module (e.g. like `fs-lib` or `net-lib`) from crypt dracut module, so that this module can explicitly be dependent on  crypt-lib.

nadzyah: The file is always installed via `inst_simple` in `module-setup.sh`, so `dracut -m "overlayfs-crypt"` works without 70crypt. Adding 70crypt as a dependency would pull in crypttab handling and udev rules, which I don't think is needed

I also think splitting `crypt-lib.sh` into its own module like `fs-lib` is the right long-term fix, but IMO that's better done in a follow-up PR

devkontrol (approved):
ok, with follow-up improvements.

Conan-Kudo (approved):
Okay, please follow up with splitting crypt-lib and cleaning up the dependencies for this.

modules.d/71overlayfs-crypt/overlayfs-crypt-lib.sh
Conan-Kudo: Okay, please do this as a followup ASAP.

jozzsi:
> Okay, please follow up with splitting crypt-lib and cleaning up the dependencies for this.

https://github.com/dracut-ng/dracut-ng/pull/2375

bdrung:
Oh, we forgot to add the new module to `.github/labeler.yml` and `doc_site/modules/ROOT/pages/modules/core.adoc`

PR #2163, opened 2026-01-30
test: always test erofs and xorriso

bdrung:
### Checklist


bdrung:
Note: The docker file for opensuse says:

> - erofs is blacklisted by default and not supported in openSUSE

PR #2168, opened 2026-01-31
fix(dracut-logger): do not export maxloglvl and syslogfacility

bdrung:
Commit 46265a9d3e0a ("dracut-logger: make $maxloglvl public") exports the `maxloglvl` variable, but this is not needed. Exporting variables is only needed when subprocesses need to access those variables.

`maxloglvl` is only used in `dracut-logger.sh` and `dracut.sh`. `dracut.sh` sources `dracut-logger.sh`. So exporting `maxloglvl` is not needed. `syslogfacility` is only used in `dracut-logger.sh`.

So do not export `maxloglvl` and `syslogfacility`.

### Checklist

### DIFF
--- a/dracut-logger.sh
+++ b/dracut-logger.sh
@@ -166,7 +166,6 @@ dlog_init() {
         else
             readonly syslogfacility=daemon
         fi
-        export syslogfacility
     fi
 
     local lvl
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Don't we export these for out of tree modules?

PR #2175, opened 2026-02-02
refactor(overlayfs): move legacy option handing into a cmdline hook

bdrung:
Reading the code and understanding the logic gets complicated by supporting legacy kernel command line options. This will be especially be true once `rd.overlayfs` gets merged into `rd.overlay`.

So move all the legacy kernel command line option handling into the separate `parse-legacy-overlayfs-options.sh` script that is executed during the cmdline stage.

Check that no legacy option is queried directly:

```
git grep getarg -- modules.d/ | grep -w -- -d | grep live
```

### Checklist

### DIFF
--- a/modules.d/70overlayfs/module-setup.sh
+++ b/modules.d/70overlayfs/module-setup.sh
@@ -14,6 +14,7 @@ installkernel() {
 }
 
 install() {
+    inst_hook cmdline 70 "$moddir/parse-legacy-overlayfs-options.sh"
     inst_hook pre-mount 01 "$moddir/prepare-overlayfs.sh"
     inst_hook mount 01 "$moddir/mount-overlayfs.sh"       # overlay on top of block device
     inst_hook pre-pivot 00 "$moddir/prepare-overlayfs.sh" # prepare for network root (e.g. nfs)
--- a/modules.d/70dmsquash-live-autooverlay/create-overlay.sh
+++ b/modules.d/70dmsquash-live-autooverlay/create-overlay.sh
@@ -9,7 +9,7 @@ if getargbool 0 rd.live.debug; then
 fi
 
 gatherData() {
-    overlay=$(getarg rd.overlay -d rd.live.overlay)
+    overlay=$(getarg rd.overlay)
     if [ -z "$overlay" ]; then
         info "Skipping overlay creation: kernel command line parameter 'rd.overlay' is not set"
         exit 0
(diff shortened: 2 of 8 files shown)


modules.d/70overlayfs/module-setup.sh
```diff
@@ -14,6 +14,7 @@ installkernel() {
 }
 
 install() {
+    inst_hook cmdline 70 "$moddir/parse-legacy-overlayfs-options.sh"
```
jozzsi, line 17: Would this hook work for out of tree dracut modules as well that sets legacy command line in the following way  (which I think gets executed _after_ _all_ the cmdline hooks are executed regardless of the cmdline hook ordering).

https://github.com/dracut-ng/dracut-ng/blob/3f3c377bbe58f4e05358492537f777ed2b25546d/modules.d/70dmsquash-live/dmsquash-live-root.sh#L419

We should at least check if https://github.com/OSInside/kiwi/tree/main/dracut/modules.d does something like this, but in general this mechanism to support legacy options are not entirely safe to do IMHO.

CC @Vogtinator

Vogtinator: Yeah, it wouldn't work if anything writes to cmdline.d after `parse-legacy-overlayfs-options.sh` ran.

IMO if you want centralized handling of legacy cmdline options, it should be done in getarg or dracut-util.

May even make sense to just remove deprecated options when they become a burden?

Conan-Kudo (approved):
THANK YOU for this!

modules.d/70overlayfs/module-setup.sh
Conan-Kudo: Ugh, I wish I had seen this thread first. Yeah, things need updating accordingly. It probably makes sense to make this an actual module that the main one is dependent on, so that the processing is more "natural". That said, kiwi does probably need to know that the flags are changing so that it can either do its own mapping or do something else.

Vogtinator: IMO this PR needs to be reverted. Either legacy option handling (fully) works or the legacy options are removed.

Conan-Kudo: I've prepared a revert PR for now, unless @bdrung comes up with fixes: https://github.com/dracut-ng/dracut-ng/pull/2182

(Sorry about this, I blame the sick...)

Vogtinator: Oh, get well soon!

bdrung: Valid points. Let's revert this change then with your points in the commit message: https://github.com/dracut-ng/dracut-ng/pull/2183

Issue #2179, opened 2026-02-03
multipath module does not include kpartx.rules


bdrung:
**Describe the bug**

In dracut-core

file: /usr/lib/dracut/modules.d/70multipath/module-setup.sh

Line 167:

```
    inst_rules 40-multipath.rules 56-multipath.rules \
        62-multipath.rules 65-multipath.rules \
        66-kpartx.rules 67-kpartx-compat.rules \
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        11-dm-mpath.rules 11-dm-parts.rules
}
```

The udev rules numbers have changed for both multipath.rules and kpartx.rules due to https://salsa.debian.org/linux-blocks-team/multipath-tools/-/blob/master/debian/patches/0004-Debian-specific-priority-of-udev-rules.patch?ref_type=heads

Effect: kpartx is not run when boooting, multipath disk partitions are not found, boot fails

**Distribution used**
 Ubuntu 25.10

**Dracut version**
108

**Init system**
systemd

**To Reproduce**
Steps or code to reproduce the behavior.

**Expected behavior**

What should happen: kpartx runs, partitions are found, boot continues

SHOULD ADD

```
+ 60-multipath.rules 60-kpartx.rules \
```

With this change: booting proceeds without multipath assembly errors

For some reason multipath disk detection (but not partition detection) happened even when multipath-rules wasn't being included but I've added it for completeness and to avoid any more issues in future

There may be other similar errors elsewhere. Pointing to the wrong (non-existent) udev rule is a fairly easy mistake to overlook until it directly affects you

It may be better to dynamically scan for *.rules filenames without numbers as this will make the build process more robust in the face of dependent package changes. As it stands dracut will happily build an initrd which fails to boot - meaning errors are likely to recur as udev or other packages are tweaked

**Additional context**
Initial report: https://launchpad.net/bugs/2137190

devkontrol:
CC @mwilck

PR #2180, opened 2026-02-03
fix(multipath): include Debian-specific priority of udev rules

bdrung:
In Debian multipath-tools 0.13.0-1 carries
0004-Debian-specific-priority-of-udev-rules.patch which changes the priority of udev rules with following reasoning:

11-dm-*.rules supposedly should apply after dmsetup rules, but in Debian these have priority 55 instead of 11. Also for historic reasons, multipath.rules and kpartx.rules have a slightly higher prio in Debian than upstream.

The result is that kpartx is not run when booting, multipath disk partitions are not found, and boot fails.

So include the renumbered udev rules similar to what has been done in the dm and lvm Dracut modules.

Fixes: https://github.com/dracut-ng/dracut-ng/issues/2179
Bug-Ubuntu: https://launchpad.net/bugs/2137190

### Checklist

### DIFF
--- a/modules.d/70dmraid/module-setup.sh
+++ b/modules.d/70dmraid/module-setup.sh
@@ -78,6 +78,8 @@ install() {
     inst "$moddir/dmraid.sh" /sbin/dmraid_scan
 
     inst_rules 66-kpartx.rules 67-kpartx-compat.rules
+    # Debian udev rules
+    inst_rules 60-kpartx.rules
 
     inst_libdir_file "libdmraid-events*.so*"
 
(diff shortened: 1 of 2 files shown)


jozzsi:
CC @mwilck for feedback. Thank you !

The referenced bug has an interesting idea from @bdrung that is perhaps good to highlight here as well

> It may be better to dynamically scan for *.rules filenames without numbers as this will make the build process more robust in the face of dependent package changes.

bdrung:
> It may be better to dynamically scan for *.rules filenames without numbers as this will make the build process more robust in the face of dependent package changes.

That could be a feature implemented after the 110 release.

bdrung:
Changing lvm2 is the requisite to dropping https://salsa.debian.org/linux-blocks-team/multipath-tools/-/blob/master/debian/patches/0004-Debian-specific-priority-of-udev-rules.patch

PR #2187, opened 2026-02-05
Switch to /usr/lib from /lib for hooks

vittyvk:
As suggested by @Conan-Kudo on https://github.com/dracut-ng/dracut-ng/pull/1752 we
can use the opportunity and remove one more hardcoded /lib thing by moving distro shipped 
hooks location to /usr/lib.

### Checklist

### DIFF
--- a/test/TEST-70-ISCSI/test.sh
+++ b/test/TEST-70-ISCSI/test.sh
@@ -137,7 +137,7 @@ test_setup() {
         --add-confdir test-makeroot \
         -a "crypt lvm mdraid" \
         -I "setsid blockdev" \
-        -i ./create-client-root.sh /lib/dracut/hooks/initqueue/01-create-client-root.sh \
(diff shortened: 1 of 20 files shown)


test/TEST-70-ISCSI/test.sh
```diff
@@ -137,7 +137,7 @@ test_setup() {
         --add-confdir test-makeroot \
         -a "crypt lvm mdraid" \
         -I "setsid blockdev" \
-        -i ./create-client-root.sh /lib/dracut/hooks/initqueue/01-create-client-root.sh \
+        -i ./create-client-root.sh /var/var/lib/dracut/hooks/initqueue/01-create-client-root.sh \
```
jozzsi: Perhaps this is why the test is failing as this should be /var/lib ?

vittyvk: Sorry, I botched my sed completely. I'm fixing this and combining it with Benjamin's suggestion to use /usr/lib for the testsuite hooks as well.

LaszloGombos (approved):

PR #2191, opened 2026-02-05
ci: no need to explicitly add qemu-net anymore

jozzsi:
Commit 94b21ab enables dracut to automatically detect when qemu-net is required.

Ensure this condition is covered by CI by removing qemu-net from the dracut module lists.

### Checklist


LaszloGombos:
I think this might regress some tests if hostonly_mode is set strict.

bdrung (approved):

Issue #2193, opened 2026-02-05
TEST-45-SYSTEMD-IMPORT fails: libxml2.so.16 and libxxhash.so.0 missing in initrd


bdrung:
**Describe the bug**
I tried to run TEST-45-SYSTEMD-IMPORT as autopkgtest during package build, but in failed on Debian unstable:

```
[    8.852390] systemd-importd[389]: (transfer1) Pulling 'http://10.0.2.2:33787/root.tar.zst', saving as 'root'.
[    8.853165] systemd-importd[389]: (transfer1) Operating on image directory '/run/machines'.
[    9.177565] systemd-importd[389]: (transfer1) Downloading 79B for http://10.0.2.2:33787/root.tar.zst.sha256.
[    9.185045] systemd-importd[389]: (transfer1) Acquired 79B.
[    9.185784] systemd-importd[389]: (transfer1) Download of http://10.0.2.2:33787/root.tar.zst.sha256 complete.
[    9.187019] systemd-importd[389]: (transfer1) Downloading 8.2M for http://10.0.2.2:33787/root.tar.zst.
[    9.192564] systemd-importd[389]: (transfer1) Failed to retrieve image file.
[    9.193139] systemd-importd[389]: (transfer1) Exiting.
[    9.209151] systemd-importd[389]: Transfer process failed with exit code 1.
[    9.209898] systemd-importd[389]: Cannot emit varlink reply, ignoring: Invalid request descriptor
[    9.211991] varlinkctl[382]: Method call failed: io.systemd.Import.TransferFailed
[    9.215552] systemd[1]: systemd-import@run-machines-root.service: Main process exited, code=exited, status=1/FAILURE
[    9.216969] systemd[1]: systemd-import@run-machines-root.service: Failed with result 'exit-code'.
[FAILED] Failed to start systemd-import@run…http://10.0.2.2:33787/root.tar.zst.
[    9.226300] systemd[1]: Failed to start systemd-import@run-machines-root.service - Download of http://10.0.2.2:33787/root.tar.zst.
See 'systemctl status systemd-import@run-machines-root.service' for details.
```

The test starts working when calling dracut with `--install /usr/lib/x86_64-linux-gnu/libxxhash.so.0` which includes these two files:
```
/usr/lib/x86_64-linux-gnu/libxxhash.so.0
/usr/lib/x86_64-linux-gnu/libxxhash.so.0.8.3
```

@aafeijoo-suse 

**Distribution used**
Debian unstable

[...]

PR #2197, opened 2026-02-06
fix(dracut): also check libraries when resolving lazy

bdrung:
Dracut might resolve executable dependencies lazy. In this case all libraries are installed without resolving their dependencies.

Later Dracut will resolve dependencies of the included executables, but it will not explicitly check libraries. This will miss resolving dependencies of libraries that are used via `dlopen()`.

Fixes: https://github.com/dracut-ng/dracut-ng/issues/2193

### Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2860,15 +2860,15 @@ if [[ $kernel_only != yes ]]; then
     fi
 
     if [[ ${DRACUT_RESOLVE_LAZY-} ]] && [[ $DRACUT_INSTALL ]]; then
-        dinfo "*** Resolving executable dependencies ***"
+        dinfo "*** Resolving dependencies for executables and libraries ***"
         # shellcheck disable=SC2086
-        find "$initdir" -type f -perm /0111 -not -name '*.ko*' -print0 \
+        find "$initdir" -type f \( -perm /0111 -or -name '*.so*' \) -not -name '*.ko*' -print0 \
             | xargs -r -0 $DRACUT_INSTALL ${initdir:+-D "$initdir"} ${dracutsysrootdir:+-r "$dracutsysrootdir"} -R ${DRACUT_FIPS_MODE:+-f} --
         # shellcheck disable=SC2181
         if (($? == 0)); then
-            dinfo "*** Resolving executable dependencies done ***"
+            dinfo "*** Resolving dependencies for executables and libraries done ***"
         else
-            dfatal "Resolving executable dependencies failed"
+            dfatal "Resolving dependencies for executables and libraries failed"
             exit 1
         fi
     fi


dracut.sh
```diff
@@ -2860,15 +2860,15 @@ if [[ $kernel_only != yes ]]; then
...
-        find "$initdir" -type f -perm /0111 -not -name '*.ko*' -print0 \
+        find "$initdir" -type f \( -perm /0111 -or -name '*.so*' \) -not -name '*.ko*' -print0 \
             | xargs -r -0 $DRACUT_INSTALL ${initdir:+-D "$initdir"} ${dracutsysrootdir:+-r "$dracutsysrootdir"} -R ${DRACUT_FIPS_MODE:+-f} --
         # shellcheck disable=SC2181
         if (($? == 0)); then
-            dinfo "*** Resolving executable dependencies done ***"
+            dinfo "*** Resolving executable and library dependencies done ***"
         else
-            dfatal "Resolving executable dependencies failed"
+            dfatal "Resolving executable and library dependencies failed"
```
Conan-Kudo: `Resolving` -> `Resolution of`

bdrung: Is "Resolving dependencies failed" bad grammar?

How about this alternative: "Failed to resolve executable and library dependencies"?

Conan-Kudo: That works too as it fixes the mixed tense problem. "Resolution of" works because it indicates the correct past tense rather than the current mixed tense sentence structure.

Conan-Kudo: But this also applies to the "done" (success) case too.

bdrung: There are quite some cases of that:
```
dracut.sh:dinfo "*** Including modules done ***"
dracut.sh:    dinfo "*** Installing kernel module dependencies done ***"
dracut.sh:            dinfo "*** Resolving executable dependencies done ***"
dracut.sh:    dinfo "*** Stripping files done ***"
dracut.sh:    dinfo "*** Squashing the files inside the initramfs done ***"
dracut.sh:    dinfo "*** Hardlinking files done ***"
dracut.sh:                dinfo "*** Creating UEFI image file '$outfile' done ***"
dracut.sh:                    dinfo "*** Creating signed UEFI image file '$outfile' done ***"
dracut.sh:                    dinfo "*** Creating UEFI image file '$outfile' done ***"
dracut.sh:        dinfo "*** Creating initramfs image file '$outfile' done ***"
dracut.sh:        dinfo "*** Moving image file '$outfile' to '$outfile_final' done ***"
```
We should also be consistent. How about you creating a PR to fix the wording?

Conan-Kudo: Uggh. Can we just fix this instance since we're changing these strings anyway? I can follow up with the rest afterward.

bdrung: What i don't like with the rewording is that the beginning of the string changes. We have a bunch of:
```
Doing something
Doing something done
```
or "Doing something failed". "Resolution of" will break this pattern.

bdrung: How about "Resolving dependencies for executables and libraries (|done|failed)"?

Conan-Kudo: Sure, I'll take it.

Conan-Kudo (approved):

PR #2201, opened 2026-02-06
test: stop mdadm after creating client rootfs

bdrung:
To be on the safe side, stop mdadm after creating client rootfs in the iSCSI tests 70 and 71 (similar to `create-root.sh` in test 26).

### Checklist


LaszloGombos (approved):
LGTM

PR #2208, opened 2026-02-07
fix(dmsquash-live): load kernel module before mount

jozzsi:
The mount call may require the kernel module for the filesystem type to be loaded in order to succeed.

Follow the already established sequence of calls for mounting.

```
det_img_fs
load_fstype
mount
```

### Checklist

### DIFF
--- a/modules.d/70dmsquash-live/dmsquash-live-root.sh
+++ b/modules.d/70dmsquash-live/dmsquash-live-root.sh
@@ -341,6 +341,8 @@ if [ -e "$SQUASHED" ]; then
     SQUASHED_LOOPDEV=$(losetup -f)
     losetup -r "$SQUASHED_LOOPDEV" "$SQUASHED"
     mkdir -m 0755 -p /run/initramfs/squashfs
+    fstype=$(det_img_fs "$SQUASHED_LOOPDEV")
+    load_fstype "$fstype"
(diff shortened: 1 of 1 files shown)


bdrung:
Where have you seen this problem?

jozzsi:
> Where have you seen this problem?

This issue can be reproduced with `busybox mount` (when the kernel module that is not yet loaded before the mount is not built into the kernel).

TEST-30 and TEST-31 on the `alpine:busybox` would need this change to pass.

> load_fstype is already called in lines 119 & 122.

Yes indeed, this is an established pattern, which is why I hope this is a good solution to move forward. Unfortunatelly those `load_fstype` calls are for different kernel modules (e.g. in the case of TEST-30, these are for `ext4`, as the `squashfs` is wrapped inside and `ext4` image).

Without this change squashfs kernel module would not be loaded and `busybox mount` would not load that kernel module either.

With this change (and some additional workarounds) the following test-run should succeed (and without it, it would fail). 

```
TEST_CONTAINER_COMMAND="apk fix blkid; cp /sbin/blkid /usr/sbin/" test/test.sh alpine:busybox-edge 30
```

devkontrol (approved):
Thanks for the review @FGrose

bdrung (approved):
LGTM since we already have that pattern.

PR #2212, opened 2026-02-07
test: introduce make_client_rootfs/make_server_rootfs and make_test_rootfs

bdrung:
Test 72 NBD has the helper functions `make_client_rootfs` and `make_server_rootfs` which makes the test easier to read.

Refactor the other tests and introduce the helper functions `make_client_rootfs` and `make_server_rootfs` there as well. Introduce `make_test_rootfs` for the simpler tests (where there are no client/server).

### Checklist

### DIFF
--- a/test/TEST-11-USR-MOUNT/test.sh
+++ b/test/TEST-11-USR-MOUNT/test.sh
@@ -41,7 +41,7 @@ test_run() {
     client_run "readonly root snapshot" "rd.fstab=0 subvol=snapshot-root"
 }
 
-test_setup() {
+make_test_rootfs() {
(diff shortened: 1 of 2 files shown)


jozzsi:
I would like to suggets to only make this change for those tests that require a server initramfs.

bdrung:
> For me, the word client implies that there is a server, so when I see `make_client_root` in test 11, I would be looking for make_server_root as well as part of test 11.

Ah, so you connect client to server. So how about using `make_test_rootfs` then for test 11 and 26?

Conan-Kudo (approved):

PR #2214, opened 2026-02-07
test: do not require pkg-config

bdrung:
The command `pkg-config` might not be installed in the test environment (for example on Debian autopkgest).

Replace calling `pkg-config` by calling `systemctl --version`.

### Checklist

### DIFF
--- a/test/TEST-40-SYSTEMD/test.sh
+++ b/test/TEST-40-SYSTEMD/test.sh
@@ -20,6 +20,14 @@ test_run() {
     check_qemu_log
 }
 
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
Is it reasonable to avoid `pkg-config` in the test environment like this? I'm wondering if it is a hardship to have it available in the autopkgtest environment...

bdrung:
I want to test dracut with the minimal needed dependencies in autopkgtest to catch missing dependencies. That's why pkg-config (plus its dependencies) is not installed in the autopkgtest.

PR #2226, opened 2026-02-09
ci: enable TEST-45-SYSTEMD-IMPORT in openSUSE

aafeijoo-suse:
openSUSE Tumbleweed ships systemd-v258 since snapshot 20260205.

### DIFF
--- a/.github/workflows/daily-systemd.yml
+++ b/.github/workflows/daily-systemd.yml
@@ -82,9 +82,6 @@ jobs:
                     - container: debian:latest
                       test: "45"
                     # TEST-45-SYSTEMD-IMPORT requires systemd >= v258
-                    - container: opensuse:latest
-                      test: "45"
-                    # TEST-45-SYSTEMD-IMPORT requires systemd >= v258
                     - container: ubuntu:rolling
                       test: "45"
         container:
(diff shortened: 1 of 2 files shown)


.github/workflows/daily-systemd.yml
```diff
@@ -83,6 +83,7 @@ jobs:
                       test: "45"
                     # TEST-45-SYSTEMD-IMPORT requires systemd >= v258
                     - container: opensuse:latest
+                      architecture: {runner: 'ubuntu-24.04-arm', tag: 'arm'}
```
bdrung: Why do you exclude arm here? Do you have different systemd versions on different architectures?

aafeijoo-suse: Yes, the change has not yet propagated to aarch64, still in v257.9: https://download.opensuse.org/ports/aarch64/tumbleweed/repo/oss/aarch64/

bdrung: Should we wait for the propagation or will that take multiple days?

aafeijoo-suse: It depends on the specific errors found in ARM... the latest snapshot is 20260202. I guess we can wait.

PR #2230, opened 2026-02-10
fix(dracut): remove udevrulesconfdir variable

jozzsi:
There is about 50 places in the dracut source code where the `/etc/udev/rules.d` path is hardcoded.

Having a `udevrulesconfdir` dracut variable gives an impression that dracut and dracut configuration files can configure and overwrite udevrulesconfdir variable, this functionality however never was supported and never worked.

Instead of making an attempt to fix the usage of udevrulesconfdir variable, let's just remove it and require that udevrulesconfdir always maps to /etc/udev/rules.d for dracut.

### Checklist

### DIFF
--- a/modules.d/74udev-rules/module-setup.sh
+++ b/modules.d/74udev-rules/module-setup.sh
@@ -101,6 +101,6 @@ install() {
         inst_multiple -H -o \
             "$udevconfdir"/udev.conf \
             "$udevconfdir/udev.conf.d/*.conf" \
-            "$udevrulesconfdir/*.rules"
+            "$udevconfdir/rules.d/*.rules"
     fi
 }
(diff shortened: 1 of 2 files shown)


modules.d/74udev-rules/module-setup.sh
```diff
@@ -101,6 +101,6 @@ install() {
         inst_multiple -H -o \
             "$udevconfdir"/udev.conf \
             "$udevconfdir/udev.conf.d/*.conf" \
-            "$udevrulesconfdir/*.rules"
+            "/etc/udev/rules.d/*.rules"
```
Conan-Kudo: shouldn't this just be `$udevconfdir/rules.d`?

jozzsi: Yes, it should. I changed it, thanks @Conan-Kudo .

That said, the eventual goal is also remove `udevconfdir` - that is going to be another PR, so let's leave it out from this PR.

For additional context - some distributions - e.g. Fedora (use to) "overwritre" these default variable - abfa116. This is very misleading as dracut actually only seem to support `udevconfdir` set to `/etc/udev`, and everything else would lead to error as far as I can tell.

Issue #2227, opened 2026-02-09
pre-pivot and cleanup hooks were not run


bdrung:
**Describe the bug**

The test 14 (hooks) fails in Ubuntu autopkgtest with:
```
executing /assertion.sh
Checked 30 hooks that should have been run.
Checked 18 hooks that should not have been run.
**************************FAILED**************************
Hook '/var/lib/dracut/hooks/pre-pivot/testhook-difname-var.sh' was not run in the initrd, but should have.
Hook '/etc/dracut/hooks/pre-pivot/testhook-difname-etc.sh' was not run in the initrd, but should have.
Hook '/usr/lib/dracut/hooks/pre-pivot/testhook-difname-lib.sh' was not run in the initrd, but should have.
Hook '/var/lib/dracut/hooks/pre-pivot/testhook-samename3.sh' was not run in the initrd, but should have.
Hook '/etc/dracut/hooks/pre-pivot/testhook-samename2.sh' was not run in the initrd, but should have.
Hook '/var/lib/dracut/hooks/cleanup/testhook-difname-var.sh' was not run in the initrd, but should have.
Hook '/etc/dracut/hooks/cleanup/testhook-difname-etc.sh' was not run in the initrd, but should have.
Hook '/usr/lib/dracut/hooks/cleanup/testhook-difname-lib.sh' was not run in the initrd, but should have.
Hook '/var/lib/dracut/hooks/cleanup/testhook-samename3.sh' was not run in the initrd, but should have.
Hook '/etc/dracut/hooks/cleanup/testhook-samename2.sh' was not run in the initrd, but should have.
**************************FAILED**************************
```
Full log: https://autopkgtest.ubuntu.com/results/autopkgtest-resolute/resolute/amd64/d/dracut/20260209_204205_23b3f@/log.gz

**Distribution used**
Ubuntu

**Dracut version**
110

**Init system**
systemd

PR #2238, opened 2026-02-11
fix: use return instead of exit in hooks

bdrung:
The `modules-export.sh` pre-pivot hook calls `exit 0`. This causes the sourcing script (`dracut-pre-pivot.sh` on systemd or `init.sh` from the base Dracut module) to exit. The subsequent hooks are not executed.

Use `return` instead of `exit` in kernel-modules-export pre-pivot hook.

Apply a similar fix to other cases of `exit` and also update the `dracut.modules` man page to warn about this behaviour.

Fixes: https://github.com/dracut-ng/dracut-ng/issues/2227

### Checklist

### DIFF
--- a/modules.d/77syslog/rsyslogd-start.sh
+++ b/modules.d/77syslog/rsyslogd-start.sh
@@ -7,7 +7,7 @@ command -v getarg > /dev/null || . /lib/dracut-lib.sh
 # prevent starting again if already running
 if [ -f /var/run/syslogd.pid ]; then
     read -r pid < /var/run/syslogd.pid
-    kill -0 "$pid" && exit 0
+    kill -0 "$pid" && return 0
 fi
 
 rsyslog_config() {


man/dracut.modules.7.adoc
```diff
@@ -44,7 +44,12 @@ scripts and udev rules.
 dracut modules can insert custom script at various points, to control the boot
 process.
 These hooks are plain directories containing shell scripts ending with ".sh",
-which are sourced by init.
+which are sourced by init or other scripts.
+Because the hook scripts are sourced, use `return` instead of `exit` to quit
+a hook.
+Otherwise, the sourcing script will exit.
+Please unset any used variables (or, better yet, use local variables within
+functions) to avoid influencing subsequent code.
```
Conan-Kudo:
```suggestion
Since hook scripts are sourced instead of directly executed, `return` must
be used instead of `exit` at the end of a hook to avoid having dracut quit prematurely.

Please unset any used variables (or, better yet, use local variables within
functions) to avoid influencing subsequent code.
```

man/dracut.modules.7.adoc
```diff
@@ -44,7 +44,12 @@ scripts and udev rules.
 dracut modules can insert custom script at various points, to control the boot
 process.
 These hooks are plain directories containing shell scripts ending with ".sh",
-which are sourced by init.
+which are sourced by init or other scripts.
+Since hook scripts are sourced instead of directly executed, `return` must
+be used instead of `exit` at the end of a hook to avoid having dracut quit prematurely.
```
Conan-Kudo: Switch backticks for quotes and small reflow for prematurely to go on the next line.

Conan-Kudo (changes requested):

Conan-Kudo (approved):

PR #2243, opened 2026-02-15
fix(rngd): install rngd service dependency files

jozzsi:
Arch package requires the service configuration file.

Fixes: https://gitlab.archlinux.org/archlinux/packaging/packages/dracut/-/issues/12

### Checklist

CC @killermoehre @nhorman @mtorromeo 

Reference: https://gitlab.archlinux.org/archlinux/packaging/packages/rng-tools/-/blob/main/rngd.service?ref_type=heads

### DIFF
--- a/modules.d/16rngd/module-setup.sh
+++ b/modules.d/16rngd/module-setup.sh
@@ -39,6 +39,12 @@ install() {
         inst_simple "${moddir}/sysconfig" "/etc/sysconfig/rngd"
     fi
 
+    # Install the hosts local configurations
+    if [[ $hostonly ]]; then
+        inst_multiple -H -o \
+            /etc/conf.d/rngd
+    fi
+
     # make sure dependent libs are installed too
     inst_libdir_file opensc-pkcs11.so
 


jozzsi:
> The environment path is selected by the distribution. The upstream project has no way to enforce that

This is a very sad and unfortunate point of view. I do not remember asking to enforce - I have hoped for making a recommendation so for packaging/distributions are not forced to come up with a path (the same way dracut recommends `/etc/dracut.conf.d`).

jozzsi:
> Such a recommendation would just be in the form of adding an EnvironmentFile key to rngd.service.in in the repo.

Yes, exactly. 

> You're welcome to propose such a patch to do that,

Let me wait for others to chime in as well, but I plan to make a PRunless someone else beats me to it.

> but I guarantee you that distributions won't just blindly follow it.

That may be so, but in that case those distributions that are intentionally diverging from upstream projects without documented and good reasons might not get upstream support from dracut either, and this is how the two projects can help each other.

jozzsi:
> As the upstream maintainer for rng-tools, distributions and users always get support, regardless of where they choose to put environment configuration, thats their choice

Ok, I should not have made statements regarding rng-tools. I have uploaded a PR to rng-tools. Thank you very much for engaging in this conversation. 

https://github.com/nhorman/rng-tools/pull/225

modules.d/16rngd/module-setup.sh
```diff
@@ -39,6 +39,12 @@ install() {
         inst_simple "${moddir}/sysconfig" "/etc/sysconfig/rngd"
     fi
 
+    # Install the hosts local configurations
+    if [[ $hostonly ]]; then
+        inst_multiple -H -o \
+            /etc/conf.d/rngd
```
Conan-Kudo, line 45: The `/etc/conf.d` path doesn't exist in almost every distribution family, so this would need to be adjusted to handle the `/etc/sysconfig/rngd` and `/etc/default/rngd` paths most distributions use.

Conan-Kudo (changes requested):

modules.d/16rngd/module-setup.sh
jozzsi: > The /etc/conf.d path doesn't exist in almost every distribution family

Thanks you @Conan-Kudo . Do you have a data source for this statement ? If not I can look around. It exists in Arch and Gentoo. 

As I pointed out in the commit message this fix is primary for Arch.

nhorman: The reasoning you gave [here](https://github.com/nhorman/rng-tools/pull/225#issue-3944953776) for this change wasn't a fix for anything at all, but rather, to repeat your statement:

```
In an effort to provide a recommendation for the service configuration add an EnvironmentFile key to rngd.service.in in the repo
```

i.e., by my read, you're hoping that, by adding a default-ish environment file path, that distributions would just start using it.  

Arch apparently chose their own service environment file path, independently of RHEL/Fedora/Ubuntu/Mint/Debian/etc.  Gentoo did the same thing, and picked something different.  Different distributions suggest where these things should go, and generally they delegate that choice of locations to the packages themselves (i.e. each pacakge makes its own unit file, which specifies the location of an optional EnvironmentFile, which can be wherever that package decides it wants it to be).

This is the point I've been trying to make.  Distributions pick where environmental files like this go, and often they delegate that decision to individual packages, which often (but not always) consent to putting in the same directory path (/etc/sysconfig, /etc/default, /var/share, etc).  Even if some distributions decide to follow any upstream suggestion here, systemd still provides the ability to override it, and distributions are free to take advantage of that, so anything that works by altering the upstream unit file, does so by pure luck.

They're not going to stop picking their own location and allowing packages to pick something different, just because some upstream package put an extra line in their sample service unit file.

So, what exactly are you trying to fix here?  It seems to me like you want the upstream project to look more like arch/gentoo, which doesn't really fix anything.  It might make your life easier (though I don't think so).  Other distributions are just going to write their own unit file (which they mostly do now anyway).

Do you want dracut to pick up environment files from an alternate directory, other than the assumed /etc/sysconfig?  Thats  reasonable, but changing the rng-tools provided unit file isn't going to move the needle on that.  If you really want to standardize these locations so that all distributions use the same location, your best bet is to get involved with the Linux Foundations Filesystem Hierarchy Standard, to include a definition of this location, so that slowly  distributions start doing it (though that's a huge lift).

The better solution (if the above is your goal) is to make dracut more generic in where it finds these environment variable files.  That probably amounts to modifying your above patch to something like:

```
inst_simple "${systemdsystemunitdir}/rngd.service "
ENV_PATH=$(grep EnvironmentFile ${systemdsystemunitdir}/rngd.service | sed -e's/.*=[-]*//')
ENV_FILE=$(basename $ENV_PATH)
inst_simple "${moddir}/sysconfig" $ENV_PATH
sed -e"s/EnvironmentFile.*/EnvironmentFile=${moddir}/sysconfig/$ENV_FILE/"  "${systemdsystemunitdir}/rngd.service"
```

That may not be exactly right, but you get the idea.  Use the systemd unit file provided by the distribution to locate the environment file that it specifies, and copy that location into the initramfs file, then re-write the initramfs file (if thats needed) to point to the internal-initramfs location.  That should work for all distributions, regardless of the location of the EnvironmentFile.

jozzsi:

PR #2245, opened 2026-02-16
fix(systemd-tmpfiles): do not include tmpfiles configured in /etc

### LINKED ISSUE #1706: polkitd user / group missing in systemd-tmpfiles
As part of systemd-tmpfiles dracut module contents of _/etc/tmpfiles.d_ is copied to initramfs ("$tmpfilesconfdir/*.conf"). On my system I have following file for polkit there:
`D /run/polkit-1/rules.d 0755 polkitd root 0 -`

Which demands that user _polkitd_ exists (created from /usr/lib/sysusers.d/polkit.conf). But module systemd-sysusers includes only /usr/lib/sysusers.d/basic.conf.
So we have following error on early boot stage:
`systemd-tmpfiles[192]: /etc/tmpfiles.d/polkit.conf:1: Failed to resolve user 'polkitd': No such process`

To fix that either add `inst_sysusers polkit.conf` to https://github.com/dracut-ng/dracut-ng/blob/main/modules.d/78systemd-sysusers/module-setup.sh or delete `"$tmpfilesconfdir/*.conf"` from https://github.com/dracut-ng/dracut-ng/blob/6e2af855dce969dedcf7ffe01773b4d8f4d9dc0b/modules.d/11systemd-tmpfiles/module-setup.sh#L48 as it contains user-defined configuration and may include users / groups as well.
I'd prefer 2nd method (so we don't copy custom modifications).

aafeijoo-suse:
Users may have tmpfiles configured in /etc to be created after switching root, always including this user configuration in the initrd may break their setups.

Fixes #1706

### Checklist

### DIFF
--- a/modules.d/11systemd-tmpfiles/module-setup.sh
+++ b/modules.d/11systemd-tmpfiles/module-setup.sh
@@ -50,7 +50,6 @@ install() {
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
-            "$tmpfilesconfdir/*.conf" \
             "$systemdsystemconfdir"/systemd-tmpfiles-clean.service \
             "$systemdsystemconfdir/systemd-tmpfiles-clean.service.d/*.conf" \
             "$systemdsystemconfdir"/systemd-tmpfiles-setup.service \


devkontrol:
Should the variable `tmpfilesconfdir` removed from dracut as part of this PR as this PR removes the only usage of this variable ?

aafeijoo-suse:
> Should the variable `tmpfilesconfdir` removed from dracut as part of this PR as this PR removes the only usage of this variable ?

Maybe it's not used in upstream modules, but we don't know elsewhere. Also, there are currently other unused variables:

```
$ confdirvars=($(grep -r set_global_var.*confdir dracut.sh | awk '{ print $7 }' | cut -d '"' -f2))
$ for v in "${confdirvars[@]}"; do grep -q -r --exclude=dracut.sh $v || echo "$v not used"; done
environmentconfdir not used
systemdportableconfdir not used
systemduserconfdir not used
```

bdrung:
I searched for all those four variables on https://codesearch.debian.net/ - and they are not used anywhere in Debian.

devkontrol:
> Maybe it's not used in upstream modules, but we don't know elsewhere

We should remove it precisely for this reason - we want to know for sure that no dracut module is looking at files under `tmpfilesconfdir`, so that we can make conclusions about how dracut works.

To maintain code long term, we should have a reason why the code exists and ideally we should have some code coverage. Before this PR, we had some explanation for the why and some code coverage for this variable. After this PR, both the explanation and code coverage is gone for this variable, so PR arguably increases the maintenance cost.

I like to measure maintenance cost not based on the number of lines of code, but how well we understand why the code is there to begin with. 

> Also, there are currently other unused variables

I hope we do not try to address this as part of this PR.

devkontrol (approved):
I do not mind `tmpfilesconfdir` removal as a follow-up PR.

LGTM

devkontrol:
> I do not mind `tmpfilesconfdir` removal as a follow-up PR.

See https://github.com/dracut-ng/dracut/pull/2544

PR #2247, opened 2026-02-16
test(SYSTEMD-SYSEXT): introduce tests for systemd-sysext

aafeijoo-suse:
See individual commits.

It requires to land #2246 first in order to test this on the CI.

### Checklist

### DIFF
--- a/.github/workflows/daily-systemd.yml
+++ b/.github/workflows/daily-systemd.yml
@@ -47,6 +47,7 @@ jobs:
                     - "43"
                     - "44"
                     - "45"
+                    - "46"
                 exclude:
                     - container: arch:latest
                       architecture: {runner: 'ubuntu-24.04-arm', tag: 'arm'}
@@ -84,6 +85,18 @@ jobs:
                     # TEST-45-SYSTEMD-IMPORT requires systemd >= v258
                     - container: ubuntu:rolling
                       test: "45"
+                    # TEST-46-SYSTEMD-SYSEXT requires systemd >= v258
+                    - container: azurelinux:3.0
+                      test: "46"
+                    # TEST-46-SYSTEMD-SYSEXT requires systemd >= v258
+                    - container: centos:latest
+                      test: "46"
+                    # TEST-46-SYSTEMD-SYSEXT requires systemd >= v258
+                    - container: debian:latest
+                      test: "46"
+                    # TEST-46-SYSTEMD-SYSEXT requires systemd >= v258
+                    - container: ubuntu:rolling
+                      test: "46"
         container:
             image: ghcr.io/dracut-ng/${{ matrix.container }}
             options: '--device=/dev/kvm --privileged'
--- a/modules.d/11systemd-sysext/module-setup.sh
+++ b/modules.d/11systemd-sysext/module-setup.sh
@@ -17,7 +17,7 @@ check() {
 depends() {
 
     # This module has external dependency on other module(s).
-    echo systemd
+    echo systemd systemd-veritysetup
     # Return 0 to include the dependent module(s) in the initramfs.
     return 0
 
--- a/modules.d/11systemd-sysext/module-setup.sh
+++ b/modules.d/11systemd-sysext/module-setup.sh
@@ -25,7 +25,7 @@ depends() {
 
 # Install kernel module(s).
 installkernel() {
-    hostonly=$(optional_hostonly) instmods -s erofs
+    hostonly=$(optional_hostonly) instmods -s erofs ext4 squashfs
     hostonly='' instmods loop
 }
 
--- a/modules.d/11systemd-sysext/module-setup.sh
+++ b/modules.d/11systemd-sysext/module-setup.sh
@@ -41,7 +41,6 @@ install() {
     # regular files, but not with directory trees.
 
     inst_multiple -o \
-        "/usr/lib/extension-release.d/extension-release.*" \
         "$systemdsystemunitdir"/systemd-confext${_suffix}.service \
         "$systemdsystemunitdir/systemd-confext${_suffix}.service.d/*.conf" \
         "$systemdsystemunitdir"/systemd-confext-sysroot.service \
@@ -70,7 +69,6 @@ install() {
     # Install the hosts local user configurations if enabled.
     if [[ $hostonly ]]; then
         inst_multiple -H -o \
-            "/etc/extension-release.d/extension-release.*" \
             "$systemdsystemconfdir"/systemd-confext${_suffix}.service \
             "$systemdsystemconfdir/systemd-confext${_suffix}.service.d/*.conf" \
             "$systemdsystemconfdir"/systemd-sysext${_suffix}.service \
--- a/test/TEST-46-SYSTEMD-SYSEXT/test.sh
+++ b/test/TEST-46-SYSTEMD-SYSEXT/test.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+set -eu
+
+# shellcheck disable=SC2034
+TEST_DESCRIPTION="root filesystem on a ext4 filesystem with systemd and extensions"
+
+# Uncomment these to debug failures
+#DEBUGFAIL="systemd.show_status=1 systemd.log_level=debug"
+
+test_check() {
+    local binary
+
+    for binary in systemd-repart openssl; do
+        if ! type -p "$binary" &> /dev/null; then
+            echo "Test needs $binary... Skipping"
+            return 1
+        fi
+    done
+}
+
+test_run() {
+    declare -a disk_args=()
+    qemu_add_drive disk_args "$TESTDIR"/root.img root
+
+    "$testdir"/run-qemu \
+        "${disk_args[@]}" \
+        -append "root=LABEL=dracut $TEST_KERNEL_CMDLINE" \
+        -initrd "$TESTDIR"/initramfs.testing
+
+    # Check if the message "All OK" is in QEMU logs
+    check_qemu_log
+
+    # Also check if the message "dracut-sysext-success" is in QEMU logs
+    check_qemu_log "$QEMU_LOGFILE" "dracut-sysext-success"
+}
+
+test_setup() {
+    local crt_dir fs_type confext_name sysext_name
+
+    # Create plain root filesystem
+    build_client_rootfs "$TESTDIR/rootfs"
+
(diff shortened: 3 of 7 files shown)


test/TEST-46-SYSTEMD-SYSEXT/test.sh
```diff
@@ -0,0 +1,129 @@
...
+prompt = no
+distinguished_name = dracut_distinguished_name
+[dracut_distinguished_name]
+C = ES
+ST = State
+L = Locality
+O = Org
+OU = Org Unit
+CN = Common Name
+emailAddress = dracut@email.com
```
bdrung: Use something that is reserved for testing/documenting:
```suggestion
emailAddress = dracut@exampe.com
```

modules.d/11systemd-sysext/module-setup.sh
```diff
@@ -17,7 +17,7 @@ check() {
 depends() {
 
     # This module has external dependency on other module(s).
-    echo systemd
+    echo systemd systemd-veritysetup
```
devkontrol, line 20: This change would imply that `systemd-sysext` dracut module is not functional without `systemd-veritysetup` module. Can you please elaborate on this new dependency ? 

The commit message explains a scenario where `systemd-sysext`  and `systemd-veritysetup` used together for a particular use-case but it does not explain why `systemd-sysext` could not be used without `systemd-veritysetup` any more.

aafeijoo-suse: > This change would imply that systemd-sysext dracut module is not functional without systemd-veritysetup module. Can you please elaborate on this new dependency ?

Adding the `systemd-veritysetup` dracut module as a dependency avoids having to duplicate the following:
- The `systemd-sysext` binary dlopens `libcryptsetup`. It also dlopens `libmount` (covered with the `systemd` dracut module) and `libblkid` (covered with the `systemd-udevd` dracut module).
- For verity auth it needs the `dm-verity` kernel module.
- It doesn't work without the stuff added by the `dm` dracut module either.
- It needs `/{etc,usr/lib}/verity.d/*.crt` for verity userspace verification.

It does not need the other units/binaries/conf pulled by the `systemd-veritysetup` dracut module, but it makes more sense to me to have all the verity stuff only in one place.

> The commit message explains a scenario where systemd-sysext and systemd-veritysetup used together for a particular use-case but it does not explain why systemd-sysext could not be used without systemd-veritysetup any more.

No one said it can't, the commit message explains why the dependency was added. People are free to add extensions that are not trusted (but not in `/.extra`).

test/TEST-46-SYSTEMD-SYSEXT/test.sh
```diff
@@ -0,0 +1,129 @@
...
+    mkdir -p "$TESTDIR/$confext_name/etc/extension-release.d"
+    pushd "$TESTDIR/$confext_name/etc"
+    echo "dracut-sysext-success" > "$confext_name.marker"
+    {
+        grep -e "^ID=" -e "^VERSION_ID=" /etc/os-release
+        echo "CONFEXT_SCOPE=initrd"
+    } > "extension-release.d/extension-release.$confext_name"
+    popd
+    # systemd-repart creates erofs partitions by default: override it to use
+    # a more widely supported format like squashfs
```
vittyvk: Given that erofs is the default I'd suggest we do the test twice: for EROFS and for squashfs. E.g. RHEL10 has already deprecated squashfs/squashfs-tools package.

.github/workflows/daily-systemd.yml
```diff
@@ -84,6 +85,18 @@ jobs:
...
+                      test: "46"
+                    # TEST-46-SYSTEMD-SYSEXT requires systemd >= v258
+                    - container: centos:latest
+                      test: "46"
+                    # TEST-46-SYSTEMD-SYSEXT requires systemd >= v258
+                    - container: debian:latest
+                      test: "46"
+                    # TEST-46-SYSTEMD-SYSEXT requires systemd >= v258
+                    - container: ubuntu:rolling
+                      test: "46"
```
vittyvk, line 99: Given that systemd-258 should be available in Fedora since F43, would it be possible to also add fedora:latest and fedora:rawhide to the list?

test/TEST-46-SYSTEMD-SYSEXT/test.sh
```diff
@@ -0,0 +1,129 @@
...
+EOF
+    openssl req \
+        -config "dracut.cnf" \
+        -new -x509 \
+        -newkey rsa:1024 \
+        -keyout "dracut.key" \
+        -out "dracut.crt" \
+        -days 365 \
+        -nodes
+    popd
```
vittyvk, line 60: There is nothing wrong with using the config but we don't really care about C/ST/L/... here in the test so I guess we can simplify this to a one-liner like

`openssl req -quiet -newkey rsa:1024 -nodes -keyout dracut.key -new -x509 -days 365 -subj "/CN=Dracut Test Key/" --outform PEM -out dracut.pem`

(just to shorten the code).

devkontrol:
@vittyvk - is this a better time to try to advance this PR ? Thank you !

vittyvk:
@devkontrol yes, we certainly want tests for sysext :-) I only had a few minor comments on the MR for @aafeijoo-suse . When this one lands, I can probably add more tests, especially for systemd-{sys,conf}ext-sysroot services which I'm adding in #2391

aafeijoo-suse:
> I only had a few minor comments on the MR for @aafeijoo-suse .

@vittyvk sorry, I just realized that you said something about some comments, but I cannot see any... did you forget to make them visible?

vittyvk:
> > I only had a few minor comments on the MR for @aafeijoo-suse .
> 
> @vittyvk sorry, I just realized that you said something about some comments, but I cannot see any... did you forget to make them visible?

Indeed, silly me forgot to finish the review and comments were pending forever. Pushed now, sorry for the confusion!

.github/workflows/daily-systemd.yml
aafeijoo-suse: They are already being tested, this is the exclusion list.

test/TEST-46-SYSTEMD-SYSEXT/test.sh
aafeijoo-suse: I'd not overload the CI testing the same thing with different filesystems. So I changed the code to test erofs if it's available, and squashfs if erofs is not available. What do you think?

test/TEST-46-SYSTEMD-SYSEXT/test.sh
aafeijoo-suse: Ok, changed to PEM. If you're wondering why I didn't use the name "dracut.pem":

```
[    2.972948] localhost (sd-merge)[510]: Validation of dm-verity signature failed via the kernel, trying userspace validation instead: Required key not available
[    2.972993] localhost (sd-merge)[510]: Skipping file '/etc/verity.d/dracut.pem', suffix is not '.crt'.
[    2.973045] localhost (sd-merge)[510]: No userspace dm-verity certificates found.
[    2.973047] localhost (sd-merge)[510]: Activation of signed Verity volume worked neither via the kernel nor in userspace, can't activate.
[    2.973175] localhost (sd-merge)[510]: Failed to activate verity device /dev/mapper/loop2p1-18-verity: Destination address required
[    2.973272] localhost (sd-merge)[510]: Releasing crypt device /dev/loop2p2 context.
[    2.974187] localhost (sd-merge)[510]: Releasing device-mapper backend.
[    2.977940] localhost (sd-merge)[510]: Closing read only fd for /dev/loop2p2.
[    2.984173] localhost (sd-merge)[510]: Removed partition /dev/loop2p1
[    2.984292] localhost (sd-merge)[510]: Removed partition /dev/loop2p2
[    2.985818] localhost (sd-merge)[510]: Removed partition /dev/loop2p3
[    3.000095] localhost systemd-confext[255]: '(sd-merge)' failed with exit status 1.
[    3.000105] localhost systemd-confext[255]: Failed to merge hierarchies
```

test/TEST-46-SYSTEMD-SYSEXT/test.sh
vittyvk: Sounds good to me, I think that at least centos stream runners should benefit from this.

test/TEST-46-SYSTEMD-SYSEXT/test.sh
vittyvk: Oh, I didn't even know there's such limitation for /etc/verity.d.

aafeijoo-suse: Me neither, I don't think it's documented anywhere.: https://github.com/systemd/systemd/blob/da8a8b5de86f28705d681d720741a2d3fcb4b3a1/src/shared/dissect-image.c#L3093

modules.d/11systemd-sysext/module-setup.sh
devkontrol: > No one said it can't

I do not understand this response. 

On debian if I do not have `systemd-cryptsetup` debian package installed, after this change I would not be able to use the systemd-sysext dracut module.

Edit: 

I understand this better now.

> The systemd-sysext binary dlopens libcryptsetup

Would this mean that on debian `systemd-cryptsetup` debian package is already a dependency for `systemd-sysext` dracut module ? @bdrung what do you think ? Is this for all recent systemd versions that we support ?

aafeijoo-suse: FTR: `libcryptsetup` != `systemd-cryptsetup` (the systemd binary) != `systemd-cryptsetup` (the dracut module).

The `systemd-veritysetup` dracut module already installs `libcryptsetup`, which is what system extensions need:

https://github.com/dracut-ng/dracut-ng/blob/4369d92365ad0998a2bd2da42ad465b73209d0fc/modules.d/11systemd-veritysetup/module-setup.sh#L68

devkontrol: What confused me is that I assumed that `libcryptsetup` gets installed by `systemd-cryptsetup` package on Debian - which is not the case. 

Given
> systemd-sysext binary dlopens libcryptsetup

I have no further objection for this change. 

I would have preferred more granular dependencies, but I do not want to block this PR from advancing if there are no concerns on this from others.

devkontrol (approved):

bdrung (approved):

PR #2244, opened 2026-02-15
fix(systemd-pcrextend): add NVPCR definition files

jozzsi:
According to https://github.com/systemd/systemd/issues/40159#issuecomment-3760597003 the initramfs is expected to include the nvpcr definition files from systemd for v259 compatibility.

### Checklist
- [ ] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #2249, opened 2026-02-16
nvmf: dynamic subsystem configuration and default interface naming

mwilck:
This pull request contains enhancements to the nvmf module.

1. Instead of hard-coding the NVMe subsystems to connect to with `rd.nvmf.discover=` cmdline options, enable just reading the connections at boot time from the NBFT. The behavior can be customized at initramfs build time using the dracut.conf option `nvmf_nbft_mode={static,nbft,match}`.
2. Instead of always using the interface names `nbft0`, `nbft1` etc., allow using regular interface names as determined by the system's general interface naming policy. This requires NetworkManager 1.54.0 or newer. The feature can be enabled using the command line option `rd.nvmf.nm=1`, which is automatically enabled with `hostonly_cmdline` if NM 1.54.0 or newer is detected on the system.

The set also contains a minor man page fix.

### Checklist

Cc: @aafeijoo-suse, @tbzatek, @johnmeneghini

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1186,6 +1186,7 @@ drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
 [[ $kernel_image_l ]] && kernel_image=$(path_rel_to_abs "$kernel_image_l")
 [[ $sbat_l ]] && sbat="$sbat_l"
 [[ $machine_id_l ]] && machine_id="$machine_id_l"
+[[ $nvmf_nbft_mode ]] || nvmf_nbft_mode=match
 
 # shellcheck source=./dracut-functions.sh
 . "$dracutbasedir"/dracut-functions.sh
@@ -2055,6 +2056,14 @@ else
     unset enhanced_cpio
 fi
 unset threecpio_help_output
+
+case $nvmf_nbft_mode in
+    nbft | match | static) ;;
+    *)
+        dwarn "Invalid value \"$nvmf_nbft_mode\" for nvmf_nbft_mode. Assuming \"match\"."
+        nvmf_nbft_mode=match
+        ;;
+esac
 
 if [[ $no_kernel != yes ]] && ! [[ -d $srcmods ]]; then
     dfatal "Cannot find module directory $srcmods"
--- a/dracut.sh
+++ b/dracut.sh
@@ -1194,6 +1194,7 @@ drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
 [[ $kernel_image_l ]] && kernel_image=$(path_rel_to_abs "$kernel_image_l")
 [[ $sbat_l ]] && sbat="$sbat_l"
 [[ $machine_id_l ]] && machine_id="$machine_id_l"
+[[ $nvmf_nbft_mode ]] || nvmf_nbft_mode=match
 
 TMPDIR="$(realpath -e "$tmpdir")"
 readonly TMPDIR
@@ -2063,6 +2064,14 @@ else
     unset enhanced_cpio
 fi
 unset threecpio_help_output
+
+case $nvmf_nbft_mode in
+    nbft | match | static) ;;
+    *)
+        derror "Invalid value \"$nvmf_nbft_mode\" for nvmf_nbft_mode. Assuming \"match\"."
+        nvmf_nbft_mode=match
+        ;;
+esac
 
 if [[ $no_kernel != yes ]] && ! [[ -d $srcmods ]]; then
     dfatal "Cannot find module directory $srcmods"
--- a/man/dracut.cmdline.7.adoc
+++ b/man/dracut.cmdline.7.adoc
@@ -946,6 +946,13 @@ Requires the dracut 'nvmf' module.
     the initramfs was built. Targets specified with rd.nvmf.discover on the
     kernel command line will still be tried.
 
+**rd.nvmf.nm**::
+    Rely on NetworkManagers nm-initrd-generator tool to set up network
+    interfaces from the NVMe Boot Firmware Table (NBFT). This has the effect
+    that the network interfaces generated from the NBFT will have regular
+    interface names rather than ++nbft0++ etc. *This requires NetworkManager
+    1.54.0 or higher*.
+
 **rd.nvmf.hostnqn=**__<hostNQN>__::
     NVMe host NQN to use
 
--- a/man/dracut.conf.5.adoc
+++ b/man/dracut.conf.5.adoc
@@ -400,6 +400,21 @@ Defaults to _initramfs-$\{kernel\}.img_.
 Let kernel-install create a rescue image (default=no).
 This configuration option applies only to kernel-install.
 
+*nvmf_nbft_mode*="__{match|nbft|static}__"::
+For systems that use the ++nvmf++ module for booting from NVMe over Fabrics
+(NVMeoF), specify how to set command line options for NVMe subsystems that are
+listed in the NVMe Boot Firmware Table (NBFT). This option only has an effect
+if ++hostonly_cmdline++ is set to ++yes++.  If set to ++static++, generate
+cmdline entries for all NVMe subsystems detected while the initramfs was
+built. If set to ++nbft++, don't generate any entries for NVMe subsystems using
+the ++tcp++ transport (i.e. rely on NBFT parsing for detecting NVMe subsystems
+at boot time). If set to ++match++ (default), only add cmdline entries for
+++tcp++ subsystems that do not match any entry in the NBFT, comparing subsystem
+IP address (traddr) and port (trsvcid).  Use ++nbft++ or ++match++ if you want
+the system to adapt if the content of the NBFT changes between reboots. Use
+++static++ if you want connect to the same subsystems, always, no matter what
+the content of the NBFT is.  The default is ++match++.
+
 == Files
 _/etc/dracut.conf_::
 Old configuration file. It is recommended to use individual files in
--- a/man/dracut.conf.5.adoc
+++ b/man/dracut.conf.5.adoc
@@ -192,6 +192,9 @@ in its current state will not be included, and modules may make additional
 efforts to save more space. Minor changes in hardware or environment can make
 the image unbootable.
 
+WARNING: If chrooted to another root other than the real root device, use
+`--fstab` and provide a valid _/etc/fstab_.
+
 *hostonly_cmdline=*"__{yes|no}__"::
 If set to "yes", store the kernel command line arguments needed in the
 initramfs. If **hostonly="yes"** and this option is not configured, it's
@@ -209,9 +212,6 @@ _<policy>_ can be any directory name found in /dev/disk (e.g. "by-uuid",
 *tmpdir=*"__<temporary directory>__"::
 Specify temporary directory to use.
 +
-WARNING: If chrooted to another root other than the real root device, use
-`--fstab` and provide a valid _/etc/fstab_.
-
 *use_fstab=*"__{yes|no}__"::
 Use _/etc/fstab_ instead of _/proc/self/mountinfo_ (default=no).
 
--- a/modules.d/74nvmf/module-setup.sh
+++ b/modules.d/74nvmf/module-setup.sh
@@ -98,6 +98,7 @@ cmdline() {
         local _address
         local -a _address_parts
         local nbft_entry
+        local -a version
 
         [[ -L "/sys/dev/block/$_dev" ]] || return 0
         cd -P "/sys/dev/block/$_dev" || return 0
@@ -150,6 +151,13 @@ cmdline() {
     if [ -f /etc/nvme/hostid ]; then
         read -r _hostid < /etc/nvme/hostid
         echo -n " rd.nvmf.hostid=${_hostid}"
+    fi
+
+    if dracut_module_included network-manager; then
+        mapfile -t -d . version < <(NetworkManager --version)
+        [[ ${#version[@]} == 3 &&
+            $((10000 * version[0] + 100 * version[1] + version[2])) -ge 15400 ]] \
+            && echo -n " rd.nvmf.nm=1 "
     fi
 
     [[ $hostonly ]] || [[ $mount_needs ]] && {
--- a/modules.d/74nvmf/module-setup.sh
+++ b/modules.d/74nvmf/module-setup.sh
@@ -86,6 +86,7 @@ installkernel() {
 cmdline() {
     local _hostnqn
     local _hostid
+    local -a _nbft_subsystems
 
     # shellcheck disable=SC2317,SC2329  # called later by for_each_host_dev_and_slaves
     gen_nvmf_cmdline() {
@@ -96,6 +97,7 @@ cmdline() {
         local trsvcid
         local _address
         local -a _address_parts
+        local nbft_entry
 
         [[ -L "/sys/dev/block/$_dev" ]] || return 0
         cd -P "/sys/dev/block/$_dev" || return 0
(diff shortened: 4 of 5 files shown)


jozzsi:
> The behavior can be customized at initramfs build time using the dracut.conf option nvmf_nbft_mode={static,nbft,match}.

@mwilck would you be able to share more information why do you think this level of customization is necessary at initramfs build time ?

modules.d/74nvmf/module-setup.sh
```diff
@@ -136,7 +164,18 @@ cmdline() {
         echo -n " rd.nvmf.hostid=${_hostid}"
     fi
 
+    if dracut_module_included network-manager; then
+        mapfile -t -d . version < <(NetworkManager --version)
```
aafeijoo-suse, line 164: Arg, the arithmetic below to get the version is so ugly.. too bad `/usr/libexec/nm-initrd-generator -h` was not updated when nbft support was added... but it's easy to fix it so we can replace this with a simple grep.

modules.d/74nvmf/module-setup.sh
```diff
@@ -86,6 +87,18 @@ installkernel() {
 cmdline() {
     local _hostnqn
     local _hostid
+    local -a _nbft_subsystems
+
+    # Same as in check() above
+    has_nbft() {
```
aafeijoo-suse: This function is duplicated, I'd only have one outside.

modules.d/74nvmf/module-setup.sh
mwilck: I don't understand what you mean here. Do you want me to change this into a grep? IMO grep never works well for version comparisons.
I actually found this way to code the version check quite elegant :stuck_out_tongue_winking_eye: but tastes differ, of course.

modules.d/74nvmf/module-setup.sh
mwilck: I was afraid of polluting the global namespace. But i can rename the function of course.

mwilck:
> > The behavior can be customized at initramfs build time using the dracut.conf option nvmf_nbft_mode={static,nbft,match}.
> 
> @mwilck would you be able to share more information why do you think this level of customization is necessary at initramfs build time ?

I hope so :smile: 

It has been an ongoing debate for years (mostly in the iSCSI context) whether or not target information like this should be hardcoded in the initramfs. My personal opinion is "no", we should hardcode as little as possible and rely on the information from the firmware (NBFT)[^hostid]. The advantage is, as stated above, that the system will still be able to boot when the NVMe namespaces have been moved to a different subsystem or network segment. But there are other people who hold the opposite opinion. They tend to argue that the firmware information might be stale or wrong during a future boot attempt, and prefer to hardcode everything.

These two extremes correspond to the "nbft" and "static" setups. The "match" option is an attempt to get the "best of both worlds". Some NVMeoF subsystems/namespaces may be necessary for booting but not mentioned in the NBFT. That's not very likely, but possible. The "match" option would still hard-code these, while omitting those subsystems that are listed in the NBFT. In practice, more often than not, "match" will have same effect as "nbft". Still, I found it not too hard to implement, and thought might be less of a breaking change than "nbft". 

[^hostid]: In principle we could hardcode even less and omit also the hostnqn and hostid fields, as these will be taken from the NBFT, too.

jozzsi:
> These two extremes correspond to the "nbft" and "static" setups. The "match" option is an attempt to get the "best of both worlds". 

Thank you very much @mwilck . 

One mechanism for including/excluding components during generation time are dracut modules. It is a little extremw idea, but would it be possible to create new dracut module for some (maybe not all) nvmf_nbft_mode modes ? 

E.g. the current nvmf dracut module would be  "nbft" and to opt in to "static" mode, you would need to add the nvmf-static dracut module. We already have very good mechanisms for adding and omitting dracut modules and dependencies between modules (and some distros can even decide not to ship/support e.g. nvmf-static module)

Creating a new dracut module is cheap in terms of overhead during initramfs generation and this idea would avoid polluting dracut.sh, which is meant to be dracut module independent. 

Let's see how strongly others feel about adding a new global dracut.sh mode vs creating a new dracut module. I just wanted to add an alternative into the discussion.

modules.d/74nvmf/module-setup.sh
aafeijoo-suse: > Do you want me to change this into a grep?

No, I didn't mean that, you did the best you could with the info it provides. I meant that a simple missing change in the help text [1] would allow to simplify this check with `/usr/libexec/nm-initrd-generator -h | grep -q rd.nvmf.nonbft`.

[1] https://gitlab.freedesktop.org/NetworkManager/NetworkManager/-/blob/main/src/nm-initrd-generator/nm-initrd-generator.c?ref_type=heads#L162

man/dracut.conf.5.adoc
```diff
@@ -400,6 +400,21 @@ Defaults to _initramfs-$\{kernel\}.img_.
 Let kernel-install create a rescue image (default=no).
 This configuration option applies only to kernel-install.
 
+*nvmf_nbft_mode*="__{match|nbft|static}__"::
+For systems that use the ++nvmf++ module for booting from NVMe over Fabrics
+(NVMeoF), specify how to set command line options fore NVMe subsystems that are
```
aafeijoo-suse: Minor typo: `s/fore/for/`

aafeijoo-suse (approved):
It looks ok, thanks.

man/dracut.cmdline.7.adoc
```diff
@@ -946,6 +946,13 @@ Requires the dracut 'nvmf' module.
     the initramfs was built. Targets specified with rd.nvmf.discover on the
     kernel command line will still be tried.
 
+**rd.nvmf.nm**::
+    Rely on NetworkManagers nm-initrd-generator tool to set up network
+    interfaces from the NVMe Boot Firmware Table (NBFT). This has the effect
+    that the network interfaces generated from the NBFT will have regular
+    interface names rather than ++nbft0++ etc. *This requires NetworkManager
+    1.54.0 or higher*.
```
tbzatek, line 954: Nice, glad that you managed to find the NM NBFT parser useful! So this effectively means we'll need to rework our userspace NM hooks that do match `nbftX` interface names and finally upstream them - see https://github.com/linux-nvme/nvme-cli/pull/2620/changes#diff-9f2319cde42d7021165f379d55e1bd6321102d00d3adc15de85ce98174718a5d

Or perhaps you have an alternative solution in works?

mwilck: > Or perhaps you have an alternative solution in works?

I admit I haven't considered that part yet. This is about activating NBFT connections that for whatever reason haven't been activated in the initramfs, right?

NM will generate entries under `/run/NetworkManager/system-connections` for those using the initrd generator. Could we set some other property in these connection files that could be tested in these hooks?

mwilck: Note that this PR does _not_ imply that `nbft$X` interface names _cannot_ be used any more. It is still possible to add `ifname=` parameters on the dracut cmdline. It is also possible to add systemd `.link` files or udev rules that modify the interface name as the user sees fit.

This PR just avoids **enforcing** the `nbft$X` name for booting.

mwilck: FTR, I have created https://github.com/linux-nvme/nvme-cli/pull/3117 for a new NM hook.

That's an add-on fix though, and not directly related to this PR.

tbzatek: Thanks, that looks good!

tbzatek (approved):
> As a matter of fact, current dracut (in hostonly mode, without this PR) **does** hard-code the subsystem connection in cmdline parameters. Why is this so? Because both you and I previously overlooked it?

You're right, it was there for a while. Funny that I never noticed :-o

> Yes, but also because "hostonly" can mean different things depending on who you ask. For me, in this case, it means "this host is configured to boot from the NBFT". But in principle, it might as well mean "this host is configured to boot from exactly this subsystem and namespace". It comes down to "are the settings in the system BIOS, and possibly on remote DHCP and discovery servers, part of 'hostonly' configuration, or not?".
> 
> DHCP is actually a good analogy. I believe most people would agree that if a system is configured to use DHCP, it would be wrong to hard-code static IP address configuration (matching the current DHCP lease) in the initrd. I'd argue that NVMe (and iSCSI) discovery responses should be treated in the same way as DHCP. But dracut currently doesn't.

Got it, fully agree. I always had the volatile NBFT table in mind and never really considered extracting current connections for the purpose of hardcoding them.

Speaking of that, there's perhaps a second level of NBFT table volatility - discovery descriptors without corresponding subsystem entries - i.e. the point where nvme-cli performs discovery and connect-all on its own. Arguably this is a corner case and perhaps out of the scope of this PR, but the possibility is there - and admins may intentionally use this concept to seamlessly reconfigure clients on next reboot just by changing the discovery entries on the target.

Although we have formalized typical use cases in the Boot Specification, I believe we might need something similar in the context of the OS. Best practices and/or use cases coming from the dracut nvmf module implementation.

> > If true, that kinda breaks our (downstream) userspace reconnection concept that only takes the volatile NBFT table in account.
> 
> We don't do that, do we? Any system that boots from NBFT is free to configure as many _additional_ NVMeoF targets as it sees fit. And I am very certain that most of these systems do. We just need to distinguish between NVMeoF targets **for booting** (i.e. for the OS) and **for data**. We aren't concerned about the latter here at all, only about the connections we need to make to get the OS up.
> 
> It's true that users who convert configuration from the NBFT into static `config.json` entries might also face inconsistency issues if NBFT content changes. But that's far beyond the scope of this PR and up to the nvme-cli maintainers to think about.

I'm not concerned about additional targets for data either. My concern here would be again re-discovery from NBFT and connect-all, possibly working with different targets. But again, it's just probably an administrator's concern and outside of the scope of this PR as well.

> But you're of course right wrt the "disparity". I have to confess I haven't yet tested what happens if the NBFT content and the static entries differ. I suppose dracut would try to connect to both. In the "good" case, the connection to either would fail, causing delays and confusing error messages. In the "bad" case, both would succeed, and the system might boot from a wrong or outdated file system.

Right, I think this falls to the same category as my points above. Not a blocker I'd say.

> > One TODO I had deep down in my list is to check how does this all (NBFT) cope with provided `discovery.conf`/`config.json` as this dracut module basically packs three distinct functionalities altogether and not all are being tested thoroughly.
> 
> IMO we need to treat these cases separately. You can configure everything statically and use `discovery.conf` or `config.json`. That's possible and I'm pretty sure that it works; but people who configure their hosts this way will _not_ want to use the NBFT. How much sense does it make to configure the NBFT in the BIOS and install/boot from it, while not trusting the NBFT to provide correct data on subsequent boots? NBFT users will want to refer to the NBFT for booting, and to `config.json` or `discovery.conf` for everything else.

True, that falls more into the "arbitrary data targets" category I guess. 

Anyway, I'm fine with these changes, go ahead!

devkontrol (dismissed):

mwilck:
@josszi:

First of all, I appreciate your volunteer efforts a lot.

I am just wondering what hinders merging of this PR. I was asking for open issues with it. There is [your proposal to split the nvmf module](https://github.com/dracut-ng/dracut-ng/pull/2249#issuecomment-3930758061), which I and @aafeijoo-suse have responded to. You didn't follow up, so I thought that you were satisfied with our response. Nobody else objected to our comments, either.

If the additional configuration options are really a problem, I can rework this PR to use less or even no parameters, at the risk that issues may arise in some environments. But honestly, the idea of having multiple `nvmf` modules that are identical except for such tiny policy differences still makes no sense to me. It's going to confuse users, too.

> If contributors are only uploading PR of their own interest

As you may be aware, this doesn't apply to me. I have spent considerable time reviewing dracut PRs, obviously mostly in my area of expertise (booting from network devices and complex block device stacks). Other SUSE employees, in particular @aafeijoo-suse, are regular contributors to this project.

> So here is an example - https://github.com/dracut-ng/dracut-ng/pull/2244

Sorry, I have no clue about these `.nvpcr` files. Maybe @aplanas can comment on that PR.

jozzsi:
> You didn't follow up, so I thought that you were satisfied with our response. 

Yes, I am satisfied with your responses. Thanks @mwilck .

> in particular @aafeijoo-suse, are regular contributors to this project.

I try to review patches in areas where I am confirmable giving feedback. I am blocked from reviewing PR's from @aafeijoo-suse .

modules.d/74nvmf/module-setup.sh
```diff
@@ -136,7 +153,17 @@ cmdline() {
         echo -n " rd.nvmf.hostid=${_hostid}"
     fi
 
+    if dracut_module_included network-manager; then
+        mapfile -t -d . version < <(NetworkManager --version)
+        [[ ${#version[@]} == 3 &&
+            $((10000 * version[0] + 100 * version[1] + version[2])) -ge 15400 ]] \
```
pvalena, line 166: This probably deserves a comment; I get that it just tests on a specific version (or greater), but it might also be good to know why.

dracut.sh
```diff
@@ -2056,6 +2057,14 @@ else
 fi
 unset threecpio_help_output
 
+case $nvmf_nbft_mode in
+    nbft | match | static) ;;
+    *)
+        dwarn "Invalid value \"$nvmf_nbft_mode\" for nvmf_nbft_mode. Assuming \"match\"."
```
pvalena: This should IMO make dracut fail instead.

pvalena (approved):
LGTM, a I think splitting into multiple modules doesn't make much sense from design POV, and although module-specific configuration in `dracut.sh` are undesired, it makes full sense here (having more module-friendly approach would be better, but that's not in scope for this PR). 
The added granularity seems great. I have only one question -- when adding a configuration option, would it make sense to add it as a commandline argument as well (like [the others](https://github.com/mwilck/dracut-ng/blob/d0f425d0bdae48deb666650bc449cefe86b76731/dracut.sh#L142))?

modules.d/74nvmf/module-setup.sh
mwilck: @pvalena,

I mentioned in the commit description of 1f8c8b6 that NetworkManager 1.54 is required to read the NBFT. Is that explanation not sufficient? Do you want a comment in the code?

dracut.sh
mwilck: I beg to disagree. dracut should not fail easily, because it is frequently called non-interactively from post-installation scripts. What's the benefit of failing here instead of using a sane default?

modules.d/74nvmf/module-setup.sh
pvalena: Yes, a brief comment in the code would bring more clarity, since there's a hardcoded number (and I doubt people read documentation or commit history much).

dracut.sh
pvalena: Since the configuration is really invalid, the behaviour might be undesired (and leaving the user with just a warning that might be left unseen) ... I think there's a difference between default and "ignore invalid configuration". But this decision is quite arbitrary; and very much a corner case, so feel free to leave it as-is.

mwilck: I'm going to change this from `dwarn` to `derror` to increase visibility. But I still don't think that dracut should fail. This will most probably not cause a non-functional initrd to be built, whereas aborting while building the initrd is always fatal.

mwilck:
Rebased to latest main, and fixed the remaining minor issues.

mwilck:
While I fixed the minor issues here, I created #2496 now as a superset of this PR. It implements the command line option `--nvmf-nbft-mode` as suggested by @pvalena, and adds more fixes.

I'll leave it up to the maintainers if they'd rather merge this PR (which should be fully reviewed now) or #2496.

devkontrol:
I believe this PR can be closed now that https://github.com/dracut-ng/dracut/pull/2496 landed. 

Please reopen (and rebase) if still needed.

PR #2260, opened 2026-02-24
feat(overlayfs): support tmpfs size parameter in rd.overlay

nadzyah:
When `rd.overlay=tmpfs:size=<size>` is passed on the kernel command line, mount a dedicated tmpfs with the specified options for the overlayfs upper directory, instead of using the default `/run` tmpfs.

Supported options (comma-separated after `tmpfs:`) are `size=`, `nr_blocks=` and `nr_inodes=`. If mounting fails, the script falls back gracefully to the default `/run` tmpfs overlay.

### Checklist

### DIFF
--- a/man/dracut.cmdline.7.adoc
+++ b/man/dracut.cmdline.7.adoc
@@ -1109,6 +1109,29 @@ The **rd.overlay.readonly** option, which allows a persistent overlayfs to
 be mounted read-only through a higher level transient overlay directory, has
 been implemented through the multiple lower layers feature of OverlayFS.
 
+**rd.overlay=**__tmpfs:[size=<size>][,nr_blocks=<n>][,nr_inodes=<n>]__::
+Mounts a dedicated tmpfs with the specified options for the OverlayFS upper
+directory, instead of using the default `/run` tmpfs.  This allows controlling
+the size of the writable overlay at boot time, which is useful for deployments
+where the available RAM is known in advance (e.g. MAAS ephemeral environments).
++
+Supported options (comma-separated):
++
+--
+* _size=<size>_ - Maximum size of the tmpfs (e.g. `4G`, `512M`, `25%`).
+  Accepts the same values as the tmpfs `size=` mount option.
+* _nr_blocks=<n>_ - Maximum number of blocks allocated to the tmpfs.
+* _nr_inodes=<n>_ - Maximum number of inodes in the tmpfs.
+--
++
+[listing]
+.Examples
+----
+rd.overlay=tmpfs:size=4G
+rd.overlay=tmpfs:size=25%
+rd.overlay=tmpfs:size=512M,nr_inodes=100000
+----
+
 **rd.overlay=**__{LABEL=<label>|UUID=<uuid>|PARTUUID=<uuid>|PARTLABEL=<label>|/dev/<device>}__::
 Specifies the storage device for persistent overlay (for both live images and
 regular OverlayFS).  When a block device is specified, it will be mounted and
--- a/modules.d/70overlayfs/prepare-overlayfs.sh
+++ b/modules.d/70overlayfs/prepare-overlayfs.sh
@@ -10,6 +10,7 @@ overlay=$(get_rd_overlay)
 
 overlay_mode="tmpfs"
 overlay_device=""
+overlay_tmpfs_opts=""
 
 case "$overlay" in
     LABEL=* | UUID=* | PARTLABEL=* | PARTUUID=* | /dev/*)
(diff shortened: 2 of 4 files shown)


modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -20,6 +21,27 @@ case "$overlay" in
             overlay_mode="tmpfs"
         fi
         ;;
+    tmpfs:*)
+        overlay_mode="tmpfs"
+        # Parse comma-separated key=value pairs from rd.overlay=tmpfs:key=val,
+        _remainder="${overlay#tmpfs:}"
+        while [ -n "$_remainder" ]; do
```
bdrung: Why not using `IFS=, while read` instead?

man/dracut.cmdline.7.adoc
```diff
@@ -1109,6 +1109,29 @@ The **rd.overlay.readonly** option, which allows a persistent overlayfs to
 be mounted read-only through a higher level transient overlay directory, has
 been implemented through the multiple lower layers feature of OverlayFS.
 
+**rd.overlay=**__tmpfs:{size=<size>[,nr_blocks=<n>,nr_inodes=<n>]}__::
```
bdrung: The placement of brackets doesn't look right to me. I am not exactly sure what the correct format is. Something like:
```
**rd.overlay=**__tmpfs:[size=<size>],[nr_blocks=<n>],[nr_inodes=<n>]__::
**rd.overlay=**__tmpfs:[size=<size>][,nr_blocks=<n>][,nr_inodes=<n>]__::
```

modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -20,6 +21,24 @@ case "$overlay" in
             overlay_mode="tmpfs"
         fi
         ;;
+    tmpfs:*)
+        overlay_mode="tmpfs"
+        # Parse comma-separated key=value pairs from rd.overlay=tmpfs:key=val,
+        old_IFS="$IFS"
```
bdrung: `git grep` shows that `OLDIFS` is the common name used for it.

bdrung: I am not a big fan of setting IFS + resetting and `set -f` plus `set -f`. I tried different approaches to make the code more compact and readable, but I failed. I found no better solution.

modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -20,6 +21,24 @@ case "$overlay" in
...
+        IFS=,
+        set -f
+        for _param in ${overlay#tmpfs:}; do
+            _key="${_param%%=*}"
+            _val="${_param#*=}"
+            case "$_key" in
+                size | nr_blocks | nr_inodes)
```
bdrung, line 34: Can you add a warning in case the key does not match saying that this is unknown and ignored?

modules.d/70overlayfs/prepare-overlayfs.sh
```diff
@@ -61,10 +80,20 @@ if [ "$overlay_mode" = "tmpfs" ]; then
...
     [ -d /run/overlayfs ] || {
-        mkdir -m 0755 -p /run/initramfs/overlay/overlayfs
-        mkdir -m 0755 -p /run/initramfs/overlay/ovlwork
-        ln -sf /run/initramfs/overlay/overlayfs /run/overlayfs
-        ln -sf /run/initramfs/overlay/ovlwork /run/ovlwork
+        _overlay_dir="/run/initramfs/overlay"
+        if [ -n "$overlay_tmpfs_opts" ]; then
+            info "Mounting tmpfs overlay with options: ${overlay_tmpfs_opts}"
+            mkdir -m 0755 -p /run/overlayfs-tmpfs
+            if mount -t tmpfs -o "${overlay_tmpfs_opts}" tmpfs /run/overlayfs-tmpfs; then
```
bdrung: Is there a reason why you use `/run/overlayfs-tmpfs` instead of `/run/initramfs/overlay` here?

test/TEST-21-OVERLAYFS/test.sh
```diff
@@ -38,6 +38,8 @@ test_run() {
     client_run "persistent device overlay (device path)" \
         "rd.overlay=/dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_overlay test.expect=device"
     client_run "fallback to tmpfs (non-existent LABEL)" "rd.overlay=LABEL=NONEXISTENT test.expect=tmpfs"
+    client_run "tmpfs overlay with size (rd.overlay=tmpfs:size=32M)" \
+        "rd.overlay=tmpfs:size=32M test.expect=tmpfs-sized"
```
bdrung: Could you set another option (to include testing the comma-splitting)?

test/TEST-21-OVERLAYFS/assertion.sh
```diff
@@ -26,6 +26,12 @@ else
     fi
 fi
 
+if grep -q 'test.expect=tmpfs-sized' /proc/cmdline; then
+    if ! grep -q " /run/overlayfs-tmpfs tmpfs " /proc/mounts; then
```
bdrung: Can you test that the overlay mount has the correct size configured?

bdrung (commented):
I did review the complete PR. Thanks for updating the documentation (and including examples) and adding a test case for it. I only have some smaller remarks.

modules.d/70overlayfs/prepare-overlayfs.sh
nadzyah: Fixed in https://github.com/nadzyah/dracut-ng/blob/a1b82dc4ae77322bd12fba5e33beb69a399430c3/modules.d/70overlayfs/prepare-overlayfs.sh#L38

Conan-Kudo (approved):

bdrung (approved):

PR #2261, opened 2026-02-25
test/run-qemu: force cpu to neoverse-n1 for arm64 when QEMU_CPU not set

hector-cao:
using cpu=max might expose new unstable features with qemu/edk2 upgrades, this unstability might cause test failures to happen , we are experiencing that in Debian/Ubuntu with edk2 latest version that enables LPA2 that is still unstable.

since using cpu=max is not required for dracut tests, using a named model will offer a better stability for the tests for future qemu/edk2 upgrades. the choosen cpu model for arm64 is neoverse-n1 since it is one of the most stable and proven CPU model from a virtualization standpoint.

### Checklist

### DIFF
--- a/test/run-qemu
+++ b/test/run-qemu
@@ -5,7 +5,16 @@ set -eu
 
 export PATH=/usr/sbin:/usr/bin:/sbin:/bin
 ARCH="${ARCH-$(uname -m)}"
-QEMU_CPU="${QEMU_CPU:-max}"
+if [ -z "${QEMU_CPU-}" ]; then
+    case "$ARCH" in
+        aarch64 | arm64)
+            QEMU_CPU="neoverse-n1"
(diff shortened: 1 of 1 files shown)


test/run-qemu
```diff
@@ -5,7 +5,16 @@ set -eu
 
 export PATH=/usr/sbin:/usr/bin:/sbin:/bin
 ARCH="${ARCH-$(uname -m)}"
-QEMU_CPU="${QEMU_CPU:-max}"
+if [ -z "${QEMU_CPU+x}" ]; then
```
bdrung:
```suggestion
if [ -z "${QEMU_CPU-}" ]; then
```

Conan-Kudo:
Is there a bug report with QEMU about this?

PR #2266, opened 2026-03-01
Add 70devicetree-fw module to add fw files from devicetree firmware-name properties

jwrdegoede:
Devicetree based platforms may need board / model specific firmwares in the initrd.

E.g. on Qualcomm Snapdragon X1e Windows (WoA) laptops the qcom_q6v5_pas ADSP driver gets added to the initrd because dracut includes all pinctrl drivers in the initrd and the 6e80000.pinctrl lpass-lpi-pinctrl device has a supplier:platform:6800000.remoteproc symlink, dragging the ADSP driver into the initrd.

The ADSP driver tries to load model-specific firmware files with the filenames to use specified in a devicetree firmware-name properties.

This module adds support for retrieving the firmware-names from these devicetree properties and adding the firmware files to the initrd.

The ADSP driver, being a remoteproc driver does bind even without the firmware files, operating in a degraded mode. This means e.g. no battery readings and no working audio, since both depend on the ADSP. These firmware files need to be in the initrd to avoid this degraded mode.

The ADSP driver getting brought into the initrd is actually a good thing, since the ADSP also controls the Type-C PD-controller(s) and when the firmware is present and loaded this results in a reset of all Type-C USB ports, so this needs to happen early on, to e.g. avoid USB-storage devices getting disconnected and filesystems on top getting in a broken state.

### Checklist

This is new code, but there are no tests for this. I've thought about this, but I don't see a good way to add a test for this.

### DIFF
--- a/modules.d/70devicetree-firmware/module-setup.sh
+++ b/modules.d/70devicetree-firmware/module-setup.sh
@@ -1,0 +1,71 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# Devicetree based platforms may need board / model specific firmwares in
+# the initrd. E.g. Qualcomm Snapdragon Windows-on-Arm (WoA) laptops need model
+# specific firmware files in the initrd for the ADSP/Type-C controller.
+# The filenames for these firmware files are specified in "firmware-name"
+# devicetree properties, this module copies any such files to the initrd.
+
+# called by dracut
+check() {
+    if [[ $hostonly ]]; then
+        [ -d /sys/firmware/devicetree/base ] || return 1
+    else
+        # ATM install_generic only copies aarch64 qcom model specific firmwares
+        [ "$DRACUT_ARCH" = "aarch64" ] || return 1
+    fi
+    return 0
+}
+
+# hostonly install (hostonly / generic use completely different approaches)
+install_hostonly() {
+    local _fw_names=()
+    # contents of firmware-name file is 0 delimited
+    while read -r -d $'\0' _fw; do
+        _fw_names+=("$_fw")
+    done < <(find /sys/firmware/devicetree/base -name firmware-name -exec cat {} \;)
+
+    local _fw_files=()
+    for _fw_name in "${_fw_names[@]}"; do
+        # shellcheck disable=SC2154 # fw_dir is set by dracut.sh
+        for _fwdir in $fw_dir; do
+            # add '*' after firmware-name for .gz, etc. compression
+            for _fw in "$_fwdir/$_fw_name"*; do
+                [ -f "$_fw" ] || continue
+                _fw=${_fw#"${dracutsysrootdir-}"}
+                _fw_files+=("$_fw")
+                break 2
+            done
+        done
+    done
+    inst_multiple -o "${_fw_files[@]}"
+}
+
+# generic install (hostonly / generic use completely different approaches)
+install_generic() {
+    local _fw_files=()
+    # ATM only qcom WoA laptops need this
+    for _soc in qcom/sc8280xp qcom/x1e80100; do
+        # shellcheck disable=SC2154 # fw_dir is set by dracut.sh
+        for _fwdir in $fw_dir; do
+            # add '*' after mbn, elf for .gz, etc. compression
+            for _fw in "$_fwdir/$_soc"/*/*/*.mbn* "$_fwdir/$_soc"/*/*/*.elf*; do
+                [ -f "$_fw" ] || continue
+                _fw=${_fw#"${dracutsysrootdir-}"}
+                _fw_files+=("$_fw")
+            done
+        done
+    done
+    inst_multiple -o "${_fw_files[@]}"
+}
+
+# called by dracut
+install() {
+    if [[ $hostonly ]]; then
+        install_hostonly
+    else
+        install_generic
+    fi
+}
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -251,6 +251,10 @@ connman:
     - changed-files:
           - any-glob-to-any-file: 'modules.d/[0-9][0-9]connman/*'
 
+devicetree-firmware:
+    - changed-files:
+          - any-glob-to-any-file: 'modules.d/[0-9][0-9]devicetree-firmware/*'
+
 network-legacy:
     - changed-files:
           - any-glob-to-any-file: 'modules.d/[0-9][0-9]network-legacy/*'
--- a/doc_site/modules/ROOT/pages/modules/core.adoc
+++ b/doc_site/modules/ROOT/pages/modules/core.adoc
@@ -69,6 +69,10 @@ code, there are no specific types or categories for dracut modules.
 | debug features
 |
 
+| devicetree-firmware
+| add firmware files from DT firmware-name properties
+|
+
 | dm
 | device-mapper
 | library


jwrdegoede:
Note the [Integration Test / 20 on fedora:latest](https://github.com/dracut-ng/dracut-ng/actions/runs/22550921784/job/65321214924?pr=2266) failing seems to be unrelated to this PR.

This runs on x86_64 and this new module has a `check()` to only be active on aarch64:

```
check() {
    local _arch=${DRACUT_ARCH:-$(uname -m)}
    # Restrict to aarch64 for now, may be used on other archs in the future
    [ "$_arch" = "aarch64" ] || return 1
    ...
}
```

modules.d/70devicetree-fw/module-setup.sh
```diff
@@ -0,0 +1,64 @@
...
+
+    # shellcheck disable=SC2086 # globbing is wanted here
+    inst_multiple -o ${_fw_files}
+}
+
+# generic install (hostonly / generic use completely different approaches)
+install_generic() {
+    # ATM only qcom WoA laptops need this
+    for soc in qcom/sc8280xp qcom/x1e80100; do
+        for vendor in ASUSTeK dell hp microsoft LENOVO; do
```
Conan-Kudo: Is there a reason for a vendor iteration list here? Wouldn't just pulling all the subdirectories make sense?

Conan-Kudo (changes requested):

modules.d/70devicetree-fw/module-setup.sh
jwrdegoede: That is a good idea, I'll prepare a new version with this changed.

jwrdegoede: I've just pushed a new version reworking install_generic() to drop the list of vendors. Note the indentation of the 2 globs used in the new for-loop is a bit weird. This is required to make shfmt happy.

jwrdegoede:
I've just submitted a related PR which fixes the issue mentioned in the description with the ADSP firmware-loading resetting Type-C USB busses causing issues with booting from USB drivers: https://github.com/dracut-ng/dracut-ng/pull/2302

Conan-Kudo:
@jlinton @nullr0ute, could you please sanity-check this for me?

modules.d/70devicetree-fw/module-setup.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Devicetree based platforms may need board / model specific firmwares in
+# the initrd. E.g. Qualcomm Snapdragon Windows-on-Arm (WoA) laptops need model
+# specific firmware files in the initrd for the ADSP/Type-C controller.
+# The filenames for these firmware files are specified in "firmware-name"
+# devicetree properties, this module copies any such files to the initrd.
+
+# called by dracut
+check() {
+    local _arch=${DRACUT_ARCH:-$(uname -m)}
```
bdrung: `DRACUT_ARCH` is always set nowadays. So calling `uname` is not needed any more.

jozzsi:
This PR also needs a doc change (that categorizes all dracut module with a brief description) and a GitHub action labeler change (not sure the usefulness of this, but the project seem to do this for consistency) - here is a blueprint PR that introduces a new dracut module - https://github.com/dracut-ng/dracut-ng/pull/2060 

I would also spell out fw; devicetree-fw --> devicetree-firmware.

'fw' could mean too many other things (framework to name one).

jozzsi:
CC @classabbyamp @alpernebbi - linking an earlier discussion for visibility, hope it does not deflect the discussion from reviewing this PR at hand - https://github.com/dracut-ng/dracut-ng/pull/952

jwrdegoede:
I just did a forced push with a new rebased version which honors the `$fw_dirs` dracut variable as I noticed that this was not picking up firmwares installed under `/lib/firmware/updates` .

I missed that there were some review comments added here in the last 24h. I'll go and take a look at those now and I'll push a new version addressing those.

jwrdegoede:
> @jwrdegoede - this PR (partially) resolves https://bugzilla.redhat.com/show_bug.cgi?id=2455055 ?
> 
> How about this one - https://bugzilla.redhat.com/show_bug.cgi?id=2275920

Yes that should be resolved by this too.

jwrdegoede:
@bdrung @jozzsi thank you for your review / comments. The main reason for this functionality is including the firmware for Qualcomm Snapdragon ADSP and CDSP co-processors which will typically have their modules in the initramfs for various valid reasons.

I've a second PR #2302 open adding a second module which also is related to QC Snapdragon DSP usage. I'm wondering if I should just merge these 2 dracut-modules into a single module ?

We could still review the 2 PRs one at a time, starting with this one and then extending the module with the functionality from the second PR #2302 .

jozzsi:
> I'm wondering if I should just merge this 2 dtacut-modules into a single module ?

I prefer the two separate dracut module (as it is now).

modules.d/70devicetree-fw/module-setup.sh
jwrdegoede: Ack, I've fixed this locally for the next version.

jwrdegoede:
> here is a blueprint PR that introduces a new dracut module - https://github.com/dracut-ng/dracut-ng/pull/2060

Thanks, I'll add the expected `.github/labeler.yml` and `doc_site/modules/ROOT/pages/modules/core.adoc` changes for the next version.

> I would also spell out fw; devicetree-fw --> devicetree-firmware.

Ack, will fix for the next version.

jwrdegoede:
Just pushed a new (and rebased) version.

Changes from initial PR version:
- Honor the $fw_dirs dracut variable so that e.g. /lib/firmware/updates gets checked
- Directly use $DRACUT_ARCH instead of defining _arch local var
- Add the new module to labeler.yml and core.adoc
- Rename module from devicetree-fw to devicetree-firmware

jozzsi (approved):
non-owner LGTM

modules.d/70devicetree-firmware/module-setup.sh
```diff
@@ -0,0 +1,75 @@
...
+# Devicetree based platforms may need board / model specific firmwares in
+# the initrd. E.g. Qualcomm Snapdragon Windows-on-Arm (WoA) laptops need model
+# specific firmware files in the initrd for the ADSP/Type-C controller.
+# The filenames for these firmware files are specified in "firmware-name"
+# devicetree properties, this module copies any such files to the initrd.
+
+# called by dracut
+check() {
+    # Restrict to aarch64 for now, may be used on other archs in the future
+    [ "$DRACUT_ARCH" = "aarch64" ] || return 1
```
bdrung: I would drop that check and rely on the checks below (for devicetree and qcom).

modules.d/70devicetree-firmware/module-setup.sh
```diff
@@ -0,0 +1,75 @@
...
+    if [[ $hostonly ]]; then
+        [ -d /sys/firmware/devicetree/base ] || return 1
+    else
+        # ATM install_generic only copies qcom model specific firmwares
+        [ -d /lib/firmware/qcom ] || return 1
```
bdrung: Checking `/lib/firmware/qcom` could be wrong in case the firmware files are in other paths in `fw_dir`.

modules.d/70devicetree-firmware/module-setup.sh
```diff
@@ -0,0 +1,75 @@
...
+    fi
+    return 0
+}
+
+# hostonly install (hostonly / generic use completely different approaches)
+install_hostonly() {
+    local _fw_names=""
```
bdrung: Could you make `_fw_names` an array? Then that would prevent word splitting.

modules.d/70devicetree-firmware/module-setup.sh
```diff
@@ -0,0 +1,75 @@
...
+    # contents of firmware-name file is 0 delimited
+    while read -r -d $'\0' _fw; do
+        _fw_names="$_fw_names $_fw"
+    done < <(find /sys/firmware/devicetree/base -name firmware-name -exec cat {} \;)
+
+    local _fw_files=""
```
bdrung: Same for `_fw_files`. That could be an array.

modules.d/70devicetree-firmware/module-setup.sh
```diff
@@ -0,0 +1,75 @@
...
+            for _fw in "$_fwdir/$_fw_name"*; do
+                [ -f "$_fw" ] || continue
+                _fw=${_fw#"${dracutsysrootdir-}"}
+                _fw_files="$_fw_files $_fw"
+                break 2
+            done
+        done
+    done
+    # shellcheck disable=SC2086 # word splitting is wanted here
+    inst_multiple -o ${_fw_files}
```
bdrung: Then the array could avoid the shellcheck:

```suggestion
    inst_multiple -o "${_fw_files[@]}"
```

bdrung (changes requested):

bdrung:
Can you reference https://bugzilla.redhat.com/show_bug.cgi?id=2455055 in your git commit message?

modules.d/70devicetree-firmware/module-setup.sh
jwrdegoede: I agree that the check should be dropped for the hostonly case, so that this can e.g. also work on Risc-V.

But I think it makes sense to move it to the generic/non-hostonly path instead of dropping it. More on this in my next comment.

modules.d/70devicetree-firmware/module-setup.sh
jwrdegoede: That is a good point. Rather then doing something complicated with `$fw_dir` I think it would be best to move the `"$DRACUT_ARCH" = "aarch64"` check here.

linux-firmware upstream has some qcom firmwares so depending on how distros split linux-firmware we could end up with a `/lib/firmware/qcom` on non aarch64 but `install_generic()` only deals with qcom aarch64 firmwares in its current version so it would be best to not run this on other archs. All it can do there is add unnecessary firmwares to the initramfs.

As for no longer checking for `/lib/firmware/qcom` if for some reason there is no `qcom` subdir in any of the `$fw_dir` path dirs, then `install_generic()`  is a no-op, so I think completely dropping the dir check and instead using an `$DRACUT_ARCH` check in the generic/non-hostonly path is fine.

modules.d/70devicetree-firmware/module-setup.sh
jwrdegoede: Ack, I'll make this an array for the next version.

modules.d/70devicetree-firmware/module-setup.sh
jwrdegoede: Ack, I'll make this an array for the next version.

modules.d/70devicetree-firmware/module-setup.sh
jwrdegoede: Ack, I'll drop the `shellcheck disable=SC2086` after making `_fw_files` an array for the next version.

jwrdegoede:
> Can you reference https://bugzilla.redhat.com/show_bug.cgi?id=2455055 in your git commit message?

Ack, I'll add a 'Resolves #722' + links to the 2 relevant RH bugzillas in the commit message.

jwrdegoede:
I've just pushed a new version with the following changes:
- Move the `"$DRACUT_ARCH" = "aarch64"` check to the generic/non-hostonly path in `check()`
- Drop the `/lib/firmware/qcom` check from the generic/non-hostonly path in `check()`
- Make _fw_names and _fw_files arrays and drop the shellcheck disable=SC2086
- Reference issue #722 and 2 Red Hat bugzillas in commit message

modules.d/70devicetree-firmware/module-setup.sh
```diff
@@ -0,0 +1,72 @@
...
+
+# Devicetree based platforms may need board / model specific firmwares in
+# the initrd. E.g. Qualcomm Snapdragon Windows-on-Arm (WoA) laptops need model
+# specific firmware files in the initrd for the ADSP/Type-C controller.
+# The filenames for these firmware files are specified in "firmware-name"
+# devicetree properties, this module copies any such files to the initrd.
+
+# called by dracut
+check() {
+    # Restrict to aarch64 for now, may be used on other archs in the future
```
bdrung: That comment is not correct any more. in hostonly mode it only checks for the devicetree. IMO this comment can be dropped.

jwrdegoede: Ack, I've just pushed a new version dropping this.

jwrdegoede:
I've just pushed a new version dropping the obosolete `# Restrict to aarch64 for now, may be used on other archs in the future` comment, no other changes.

bdrung (approved):

devkontrol (approved):

Conan-Kudo (approved):

devkontrol:
https://bugzilla.redhat.com/show_bug.cgi?id=2487066

jwrdegoede:
> https://bugzilla.redhat.com/show_bug.cgi?id=2487066

Thank you for letting me know about this, this should be fixed by #2598, review of #2598 would be appreciated.

Issue #1815, opened 2025-11-01
TEST-72-NBD fails in ubuntu:devel container: dbus service to start too late


bdrung:
**Describe the bug**

TEST-72-NBD fails in Ubuntu container:

```
CLIENT TEST START: NBD root=LABEL=dracut netroot=nbd:IP:port
run-qemu: /usr/bin/qemu-system-x86_64 -enable-kvm -cpu host -M q35 -smp 2 -m 1024 -nodefaults -vga none -display none -no-reboot -watchdog-action poweroff -device virtio-rng-pci -serial stdio -device i6300esb -kernel /boot/vmlinuz-6.17.0-6-generic -device virtio-scsi-pci,id=scsi0 -drive if=none,format=raw,file=/var/tmp/dracut-test.Y9w1qw/marker.img,id=drive-data0 -device scsi-hd,bus=scsi0.0,drive=drive-data0,id=data0,serial=marker -net nic,macaddr=52:54:00:12:34:00,model=virtio -net socket,connect=127.0.0.1:12340 -append 'root=LABEL=dracut rd.luks.uuid=39658691-f797-4168-a960-fe0e52a0209f rd.lv.vg=dracut ip=dhcp netroot=nbd:192.168.50.1:encrypted rd.auto ro console=ttyS0,115200n81' -initrd /var/tmp/dracut-test.Y9w1qw/initramfs.testing
[...]
[ TIME ] Timed out waiting for device dev-nbd0.device - /dev/nbd0.
[DEPEND] Dependency failed for systemd-cryp…658691-f797-4168-a960-fe0e52a0209f.
[DEPEND] Dependency failed for cryptsetup.target - Local Encrypted Volumes.
```

**Distribution used**
Ubuntu devel

**Dracut version**
main branch

**To Reproduce**
Enable TEST-72-NBD in the CI for ubuntu:devel

bdrung:
I created https://bugs.launchpad.net/dracut/+bug/2141603 to fix this in Ubuntu's dbus package.

PR #2273, opened 2026-03-03
fix(dbus): remove After/Requires from dbus service/socket

bdrung:
Following dependencies are added to sockets unless `DefaultDependencies=no` is set (according to systemd.socket man page):

```
Before=sockets.target shutdown.target
After=sysinit.target
Requires=sysinit.target
Conflicts=shutdown.target
```

Following dependencies are added to units unless `DefaultDependencies=no` is set (according to systemd.service man page):

```
Requires=sysinit.target
After=sysinit.target basic.target
Conflicts=shutdown.target
Before=shutdown.target
```

The dbus-broker and dbus-daemon Dracut modules want to remove `After=sysinit.target` and `Requires=sysinit.target` from `dbus.service` and `dbus.socket`. This is done by setting `DefaultDependencies=no` and specifying the remaining default dependencies. The relevant code has been there since the introduction (see commit e1845955ff3d).

The Ubuntu package of dbus already removes some of the default dependencies from the `dbus.service` and `dbus.socket` (see https://launchpad.net/bugs/1438612):

```
DefaultDependencies=no
Wants=sysinit.target
After=sysinit.target basic.target
```

This results in not removing the `sysinit.target` dependency on Ubuntu and causes the dbus service to start too late. Since NetworkManager depends on dbus, it will start too late as well. That causes the NBD test cases to fail.

So remove the `After` and `Requires` entries from the `dbus.service` and `dbus.socket` before setting `DefaultDependencies=no`. Also restore `Before=sockets.target` in `dbus.socket`.

This is a follow-up on https://github.com/dracut-ng/dracut-ng/pull/2178 after trying to fix in Ubuntu.

Fixes: https://github.com/dracut-ng/dracut-ng/issues/1815
Bug-Ubuntu: https://launchpad.net/bugs/2141603

### Checklist

### DIFF
--- a/.github/workflows/daily-network.yml
+++ b/.github/workflows/daily-network.yml
@@ -48,12 +48,6 @@ jobs:
                       architecture: {runner: 'ubuntu-24.04-arm', tag: 'arm'}
                     - container: fedora:rawhide
                       architecture: {runner: 'ubuntu-24.04-arm', tag: 'arm'}
-                    # https://github.com/dracut-ng/dracut-ng/issues/1815
-                    - container: ubuntu:devel
-                      test: "72"
-                    # https://github.com/dracut-ng/dracut-ng/issues/1815
-                    - container: ubuntu:rolling
-                      test: "72"
                     # https://github.com/dracut-ng/dracut-ng/issues/1988
                     - container: debian:sid
                       architecture: {runner: 'ubuntu-24.04-arm', tag: 'arm'}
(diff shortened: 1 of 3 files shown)


bdrung:
Ubuntu has a patch that sets `DefaultDependencies=no`. So that is Ubuntu specific.

Conan-Kudo (approved):

PR #2302, opened 2026-03-13
feat(dracut): add module to load Qualcomm ADSP module pre-udev

jwrdegoede:
On Qualcomm sc8280xp and x1e laptops the kernel reboots the ADSP with new firmware because the BIOS loads ADSP firmware with limited functionality without sound or battery charge/status reading support.

Unfortunately the ADSP also controls the TCPM (Type-C Port Manager) and rebooting the ADSP also resets the TCPM, causing any USB devices connected over Type-C ports to get disconnected as all devices on the USB bus are removed and re-enumerated.

This breaks booting from USB-drives as the drive gets disconnected and re-enumerated as a new block device, leaving any filesystems mounted before the ADSP reset without any backing device.

To workaround this add a module to load the ADSP driver from a pre-udev hook so that the USB re-enumeration happens before the rootfs is mounted.

### Checklist
- [x] I am providing new code for it

### DIFF
--- a/modules.d/70qcom-adsp/module-setup.sh
+++ b/modules.d/70qcom-adsp/module-setup.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# On Qualcomm sc8280xp and x1e laptops the kernel reboots the ADSP with new
+# firmware because the BIOS loads ADSP firmware with limited functionality
+# without sound or battery charge/status reading support.
+#
+# Unfortunately the ADSP also controls the TCPM (Type-C Port Manager) and
+# rebooting the ADSP also resets the TCPM, causing any USB devices connected
+# over Type-C ports to get disconnected as all devices on the USB bus are
+# removed and re-enumerated.
+#
+# This breaks booting from USB-drives as the drive gets disconnected and
+# re-enumerated as a new block device, leaving any filesystems mounted
+# before the ADSP reset without any backing device.
+#
+# To workaround this, this module load the ADSP driver from a pre-udev hook
+# so that the USB re-enumeration happens before the rootfs is mounted.
+
+# called by dracut
+check() {
+    [ "$DRACUT_ARCH" = "aarch64" ] || return 1
+    if [[ $hostonly ]]; then
+        grep -q -E 'qcom,sc8280xp-adsp-pas|qcom,x1e80100-adsp-pas' \
+            /sys/bus/platform/devices/*.remoteproc/modalias 2> /dev/null || return 1
+    fi
+    return 0
+}
+
+# called by dracut
+installkernel() {
+    instmods qcom_q6v5_pas
+}
+
+# called by dracut
+install() {
+    inst_hook pre-udev 30 "$moddir/qcom-adsp-pre-udev.sh"
+}
--- a/modules.d/70qcom-adsp/module-setup.sh
+++ b/modules.d/70qcom-adsp/module-setup.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+# -*- mode: shell-script; indent-tabs-mode: nil; sh-basic-offset: 4; -*-
+# ex: ts=8 sw=4 sts=4 et filetype=sh
+
+# On Qualcomm sc8280xp and x1e laptops the kernel reboots the ADSP with new
+# firmware because the BIOS loads ADSP firmware with limited functionality
+# without sound or battery charge/status reading support.
+#
+# Unfortunately the ADSP also controls the TCPM (Type-C Port Manager) and
+# rebooting the ADSP also resets the TCPM, causing any USB devices connected
+# over Type-C ports to get disconnected as all devices on the USB bus are
+# removed and re-enumerated.
+#
+# This breaks booting from USB-drives as the drive gets disconnected and
+# re-enumerated as a new block device, leaving any filesystems mounted
+# before the ADSP reset without any backing device.
+#
+# To workaround this, this module load the ADSP driver from a pre-udev hook
+# so that the USB re-enumeration happens before the rootfs is mounted.
+
+# called by dracut
+check() {
+    [ "$DRACUT_ARCH" = "aarch64" ] || return 1
+    if [[ $hostonly_mode == "strict" ]]; then
+        grep -q -E 'qcom,sc8280xp-adsp-pas|qcom,x1e80100-adsp-pas' \
+            /sys/bus/platform/devices/*.remoteproc/modalias 2> /dev/null || return 1
+    fi
+    return 0
+}
+
+# called by dracut
+installkernel() {
+    instmods qcom_q6v5_pas
+}
+
+# called by dracut
+install() {
+    inst_multiple grep modprobe
+    inst_hook pre-udev 30 "$moddir/qcom-adsp-pre-udev.sh"
+}
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -379,6 +379,10 @@ qemu-net:
     - changed-files:
           - any-glob-to-any-file: 'modules.d/[0-9][0-9]qemu-net/*'
 
+qcom-adsp:
+    - changed-files:
+          - any-glob-to-any-file: 'modules.d/[0-9][0-9]qcom-adsp/*'
+
 systemd-cryptsetup:
     - changed-files:
           - any-glob-to-any-file: 'modules.d/[0-9][0-9]systemd-cryptsetup/*'
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -372,6 +372,10 @@ ppcmac:
     - changed-files:
           - any-glob-to-any-file: 'modules.d/[0-9][0-9]ppcmac/*'
 
+qcom-adsp:
+    - changed-files:
+          - any-glob-to-any-file: 'modules.d/[0-9][0-9]qcom-adsp/*'
+
 qemu:
     - changed-files:
           - any-glob-to-any-file: 'modules.d/[0-9][0-9]qemu/*'
--- a/doc_site/modules/ROOT/pages/modules/core.adoc
+++ b/doc_site/modules/ROOT/pages/modules/core.adoc
@@ -201,6 +201,10 @@ code, there are no specific types or categories for dracut modules.
 | polls CD-ROM
 |
 
+| qcom-adsp
+| Load Qualcomm ADSP module pre-udev
+|
+
 | rescue
 | utilities for rescue mode (such as ping, ssh, vi, fsck.*)
 | utils
--- a/doc_site/modules/ROOT/pages/modules/core.adoc
+++ b/doc_site/modules/ROOT/pages/modules/core.adoc
@@ -193,6 +193,10 @@ code, there are no specific types or categories for dracut modules.
 | polls CD-ROM
 |
 
+| qcom-adsp
+| Load Qualcomm ADSP module pre-udev
+|
+
 | rescue
 | utilities for rescue mode (such as ping, ssh, vi, fsck.*)
 | utils
--- a/modules.d/70qcom-adsp/qcom-adsp-pre-udev.sh
+++ b/modules.d/70qcom-adsp/qcom-adsp-pre-udev.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+# Ensure ADSP module is loaded and Type-C USB busses are reset before
+# filesystems are mounted
+
+if grep -q -E 'qcom,sc8280xp-adsp-pas|qcom,x1e80100-adsp-pas' \
+    /sys/bus/platform/devices/*.remoteproc/modalias 2> /dev/null; then
+    modprobe qcom_q6v5_pas
+fi


jwrdegoede:
> The new shell script (qcom-adsp-pre-udev.sh) could be removed and replaced with rd.driver.pre 

The shell script checks if the hw is there to avoid needlessly loading the module on systems without the hw. This is necessary on generic (non-hostonly) builds such as the Fedora live ISOs initramfs.

The hostonly mode code could indeed instead use `rd.driver.pre`, but I would rather keep the 2 modes in sync and use a pre-udev hook in both modes.

jwrdegoede:
Just pushed a new (and rebased) version.

Changes from initial PR version:

- Directly use $DRACUT_ARCH instead of defining _arch local var
- Add the new module to labeler.yml and core.adoc

jozzsi:
> The new shell script (qcom-adsp-pre-udev.sh) could be removed and replaced with rd.driver.pre 

What do you think of this suggestion ?

bdrung:
Can't this handled by the kernel? I would prefer to only change dracut to include the needed module and rely on the kernel/udev to load the module at the correct time.

jwrdegoede:
> > The new shell script (qcom-adsp-pre-udev.sh) could be removed and replaced with rd.driver.pre
> 
> What do you think of this suggestion ?

I've already responded to that suggestion in: https://github.com/dracut-ng/dracut-ng/pull/2302#issuecomment-4225862249

jwrdegoede:
> Can't this handled by the kernel? I would prefer to only change dracut to include the needed module and rely on the kernel/udev to load the module at the correct time.

The USB bus running over the Type-C connectors looks like just another USB bus to the kernel XHCI driver.

The problem is that the XHCI controller sees a disconnect of all USB devices because of the Type-C mux (which switches pins between USB / DP / thunderbolt / xxx functions) getting put in high-z mode for all muxed pins while the TCPM is rebooting and booting the ADSP with the correct firmware resets/reboots the TCPM.

Basically from the point of the kernels USB subsystem it looks like someone yanks out the Type-C cable and then plugs it back in a second or so later when the ADSP firmware boots. And we must load the ADSP firmware to be able to have audio and battery monitoring.

Note that electrically speaking the cable physically actually does get disconnected by the mux going in high-Z mode.

Hence this workaround to make sure that this software-triggered USB disconnect happens before dracut tries to use any USB-storage devices.

This actually also stops the laptop from charging for that 1 second. Which means these laptops also cannot be booted on just a charger without a battery being connected. At least not without disabling the ADSP driver.

I've been told that the next generation of Qualcomm Snapdragon WoA laptops will not have this behavior where the TCPM is reset on booting the ADSP, which is why the workaround here is restricted to just the 2 current generations known to have this issue.

jwrdegoede:
@bdrung @jozzsi first of all thank you for all your help with getting #2266 reviewed.

I was wondering if you can also help with reviewing this PR? Note that I've already replied to / answered your previous comments here.

modules.d/70qcom-adsp/module-setup.sh
```diff
@@ -0,0 +1,39 @@
...
+}
+
+# called by dracut
+installkernel() {
+    instmods qcom_q6v5_pas
+}
+
+# called by dracut
+install() {
+    inst_hook pre-udev 30 "$moddir/qcom-adsp-pre-udev.sh"
```
aafeijoo-suse, line 39: This hook calls `grep`, so please install it via `inst grep` here.

jwrdegoede: Thank you for the review. I've fixed this and I'm pushing a new version adding a "inst grep" line to install() now.

This is rebased on the latest main, so it also fixes the reported conflicts in the previous version.

jwrdegoede:
Just pushed a new (and rebased) version adding a "inst grep" line to install() as requested during review.

aafeijoo-suse:
> Just pushed a new (and rebased) version adding a "inst grep" line to install() as requested during review.

Thanks. If I'm not mistaken, this PR could also fix this downstream bug, right? https://bugzilla.suse.com/show_bug.cgi?id=1261649#c5 (ThinkPad X13s has Qualcomm SC8280X)

jwrdegoede:
> Thanks. If I'm not mistaken, this PR could also fix this downstream bug, right? https://bugzilla.suse.com/show_bug.cgi?id=1261649#c5 (ThinkPad X13s has Qualcomm SC8280X)

That one should already be fixed by: #2266

The problem this one tries to fix is related to booting from USB storage on a Type-C port. Loading the ADSP remoteproc driver resets the TCPM (which is controlled by the ADSP firmware) which drops Vbus on the Type-C port, causing any USB storage to drop of the bus and then get re-enumerated when Vbus gets turned back on.

Having the USB storage disappear when booting from it is a problem. This is why, e.g.: https://fedoraproject.org/wiki/Snapdragon_WoA_Laptop_Install advices adding ` modprobe.blacklist=qcom_q6v5_pas` to the kernel commandline to boot a liveusb to install. Note in this case qcom_q6v5_as must *not* be modprobe-d later. The workaround is to simply never load it and live with the missing sound and battery status info while doing the live install.

This PR fixes this by loading qcom_q6v5_pas and thus doing the TCPM reset before looking for storage devices and thus before mounting any storage from the USB drive. Which avoids the need to blacklist the driver.

aafeijoo-suse:
> > Thanks. If I'm not mistaken, this PR could also fix this downstream bug, right? https://bugzilla.suse.com/show_bug.cgi?id=1261649#c5 (ThinkPad X13s has Qualcomm SC8280X)
> 
> That one should already be fixed by: #2266
> 
> The problem this one tries to fix is related to booting from USB storage on a Type-C port. Loading the ADSP remoteproc driver resets the TCPM (which is controlled by the ADSP firmware) which drops Vbus on the Type-C port, causing any USB storage to drop of the bus and then get re-enumerated when Vbus gets turned back on.
> 
> Having the USB storage disappear when booting from it is a problem. This is why, e.g.: https://fedoraproject.org/wiki/Snapdragon_WoA_Laptop_Install advices adding ` modprobe.blacklist=qcom_q6v5_pas` to the kernel commandline to boot a liveusb to install. Note in this case qcom_q6v5_as must _not_ be modprobe-d later. The workaround is to simply never load it and live with the missing sound and battery status info while doing the live install.
> 
> This PR fixes this by loading qcom_q6v5_pas and thus doing the TCPM reset before looking for storage devices and thus before mounting any storage from the USB drive. Which avoids the need to blacklist the driver.

Thanks for your feedback,  and even for providing such helpful information in our bug report.

modules.d/70qcom-adsp/module-setup.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+# called by dracut
+installkernel() {
+    instmods qcom_q6v5_pas
+}
+
+# called by dracut
+install() {
+    inst grep
```
bdrung: You also use `modprobe`.

modules.d/70qcom-adsp/module-setup.sh
```diff
@@ -0,0 +1,40 @@
...
+    inst_hook pre-udev 30 "$moddir/qcom-adsp-pre-udev.sh"
```
bdrung, line 39: Have you considered using a systemd unit in case systemd is used?

modules.d/70qcom-adsp/module-setup.sh
```diff
@@ -0,0 +1,40 @@
...
+# re-enumerated as a new block device, leaving any filesystems mounted
+# before the ADSP reset without any backing device.
+#
+# To workaround this, this module load the ADSP driver from a pre-udev hook
+# so that the USB re-enumeration happens before the rootfs is mounted.
+
+# called by dracut
+check() {
+    [ "$DRACUT_ARCH" = "aarch64" ] || return 1
+    if [[ $hostonly ]]; then
```
bdrung: Should we relax it to only do this check in hostonly strict mode?

bdrung (commented):
The PR looks good overall, but I do not like having a separate module just for this fix. I looked where this code could live, but I found no good place (doesn't really fit into the systemd-udevd module). Do we have other places where we do something similar?

modules.d/70qcom-adsp/module-setup.sh
jwrdegoede: Ack, will fix in a new version.

jwrdegoede:
Thank you for your review.

> The PR looks good overall, but I do not like having a separate module just for this fix. I looked where this code could live, but I found no good place (doesn't really fit into the systemd-udevd module). Do we have other places where we do something similar?

This could maybe be moved to the "70devicetree-firmware" module which was mainly written to deal with loading the device-model specific firmwares for the various DSP co-procs on Qualcomm platforms. That would make that module somewhat less generic though.

modules.d/70qcom-adsp/module-setup.sh
jwrdegoede: No I've not considered that. This seems to work fine with systemd though. I've tested this with a locally generated Fedora liveusb image, and AFAIK Fedora uses systemd in the initramfs.

So I believe doing things this way works for both the with and without systemd cases?

modules.d/70qcom-adsp/module-setup.sh
jwrdegoede: I believe non strict hostonly mode is meant to allow moving a drive from one machine to another similar (but not necessary the same model) machine and hopefully still have a working initramfs, right?

This PR is only necessary when booting from USB storage connected to one of the Type-C ports on Qualcomm Windows on ARM laptops. So for moving the drive from one laptop to another this does not matter. OTOH for removing the drive and then trying to boot it from an USB enclosure this does matter.

I'm fine handling this either way, please let me know what you think is best.

jwrdegoede:
I've replied to all review remarks. I'll prepare a new version once the strict hostonly mode and the move to the "70devicetree-firmware" module questions are settled.

modules.d/70qcom-adsp/module-setup.sh
bdrung: > I believe non strict hostonly mode is meant to allow moving a drive from one machine to another similar (but not necessary the same model) machine and hopefully still have a working initramfs, right?

Yes.

> OTOH for removing the drive and then trying to boot it from an USB enclosure this does matter.

This is a case we should support for the sloppy hostonly mode IMO. At least I have done such things in the past.

bdrung:
> Thank you for your review.
> 
> > The PR looks good overall, but I do not like having a separate module just for this fix. I looked where this code could live, but I found no good place (doesn't really fit into the systemd-udevd module). Do we have other places where we do something similar?
> 
> This could maybe be moved to the "70devicetree-firmware" module which was mainly written to deal with loading the device-model specific firmwares for the various DSP co-procs on Qualcomm platforms. That would make that module somewhat less generic though.

 "70devicetree-firmware" has "devicetree" and "firmware" in the name. Loading this module is unrelated to devicetree and unrelated to firmware IMO. So it does not really fit in there, does it?

@devkontrol @Conan-Kudo what is your opinion on this topic? Do you like having this small additional module or should this code live in an existing module?

devkontrol:
> Do you like having this small additional module or should this code live in an existing module?

I do prefer separate modules when possible - including here. Dracut's strength is modularity and modularity suffers when we do not separate functionality into separate modules when it is meaningful.

I certainly hope distributions will just package all upstream modules, but often some distributions decide to not support some upstream dracut modules (e.g. network-legacy).

This module is only for arm - so a distribution not supporting arm could for example decide not to package this module. This is a benefit that we should consider.

A separate module also greatly improves debugging both for us and end-users. It is very easy to tell if a module is included in the initramfs or not, but it is much harder to tell if a particular condition is met buried within a dracut module code. 

There is no significant overhead during initramfs creation that I am aware of for breaking out dracut modules into more submodules.

modules.d/70qcom-adsp/module-setup.sh
jwrdegoede: Ok, I've switch the check to `[[ $hostonly_mode == "strict" ]]` now.

jwrdegoede:
Based on @devkontrol 's comment I'm keeping this as a separate module.

devkontrol (dismissed):

jwrdegoede:
@devkontrol thank you for the approval, but I had not yet pushed the discussed changes. Can you re-approve?

The version which I just pushed has the following changes:

- Change host-only check to `[[ $hostonly_mode == "strict" ]]` so that `30-qcom-adsp-pre-udev.sh` gets installed in sloppy hostonly mode
- Add modprobe to list of binaries to install from install()

devkontrol (dismissed):

modules.d/70qcom-adsp/module-setup.sh
```diff
@@ -0,0 +1,40 @@
...
+}
+
+# called by dracut
+installkernel() {
+    instmods qcom_q6v5_pas
+}
+
+# called by dracut
+install() {
+    inst_multiple modprobe grep
```
bdrung: Unless there is a reason, the list could be sorted:
```suggestion
    inst_multiple grep modprobe
```

jwrdegoede:
I've used github's apply suggestion feature to fix @bdrung's review remark about sorting the `inst_multiple` args. This is now a separate commit for easy reviewing of just the changes.

AFAIK you can squash this while merging? Let me know if you prefer me to do a forced push with the fix squashed.

jwrdegoede:
I see this also causes the "[Commisery / Conventional Commit Message Checker (Commisery) (pull_request)](https://github.com/dracut-ng/dracut/actions/runs/30103540917/job/89515286807?pr=2302)" CI check to fail.

Doing a forced push with things squashed to fix this now.

devkontrol (approved):

bdrung (approved):
Yes, squashing the commit is the way to go.

devkontrol:
https://bugzilla.redhat.com/show_bug.cgi?id=2487066

jwrdegoede:
> https://bugzilla.redhat.com/show_bug.cgi?id=2487066

Thank you for letting me know about this, this should be fixed by #2598, review of #2598 would be appreciated.

PR #2308, opened 2026-03-19
fix(crypt): use systemd generator for setting the timeout

bdrung:
Test 72 "NBD root=LABEL=dracut netroot=nbd:IP:port" can fail on slow architectures.

```
603s [   36.841605] dracut-cmdline[178]: + set_systemd_timeout_for_dev /dev/nbd0
[...]
603s [   37.015613] dracut-cmdline[178]: + mkdir -p /etc/systemd/system/dev-nbd0.device.d
603s [   37.020616] dracut-cmdline[178]: + echo [Unit]
603s [   37.025872] dracut-cmdline[178]: + echo JobTimeoutSec=infinity
603s [   37.031417] dracut-cmdline[178]: + echo JobRunningTimeoutSec=infinity
603s [   37.036838] dracut-cmdline[178]: + type mark_hostonly
603s [   37.043508] dracut-cmdline[178]: + _needreload=1
603s [   37.047966] dracut-cmdline[178]: + [ -z  ]
603s [   37.055568] dracut-cmdline[178]: + [ 1 = 1 ]
603s [   37.060790] dracut-cmdline[178]: + [ -z  ]
603s [   37.065785] dracut-cmdline[178]: + /sbin/initqueue --onetime --unique --name daemon-reload systemctl daemon-reload
[...]
[ TIME ] Timed out waiting for device dev-nbd0.device - /dev/nbd0.
679s [DEPEND] Dependency failed for systemd-cryp…0e2d54-ab3c-44b1-b289-7e1921480385.
679s [DEPEND] Dependency failed for cryptsetup.target - Local Encrypted Volumes.
679s [  113.311607] systemd[1]: dev-nbd0.device: Job dev-nbd0.device/start timed out.
679s [  113.332671] systemd[1]: Timed out waiting for device dev-nbd0.device - /dev/nbd0.
[...]
751s          Starting dracut-initqueue.service - dracut initqueue hook...
[...]
822s [  255.665608] dracut-initqueue[495]: + /usr/bin/systemctl daemon-reload
[...]
835s [  269.333700] systemd[1]: Reload requested from client PID 593 ('systemctl') (unit dracut-initqueue.service)...
835s [  269.378921] systemd[1]: Reloading...
894s [  328.119996] systemd[1]: Reloading finished in 67715 ms.
```

[...]

PR #2310, opened 2026-03-19
fix(base): move initrd.target.wants symlink creation to wait_for_dev

bdrung:
Commit 4d9d767da2e5 ("crypt/parse-crypt.sh: hide encrypted devices from systemd timeout warnings.") split parts of `wait_for_dev` into `set_systemd_timeout_for_dev` to allow setting the systemd timeout separately. `parse-crypt.sh` is the only direct caller of `set_systemd_timeout_for_dev`.

Fixes: 4d9d767da2e5 ("crypt/parse-crypt.sh: hide encrypted devices from systemd timeout warnings.")

### Checklist

### DIFF
--- a/modules.d/80base/dracut-dev-lib.sh
+++ b/modules.d/80base/dracut-dev-lib.sh
@@ -75,12 +75,6 @@ set_systemd_timeout_for_dev() {
     _timeout=${_timeout:-infinity}
 
     _name=$(dev_unit_name "$1")
-    if ! [ -L "${PREFIX-}/etc/systemd/system/initrd.target.wants/${_name}.device" ]; then
-        [ -d "${PREFIX-}"/etc/systemd/system/initrd.target.wants ] || mkdir -p "${PREFIX-}"/etc/systemd/system/initrd.target.wants
-        ln -s ../"${_name}".device "${PREFIX-}/etc/systemd/system/initrd.target.wants/${_name}.device"
-        type mark_hostonly > /dev/null 2>&1 && mark_hostonly /etc/systemd/system/initrd.target.wants/"${_name}".device
-        _needreload=1
-    fi
 
     if ! [ -f "${PREFIX-}/etc/systemd/system/${_name}.device.d/timeout.conf" ]; then
         mkdir -p "${PREFIX-}/etc/systemd/system/${_name}.device.d"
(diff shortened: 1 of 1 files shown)


Conan-Kudo (dismissed):

modules.d/80base/dracut-dev-lib.sh
```diff
@@ -125,6 +119,15 @@ wait_for_dev() {
         printf 'warn "\"%s\" does not exist"\n' "$1"
     } >> "${PREFIX-}$hookdir/emergency/80-${_name}.sh"
 
+    [ -n "${DRACUT_SYSTEMD-}" ] || return 0
+
+    if ! [ -L "${PREFIX-}/etc/systemd/system/initrd.target.wants/${_name}.device" ]; then
+        [ -d "${PREFIX-}"/etc/systemd/system/initrd.target.wants ] || mkdir -p "${PREFIX-}"/etc/systemd/system/initrd.target.wants
+        ln -s ../"${_name}".device "${PREFIX-}/etc/systemd/system/initrd.target.wants/${_name}.device"
```
aerusso: `${_name}` means something different in this context.  In particular: `_name="$(str_replace "$1" '/' '\x2f')"` here, but `_name=$(dev_unit_name "$1")` above.

This also affects the Debian version you uploaded yesterday.

Obviously, you can just, e.g., set `_sdname=$(dev_unit_name "$1")` and use that instead.  I'm current running with that fix.  (In my case, the bug appears as a 90 second delay during boot, since that `.device` unit waits for a file that will never show up.)

bdrung: Good catch. I fixed this and reworked the code a bit.

Once we fixed https://github.com/dracut-ng/dracut-ng/pull/2308 I want to get rid of `set_systemd_timeout_for_dev` to avoid calling `dev_unit_name` twice.

aerusso (approved):
I booted this without issue on the Debian machine, so my review is that this works (i.e., backported to 110, as it is packaged in Debian).  I will be the first to admit, however, that I do not understand what this is trying to accomplish, or why we would ever need to wait for a device after having root mounted (i.e., can't the post-pivot real init system perform the wait)?  But, this is of course due to my ignorance of the broader issues at hand.

In any case, unlike the previous version, I have no complaints!

devkontrol (approved):

Conan-Kudo (approved):

PR #2316, opened 2026-03-24
fix(zipl): Fix parsing rd.zipl=LABEL|UUID|...=

Vogtinator:
label_uuid_to_dev needs the LABEL=|UUID=|... prefix as well to determine the full path.

Without this fix it hangs waiting for just the UUID/LABEL/... without the /dev path, breaking boot, e.g.:

Warning: 200c78bc-7099-3c44-f3b5-6adf477ffd15 does not exist

Downstream bug: https://bugzilla.opensuse.org/show_bug.cgi?id=1259587

### Checklist

### DIFF
--- a/modules.d/73zipl/parse-zipl.sh
+++ b/modules.d/73zipl/parse-zipl.sh
@@ -9,22 +9,22 @@ if [ -n "$zipl_arg" ]; then
         LABEL=*)
             zipl_env="ENV{ID_FS_LABEL}"
             zipl_val=${zipl_arg#LABEL=}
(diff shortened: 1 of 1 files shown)


bdrung (approved):

aafeijoo-suse (approved):
Argh... we had already fixed this in our old fork (https://github.com/openSUSE/dracut/commit/1b86da026b8e60275a59b1453b15329aa09c4373), but the fix was not sent upstream, so we lost it after the migration.

devkontrol (approved):

PR #2323, opened 2026-03-27
test(FULL-SYSTEMD): increase device timeout to infinity

bdrung:
Test 41 fails to boot on arm64 on Debian in the autopkgtest because the service waiting for `/dev/disk/by-label/dracutusr` times out after 90 seconds. This is just caused by the arm64 runner being slow.

So increase the device timeout to infinity to make the test succeed on slow systems.

See also: d62de66d35d ("test(SYSTEMD-INITRD): increase device timeout to infinity")

### Checklist


Conan-Kudo (approved):
Not a fan, but ehh...

devkontrol (approved):

PR #2326, opened 2026-03-28
refactor(base): merge set_systemd_timeout_for_dev into wait_for_dev

bdrung:
The function `set_systemd_timeout_for_dev` is only used by `wait_for_dev`. Merge this function back into `wait_for_dev` to avoid calling `dev_unit_name` twice.

### Checklist


devkontrol (approved):

Conan-Kudo (approved):
We can always bring it back if we need it. Let's land this.

PR #2329, opened 2026-03-30
ci: switch from ubuntu:devel to a more stable debian:latest

jozzsi:
Switch some of the daily runners over to a mores table container to avoid running into non-dracut related issues and regressions.

### Checklist


devkontrol (approved):
Sad to see this change, but confidence in dracut CI should take priority over distribution-specific fluctuations.

PR #2331, opened 2026-03-30
ci: increase timeout for tests

jozzsi:
Tests seems to need more time to pass.

Give 30 min timeout for non-networking tests and 40 min for networked tests.

### Checklist


devkontrol (approved):
nice

PR #2332, opened 2026-03-30
ci: do not force Void CI continer to test zfs

jozzsi:
Running some tests with zfs enabled fails on Void. Disable testing with zfs until the issue is resolved.

### Checklist


devkontrol (approved):
ok, but let's plan on reintroducing test coverage somehow for zfs.

PR #2341, opened 2026-04-01
perf: export DRACUT_ARCH globally

aafeijoo-suse:
Usually `DRACUT_ARCH` is not set, so we are calling `uname -m` multiple times to get always the same value. Hence, get its external value at the beginning and export it globally, in the same way we do with other global variables.

### Checklist

### DIFF
--- a/man/dracut.modules.7.adoc
+++ b/man/dracut.modules.7.adoc
@@ -177,6 +177,7 @@ drauct modules. These shell variables are always declared and assigned
 values (the assigned value might be an empty string). There is a
 compatibility promise on these variables between dracut releases.
 Module should only read (and not write) these variables.
+ * `$DRACUT_ARCH` (not an empty string)
  * `$hostonly` (could be an empty string)
  * `$hostonly_mode` (could be an empty string)
  * `$hostonly_cmdline` (not an empty string)
--- a/dracut.sh
+++ b/dracut.sh
@@ -970,6 +970,8 @@ export DRACUT_LOG_LEVEL=warning
 
 [[ ${dracutbasedir-} ]] || dracutbasedir="${dracutsysrootdir-}"/usr/lib/dracut
 
+export DRACUT_ARCH=${DRACUT_ARCH:-$(uname -m)}
+
 # These config variables needs to be exported for dracut-install.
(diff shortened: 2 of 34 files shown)


man/dracut.modules.7.adoc
```diff
@@ -178,10 +178,12 @@ value might be an empty string). There is a compatibility promise on these
 variables between dracut releases.
 Modules should only read (and not write) these variables.
  * `$DRACUT_ARCH` (not an empty string)
+ * `$DRACUT_VERSION` (not an empty string)
```
bdrung: I would not document `DRACUT_VERSION` here, because modules should not rely on it.

man/dracut.modules.7.adoc
```diff
@@ -178,10 +178,12 @@ value might be an empty string). There is a compatibility promise on these
...
  * `$hostonly` (could be an empty string)
  * `$hostonly_mode` (could be an empty string)
  * `$hostonly_cmdline` (not an empty string)
  * `$moddir` (not an empty string)
+ * `$SYSTEMCTL` (not an empty string)
```
bdrung: I would not document `SYSTEMCTL` here. Instead we should add helpers functions to call `$SYSTEMCTL -q --root "$initdir" enable` or `$SYSTEMCTL -q --root "$initdir" add-wants`.

man/dracut.modules.7.adoc
```diff
@@ -172,11 +172,11 @@ after udev has settled and the basic device drivers have been loaded.
 
 All module installation information is in the file module-setup.sh.
 
-The following are the most commonly used shell variables used inside
-drauct modules. These shell variables are always declared and assigned
-values (the assigned value might be an empty string). There is a
-compatibility promise on these variables between dracut releases.
-Module should only read (and not write) these variables.
+The following are the most commonly used shell variables within dracut modules.
+These shell variables are always declared and assigned values (the assigned
```
bdrung: Is that good English: "These shell variables are always assigned values"? This sounds strange to me.

bdrung (approved):

devkontrol (approved):

PR #2344, opened 2026-04-01
fix(crypt): standardise password retry behaviour with rd.luks.tries

nadzyah:
Without an explicit tries= in the crypttab entry, systemd-cryptsetup uses its default of 3 attempts and then gives up, causing dracut-initqueue to hang silently before dropping to an emergency shell.

Set tries=0 across all paths that generate crypttab entries in the initrd. Entries that already specify tries= are left unchanged.

Add a test to TEST-41-FULL-SYSTEMD that sends three wrong passwords, then the correct one, verifying the system unlocks successfully.

The bug was originally reported on Launchpad: https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/2146686

### Checklist
- [x] I have reviewed and updated any documentation if relevant - not needed
- [x] I am providing new code and test(s) for it - a bugfix with a test


modules.d/70crypt/crypt-run-generator.sh
```diff
@@ -22,7 +22,7 @@ if [ -n "${DRACUT_SYSTEMD-}" ] || strstr "$(cryptsetup --help)" "allow-discards"
     fi
 fi
 
-echo "$luks $dev - timeout=0,$allowdiscards" >> /etc/crypttab
+echo "$luks $dev - timeout=0,tries=0,$allowdiscards" >> /etc/crypttab
```
bdrung: Should we make it configurable how many times we retry?

nadzyah: I was thinking that the whole point of the fix is to match the non-systemd path and let the user decide when to give up, which they already can via `rd.luks.timeout=` rather than a fixed attempt count

Conan-Kudo: I think I'd be more comfortable if we changed it so there's a fixed default set of tries regardless of systemd/non-systemd path.

modules.d/70crypt-lib/crypt-lib.sh
```diff
@@ -107,9 +107,16 @@ ask_for_password() {
         flock -s 9
         # Prompt for password with plymouth, if installed and running.
         if type plymouth > /dev/null 2>&1 && plymouth --ping 2> /dev/null; then
-            plymouth ask-for-password \
-                --prompt "$ply_prompt" --number-of-tries="$ply_tries" \
-                --command="$ply_cmd"
+            if [ "$ply_tries" -eq 0 ]; then
```
bdrung: plymouth.c says:
```c
        ply_command_parser_get_command_options (state->command_parser,
                                                command,
                                                "command", &program,
                                                "prompt", &prompt,
                                                "number-of-tries", &number_of_tries,
                                                "dont-pause-progress", &dont_pause,
                                                NULL);
        if (number_of_tries <= 0)
                number_of_tries = INT_MAX;
```

modules.d/70crypt-lib/crypt-lib.sh
```diff
@@ -118,9 +125,15 @@ ask_for_password() {
             fi
 
             local i=1
-            while [ $i -le "$tty_tries" ]; do
-                [ -n "$tty_prompt" ] \
-                    && printf "%s" "$tty_prompt [$i/$tty_tries]:" >&2
+            # 0 = unlimited
+            while [ "$tty_tries" -eq 0 ] || [ $i -le "$tty_tries" ]; do
+                [ -n "$tty_prompt" ] && {
```
bdrung: Can we use a normal `if` condition here?

modules.d/70crypt-lib/crypt-lib.sh
```diff
@@ -144,17 +157,19 @@ luks_open_interactive() {
...
     luks_open="$(command -v cryptsetup) $cryptsetupopts luksOpen"
     ask_for_password \
-        --ply-tries 5 \
+        --ply-tries "$_tries" \
         --ply-cmd "$luks_open -T1 $dev $mapname" \
         --ply-prompt "$prompt" \
-        --tty-tries 1 \
-        --tty-cmd "$luks_open -T5 -t $_timeout $dev $mapname"
+        --tty-tries "$_tries" \
+        --tty-cmd "$luks_open -T1 -t $_timeout $dev $mapname"
```
bdrung, line 193: Have you a reason to move the tries from the `tty-cmd` to `tty-tries`?

modules.d/70crypt-lib/crypt-lib.sh
```diff
@@ -107,8 +107,12 @@ ask_for_password() {
         flock -s 9
         # Prompt for password with plymouth, if installed and running.
         if type plymouth > /dev/null 2>&1 && plymouth --ping 2> /dev/null; then
+            # According to plymouth.c, number_of_tries <= 0 means
+            # number_of_tries = INT_MAX, so we can skip the check for omiting
+            # the parameter
```
bdrung: I am not sure how useful this comment is here. It might be better to have the "tries=0 means unlimited" in the description of the function instead.

modules.d/70crypt/crypt-generator.sh
```diff
@@ -25,12 +25,45 @@ if ! getargbool 1 rd.luks; then
...
+        _escaped=$(dev_unit_name "luks-${_luksid}")
+        _dropin="${GENERATOR_DIR}/systemd-cryptsetup@${_escaped}.service.d"
+        mkdir -p "$_dropin"
+        {
+            echo "# Automatically generated by ${0##*/}"
+            echo "[Service]"
+            echo "ExecStart="
+            printf 'ExecStart=%s attach luks-%s /dev/disk/by-uuid/%s - tries=%s\n' \
+                "$_cryptsetup_bin" "$_luksid" "$_luksid" "$_tries"
+        } > "$_dropin/tries.conf"
```
bdrung, line 62: Instead of fiddling with systemd-cryptsetup-generator, we could decide to unify with systemd-cryptsetup-generator.

Current systemd-cryptsetup-generator supports retries by setting `rd.luks.options=retries=0`.

bdrung (commented):
rd.luks.tries should be documented in the man page.

modules.d/70crypt-lib/crypt-lib.sh
nadzyah: I've moved the retries to the outer loop so that `tries=0` (unlimited) works, since `cryptsetup -T0` means zero attempts rather than unlimited. Would you like it to be defined somehow in the inner loop instead?

bdrung:
Created https://github.com/systemd/systemd/issues/41984 to address the systemd-cryptsetup default in the tool itself.

devkontrol:
Merge conflict, marking it draft for now.

Issue #1419, opened 2025-07-05
feature: copy or reflink cpio data directly from source instead of staging in tmpdir


bdrung:
When generating initramfs images, dracut currently recreates the filesystem tree in a staging area under `tmpdir`. This is mostly needed due to GNU cpio's unfortunate source-path-is-dest-path behaviour.

[3cpio](https://github.com/bdrung/3cpio) is a Rust implementation for cpio files for initrds. It just learned how to build cpios from a manifest file where the source and destination file paths can differ. See https://github.com/bdrung/3cpio/issues/3 for the code, manifest format, and benchmarks. Feedback there will be welcome.

Eliminating `tmpdir` staging for cpio data should provide a not-insignificant performance boost for both reflinked (via copy_file_range) and non-reflinked cpio archives.

This is a follow-up on https://github.com/dracutdevs/dracut/issues/1662.

bdrung:
@alpernebbi or @chewi touched `dracut-install` last. Are you interested in working on `dracut-install`?

chewi:
Yeah, I thought this was already solved, but I'm not familiar with the approach because my distro, Flatcar, cannot benefit from it.

bdrung:
Yes, the benefit of dracut-cpio is to benefit from reflinks (due to the alignment). This benefit only exist as long as the file stays on the same partition and is not compressed.

The benefit of 3cpio is faster creation time and memory safety due to being written in Rust.

bdrung:
I did a benchmark on my laptop to compare 3cpio and dracut-cpio and they have the same speed on a cold cache:
```
$ hyperfine -w 2 -r 100 -p "rm -f initrd.img && sync && echo 3 > /proc/sys/vm/drop_caches" "3cpio -c initrd.img -C initrd < files" "cd initrd && dracut-cpio < ../files ../initrd.img" "cd initrd && bsdcpio -o -H newc > ../initrd.img < ../files" "cd initrd && cpio -o -H newc --reproducible > ../initrd.img < ../files" --export-markdown create-cold.md
Benchmark 1: 3cpio -c initrd.img -C initrd < files
  Time (mean ± σ):      98.2 ms ±   5.3 ms    [User: 4.4 ms, System: 87.8 ms]
  Range (min … max):    90.8 ms … 109.0 ms    100 runs
 
Benchmark 2: cd initrd && dracut-cpio < ../files ../initrd.img
[...]

PR #2353, opened 2026-04-07
perf(dracut-functions): optimize inst_libdir_file()

aafeijoo-suse:
Let's start analyzing the output of a non-hostonly build, focusing on the `inst_libdir_file()` calls:

```
$ dracut -f -N --debug testn.img 2>&1 &> testn.txt
$ grep -r -c "(inst_libdir_file): inst_multiple" testn.txt
17
$ grep -r "(inst_libdir_file): inst_multiple" testn.txt | awk -F"inst_multiple " '{ print $2 }' | tr ' ' '\n' | sort | wc -l
295
$ grep -r "(inst_libdir_file): inst_multiple" testn.txt | awk -F"inst_multiple " '{ print $2 }' | tr ' ' '\n' | sort -u | wc -l
201
```

There are 94 files that are attempted to be installed multiple times. After adding a simple cache:

```
$ grep -r -c "(inst_libdir_file): inst_multiple" testn2.txt
13
$ grep -r "(inst_libdir_file): inst_multiple" testn2.txt | awk -F"inst_multiple " '{ print $2 }' | tr ' ' '\n' | sort | wc -l
201
```

### Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1336,6 +1336,13 @@ inst_libdir_dir() {
 # -n <pattern> install matching files
 inst_libdir_file() {
     local -a _files=()
+    local _path
+
+    # cache installed libdir files
+    if ! declare -p _libdir_file_cache 2> /dev/null | grep -q "declare -A"; then
+        declare -gxA _libdir_file_cache=()
+    fi
+
     if [[ $1 == "-n" ]]; then
         local _pattern=$2
         shift 2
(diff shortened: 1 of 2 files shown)


Conan-Kudo (approved):

bdrung:
I would prefer to aim for a bigger improvement: Instead of caching on individual file level, I would go one step further for https://github.com/dracut-ng/dracut-ng/issues/1419: Add staging functions similar to the inst_* functions. These functions would just collect what should be added. Then call an dracut-intsall equivalent to only write a manifest for 3cpio.

PR #2356, opened 2026-04-08
refactor: remove unneeded extra slash passed in some args to inst_libdir_file()

aafeijoo-suse:
### Checklist

### DIFF
--- a/modules.d/70crypt/module-setup.sh
+++ b/modules.d/70crypt/module-setup.sh
@@ -158,6 +158,6 @@ install() {
 
     # Install required libraries.
     inst_libdir_file \
-        {"tls/$DRACUT_ARCH/",tls/,"$DRACUT_ARCH/",}"/ossl-modules/fips.so" \
-        {"tls/$DRACUT_ARCH/",tls/,"$DRACUT_ARCH/",}"/ossl-modules/legacy.so"
+        {"tls/$DRACUT_ARCH/",tls/,"$DRACUT_ARCH/",}"ossl-modules/fips.so" \
+        {"tls/$DRACUT_ARCH/",tls/,"$DRACUT_ARCH/",}"ossl-modules/legacy.so"
 }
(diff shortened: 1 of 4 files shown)


modules.d/70crypt/module-setup.sh
```diff
@@ -158,6 +158,6 @@ install() {
 
     # Install required libraries.
     inst_libdir_file \
-        {"tls/$DRACUT_ARCH/",tls/,"$DRACUT_ARCH/",}"/ossl-modules/fips.so" \
-        {"tls/$DRACUT_ARCH/",tls/,"$DRACUT_ARCH/",}"/ossl-modules/legacy.so"
+        {"tls/$DRACUT_ARCH/",tls/,"$DRACUT_ARCH/",}"ossl-modules/fips.so" \
+        {"tls/$DRACUT_ARCH/",tls/,"$DRACUT_ARCH/",}"ossl-modules/legacy.so"
```
bdrung, line 162: Good catch. To shorten the string, we could specify `/` only once:
```suggestion
        {"tls/$DRACUT_ARCH",tls,"$DRACUT_ARCH",}"/ossl-modules/fips.so" \
        {"tls/$DRACUT_ARCH",tls,"$DRACUT_ARCH",}"/ossl-modules/legacy.so"
```
What do you think?

aafeijoo-suse: You're missing the default case:

```
{"tls/$DRACUT_ARCH",tls,"$DRACUT_ARCH",}"/ossl-modules/legacy.so"
                                      ^
```

The function prepends the `/`, so in the default case, we would have `//`:

https://github.com/dracut-ng/dracut-ng/blob/e4a2555e61d5041747357e7350a5edcea2acd407/dracut-functions.sh#L1353

Besides that, with your suggestion, we would need to prepend a `/` to _every_ argument in _every_ function call, tree-wide...

bdrung: Oh, you are right. I missed the default case. In this case my suggestions wouldn't work.

Maybe we could improve the helper in the future to specify the additional paths to look for.

bdrung (approved):
Looks good. Found no more such cases.

devkontrol (approved):

PR #2363, opened 2026-04-09
ci: disable running failing tests for daily network GA

jozzsi:
`ubuntu:devel` and `ubuntu:rolling` needs over an hour to run test 72 and it times out. Other Linux distributions finish the test in less than 30 min.
    

### Checklist


devkontrol (approved):
ok, I prefer to get back a green CI baseline for daily GA as well.

PR #2364, opened 2026-04-09
fix(systemd-ask-password): no graphical output in aarch64

jozzsi:
Depend on the drm module on aarch64 in non-hostonly (generic) mode.

Follow-up to 4cc962a .

Fixes: https://github.com/dracut-ng/dracut-ng/issues/2362

CC @bdrung @Rolv-Apneseth @cgwalters @aafeijoo-suse  @travier 

### Checklist

### DIFF
--- a/modules.d/11systemd-ask-password/module-setup.sh
+++ b/modules.d/11systemd-ask-password/module-setup.sh
@@ -19,15 +19,15 @@ check() {
 # Module dependency requirements.
 depends() {
 
-    if [[ $hostonly ]]; then
+    if [[ ${DRACUT_ARCH} == arm* || ${DRACUT_ARCH} == aarch64 ]]; then
         # A password cannot be entered if there is no graphical output during boot,
         # as is the case in aarch64, where efifb does not work with qemu-system-aarch64:
         # - virtio-gpu-pci does not expose a linear framebuffer
         # - virtio-vga is not supported
         # - ramfb is not enough
         # Therefore, depend on the drm module if virtio_gpu is loaded on the system.
-        if [[ ${DRACUT_ARCH} == arm* || ${DRACUT_ARCH} == aarch64 ]] \
-            && grep -r -qs "virtio:d00000010v" /sys/bus/virtio/devices/*/modalias; then
+        if [[ ! $hostonly ]] \
+            || grep -r -qs "virtio:d00000010v" /sys/bus/virtio/devices/*/modalias; then
             echo drm
         fi
     fi


modules.d/11systemd-ask-password/module-setup.sh
```diff
@@ -19,15 +19,15 @@ check() {
 # Module dependency requirements.
 depends() {
 
-    if [[ $hostonly ]]; then
+    if [[ ${DRACUT_ARCH} == arm* || ${DRACUT_ARCH} == aarch64 ]]; then
```
Conan-Kudo, line 22: Is there a reason this needs to be arch specific? It seems like it would make sense on every architecture?

jozzsi: drm dracut module is relatively large - this would have performance implications for non-arm.

Conan-Kudo: How significant is the performance delta?

Rolv-Apneseth:
Take this with a grain of salt as it may be a misconfiguration of something on my end, but I've been trying to get an image built containing this updated module I was struggling to actually get the drm module enabled. Wasn't sure what I was doing wrong, but found this in the debug output:

```
$ dracut --debug --no-hostonly -f /tmp/test-initrd.img $(uname -r) 2>&1 | grep -n 'depends' | grep -i 'ask-password'
...
2645://usr/lib/dracut/modules.d/01systemd-ask-password/module-setup.sh@22(depends): [[ '' == arm* ]]
2646://usr/lib/dracut/modules.d/01systemd-ask-password/module-setup.sh@22(depends): [[ '' == aarch64 ]]
...
```

Looks like `DRACUT_ARCH` is not set? Do we need `${DRACUT_ARCH:-$(uname -m)}` here if it is indeed to be for aarch64 only?

*Edit*: Looks like removing the arch check does allow drm to get enabled:

```
$ sudo lsinitrd -m | grep drm
drm
```

jozzsi:
> _Edit_: Looks like removing the arch check does allow drm to get enabled:

This is expected as you do not have some of the dependent PR in your environment. 

Did this change resolved your issue ?

modules.d/11systemd-ask-password/module-setup.sh
jozzsi: > It seems like it would make sense on every architecture?

Could the actual exact condition be a follow-up discussion ? 

For example I do not think that in hostonly_mode strict we want to include drm module here.

> Is there a reason this needs to be arch specific?

@aafeijoo-suse @Vogtinator and the SUSE team is the best to comment on this, see https://github.com/dracut-ng/dracut-ng/pull/317

Rolv-Apneseth:
> Did this change resolved your issue ?

No, that's what I'm trying to say (sorry if it was worded poorly). If I, additionally to this change, remove the check for architecture, then I see `drm` enabled. I.e. the arch check fails even on the aarch64 system since it looks like that env var is not set

jozzsi:
> If I, additionally to this change, remove the check for architecture, then I see drm enabled. 

Did enabling the drm module resolved https://github.com/coreos/fedora-coreos-tracker/issues/1959 ?

I think this is important because you made this earlier statement before

> I'm not sure drm is involved here though, as on a F43 system:

Rolv-Apneseth:
> Did enabling the drm module resolved coreos/fedora-coreos-tracker#1959?

Yes, combined with other platform-specific changes, that module fixes the issue for Hetzner on aarch64.

I believe my previous problem with `drm` was because I was trying to rebuild the initramfs on a running system and due to the additional size required by the module, it would fail. On a side note, without `drm` the initramfs is at 86M but with it, it comes to 225M (~160% increase).

> I think this is important because you made this earlier statement before
> 
> > I'm not sure drm is involved here though, as on a F43 system:

This was because I noticed that aarch64 Fedora 43 on Hetzner does not have that `drm` module enabled, but _does_ have the `virtio-gpu` driver (which I have since realised is probably due to being built with `hostonly=yes`). I did not understand the relation between `drm` and `virtio-gpu` (still a bit fuzzy to be honest), and combined with the fact that I couldn't get `drm` to work, I was wondering if they may not be related after all.

Thanks again for your work on this.

modules.d/11systemd-ask-password/module-setup.sh
Vogtinator: > Is there a reason this needs to be arch specific? It seems like it would make sense on every architecture?

The base framebuffer (vga, efifb, ...) is usually enough for a boot display (x86, ppc64) or a serial console is commonly used instead (ppc64, s390). Not sure about risc-v.

> How significant is the performance delta?

In non-hostonly mode, the drm module pulls in all possible drm kernel modules and referenced firmware, which can be > 100 MBs (!). Bootloaders can have rather slow implementations for file loading, so this can be a boot time difference of > 10s in some cases.

IMO that's fine for initial initrds like used for installation or initial boot after deployment, but should be avoided in more general cases.

> For example I do not think that in hostonly_mode strict we want to include drm module here.

Agreed.

Conan-Kudo: The reason I am skeptical of this is that all of those modes are now under drm in at least Fedora, and I know it's being worked on in openSUSE too. So not having drm will actually break things even for firmwareless hostonly initrds. That's partly why this is happening on aarch64, I believe.

Vogtinator: > So not having drm will actually break things even for firmwareless hostonly initrds.

The "drm" dracut module != "drm" kernel subsystem.

The "simpledrm" infrastructure in the kernel exposes the system framebuffers like vga, vesa and efifb through drm without firmware.

jozzsi:
Thank you for the opportunity to be involved. At this time, I feel it is best for me to step away from contributing to the project. Closing this PR.

PR #2366, opened 2026-04-10
test: run all tests with --no-hostonly-cmdline

bdrung:
Several tests fail in a Ubuntu VM after commit 04af3c097797 enabled `hostonly_cmdline` in hostonly mode again, because the host disks leak into the `cmdline` config stored in the initrd. See 8c70c5bf99ea ("test(sysroot): run with --no-hostonly-cmdline") for a more detailed explanation.

So run all tests with `--no-hostonly-cmdline`.

### Checklist

### DIFF
--- a/dracut.conf.d/test/test.conf
+++ b/dracut.conf.d/test/test.conf
@@ -1,4 +1,5 @@
 add_dracutmodules+=" qemu "
+hostonly_cmdline="no"
 
 # test the dlopen dependencies support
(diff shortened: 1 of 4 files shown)


aafeijoo-suse (approved):
Thanks, I always have this problem when I run tests locally.

Conan-Kudo (approved):

devkontrol (approved):

PR #2369, opened 2026-04-10
feat(dracut): add a DRACUT_EXTRA_ARGS environment variable

jozzsi:
Some build tools (e.g. autopkgtest-build-qemu) invoke dracut indirectly through update-initramfs via dpkg triggers, making it impossible to pass extra flags directly on the command line.

Add a DRACUT_EXTRA_ARGS environment variable whose contents are word-split and prepended to the dracut argument list before option parsing.  This allows callers to inject arbitrary flags without modifying the configuration on disk:

  DRACUT_EXTRA_ARGS=--no-hostonly autopkgtest-build-qemu ...

The variable is intentionally word-split (like MAKEFLAGS), so arguments that themselves contain spaces must be shell-quoted within the value:

  DRACUT_EXTRA_ARGS='--add "mod1 mod2"' dracut
  
Changed test 13 to provide test coverage for DRACUT_EXTRA_ARGS variable.

### Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -601,6 +601,13 @@ PARMS_TO_STORE=""
 
 eval set -- "$TEMP"
 
+if [[ -n ${DRACUT_EXTRA_ARGS-} ]]; then
+    mapfile -d '' _extra_args < <(xargs printf '%s\0' <<< "$DRACUT_EXTRA_ARGS" 2> /dev/null \
+        || printf "%s\n" "dracut[W]: Ignoring malformed DRACUT_EXTRA_ARGS: $DRACUT_EXTRA_ARGS" >&2)
+    set -- "${_extra_args[@]}" "$@"
+    unset _extra_args
+fi
+
 while :; do
     case $1 in
         -h | --help)
--- a/man/dracut.8.adoc
+++ b/man/dracut.8.adoc
@@ -734,6 +734,18 @@ _DRACUT_ARCH_::
 Default:
     _empty_ (the value of **uname -m** on the host system)
 
+_DRACUT_EXTRA_ARGS_::
+    A string of additional _dracut_ command-line arguments.  The value is
+    word-split and prepended to the argument list before option parsing, so
+    explicit flags supplied directly on the command line will override the
+    values set here.
++
+Arguments that contain spaces must be shell-quoted within the value:
++
+----
+DRACUT_EXTRA_ARGS='--add "module1 module2"' dracut
+----
+
 _SYSTEMD_VERSION_::
     overrides systemd version. Used for **--sysroot**.
 
--- a/test/TEST-13-SYSROOT/test.sh
+++ b/test/TEST-13-SYSROOT/test.sh
@@ -23,7 +23,10 @@ test_setup() {
     build_ext4_image "$TESTDIR/rootfs" "$TESTDIR"/root.img dracut
 
     ln -s / "$TESTDIR"/sysroot
-    test_dracut --keep --hostonly --sysroot "$TESTDIR"/sysroot
+    # test DRACUT_EXTRA_ARGS
+    DRACUT_EXTRA_ARGS="--keep --hostonly --sysroot '$TESTDIR/sysroot'" test_dracut
+
+    grep 'hostonly' "$TESTDIR"/initrd/dracut.*/initramfs/usr/lib/dracut/build-parameter.txt
 
     if grep -q '^root:' /etc/shadow; then
         if ! grep -q '^root:' "$TESTDIR"/initrd/dracut.*/initramfs/etc/shadow; then


man/dracut.8.adoc
```diff
@@ -734,6 +734,14 @@ _DRACUT_ARCH_::
 Default:
     _empty_ (the value of **uname -m** on the host system)
 
+_DRACUT_EXTRA_ARGS_::
+    A string of additional _dracut_ command-line arguments.  The value is
+    word-split and prepended to the argument list before option parsing, so
+    explicit flags supplied directly on the command line will override the
+    values set here.
+
+    Arguments that contain spaces must be shell-quoted within the value.
```
bdrung: Could you add an example here?

jozzsi: ok. done.

man/dracut.8.adoc
```diff
@@ -734,6 +734,18 @@ _DRACUT_ARCH_::
 Default:
     _empty_ (the value of **uname -m** on the host system)
 
+_DRACUT_EXTRA_ARGS_::
+    A string of additional _dracut_ command-line arguments.  The value is
+    word-split and prepended to the argument list before option parsing, so
+    explicit flags supplied directly on the command line will override the
+    values set here.
+
+    Arguments that contain spaces must be shell-quoted within the value:
```
bdrung: This section has the wrong indentation.

man/dracut.8.adoc
```diff
@@ -734,6 +734,18 @@ _DRACUT_ARCH_::
...
+
+----
+    DRACUT_EXTRA_ARGS='--add "mod1 mod2"' dracut
```
bdrung: For consistency with the other examples:
```suggestion
    DRACUT_EXTRA_ARGS='--add "module1 module2"' dracut
```

bdrung:
I tested `DRACUT_EXTRA_ARGS='--add bash' dracut --force`  on my schroot-big-initrd which worked, but the quoting did not work:

```
$ DRACUT_EXTRA_ARGS='--add "bash qemu"' dracut --force

!Unknown option: 'qemu"'

Usage: /usr/bin/dracut [OPTION]... [<initramfs> [<kernel-version>]]

Version: 110

Creates initial ramdisk images for preloading modules

  -h, --help  Display all options

If a [LIST] has multiple arguments, then you have to put these in quotes.

For example:

    # dracut --add-drivers "module1 module2"  ...

```

man/dracut.8.adoc
jozzsi: not sure what is the right indentation here, but made some changes.

man/dracut.8.adoc
jozzsi: ok. done.

man/dracut.8.adoc
```diff
@@ -734,6 +734,17 @@ _DRACUT_ARCH_::
 Default:
     _empty_ (the value of **uname -m** on the host system)
 
+_DRACUT_EXTRA_ARGS_::
+    A string of additional _dracut_ command-line arguments.  The value is
+    word-split and prepended to the argument list before option parsing, so
+    explicit flags supplied directly on the command line will override the
+    values set here.
+
+Arguments that contain spaces must be shell-quoted within the value:
```
bdrung: This will fix the indentation:
```suggestion
+
Arguments that contain spaces must be shell-quoted within the value:
+
```
You can verify that with `make doc && man -l man/dracut.8`.

bdrung:
Successfully tested the concept that the environment variable is passed through update-initramfs to dracut: `DRACUT_EXTRA_ARGS="--add bash" update-initramfs -u`

jozzsi:
> I tested `DRACUT_EXTRA_ARGS='--add bash' dracut --force` on my schroot-big-initrd which worked, but the quoting did not work:

Pushed a new version, please check when you have a chance.

man/dracut.8.adoc
jozzsi: fixed, thanks

dracut.sh
```diff
@@ -601,6 +601,14 @@ PARMS_TO_STORE=""
 
 eval set -- "$TEMP"
 
+if [[ -n ${DRACUT_EXTRA_ARGS-} ]]; then
+    # eval is required so that quoted arguments inside DRACUT_EXTRA_ARGS are
+    # parsed correctly (plain word-splitting does not honour shell quoting).
+    # The existing positional parameters are protected via printf %q.
+    # shellcheck disable=SC2294
+    eval "set -- ${DRACUT_EXTRA_ARGS} $(printf '%q ' "$@")"
```
bdrung: We should avoid `eval` because that can execute any code defined in the env.

```
mapfile -d '' _extra_args < <(xargs printf '%s\0' <<< "$DRACUT_EXTRA_ARGS" 2>/dev/null)
set -- "${_extra_args[@]}" "$@"
unset _extra_args
```

bdrung:
Works and the arguments are recorded in `Arguments` in `lsinitrd`.

dracut.sh
jozzsi: thx

dracut.sh
```diff
@@ -601,6 +601,12 @@ PARMS_TO_STORE=""
 
 eval set -- "$TEMP"
 
+if [[ -n ${DRACUT_EXTRA_ARGS-} ]]; then
+    mapfile -d '' _extra_args < <(xargs printf '%s\0' <<< "$DRACUT_EXTRA_ARGS" 2> /dev/null)
```
bdrung: Maybe we should not use `2> /dev/null`. Then xargs will complain in case DRACUT_EXTRA_ARGS is malformed.

bdrung: Okay, we should log failures and that we ignore the extra args then:

```
xargs printf '%s\0' <<< "$DRACUT_EXTRA_ARGS" 2> /dev/null || echo "DRACUT_EXTRA_ARGS='$DRACUT_EXTRA_ARGS' is malformed" >&2
```

jozzsi: thx

dracut.sh
```diff
@@ -601,6 +601,12 @@ PARMS_TO_STORE=""
 
 eval set -- "$TEMP"
 
+if [[ -n ${DRACUT_EXTRA_ARGS-} ]]; then
+    mapfile -d '' _extra_args < <(xargs printf '%s\0' <<< "$DRACUT_EXTRA_ARGS" 2> /dev/null || echo "DRACUT_EXTRA_ARGS='$DRACUT_EXTRA_ARGS' is malformed" >&2)
```
bdrung: We should use our logging infrastructure, shouldn't we?

jozzsi: I think at this point dracut logging i not yet avaialble/initialized.

test/TEST-13-SYSROOT/test.sh
```diff
@@ -23,7 +23,8 @@ test_setup() {
     build_ext4_image "$TESTDIR/rootfs" "$TESTDIR"/root.img dracut
 
     ln -s / "$TESTDIR"/sysroot
-    test_dracut --keep --hostonly --sysroot "$TESTDIR"/sysroot
+    # test DRACUT_EXTRA_ARGS
+    DRACUT_EXTRA_ARGS="--keep --hostonly --sysroot $TESTDIR/sysroot" test_dracut
```
bdrung: Should we check that usr/lib/dracut/build-parameter.txt contains `--hostonly` from `DRACUT_EXTRA_ARGS`?

jozzsi: I think if you misspell the variable name (A="--keep --hostonly --sysroot $TESTDIR/sysroot" test_dracut) the test would fail, that is enough validation for me, but if you think this is important, I can look into it.

dracut.sh
jozzsi: value of DRACUT_EXTRA_ARGS variable could also change logging levels, so hard to get an accurate reading of what and how to log at this point of time in the code.

bdrung: Okay, then let's use our common pattern:
```suggestion
    mapfile -d '' _extra_args < <(xargs printf '%s\0' <<< "$DRACUT_EXTRA_ARGS" 2> /dev/null || printf "%s\n" "dracut[W]: Ignoring malformed DRACUT_EXTRA_ARGS: $DRACUT_EXTRA_ARGS" >&2)
```

Maybe we should wrap this long line.

test/TEST-13-SYSROOT/test.sh
bdrung: Something like:
```
grep '--hostonly` "$TESTDIR"/initrd/dracut.*/usr/lib/dracut/build-parameter.txt
```
should be enough to verify it. This will also check the code path were we record the arguments.

jozzsi: ok

dracut.sh
jozzsi: thx

test/TEST-13-SYSROOT/test.sh
```diff
@@ -23,7 +23,10 @@ test_setup() {
...
```
bdrung: This would also test the quoting:
```suggestion
    DRACUT_EXTRA_ARGS="--keep --hostonly --sysroot '$TESTDIR/sysroot'" test_dracut
```

jozzsi: ok

bdrung (approved):

devkontrol (approved):

jozzsi:
Somehow this commit regressed (or exposed) TEST-13 passing on alpine.

To reproduced

```
test/test.sh alpine:edge 13
```

As seen https://github.com/dracut-ng/dracut-ng/actions/runs/24373582467

Note that the following warning is unexpectedly present both before and after the PR

```
dracut[E]: Module 'shell-interpreter' depends on module 'busybox', which can't be installed
```

I suspect somehow this is an issue with `--sysroot` argument not working well in this Alpine comtainer.

PR #2373, opened 2026-04-11
ci(arch): ensure /etc/vconsole.conf exists

jozzsi:
`/etc/vconsole.conf` no longer gets installed as part of package installation. 

See https://bbs.archlinux.org/viewtopic.php?id=310236

### Checklist


devkontrol (approved):
ok.

I was informed this PR would help reenable running test 41 on arch.

PR #2382, opened 2026-04-13
fix(iscsi): handle empty URI in firmware boot mode

### LINKED ISSUE #1701: ip=dhcp netroot=iscsi
**Describe the bug**

```
[ 2.248948] dracut-cmdline[185]: /lib/net-lib.sh: line 432: shift: shift count out of range
[ 2.251889] dracut-cmdline[185]: /lib/net-lib.sh: line 462: shift: shift count out of range
```

Full downstream bug report - https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/2121865

Not very clear from the downstream bug report, but it seems the errors are not fatal.

CC @mwilck @bdrung @fabiomirmar

rakeshw728:
When rd.iscsi.firmware=1 is set, netroot is either 'iscsi' or 'iscsi:'                                                                              
  with no real URI — the firmware handles the iSCSI target discovery.   
  However dracut still calls parse_iscsi_root() and handle_netroot() with                                                                             
  these bare values causing two issues:                                                                                                               
                                                                                                                                                      
  1. "shift count out of range" in parse_iscsi_root() in net-lib.sh                                                                                   
     When netroot=iscsi: (with colon), set $v produces $#=0.                                                                                          
     When netroot=iscsi (no colon), after the first shift $#=0.                                                                                       
     Both cases cause subsequent shift calls to fail.                                                                                                 
                                                                                                                                                      
     Fix: add two guards to return early when $#=0 — nothing to parse                                                                                 
     in firmware mode.                                                                                                                                
                                                                                                                                                      
  2. "iscsiadm: Cannot resolve host iscsi" in iscsiroot.sh                                                                                            
     The netroot loop strips the iscsi: prefix but when netroot=iscsi                                                                                 
     (no colon) the pattern doesn't match, leaving nroot="iscsi" —                                                                                    
     non-empty — so handle_netroot is called with "iscsi" as the target                                                                               
     IP causing a spurious DNS resolution attempt.                                                                                                    
                                                                                                                                                      
     Fix: skip handle_netroot when nroot="iscsi" after stripping.                                                                                     
                                                                                                                                                      
  Tested on: OCI VM.Standard.E5.Flex, RHEL 10.1, dracut-107-3.el10.x86_64                                                                           
                                                                                                                                                      
  Fixes #1701

### DIFF
--- a/modules.d/45net-lib/net-lib.sh
+++ b/modules.d/45net-lib/net-lib.sh
@@ -426,6 +426,7 @@ parse_iscsi_root() {
     set $v
     IFS="$OLDIFS"
 
+    [ $# -ge 2 ] || return 0
     iscsi_protocol=$1
     shift # ignored
     iscsi_target_port=$1
--- a/modules.d/74iscsi/iscsiroot.sh
+++ b/modules.d/74iscsi/iscsiroot.sh
@@ -301,7 +301,7 @@ if ! [ "$netif" = "online" ]; then
         for nroot in $(getargs netroot); do
             [ "${nroot%%:*}" = "iscsi" ] || continue
             nroot="${nroot##iscsi:}"
-            if [ -n "$nroot" ]; then
+            if [ -n "$nroot" ] && [ "$nroot" != "iscsi" ]; then
                 handle_netroot "$nroot"
                 ret=$((ret + $?))
             fi


bdrung:
Thanks. Can you also add this line to your commit:
```
Bug-Ubuntu: https://launchpad.net/bugs/2121865
```

rakeshw728:
done.

modules.d/45net-lib/net-lib.sh
```diff
@@ -426,8 +426,12 @@ parse_iscsi_root() {
     set $v
     IFS="$OLDIFS"
 
+    [ $# -eq 0 ] && return 0
```
bdrung: IMO the opposite is more readable:

```suggestion
    [ $# -gt 0 ] || return 0
```

mwilck: I agree with @bdrung. But with your change further down, `parse_iscsiroot()` shouldn't be called at all any more for `netroot=iscsi` or `netroot=iscsi:`. So why is this needed at all?

bdrung: I still prefer to have this check here in case the function is called incorrectly.

Oh, and this can be changed to:
```
[ $# -ge 2 ] || return 0
```
Then the second check for `$#` can be dropped.

rakeshw728:
Thank you for the suggestions. Updated the fix - replaced both guards with [
$# -ge 2 ] || return 0 as suggested by @bdurang . This handles all cases

  in one check and drops the second guard entirely.





  @mwilck - agree that the iscsiroot.sh change already prevents
parse_iscsi_root from being called with an empty URI, but keeping this
check here as a

   defensive measure in case the function is called incorrectly from other
paths.

modules.d/45net-lib/net-lib.sh
```diff
@@ -426,8 +426,11 @@ parse_iscsi_root() {
     set $v
     IFS="$OLDIFS"
 
+    [ $# -eq 2 ] && return 0
+
     iscsi_protocol=$1
     shift # ignored
+
```
bdrung: I suggest to drop this added newline again.
```suggestion
```

bdrung:
Oh, your git commit message has a two space indentation which is not needed.

modules.d/45net-lib/net-lib.sh
rakeshw728: changes pushed as suggested

Conan-Kudo:
Wait, this was tested on RHEL but has an Ubuntu bug? Also, what?

Conan-Kudo (dismissed):

bdrung (approved):

modules.d/45net-lib/net-lib.sh
```diff
@@ -426,6 +426,7 @@ parse_iscsi_root() {
     set $v
     IFS="$OLDIFS"
 
+    [ $# -eq 2 ] && return 0
```
bdrung: I just noticed that the `-eq` condition is wrong. It should check that `$#` is greater or equal to 2.
```suggestion
    [ $# -ge 2 ] || return 0
```

bdrung (changes requested):

rakeshw728:
Fixed!

bdrung (approved):

bdrung:
There is a merge commit in your PR. Can you rebase your PR on main to get rid of the merge commit?

bdrung:
Ditch my request, Github allowed me to do the rebase.

bdrung (changes requested):
Can you remove the two-spaces indentation in the git commit message?

bdrung:
> Wait, this was tested on RHEL but has an Ubuntu bug? Also, what?

This was found on Ubuntu and affects all distributions.

bdrung (approved):

Conan-Kudo (approved):

PR #2386, opened 2026-04-13
ci: sort modules in labeler.yml

bdrung:
Sort modules in `labeler.yml` and keep them last. This is done in preparation to generate this part of `labeler.yml` with a script.

### Checklist


devkontrol (approved):
thanks @bdrung

PR #2397, opened 2026-04-17
feat(chrony): introducing the chrony module

aafeijoo-suse:
This new module relies on chrony to add support for NTP in the initrd. By default, it also uses NTP sources from DHCP.

It provides two new kernel command line options:
- `rd.ntp`: it can be used multiple times to specify server, pool and peer time sources.
- `rd.ntp.nodhcp`: it disables the use of NTP sources from DHCP.

In some environments, it's necessary to fetch live images from HTTPS URLs. The problem is that the hardware clock may be off, so the validation of the SSL certificates may fail if the time is out of sync. Also, some systems may not have access to DHCP for security reasons, that's why we need a way to configure them on the kernel command line.

Notes:
- chrony upstream only provides systemd services and configuration as [examples](https://gitlab.com/chrony/chrony/-/tree/master/examples?ref_type=heads), so there are many differences between the distributions.
- While it might make sense to use /var on the host, let's try to use /run in the initrd, so everything can persist after switching root.

### Checklist

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -88,6 +88,10 @@ caps:
     - changed-files:
           - any-glob-to-any-file: 'modules.d/[0-9][0-9]caps/*'
 
+chrony:
+    - changed-files:
+          - any-glob-to-any-file: 'modules.d/[0-9][0-9]chrony/*'
+
 cifs:
     - changed-files:
           - any-glob-to-any-file: 'modules.d/[0-9][0-9]cifs/*'
--- a/.github/workflows/daily-network.yml
+++ b/.github/workflows/daily-network.yml
@@ -40,6 +40,7 @@ jobs:
                 test:
                     - "31"
                     - "60"
+                    - "61"
                 exclude:
                     - container: arch:latest
                       architecture: {runner: 'ubuntu-24.04-arm', tag: 'arm'}
@@ -50,6 +51,17 @@ jobs:
                     # https://github.com/dracut-ng/dracut/issues/1988
                     - container: debian:sid
                       architecture: {runner: 'ubuntu-24.04-arm', tag: 'arm'}
+                    # In Debian/Ubuntu both chrony and systemd-timesyncd provide
+                    # the virtual package time-daemon, so systemd-timesyncd
+                    # would have to be removed, breaking test 41.
+                    - container: debian:latest
+                      test: "61"
+                    - container: debian:sid
+                      test: "61"
+                    - container: ubuntu:devel
+                      test: "61"
+                    - container: ubuntu:rolling
+                      test: "61"
         container:
             image: ghcr.io/dracut-ng/${{ matrix.container }}
             options: '--device=/dev/kvm --privileged'
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -190,6 +190,13 @@ jobs:
                 test:
                     - "31"
                     - "60"
+                    - "61"
+                exclude:
+                    # In Debian/Ubuntu both chrony and systemd-timesyncd provide
+                    # the virtual package time-daemon, so systemd-timesyncd
+                    # would have to be removed, breaking test 41.
+                    - container: ubuntu:devel
+                      test: "61"
         container:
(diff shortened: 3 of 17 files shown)


devkontrol:
Would it be reasonable to add test case for this new dracut module ? 

> so the validation of the SSL certificates may fail 

Does this happen already on the CI ? Can we trigger it with some test ?

aafeijoo-suse:
> Would it be reasonable to add test case for this new dracut module ?

Upstream has 143 modules and 24 tests. There is already a module providing NTP without a test (well, `systemd-timesyncd` only provides simple NTP). I can't find a rule of thumb, some of the latest additions do not have a test: `devicetree-firmware`, `memdisk`, `simpledrm`, `systemd-battery-check`, `hwdb`, `pcmcia`, `numlock`, `systemd-bsod`... Furthermore, adding tests can take ages (see my PR adding tests for `systemd-sysext`).

> > so the validation of the SSL certificates may fail
> 
> Does this happen already on the CI ?

I don't think so. The real use case for this module is an OS installer image, downloaded remotely from the initrd. We need a way to specify NTP in the initrd before fetching the image over HTTPS, otherwise, in isolated systems with an out of sync clock the fetch may fail and the OS could not be installed.

> Can we trigger it with some test ?

I guess an HTTPS server with an expired certificate could be simulated. But... I'm not sure if it makes sense to test this in the CI.

devkontrol:
> some of the latest additions do not have a test: `devicetree-firmware`, `memdisk`, `simpledrm`, `systemd-battery-check`, `hwdb`, `pcmcia`, `numlock`, `systemd-bsod`

None of these dracut modules maintain a dedicated .service file - they all just copy the .service file that is maintained by other project. This PR however introducing a new .service file (and .conf file) that this project would have to maintain.

Could we pick up the .service file and .conf file from the host instead ? 

> chrony upstream only provides systemd services and configuration as examples, so there are many differences between the distributions.

This PR suggests that dracut would now maintain distribution-independent service file without a test.

aafeijoo-suse:
> None of these dracut modules maintain a dedicated .service file - they all just copy the .service file that is maintained by other project.

And that does not mean that they work with dracut, usually because dracut is not installing all required dependencies, binaries, kernel modules, configuration... Also, there are some dracut modules that install modified versions of systemd services via `sed` without a test (`bluetooth`, `dbus-daemon`, `dbus-broker`).

> This PR however introducing a new .service file (and .conf file) that this project would have to maintain.
> Could we pick up the .service file and .conf file from the host instead ?

No, we can't, for the reasons explained in the notes of the PR description, plus although the systemd services were the same on the host in all distros, they are not implemented to work in the initrd (the Install section cannot be overridden and it has a `WantedBy=multi-user.target`).

> This PR suggests that dracut would now maintain distribution-independent service file without a test.

I don't see why that is a big deal here, most of the merged code is not being tested. But anyway, I'll add one test if that is the only objection to this PR.

Conan-Kudo:
I would prefer some tests for this module, please.

bdrung:
I would like to see code tested as well: either get those services upstream and test it there; or add a test case in dracut.

aafeijoo-suse:
Draft until the CI has #2418

devkontrol:
> There is already a module providing NTP without a test

I think this is not the case as TEST-41 includes `systemd-network-management`, which than includes `systemd-timesyncd`. Perhaps NTP functionality is not tested/confirmed, but at the CI would break without `systemd-timesyncd`

Conan-Kudo:
> Draft until the CI has #2418

This has been in for a little while now, do you want to rebase it and prepare it for review?

aafeijoo-suse:
> This has been in for a little while now, do you want to rebase it and prepare it for review?

There is an open PR for systemd-networkd that should land first in one way or another, related to the change from dash to bash and the use of `printf '%q'`.

bdrung:
Instead of requiring bash just for `printf '%q'`, I implemented a `escape` function to make in POSIX compliant: https://github.com/dracut-ng/dracut/pull/2507

devkontrol (approved):

Conan-Kudo (approved):

PR #2398, opened 2026-04-18
fix(kernel-modules): add Mediatek MTU3 USB controller

Icenowy:
This is the USB dual-role controller used by some MediaTek SoCs.

Adding it fixes booting from USB storage on Lenovo Chromebook Duet EDU G2 (mt8188-geralt-ciri). That device does not have any SD slot, so USB storage is the only usable external storage.

### Checklist

### DIFF
--- a/modules.d/70kernel-modules/module-setup.sh
+++ b/modules.d/70kernel-modules/module-setup.sh
@@ -100,6 +100,7 @@ installkernel() {
                 "=drivers/usb/host" \
                 "=drivers/usb/isp1760" \
                 "=drivers/usb/misc" \
+                "=drivers/usb/mtu3" \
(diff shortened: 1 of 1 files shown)


Conan-Kudo (approved):

Conan-Kudo (changes requested):
The commit message uses the wrong number here. Just drop the number as we don't typically use that in commit messages.

Icenowy:
Number removed now.

bdrung (approved):
initramfs-tools from Debian includes `=drivers/usb/mtu3` as well.

devkontrol (approved):

Conan-Kudo (approved):

PR #2399, opened 2026-04-19
01-gentoo.conf: disable hostonly_cmdline and improve documentation

Nowa-Ammerlaan:
Our new users are stumbling over a check we added that prevents accidentally including a cmdline for the live system in the initramfs for the new install. Since the overall majority of our new users will be setting a cmdline via the bootloader or system firmware anyway this setting is not really useful for our users and only creates confusion because there are now two knobs which control the same thing.

For that reason we would like to disable the `hostonly_cmdline` by default (like Fedora does as well).

While I am touching this file anyway I added explicitly the other configuration settings that we have been controlling implicitly via other packages. By documenting this explicitly I hope that users reading this configuration file will have an (even) easier time finding "the Gentoo way" of doing things like including microcode or building an UKI. We control these things outside the scope of Dracut since optional external dependencies are required for these functions. (linux-firmware, intel-microcode systemd-stub, ukify, etc).

See-also: https://bugs.gentoo.org/971572
See-also: https://github.com/dracut-ng/dracut-ng/issues/1326

CC @thesamesam @immolo 

tl;dr The only effective change here is disabling `hostonly_cmdline`, the rest is just more verbosity from my side on things we are already doing.

### Checklist

### DIFF
--- a/dracut.conf.d/gentoo/01-gentoo.conf
+++ b/dracut.conf.d/gentoo/01-gentoo.conf
@@ -1,2 +1,22 @@
-# Gentoo specific Dracut configuration
+# Build an initramfs specific to this machine by default. This ensures
+# local configuration files (/etc/...) are included.
+hostonly="yes"
+hostonly_mode="sloppy"
+
+# Usage of ukify is controlled via sys-kernel/installkernel[+/-ukify].
+# Automatic delegation in sys-kernel/dracut is disabled here.
 ukify="no"
+# Building an UKI is toggled by sys-kernel/installkernel[+/-uki].
+# If the ukify USE flag is also enabled then ukify is used to build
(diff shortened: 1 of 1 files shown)


devkontrol (dismissed):

thesamesam (approved):
I'd prefer it split into two commits (one PR is fine). Anyway, thanks!

Nowa-Ammerlaan:
> I'd prefer it split into two commits (one PR is fine). Anyway, thanks!

I split it into two parts:
- first commit effectively changes nothing, just sets explicitly what we are already doing.
- second commit contains the functional change: disabling hostonly_cmdline

devkontrol (approved):

thesamesam (approved):

Conan-Kudo (approved):

PR #2400, opened 2026-04-20
refactor(dracut): remove unnecessary 'realpath' call if '-k' is specified

MewtR:
No need to use 'realpath' to resolve a default directory to look for modules if one has already been specified. 'realpath' will complain if the directory doesn't exist which can lead a user to think that something went wrong.

### Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1468,11 +1468,11 @@ fi
 
 export LC_MESSAGES=C kernel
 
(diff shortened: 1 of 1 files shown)


jozzsi:
please make commit message shorter, commisery fails

```
Error: [C014] Subject should be within the line length limit (80), exceeded by 4 characters
```

MewtR:
Done

Conan-Kudo (approved):

bdrung (approved):

PR #2402, opened 2026-04-23
feat(crypt-gpg): add a parameter for GPG password max tries

ChristianAngel:
This adds the rd.luks.key.max-tries parameter that specifies the number of times that dracut will prompt for a GPG password before falling back to passphrase decryption. Previously, the number of tries was hardcoded to be 3, but this parameter lets the user customize it. In particular, if rd.luks.key.max-tries is set to 0, dracut will re-prompt for the GPG password forever and will not try to fall back to a passphrase. If rd.luks.key.max-tries is not specified, the default value is 3.

### DIFF
--- a/modules.d/70crypt-lib/crypt-lib.sh
+++ b/modules.d/70crypt-lib/crypt-lib.sh
@@ -118,12 +118,17 @@ ask_for_password() {
             fi
 
             local i=1
-            while [ $i -le "$tty_tries" ]; do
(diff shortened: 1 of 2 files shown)


Conan-Kudo:
Please squash your style change into your feature change. The style change is a modification of your feature change commit anyway, and so the separation is irrelevant.

bdrung:
And the new `rd.luks.key.max-tries` should be documented in the man page.

PR #2406, opened 2026-04-24
fix(crypt): set need_shutdown for LUKS under systemd-cryptsetup

nadzyah:
cryptroot-ask.sh calls need_shutdown() in the legacy path, but systemd-cryptsetup has no equivalent. Without the flag, dracut-initramfs-restore exits early, the initramfs is not pivoted to at shutdown, and dm-shutdown.sh never runs, leaving LUKS stacks busy and blocking poweroff.

Add a cleanup hook that sets the flag whenever a LUKS mapping is active, matching the mdraid-needshutdown.sh / multipathd-needshutdown.sh pattern.

Bug-Ubuntu: https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/2147476

### Checklist

### DIFF
--- a/test/TEST-26-ENC-RAID-LVM/test.sh
+++ b/test/TEST-26-ENC-RAID-LVM/test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -eu
 # shellcheck disable=SC2034
-TEST_DESCRIPTION="root filesystem on LVM on encrypted partitions of a RAID"
+TEST_DESCRIPTION="root filesystem on LVM on encrypted partitions (with and without RAID)"
 
 # Uncomment this to debug failures
 #DEBUGFAIL="rd.shell rd.break" # udev.log-priority=debug
@@ -11,6 +11,11 @@ TEST_DESCRIPTION="root filesystem on LVM on encrypted partitions of a RAID"
 test_check() {
     if ! type -p cryptsetup &> /dev/null; then
         echo "Test needs cryptsetup for crypt module... Skipping"
+        return 1
+    fi
+
+    if ! type -p lvm &> /dev/null; then
+        echo "Test needs lvm for lvm module... Skipping"
         return 1
     fi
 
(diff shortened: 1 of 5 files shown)


test/TEST-26-ENC-RAID-LVM/test.sh
```diff
@@ -76,6 +119,30 @@ make_test_rootfs() {
...
+    if [ -x /usr/lib/systemd/systemd-cryptsetup ] || [ -x /lib/systemd/systemd-cryptsetup ]; then
+        # Build the no-RAID rootfs and systemd-based initramfs first, then
+        # rename it aside, the RAID build reuses $TESTDIR/initramfs.testing
+        make_enc_lvm_rootfs
+
+        luks_uuid=$(grep -F -a -m 1 ID_FS_UUID "$TESTDIR"/marker-enc-lvm.img | cut -d= -f2)
+        printf 'rd.luks.uuid=luks-%s' "$luks_uuid" > "$TESTDIR"/luks-enc-lvm.txt
+        printf 'luks-%s /dev/disk/by-id/scsi-0QEMU_QEMU_HARDDISK_disk1 /etc/key timeout=0\n' "$luks_uuid" > /tmp/crypttab-enc-lvm
+
+        test_dracut -f \
```
jozzsi, line 136: Why `-f` ?

nadzyah: I use `-f` as it forces dracut to overwrite an existing `$TESTDIR/initramfs.testing`. This is needed because `$TESTDIR` is reused across reruns, so the file from a prior run is already there, and we now do two builds in one run. The first writes `initramfs.testing`, we mv it to `initramfs.testing.enc-lvm`, then the second writes a fresh `initramfs.testing` for the RAID case. Without -f, dracut errors out with `Will not override existing initramfs`

jozzsi: I see. Thanks !

bdrung:
Is this a bug in systemd-shutdown? In other words: should we make this case work with systemd-shutdown and not rely on needing dracut's exitrd?

zertrin:
I think I'm subject to the bug reported on ubuntu. I'm using Ubuntu 25.10 with the root system on LUKS, with `systemd-cryptsetup`.

```
└─nvme0n1p7                 259:7    0   590G  0 part  
  └─dm_crypt-0              252:0    0   590G  0 crypt 
    └─ubuntu--vg-ubuntu--lv 252:1    0   590G  0 lvm   /
```

As this is very annoying, could I "safely" test this fix on my end?

Should I use the hook in this PR and put in under `/usr/lib/dracut/modules.d/70crypt` (or somewhere else), then change `module-setup.sh` as per the diff?

zertrin:
Hum, while digging a bit deeper to understand what's going on, I noticed that on the unmodified system, I already have a `/run/initramfs/.need_shutdown` while being on the main system.

I'm not sure which module is the source of it, could be `lvm_scan.sh` or `cryptroot-ask.sh` most likely (although the root passphrase is being asked by some plymouth nice UI, I don't know if that matters.)

Does this indicate that the changes in this PR wouldn't really change anything? (since, as far as I can understand, this PR is supposed to call `need_shutdown` function in one more case, but in my case I'm guessing it is already called).

jefferyto:
Hello - I am the original reporter of the Launchpad issue.

I recently updated to Ubuntu 26.04 and just now tried the fix in this PR (editing /usr/lib/dracut/modules.d/70crypt/module-setup.sh, running `update-initramfs -c -k all`), and I still have the same issue of shutdown not fully completing.

Similar to what @zertin found, on my system (without the fix in this PR) I have the empty file ".need_shutdown" in /run/initramfs. I also believe this was created by lvm_scan.sh.

Not sure if these will be helpful (both without changes in this PR):

* [Journal entries](https://github.com/user-attachments/files/28552896/journal.txt) of a session where I booted my computer, then immediately triggered shutdown at the login screen
* Output of [lsinitrd](https://github.com/user-attachments/files/28552951/lsinitrd.txt)

nadzyah:
@zertrin @jefferyto thank you so much for testing these changes! I really appreciate it! I'll be sure to dive back into this when I find some time later. Thanks again!

jefferyto:
Thanks @nadzyah - just wanted to clarify, this PR, or a fix in systemd-shutdown as @bdrung asked, is probably necessary (i.e. when LUKS encryption is used but without lvm) but it doesn't appear to fix the Launchpad issue.

PR #2414, opened 2026-04-27
man: document hookdir

jozzsi:
hookdir variable is the recommended way to refer to the path of the directory used for dracut hooks during boot.

This variable is part of the API for 3rd party dracut modules.

Follow-up to 04d5e29 .

CC @vittyvk 

### Checklist


vittyvk (approved):
Certainly makes sense to mention this in the doc as it's been widely used already!

devkontrol (approved):

PR #2418, opened 2026-04-30
ci: install chrony to test the new chrony module

aafeijoo-suse:
Required to run tests for #2397

### Checklist

### DIFF
--- a/test/container/Dockerfile-arch
+++ b/test/container/Dockerfile-arch
@@ -29,6 +29,7 @@ RUN pacman --noconfirm -Syu \
     bluez \
     btrfs-progs \
     cargo \
+    chrony \
     cpio \
     dhcp \
     dmraid \
--- a/test/container/Dockerfile-fedora
+++ b/test/container/Dockerfile-fedora
@@ -49,6 +49,7 @@ RUN dnf -y install --setopt=install_weak_deps=False \
     bash-completion \
     bluez \
     cargo \
+    chrony \
     cifs-utils \
     cryptsetup \
     crypto-policies-scripts \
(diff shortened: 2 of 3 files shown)


devkontrol (approved):

devkontrol (changes requested):
This change would break Debian/Ubuntu CI
```
#5 16.00 E: Unable to satisfy dependencies. Reached two conflicting assignments:
#5 16.00    1. systemd-timesyncd:arm64=260.1-1 is selected for install
#5 16.00    2. systemd-timesyncd:arm64 is not selected for install because:
#5 16.00       1. chrony:arm64=4.8-3 is selected for install
#5 16.00       2. systemd-timesyncd:arm64 Conflicts time-daemon
#5 16.00          [selected chrony:arm64=4.8-3]
```

CC @bdrung

bdrung:
Both chrony and systemd-timesyncd provide the virtual package `time-daemon` and there is only one of those allowed. So to install chrony, systemd-timesyncd has to be dropped from the dockerfile.

aafeijoo-suse:
> Both chrony and systemd-timesyncd provide the virtual package `time-daemon` and there is only one of those allowed. So to install chrony, systemd-timesyncd has to be dropped from the dockerfile.

Ok, I didn't know that. Changed.

bdrung:
No idea, why ubuntu:rolling-amd fails. I created https://github.com/dracut-ng/dracut-ng/pull/2419 to make the build log more verbose.

devkontrol:
TEST-41 requires systemd-timesyncd

bdrung:
So separate containers? One with chrony and one with systemd-timesyncd?

aafeijoo-suse:
> So separate containers? One with chrony and one with systemd-timesyncd?

Or do not test chrony in Debian/Ubuntu.

Conan-Kudo (approved):

devkontrol (approved):
This PR does not cover all distributions we have on the CI, but at the same time it makes the project moves forward.

PR #2420, opened 2026-05-01
ci: run nbd test on debian:sid again

### LINKED ISSUE #2381: nbd test started to fail on Debian unstable with nbd 1:3.27.1-1
See https://github.com/dracut-ng/dracut-ng/actions/runs/24319608739

It is only failing on arch and debian:sid (see test 72) - which seems to suggest that there is a recent regression that only made it recent rolling versions.

CC @bdrung

bdrung:
The nbd 1:3.27.1-1 Debian package broke the nbd tests. nbd 1:3.27.1-2 fixed that regression.

So run the nbd test on `debian:sid` again.

Fixes: https://github.com/dracut-ng/dracut-ng/issues/2381

### Checklist

### DIFF
--- a/.github/workflows/daily-nbd.yml
+++ b/.github/workflows/daily-nbd.yml
@@ -31,6 +31,7 @@ jobs:
                 container:
                     - arch:latest
                     - debian:latest
+                    - debian:sid
                     - fedora:latest
                     - opensuse:latest
                     - ubuntu:devel


Conan-Kudo (approved):

devkontrol:
NBD on arm (debian:sid) does not finish in an hour, we still might need to excude it from the CI (but the amd run is a great addition).

devkontrol (approved):

PR #2433, opened 2026-05-08
fix: remove basename calls

bdrung:
https://dracut-ng.github.io/dracut/developer/bash.html says that `basename` should not be used.

So replace the usage of `basename` by variable expansion instead.

### Checklist

### DIFF
--- a/modules.d/45url-lib/url-lib.sh
+++ b/modules.d/45url-lib/url-lib.sh
@@ -160,7 +160,7 @@ nfs_fetch_url() {
         mntdir="$(mkuniqdir /run nfs_mnt)"
         mount_nfs "$nfs:$server:$filepath${options:+:$options}" "$mntdir"
         # lazy unmount during pre-pivot hook
-        inst_hook --hook pre-pivot --name 99url-lib-umount-nfs-"$(basename "$mntdir")" umount -l -- "$mntdir"
+        inst_hook --hook pre-pivot --name 99url-lib-umount-nfs-"${mntdir##*/}" umount -l -- "$mntdir"
     fi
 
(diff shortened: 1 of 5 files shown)


devkontrol (dismissed):

devkontrol:
needs rebase

bdrung:
Rebased

modules.d/45url-lib/url-lib.sh
```diff
@@ -160,7 +160,7 @@ nfs_fetch_url() {
         mntdir="$(mkuniqdir /run nfs_mnt)"
         mount_nfs "$nfs:$server:$filepath${options:+:$options}" "$mntdir"
         # lazy unmount during pre-pivot hook
-        inst_hook --hook pre-pivot --name 99url-lib-umount-nfs-"$(basename "$mntdir")" umount -l -- "$mntdir"
+        inst_hook --hook pre-pivot --name 99url-lib-umount-nfs-"${mntdir##*/}" umount -l -- "$mntdir"
```
devkontrol, line 163: Is this change safe ? 

url-lib.sh is not guaranteed to run in a bash context as far as I can tell.

bdrung: Yes. This works for all shells:

```
$ mntdir=/mnt/some/base
$ basename "$mntdir"
base
$ busybox sh -c 'mntdir=/mnt/some/base; echo "${mntdir##*/}"'
base
$ dash -c 'mntdir=/mnt/some/base; echo "${mntdir##*/}"'
base
```

devkontrol (approved):

PR #2435, opened 2026-05-08
fix: replace true by :

bdrung:
Calling the `true` command is not needed in most cases. Use the `:` no-operation command instead.

### Checklist


devkontrol (approved):

Conan-Kudo:
Ugh, this is ugly, but meh...

Conan-Kudo (approved):

PR #2440, opened 2026-05-11
fix(usrmount): filter out subvolid option for btrfs

puleglot:
Currently if rootfs mount options contain subvolid, then /usr subvolume will fail to mount with the following error:

BTRFS error (device dm-0): subvol '/usr' does not match subvolid 5

Fix this by filtering out subvolid from rootfs options.

Bug: https://bugs.gentoo.org/961622

### Checklist

### DIFF
--- a/modules.d/77usrmount/mount-usr.sh
+++ b/modules.d/77usrmount/mount-usr.sh
@@ -13,6 +13,7 @@ filtersubvol() {
     while [ $# -gt 0 ]; do
         case $1 in
             'subvol='*) : ;;
+            'subvolid='*) : ;;
             *) printf '%s' "${1}," ;;
         esac
         shift


dalto8:
From the linked bug report:
```
/dev/disk/by-label/root /sysroot/usr btrfs subvol=/usr,lazytime,nodev,noatime,compress=zstd,ssd,autodefrag,rw,relatime,lazytime,compress=zstd:3,ssd,discard=async,space_cache=v2,autodefrag,subvolid=5,ro 1 2
```

This looks invalid to me. Isn't it appropriate behavior for this to fail?

Saying you want to mount `subvolid=5` and `subvol=/usr` isn't correct, is it?

`subvolid=5` is always the root so it doesn't match `/usr`. How do you know which one is correct in this case?  Are you supposed to mount `subvolid=5` or `subvol=/usr`

puleglot:
Look at `mount_usr()` function. If `/` and `/usr` are subvolumes of the same fs, then it takes rootfs mount options and appends them to `/usr` mount options. That's why it filters out `subvol=` option and should also filter out `subvolid=`, so they don't override corresponding options needed for `/usr`.

devkontrol:
CC @Nowa-Ammerlaan @mwilck @aafeijoo-suse

devkontrol (approved):

bdrung (approved):

PR #2442, opened 2026-05-15
fix(dracut-functions.sh): do not attempt to install zero-size library files

### LINKED ISSUE #2441: dracut failed due to stray files like `/usr/lib64/libnss_myhostname.so.2;69c039c9`
**Describe the bug**
In https://discussion.fedoraproject.org/t/kernel-fails-to-update-automatically-to-7-0-x-due-to-dracut-error/191371 we got a report from a Fedora user that dracut failed with errors like `dracut-install: ERROR: cannot mmap '/var/tmp/dracut.dOlGFcB/initramfs/usr/lib64/libnss_myhostname.so.2;69c039c9': Invalid argument`.

**Distribution used**
Fedora 44

**Dracut version**
Likely 108

**Init system**
systemd

**To Reproduce**
It looks like - create a file with a name like `/usr/lib64/libnss_myhostname.so.2;69c039c9` and run dracut. Not sure if it has to be a 'real' library file

**Expected behavior**
Dracut should work, probably by ignoring the file

**Additional context**
We don't know what caused these files to exist, but it does seem like dracut should not be tripped up by the existence of this kind of orphaned temporary file or whatever it turns out to have been.

aafeijoo-suse:
It should fix #2441

### Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1342,6 +1342,8 @@ inst_libdir_file() {
         for _dir in $libdirs; do
             for _i in "$@"; do
                 for _f in "${dracutsysrootdir-}$_dir"/$_i; do
+                    # Filter temporary RPM files
+                    [[ $_f == *";"???????? ]] && continue
                     [[ ${_f#"${dracutsysrootdir-}"} =~ $_pattern ]] || continue
                     [[ -s $_f ]] && _files+=("${_f#"${dracutsysrootdir-}"}")
                 done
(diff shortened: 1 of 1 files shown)


bdrung (commented):
The change looks good to me. Can you extent the first commit with some more details/reasoning from the PR and linked bug?

dracut-functions.sh
```diff
@@ -1342,16 +1342,20 @@ inst_libdir_file() {
         for _dir in $libdirs; do
             for _i in "$@"; do
                 for _f in "${dracutsysrootdir-}$_dir"/$_i; do
+                    # Filter temporary RPM files
```
devkontrol, line 1345: Is this an issue only with RPM ? 

It is sad that dracut needs to have knowledge about the underlaying host package manager.

Is this file format/pattern for temp RPM files documented somewhere that we can reference here ?

devkontrol:
Could this be handled in a config file for RPM based distros with `--remove option` - as a compromise ? 

I am not strongly against the idea, but I do wonder if dracut is the best place to deal with this problem.

PR #2444, opened 2026-05-15
fix(dracut-rescue): use entry token instead of machine-id for path construction

### LINKED ISSUE #2443: 51-dracut-rescue.install builds broken paths with non-machine-id entry tokens
**Describe the bug**
`install.d/51-dracut-rescue.install` hardcodes `$MACHINE_ID` for path construction (lines 88, 90, 96). When `kernel-install` is configured with a non-machine-id entry token (e.g. `bootctl install --entry-token=os-id` → `fedora`), the paths diverge from what `kernel-install` actually creates.

Specifically, `BOOT_DIR` is set to `/${MACHINE_ID}/0-rescue` and `LOADER_ENTRY` to `…/${MACHINE_ID}-0-rescue.conf`, but the actual boot directory tree uses the entry token (e.g. `fedora/`). The suffix-strip `BOOT_ROOT=${BOOT_DIR_ABS%"$BOOT_DIR"}` also fails because `$BOOT_DIR_ABS` doesn't end with `/${MACHINE_ID}/…`.

`$KERNEL_INSTALL_ENTRY_TOKEN` has been exported by `kernel-install` since systemd 253 but is not used by this script.

**Distribution used**
Fedora 43

**Dracut version**
107 (also reproducible on current main)

**Init system**
systemd (systemd-boot + UKI layout)

**To Reproduce**
```bash
# /etc/kernel/install.conf:
layout=uki
initrd_generator=dracut
uki_generator=ukify

# Set entry token to os-id:
bootctl install --entry-token=os-id   # writes "fedora" to /etc/kernel/entry-token

# Then:
kernel-install add-all
# → 51-dracut-rescue.install fails building incorrect paths
```

**Expected behavior**
The script should use `$KERNEL_INSTALL_ENTRY_TOKEN` (falling back to `$MACHINE_ID`) for path construction, matching how `kernel-install` organizes the boot directory.

**Additional context**
Workaround: `ln -sf /dev/null /etc/kernel/install.d/51-dracut-rescue.install`

mxsb:
`51-dracut-rescue.install` hardcodes `$MACHINE_ID` for boot directory and loader entry paths. When `kernel-install` is configured with a non-machine-id entry token (e.g. via `bootctl --entry-token=os-id`), the paths diverge from the actual boot directory layout, breaking rescue image creation.

Use `$KERNEL_INSTALL_ENTRY_TOKEN` (exported by `kernel-install` since systemd 253) with a fallback to `$MACHINE_ID` for path construction. Rescue image filenames and BLS content fields keep using `$MACHINE_ID` as an identifier.

Fixes: https://github.com/dracut-ng/dracut/issues/2443

### Checklist

### DIFF
--- a/install.d/51-dracut-rescue.install
+++ b/install.d/51-dracut-rescue.install
@@ -66,6 +66,8 @@ if ! [[ $MACHINE_ID ]]; then
     exit 0
 fi
 
+ENTRY_TOKEN="${KERNEL_INSTALL_ENTRY_TOKEN:-${MACHINE_ID}}"
+
 if [ -n "$KERNEL_INSTALL_CONF_ROOT" ]; then
     if [ -f "$KERNEL_INSTALL_CONF_ROOT/cmdline" ]; then
         read -r -d '' -a BOOT_OPTIONS < "$KERNEL_INSTALL_CONF_ROOT/cmdline"
@@ -85,15 +87,15 @@ else
 fi
 
 if [[ -d ${BOOT_DIR_ABS%/*} ]]; then
-    BOOT_DIR="/${MACHINE_ID}/0-rescue"
+    BOOT_DIR="/${ENTRY_TOKEN}/0-rescue"
     BOOT_ROOT=${BOOT_DIR_ABS%"$BOOT_DIR"}
-    LOADER_ENTRY="$BOOT_ROOT/loader/entries/${MACHINE_ID}-0-rescue.conf"
+    LOADER_ENTRY="$BOOT_ROOT/loader/entries/${ENTRY_TOKEN}-0-rescue.conf"
     KERNEL="linux"
     INITRD="initrd"
 else
     BLS_DIR="/boot/loader/entries"
     BOOT_DIR_ABS="/boot"
-    LOADER_ENTRY="$BLS_DIR/${MACHINE_ID}-0-rescue.conf"
+    LOADER_ENTRY="$BLS_DIR/${ENTRY_TOKEN}-0-rescue.conf"
     KERNEL="vmlinuz-0-rescue-${MACHINE_ID}"
     INITRD="initramfs-0-rescue-${MACHINE_ID}.img"
 fi
@@ -150,8 +152,8 @@ case "$COMMAND" in
                 echo "initrd     $BOOT_DIR/initrd"
             } > "$LOADER_ENTRY"
         else
-            if [[ -e "${BLS_DIR}/${MACHINE_ID}-${KERNEL_VERSION}.conf" ]]; then
-                cp -aT "${BLS_DIR}/${MACHINE_ID}-${KERNEL_VERSION}.conf" "$LOADER_ENTRY"
+            if [[ -e "${BLS_DIR}/${ENTRY_TOKEN}-${KERNEL_VERSION}.conf" ]]; then
+                cp -aT "${BLS_DIR}/${ENTRY_TOKEN}-${KERNEL_VERSION}.conf" "$LOADER_ENTRY"
             else
                 cp -aT "${KERNEL_IMAGE%/*}/bls.conf" "$LOADER_ENTRY"
             fi
--- a/install.d/51-dracut-rescue.install
+++ b/install.d/51-dracut-rescue.install
@@ -66,6 +66,14 @@ if ! [[ $MACHINE_ID ]]; then
     exit 0
 fi
 
+# The entry token (systemd 253+) is used for the boot directory and loader
+# entry paths, since kernel-install derives those from it. The rescue image
+# *filenames* below deliberately keep $MACHINE_ID: it guarantees per-machine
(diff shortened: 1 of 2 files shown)


install.d/51-dracut-rescue.install
```diff
@@ -85,15 +87,15 @@ else
...
+    LOADER_ENTRY="$BOOT_ROOT/loader/entries/${ENTRY_TOKEN}-0-rescue.conf"
     KERNEL="linux"
     INITRD="initrd"
 else
     BLS_DIR="/boot/loader/entries"
     BOOT_DIR_ABS="/boot"
-    LOADER_ENTRY="$BLS_DIR/${MACHINE_ID}-0-rescue.conf"
+    LOADER_ENTRY="$BLS_DIR/${ENTRY_TOKEN}-0-rescue.conf"
     KERNEL="vmlinuz-0-rescue-${MACHINE_ID}"
     INITRD="initramfs-0-rescue-${MACHINE_ID}.img"
```
bdrung, line 100: What about `KERNEL` and `INITRD`?

bdrung (commented):
Can you rebase your changes instead of merging `main` into it?

devkontrol:
CC @aafeijoo-suse @Nowa-Ammerlaan @mrc0mmand

install.d/51-dracut-rescue.install
mxsb: Intentional — I only moved the *path/entry-lookup* bits (`BOOT_DIR`, `LOADER_ENTRY`, the source `.conf` lookup) to the entry token, and deliberately kept `KERNEL`/`INITRD` (and the `machine-id` field / the sed) on `$MACHINE_ID`:

- The entry token is what `kernel-install` uses to compute the loader entry name and the Type #1 directory (`$BOOT/$ENTRY_TOKEN/…`), so those *must* follow it or the rescue entry lands in the wrong place / the source conf isn't found — that's the actual bug.
- `KERNEL`/`INITRD` are just the basenames of the rescue image files this hook creates and references itself. Keeping `$MACHINE_ID` is the safer choice: entry tokens can be shared across parallel installs (`os-id`, `os-image-id`) — per `kernel-install(8)` those "would try to use the same entry name" — whereas the machine ID guarantees per-machine uniqueness. Switching would also orphan existing `vmlinuz-0-rescue-<machine-id>` files on upgrade.
- The `machine-id` line is a Boot Loader Spec field defined as the 32-char machine ID, so it must stay `$MACHINE_ID` regardless (an `os-id` token like `fedora` wouldn't be valid there).

Note this only concerns the flat `/boot` fallback branch; in the Type #1 layout the files are `linux`/`initrd` under the token dir, with no identifier in the name.

I've pushed a rebase (dropping the merge commits) and added a comment above `ENTRY_TOKEN=…` documenting this split so it doesn't come up again.

test/TEST-43-KERNEL-INSTALL/test.sh
```diff
@@ -62,6 +70,15 @@ test_setup() {
...
     fi
+
+    # test with a custom entry token different from machine-id
+    CUSTOM_TOKEN="custom-entry-token"
+    mkdir -p "$BOOT_ROOT/$CUSTOM_TOKEN/$KVERSION" "$BOOT_ROOT/loader/entries" "$BOOT_ROOT/$CUSTOM_TOKEN/0-rescue"
+    kernel-install add --entry-token "literal:$CUSTOM_TOKEN" "$KVERSION" "$KIMAGE"
+    if [[ ! -e "$BOOT_ROOT/$CUSTOM_TOKEN/0-rescue/initrd" ]]; then
+        echo "Error: kernel-install failed to create rescue initrd with custom entry token at $BOOT_ROOT/$CUSTOM_TOKEN/0-rescue/initrd" >&2
+        return 1
+    fi
```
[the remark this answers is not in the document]
mxsb: Good catch — added an assertion in `test_setup` that the rescue loader entry is created at the entry-token path:

```sh
if [[ ! -e "$BOOT_ROOT/loader/entries/$CUSTOM_TOKEN-0-rescue.conf" ]]; then
    echo "Error: rescue loader entry not created at entry-token path ..." >&2
    return 1
fi
```

This positively verifies the loader-entry path the fix changes, so a regression that writes the machine-id-based entry name (or the wrong nested path) is caught, independent of the direct `-initrd` boot in `test_run`.

mxsb:
**Test coverage (TEST-43)** — the custom-entry-token variant boots the rescue initrd from `<boot>/<entry-token>/0-rescue/initrd` and asserts the rescue loader entry is created at `loader/entries/<entry-token>-0-rescue.conf` (catches a regression that writes the machine-id-based path).

Not covered: the flat `/boot` fallback branch. Under `kernel-install` the entry-token directory always exists, so only the Type #1 layout is reachable here; that branch also hardcodes `/boot`, so it isn't test-isolatable in TEST-43.

PR #2447, opened 2026-05-15
fix(nvmf): parse json output from "nvme list-subsys"

stuarthayes:
Modify code that parses the output of "nvme list-subsys" to set up rd.nvmf.discover parameter to use the json output of nvme rather than the human-readable output.

A recent commit to nvme-cli changed the human-readable output to only print the transport address rather than the address line that includes all of the needed information, but the json output still contains the needed information.

### Checklist

### DIFF
--- a/modules.d/74nvmf/module-setup.sh
+++ b/modules.d/74nvmf/module-setup.sh
@@ -91,11 +91,7 @@ cmdline() {
     gen_nvmf_cmdline() {
         local _dev=$1
         local trtype
-        local traddr
-        local host_traddr
-        local trsvcid
-        local _address
-        local -a _address_parts
+        local discover
 
         [[ -L "/sys/dev/block/$_dev" ]] || return 0
         cd -P "/sys/dev/block/$_dev" || return 0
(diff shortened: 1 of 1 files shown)


devkontrol:
CC @mwilck

modules.d/74nvmf/module-setup.sh
```diff
@@ -111,20 +107,19 @@ cmdline() {
...
-        done
+        discover=$(nvme list-subsys "${PWD##*/}" -o json | jq -j '
+        if type == "array" then .[] else . end
+        | .Subsystems[]?
+        | .Paths[]?
+        | select(.Transport == "'"$trtype"'" and .Address)
+        | .Address as $a
+        | ($a | capture("traddr=(?<traddr>[^,]*)(,trsvcid=(?<trsvcid>[^,]*))?(,host_traddr=(?<host_traddr>[^,]*))?")) as $m
+        | if $m.traddr == null and $m.trsvcid == null and $m.host_traddr == null
+        then .
```
mwilck: This is advanced `jq` usage which may need some commenting. What exactly does the `.` in line 118 produce? Is it the empty string?

mwilck (commented):
This looks good in general, but it conflicts with #2249 which I'd like to resolve first.

modules.d/74nvmf/module-setup.sh
stuarthayes: Yes.  But a recent commit to nvme-cli (https://github.com/linux-nvme/nvme-cli/pull/3413/) changed the JSON output of "nvme list-subsys" in a way that will break this patch, so I will need to work on that...

stuarthayes:
@mwilck I modified the patch to also work with the recent changes to nvme-cli, but I realize this will conflict with dracut patches you are working on.  I can rework this when your patches are in (or feel free to pull this into your patches, if you like).

tbzatek:
FYI, I'll be breaking the `nvme nbft show -o json` output in nvme-cli-3.0. For dracut use it's going to be about flags, perhaps most other keys will remain untouched. The development version of nvme-cli is currently volatile in terms of the output format and will take time to settle down.

mwilck:
> FYI, I'll be breaking the `nvme nbft show -o json` output in nvme-cli-3.0.

It won't hurt as long as you just add fields. But if this is going to be volatile for some time to come, we should probably add a "version" field.

devkontrol:
I believe this PR can be closed now that https://github.com/dracut-ng/dracut/pull/2496 landed. 

Please reopen (and rebase) if still needed.

PR #2451, opened 2026-05-20
feat(busybox): run before 80base so applets replace host binaries

nadzyah:
Distros that ship large coreutils binaries (e.g. Ubuntu 25.10 with rust-coreutils) inflate the initrd because 80base copies the host's binaries like cp, ls, mv in full.

When the busybox module is included, 80base now skips any name busybox provides as an applet. 81busybox then symlinks those names to /usr/bin/busybox, restoring the smaller initrd footprint.

Bug-Ubuntu: https://bugs.launchpad.net/bugs/2150657

### Checklist


bdrung:
I would prefer a solution that would work without adding the `prefer_busybox` setting. So use busybox applets in case busybox is included.

I looked at `git log --follow --stat modules.d/81busybox/module-setup.sh` to see why busybox is included so late and found only:

Commit 4e78a8705e19562f5659e352612e50905c4be12a ("fix(busybox): install busybox symlinks later in the generation process"). After a search I found the corresponding https://github.com/dracut-ng/dracut/pull/536 PR which links to https://github.com/dracut-ng/dracut/issues/534. This bug is fixed by commit 95ba03270eb3297103cba276c2925ecc1b762926 but the renumbering is unrelated to it.

Commit 3975e26a84cd8645b15d0a863263a63da93650e0 ("feat(busybox): use busybox --install to install itself") says:

> Previous commits in this PR exposed a bug whereby some utilities (e.g. dd) had both a busybox applet version and a coreutils version.
>
> The busybox dracut module should not enforce taking always the busybox version. The `busybox --install` command is safer option that does not overwrite an existing binary of already exists.

So how about moving busybox from `81busybox` to `10busybox` and address https://github.com/dracut-ng/dracut/issues/2437 to ensure that the modules work with the busybox applets. If modules do not work with the busybox applets, they can do:
```
[ -L "$binary" ] && rm "$binary"
inst "$binary"
```

What do you think? @devkontrol @Conan-Kudo

devkontrol:
> I would prefer a solution that would work without adding the `prefer_busybox` setting. 

YES. THIS.

>  inflate the initrd because 80base copies the host's binaries like cp, ls, mv in full.

On Alpine when both coreutils and busybox are installed (as Alpine packages), I think this is not always the case today. Let's understand when exactly the "expected" initrd generation happening already. So there might be a way to trigger similar kind of code path with Ubuntu as well.

I am not against making additional fixes in dracut to make this scenario more accessible, but adding a new global flag is probably not the best option here.

modules.d/80base/module-setup.sh
```diff
@@ -15,12 +15,18 @@ depends() {
     return 0
 }
 
-# we prefer the non-busybox implementation of switch_root
-# due to the dependency, this dracut module needs to be ordered before the busybox dracut module
-# as this dracut module would install the non-busybox implementation of switch_root, if available
-
-# this dracut module needs to be ordered after the systemd-sysusers dracut module, so make sure
-# that the root password set in for the emergency console in host-only mode
```
bdrung, line 23: You don't change content in this sentence, do you? Could you change the line wrapping in a separate commit?

test/TEST-83-BUSYBOX-APPLETS/test.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+set -eu
+
+# shellcheck disable=SC2034
+TEST_DESCRIPTION="busybox applets win over host binaries when busybox module is included"
```
bdrung, line 6: This new test should be added to the CI.

Since busybox is not installed in the ubuntu container, we need a separate PR to add busybox to the container.

bdrung (commented):
The change looks good to me. I only have some minor remarks.

How long will this test take? Maybe it could be folded into an existing test in case the setup takes some time.

We should prefer busybox over dash in `shell-interpreter` (separate PR).

What kind of test have you done? Did you compare the initrd size on Ubuntu with this change?

modules.d/10busybox/module-setup.sh
```diff
@@ -7,9 +7,11 @@ check() {
     return 255
 }
 
-# we prefer the non-busybox implementation of switch_root
-# due to the dependency, the busybox dracut module needs to be order later than the base dracut module
-# as the base dracut module would install the non-busybox implementation of switch_root, if available
+# This module installs busybox and its applet symlinks first, before 80base.
```
devkontrol: Perhaps do not include the numbering for dracut modules in comments, for easy maintenance.

80base --> base

modules.d/80base/module-setup.sh
```diff
@@ -101,6 +107,10 @@ install() {
     inst_simple "$moddir/insmodpost.sh" /sbin/insmodpost.sh
 
     if ! dracut_module_included "systemd"; then
+        # Prefer the host's util-linux switch_root over the busybox applet
```
devkontrol: Can we have a test around this code block to check if busybox dracut module is included ?

test/TEST-83-BUSYBOX-APPLETS/test.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
```
devkontrol, line 1: Nit: I would drop the "-WIN" from the name of the test. Does not add much value and can be confused with Windows.

test/TEST-83-BUSYBOX-APPLETS-WIN/test.sh
```diff
@@ -0,0 +1,54 @@
...
+    return 0
+}
+
+test_run() {
+    set -x
+    local initrd="$TESTDIR/initramfs"
+
+    test_dracut \
+        --no-hostonly --no-kernel --drivers "" \
+        --modules "bash base busybox" \
```
devkontrol: Why would bash dracut module needed for this test ?

test/TEST-83-BUSYBOX-APPLETS-WIN/test.sh
```diff
@@ -0,0 +1,54 @@
...
+TEST_DESCRIPTION="busybox applets win over host binaries when busybox module is included"
+
+test_check() {
+    if ! command -v busybox &> /dev/null; then
+        echo "Test needs busybox on host... Skipping"
+        return 1
+    fi
+}
+
+test_setup() {
```
devkontrol: I think empty function can be removed .

modules.d/80base/module-setup.sh
nadzyah: I've updated the content of the sentence :)

test/TEST-83-BUSYBOX-APPLETS/test.sh
nadzyah: done

.github/workflows/integration.yml
```diff
@@ -98,6 +98,7 @@ jobs:
                     - "50"
                     - "80"
                     - "81"
+                    - "83"
```
devkontrol, line 101: In addition to running this test here, we should also run it as part of https://github.com/dracut-ng/dracut/blob/main/.github/workflows/daily-basic.yml please !

devkontrol:
Fedora failure for test 83 might be related to https://github.com/dracut-ng/dracut/issues/2454, please double-check (see also https://github.com/dracut-ng/dracut/pull/2455#issuecomment-4544020069).

challvy:
> Fedora failure for test 83 might be related to #2454, please double-check (see also [#2455 (comment)](https://github.com/dracut-ng/dracut/pull/2455#issuecomment-4544020069)).

Yes, I think so. @devkontrol Looking at TEST-83 in PR #2451, it checks cp/ls/mv/rm/mkdir/sleep/tr under `usr/bin/`. My current fix(#2455 #2454 ) only handles `sh` (the immediate kdump blocker) on Fedora 43.

Is it necessary to normalize all applet paths to `/usr/bin/`? Patch like this:

```
index 402e1fc2..b5767ea6 100755
@@ -30,9 +30,19 @@ install() {
         _path=$(find_binary "$_i")
         [ -z "$_path" ] && continue

+        # 1. Install at host's path (preserves topology, e.g. /sbin/ifconfig)
         # do not remove existing destination files
-        [ -e "${_dstdir}/$_path" ] && continue
-
-        ln_r /usr/bin/busybox "$_path"
+        [ -e "${_dstdir}/$_path" ] || ln_r /usr/bin/busybox "$_path"
+
+        # 2. Ensure canonical /usr/bin/<applet> always exists (POSIX layout).
+        # find_binary's literal path depends on $DRACUT_PATH ordering and may
+        # resolve to e.g. /usr/sbin/sh on Fedora 42+ where the busybox package
+        # places applet symlinks under /usr/sbin/. Standard scripts expect
+        # /usr/bin/<applet> (and #!/bin/sh -> /usr/bin/sh on merged-usr), so
+        # create the canonical symlink unconditionally if absent.
+        local _canonical="/usr/bin/${_path##*/}"
+        [ "$_path" = "$_canonical" ] && continue
+        [ -e "${_dstdir}${_canonical}" ] && continue
+        ln_r /usr/bin/busybox "$_canonical"
     done
 }
```

devkontrol:
> > Fedora failure for test 83 might be related to #2454, please double-check (see also [#2455 (comment)](https://github.com/dracut-ng/dracut/pull/2455#issuecomment-4544020069)).
> 
> Yes, I think so. 

One way to proceed is to exclude test 83 for Fedora in the GitHub Action and deal with it as a follow-up PR.

devkontrol (dismissed):

.github/workflows/daily-basic.yml
```diff
@@ -99,10 +100,19 @@ jobs:
                     - "80"
                     - "81"
                     - "82"
+                    - "83"
                 exclude:
                     # btrfs kernel module is not available on CentOS Stream 10
                     - container: centos:latest
                       test: "11"
+                    # busybox applet symlinks regress on Fedora/CentOS, see #2454
```
bdrung: That is not the reason for the failure. The symlinks are installed in `/usr/bin`. The test fails for me with:

```
FAIL: rm is not a busybox symlink in the initrd
```

but:

```
$ podman run -it --rm ghcr.io/dracut-ng/fedora:latest 
[root@166cbe98820e /]# busybox --list | grep ^rm$
rm
```

So please investigate why `rm` from busybox is missing.

bdrung: Nadzeya and I had a pair-debugging session and found the reason for the failure:
```
$ podman run -it --rm ghcr.io/dracut-ng/fedora:latest
[root@da60a3d3ce00 /]# grep rm /usr/lib/dracut/dracut.conf.d/01-dist.conf 
install_optional_items+=" vi /usr/libexec/vi /etc/virc ps grep cat rm "
```

modules.d/10busybox/module-setup.sh
```diff
@@ -0,0 +1,46 @@
...
+        [ -z "$_path" ] && continue
+
+        # An existing symlink (e.g. another applet alias) is left alone. A real
+        # file at this path comes from install_items, which dracut.sh processes
+        # before module install(). Replace it with the busybox applet so the
+        # module's "busybox wins" contract holds even when downstream configs
+        # pre-stage host binaries (e.g. Fedora's 01-dist.conf, see #2454).
+        # Modules that need the host binary must drop the symlink and reinstall
+        [ -L "${_dstdir}/$_path" ] && continue
+        [ -e "${_dstdir}/$_path" ] && rm -f "${_dstdir}/$_path"
```
bdrung, line 42: We should install the additional items from `install_optional_items` after we ran the modules. Then we would not need to do this workaround here. But let's do this change separately.

Conan-Kudo:
"host bloat" is not accurate for the issue. The issue is that you have two separate implementations of UNIX core utilities installed and you want to use different ones for initramfs and the real system. Whether your choices are correct or not are separate issues, but regardless, it isn't "bloat". Please don't misuse words here, because it will confuse people reviewing it later.

nadzyah:
> "host bloat" is not accurate for the issue. The issue is that you have two separate implementations of UNIX core utilities installed and you want to use different ones for initramfs and the real system. Whether your choices are correct or not are separate issues, but regardless, it isn't "bloat". Please don't misuse words here, because it will confuse people reviewing it later.

The wording in question was not introduced by me. The title and commit message were changed during review, and I kept that wording when amending the commit to address subsequent review comments.

I’m happy to change the terminology if “host bloat” is not considered accurate. However, the comment assumes I chose that wording intentionally, which wasn’t the case. A simple suggestion for alternative wording would have been sufficient.

devkontrol:
The term "bloat" carries a negative connotation. Let's try to avoid it.

> When prefer_busybox=yes and the busybox module is included

The PR no longer uses this flag, lets remove it from the PR description.

modules.d/80base/module-setup.sh
```diff
@@ -15,12 +15,18 @@ depends() {
...
 
-# we prefer the non-busybox implementation of switch_root
-# due to the dependency, this dracut module needs to be ordered before the busybox dracut module
-# as this dracut module would install the non-busybox implementation of switch_root, if available
-
-# this dracut module needs to be ordered after the systemd-sysusers dracut module, so make sure
-# that the root password set in for the emergency console in host-only mode
+# The busybox dracut module runs before this module and lays down its applet
+# symlinks. inst_multiple calls for those names then no-op because the
+# destinations already exist. That is intentional and saves the bloated host
```
bdrung: The wording here should be update as well to get rid of "bloated".

bdrung (changes requested):

modules.d/10busybox/module-setup.sh
bdrung: Done in https://github.com/dracut-ng/dracut/pull/2511

bdrung: https://github.com/dracut-ng/dracut/pull/2511 is controversial. So replacing existing binaries by symlinks would be controversial as well. So can we not do it in this PR?

bdrung: This means that the test needs to ignore `cat`, `rm`, etc that `dracut.conf.d/fedora/01-dist.conf` includes via `install_optional_items` on Fedora.

.github/workflows/daily-basic.yml
```diff
@@ -100,10 +101,14 @@ jobs:
...
                     - "81"
                     - "82"
+                    - "83"
                 exclude:
                     # btrfs kernel module is not available on CentOS Stream 10
                     - container: centos:latest
                       test: "11"
+                    # busybox is not installed in the azurelinux image
+                    - container: azurelinux:3.0
+                      test: "83"
```
bdrung, line 111: This exclusion can be dropped. We have busybox in azurelinux now: https://github.com/dracut-ng/dracut/commit/874c545f535bb689927a788a0e7a5d3bf283a63e

collinfunk:
Note that GNU coreutils has supported building as a single binary since 8.23, released 2014-07-18. If your only concern is size, I am not sure why you would need busybox. See:

```bash
$ { ./configure --enable-single-binary CFLAGS="$(rpm --eval '%{build_cflags}') -fno-strict-aliasing -fpic"; make -j 16; strip src/coreutils; } > /dev/null 2>&1
$ du --summarize --apparent-size src/coreutils /bin/busybox
1450	src/coreutils
1408	/bin/busybox
```

bdrung:
Discussion from https://github.com/dracut-ng/dracut/issues/2512: we currently cannot use blkid from busybox. The output is different and we use several arguments that are not supported by busybox.

devkontrol:
> Discussion from #2512: we currently cannot use blkid from busybox. The output is different and we use several arguments that are not supported by busybox.

It would be beneficial to have a busybox test case, independently from resolving https://bugs.launchpad.net/bugs/2150657.

Perhaps test case should be factored out into its own PR and for now the test case would check for `blkid` (or some other applet) coming from busybox (and not for the problematic cp, ls, mv binaried).

One of my concern is that distro's package `busybox` quite differently (different path, different applets, etc). This test case would allow us to have some experience with busybox in our CI.

bdrung:
I picked up this branch, rebased on main and tweaked it a little bit: https://github.com/dracut-ng/dracut/pull/2689

PR #2453, opened 2026-05-25
fix(base): use printf instead of echo for hook variable

fiftydinar:
… due to `echo` translation of `\x2f` character

Without this fix, boot was delayed around 5 minutes and a bunch of errors popped-up, like:

```
dracut-initqueue[349]: Warning: /var/lib/dracut/hooks/initqueue/finished/devexists-/dev/disk/by-uuid/734eed01-7ac0-4200-bdb7-2c91aa668d30.sh: ""
```
```
dracut-initqueue[2341]: cat: /var/lib/dracut/hooks/initqueue/finished/devexists-/dev/disk/by-uuid/734eed01-7ac0-4200-bdb7-2c91aa668d30.sh: No such file or directory
```

1.  `parse-root.sh` creates a "devexis ts" hook file to wait for the root device (/dev/disk/by-uuid/...). The / in the device path is escaped to \x2f in the filename via str_replace(), producing a file named devexis ts-\x2fdev\x2fdisk\x2fby-uuid\x2f....sh.

2. list_hooks() finds this file via glob ([ -f "$hook" ] succeeds), but then outputs it with echo "$hook".

3. Dash's echo interprets \x2f as a hex escape for / (0x2f = 47 = /), corrupting the path. The caller (check_finished(), source_hook()) receives a path with literal / instead of \x2f, which doesn't point to a real file.

4. check_finished() can never find the real hook → returns "not finished" → initqueue loops for 3.5 minutes until timeout.

### Checklist

### DIFF
--- a/modules.d/80base/dracut-lib.sh
+++ b/modules.d/80base/dracut-lib.sh
@@ -378,14 +378,14 @@ list_hooks() {
     # priority, '/etc/dracut/hooks' comes after and '/usr/lib/dracut/hooks' is the
     # least priviliged location.
     for hook in "/var/lib/dracut/hooks/$dir/"$pattern; do
-        [ -f "$hook" ] && echo "$hook"
+        [ -f "$hook" ] && printf '%s\n' "$hook"
     done
     for hook in "/etc/dracut/hooks/$dir/"$pattern; do
-        [ -f "$hook" ] && [ ! -f "/var/lib/dracut/hooks/$dir/${hook##*/}" ] && echo "$hook"
+        [ -f "$hook" ] && [ ! -f "/var/lib/dracut/hooks/$dir/${hook##*/}" ] && printf '%s\n' "$hook"
     done
     for hook in "/usr/lib/dracut/hooks/$dir/"$pattern; do
         [ -f "$hook" ] && [ ! -f "/var/lib/dracut/hooks/$dir/${hook##*/}" ] \
-            && [ ! -f "/etc/dracut/hooks/$dir/${hook##*/}" ] && echo "$hook"
+            && [ ! -f "/etc/dracut/hooks/$dir/${hook##*/}" ] && printf '%s\n' "$hook"
     done
 }
 


devkontrol:
This PR broke some CI tests, @fiftydinar please check.

CC @vittyvk

devkontrol:
Is this an issue just with a specific dash version ? - see https://github.com/dracut-ng/dracut/issues/2280

bdrung:
Could this be another case of a buggy dash (see https://github.com/dracut-ng/dracut/issues/2280)?

HastD (changes requested):
`printf` doesn't automatically include a final newline, so this should use `printf '%s\n' "$hook"` instead of `printf '%s' "$hook"`.

fiftydinar:
> Is this an issue just with a specific dash version ? - see https://github.com/dracut-ng/dracut/issues/2280

I use dash 0.5.13.4-1

> Could this be another case of a buggy dash (see https://github.com/dracut-ng/dracut/issues/2280)?

Or maybe it wasn't POSIX compliant from the beggining? From what I read, what `dash` is doing is POSIX and what `bash` as `sh` does here with `echo` is non-POSIX. But it will be good to verify further.

> `printf` doesn't automatically include a final newline, so this should use `printf '%s\n' "$hook"` instead of `printf '%s' "$hook"`.

Fixed.

fiftydinar:
> This PR broke some CI tests, @fiftydinar please check.
> 
> CC @vittyvk 

Looks like only 1 test fails now, but looks unrelated, about cpio module missing

HastD (approved):

fiftydinar:
> > This PR broke some CI tests, @fiftydinar please check.
> > CC @vittyvk
> 
> Looks like only 1 test fails now, but looks unrelated, about cpio module missing

Now all tests pass beside that commit message checker, which I don't know how to fix. You can squash and merge this with commit title that would you like if I'm asked.

devkontrol:
> Now all tests pass beside that commit message checker, which I don't know how to fix. You can squash and merge this with commit title that would you like if I'm asked.

Try this for the title:

`fix(base): for dash`

devkontrol:
> > Now all tests pass beside that commit message checker, which I don't know how to fix. You can squash and merge this with commit title that would you like if I'm asked.
> 
> Try this for the title:
> 
> `fix(base): for dash`

You would need to actually change the commit message not just the GitHub PR message.

You probably should also squash the commits into one commit.

fiftydinar:
Done! Squashed all 4 commits into one with message `fix(base): for dash` and proper body. Should pass the commit message check now.

bdrung:
The title `fix(base): for dash` does not carry much information. How about extending it to `fix(base): use printf instead of echo for hook variable` or something like this?

fiftydinar:
> The title `fix(base): for dash` does not carry much information. How about extending it to `fix(base): use printf instead of echo for hook variable` or something like this?

Done

ManuLinares:
Thanks, this fixed the issue https://github.com/dracut-ng/dracut/issues/2439

thesamesam (approved):

devkontrol (approved):

Nowa-Ammerlaan (approved):

bdrung (approved):

PR #2455, opened 2026-05-25
fix(busybox): ensure /usr/bin/sh symlink is always created #2454

### LINKED ISSUE #2454: fix(busybox): ensure /usr/bin/sh symlink is always created
**Describe the bug**

kdump initramfs with squashfs fails to boot with `Failed to execute /init (error -2)`. The squash loader's `init-squash.sh` (installed as `/init`) uses `#!/bin/sh` shebang, but `/usr/bin/sh` does not exist in the outer cpio layer — only `/usr/sbin/sh` is present.

```
...
2026-05-25 11:58:48 [    0.521438] Run  as init process
2026-05-25 11:58:48 [    0.521613] Failed to execute /init (error -2)
2026-05-25 11:58:48 [    0.521768] Run /sbin/init as init process
2026-05-25 11:58:48 can't run '/etc/init.d/rcS': No such file or directory
2026-05-25 11:58:48 can't open /dev/tty4: No such file or directory
2026-05-25 11:58:48 can't open /dev/tty2: No such file or directory
2026-05-25 11:58:48 can't open /dev/tty3: No such file or directory
2026-05-25 11:58:48 Please press Enter to activate this console.
```

The busybox module creates applet symlinks based on `find_binary()` results. Since `DRACUT_PATH` defaults to `/usr/sbin /sbin /usr/bin /bin`, and the Fedora busybox package places applet symlinks under `/usr/sbin/`, `find_binary("sh")` returns `/usr/sbin/sh`. The module only creates that symlink, leaving `/usr/bin/sh` (i.e. `/bin/sh` on merged-usr) absent.

**Distribution used**

Fedora 43 (also affects Fedora 42)

**Dracut version**

dracut-107-8.fc43

**Init system**

systemd

**To Reproduce**

1. Install Fedora 43 with `busybox` (1.37.0-3.fc43) and `kexec-tools`
2. Enable kdump with squashfs (kdump-utils adds `--add 'squash-squashfs' --squash-compressor 'zstd'`)
3. `kdumpctl rebuild`
4. `echo c > /proc/sysrq-trigger`
5. Crash kernel fails with `Failed to execute /init (error -2)`

**Expected behavior**

`/usr/bin/sh` should always exist when busybox module is installed (POSIX mandates `/bin/sh`). kdump should successfully capture vmcore.

**Additional context**

Triggered by Fedora 42's [Unify /usr/bin and /usr/sbin](https://discussion.fedoraproject.org/t/f42-change-proposal-unify-usr-bin-and-usr-sbin-system-wide/99853) transition — `/usr/sbin` still exists as a separate directory and busybox places applet symlinks there.

Diagnosis:
```
$ lsinitrd /boot/initramfs-$(uname -r)kdump.img | grep -E "usr/bin/sh|usr/sbin/sh"
lrwxrwxrwx   1 root     root           14 ... usr/sbin/sh -> ../bin/busybox
# /usr/bin/sh is ABSENT
```

Workaround:
```
echo 'DRACUT_PATH="/usr/bin /bin /usr/sbin /sbin"' > /etc/dracut.conf.d/fix-path-order.conf
kdumpctl rebuild
```

My Patch:
```
index abcdef1..1234567 100755
@@ -36,4 +36,11 @@ install() {
 
         ln_r /usr/bin/busybox "$_path"
     done
+
+    # Ensure /usr/bin/sh is always available.
+    # POSIX mandates /bin/sh and most scripts use #!/bin/sh (-> /usr/bin/sh
+    # on merged-usr). find_binary may locate 'sh' at a non-standard path
+    # (e.g. /usr/sbin/sh) depending on $DRACUT_PATH ordering, so create
+    # the symlink explicitly if absent.
+    [[ -e "${_dstdir}/usr/bin/sh" ]] || ln_r /usr/bin/busybox /usr/bin/sh
 }
```

challvy:
POSIX mandates that /bin/sh must exist. On merged-usr systems this resolves to /usr/bin/sh. The busybox module creates applet symlinks at whatever path find_binary() returns, which depends on the DRACUT_PATH ordering (default: /usr/sbin /sbin /usr/bin /bin).

On Fedora 42+, the busybox package (1.37.0+) places applet symlinks under /usr/sbin/ as part of the /usr/sbin → /usr/bin unification transition. Since DRACUT_PATH searches /usr/sbin first, find_binary("sh") returns /usr/sbin/sh and the module only creates that symlink — leaving /usr/bin/sh (and thus /bin/sh) absent.

This breaks any script using #!/bin/sh as its shebang — notably squash-lib's init-squash.sh which serves as PID 1 in the squash loader. The kernel fails with "Failed to execute /init (error -2)".

Fix this by unconditionally ensuring /usr/bin/sh exists after the applet symlink loop. If find_binary already placed sh at /usr/bin/sh, the check is a no-op.

This pull request ensures /usr/bin/sh symlink is always created by the busybox module, fixing kdump squashfs boot failure on Fedora 42/43 where busybox places sh at /usr/sbin/sh due to DRACUT_PATH ordering.

Fixes: https://github.com/dracut-ng/dracut/issues/2454

### Checklist

### DIFF
--- a/modules.d/81busybox/module-setup.sh
+++ b/modules.d/81busybox/module-setup.sh
@@ -35,4 +35,10 @@ install() {
 
         ln_r /usr/bin/busybox "$_path"
     done
+    # Ensure /usr/bin/sh is always available.
+    # POSIX mandates /bin/sh and most scripts use #!/bin/sh (-> /usr/bin/sh
+    # on merged-usr). find_binary may locate 'sh' at a non-standard path
+    # (e.g. /usr/sbin/sh) depending on $DRACUT_PATH ordering, so create
+    # the symlink explicitly if absent.
+    [[ -e "${_dstdir}/usr/bin/sh" ]] || ln_r /usr/bin/busybox /usr/bin/sh
 }


challvy:
I also created a bugreport in https://bugzilla.redhat.com/show_bug.cgi?id=2481179

devkontrol:
DRACUT_PATH has been removed - https://github.com/dracut-ng/dracut/pull/1468 .

Fedora is slow to adapt latest dracut version, so Fedora contributors should make sure they consider upstream changes that are not yet in Fedora.

CC @pvalena  @prudo1

challvy:
> DRACUT_PATH has been removed - #1468 .

@devkontrol I just tested this patch, but this change alone does not resolve the issue. The symptoms remain the same.

devkontrol:
> > DRACUT_PATH has been removed - #1468 .
> 
> @devkontrol I just tested this patch, but this change alone does not resolve the issue. The symptoms remain the same.

Indeed. I did not mean to imply that this issue is fixed, but it makes this PR inconsistent with rest of dracut. As an example, the following line needs to change

```
(e.g. /usr/sbin/sh) depending on $DRACUT_PATH ordering, so create
```

modules.d/81busybox/module-setup.sh
```diff
@@ -35,4 +35,10 @@ install() {
 
         ln_r /usr/bin/busybox "$_path"
     done
+    # Ensure /usr/bin/sh is always available.
+    # POSIX mandates /bin/sh and most scripts use #!/bin/sh (-> /usr/bin/sh
+    # on merged-usr). find_binary may locate 'sh' at a non-standard path
+    # (e.g. /usr/sbin/sh) depending on $DRACUT_PATH ordering, so create
+    # the symlink explicitly if absent.
+    [[ -e "${_dstdir}/usr/bin/sh" ]] || ln_r /usr/bin/busybox /usr/bin/sh
```
devkontrol, line 43: How about

```
[[ -L $initdir/bin/sh ]] || ln -sf busybox "${initdir}/bin/sh"
```

..to follow the established pattern for `bash` and `dash` module.

This would be less opinionated and easier to read and more consistent with rest of dracut ?

bdrung:
The directory structure looks like this on Fedora:

```
$ ls -l /usr/sbin /sbin /usr/bin /bin -d
lrwxrwxrwx 1 root root     7 Jan 16 00:00 /bin -> usr/bin
lrwxrwxrwx 1 root root     8 Jan 16 00:00 /sbin -> usr/sbin
dr-xr-xr-x 1 root root 20480 May 22 13:53 /usr/bin
lrwxrwxrwx 1 root root     3 May 22 06:47 /usr/sbin -> bin
```

The initrd should have the same structure. Could you check that?

`ln_r /usr/bin/busybox "$_path"` should end up creating the symlink inside `/usr/bin` even when path is `/usr/sbin/sh`.

challvy:
@bdrung usr/sbin is different?
```
# lsinitrd /boot/initramfs-$(uname -r)kdump.img | grep -E "^[ld].*\s(bin|sbin|usr/bin|usr/sbin)$"
lrwxrwxrwx   1 root     root            7 May 25 22:21 bin -> usr/bin
lrwxrwxrwx   1 root     root            8 May 25 22:21 sbin -> usr/sbin
drwxr-xr-x   2 root     root            0 May 25 22:21 usr/bin
drwxr-xr-x   2 root     root            0 May 25 22:21 usr/sbin
lrwxrwxrwx root/root                 8 2026-05-25 22:23 squashfs-root/sbin -> usr/sbin
lrwxrwxrwx root/root                 3 2026-05-25 22:23 squashfs-root/usr/sbin -> bin
...

**Comment by bdrung** [comment]
Then we found the issue. usr/bin should be a symlink to usr/sbin in the initrd.

File: modules.d/81busybox/module-setup.sh (hunk 1 above)
Comment (challvy) [reply modules.d/81busybox/module-setup.sh:43]: @devkontrol Thanks for the suggestion. However, using $initdir directly doesn't work in the squash case:The busybox module is invoked twice during a squash build:

- First during the main module install loop (no dstdir override) → installs into $initdir (becomes inner squashfs content)
- Then again from squash_install() with dstdir=$squashdir → installs into $squashdir (becomes outer loader content)

If we use $initdir/bin/sh, the second invocation sees the symlink already created by the first invocation and skips, leaving the outer loader without /bin/sh — which is exactly the bug we're trying to fix.

The bash/dash modules don't hit this because squash-lib only depends on busybox; if they were also pulled into the loader, they'd have the same latent issue.Adopting your style preference but keeping _dstdir:
> [[ -L "${_dstdir}/bin/sh" ]] || ln -sf busybox "${_dstdir}/bin/sh"
>
Does this look acceptable?

**Comment by challvy** [comment]
> Then we found the issue. usr/bin should be a symlink to usr/sbin in the initrd.

Do you mean "usr/sbin should be a symlink to bin in the initrd"?

https://discussion.fedoraproject.org/t/f42-change-proposal-unify-usr-bin-and-usr-sbin-system-wide

**Comment by challvy** [comment]
@devkontrol @bdrung 
As a downstream user of Fedora(42~43), can the temporary workaround be like this?
```
echo 'DRACUT_PATH="/usr/bin /bin /usr/sbin /sbin"' > /etc/kdump/dracut.conf.d/fix-path-order.conf
```

devkontrol:
> @devkontrol @bdrung As a downstream user of Fedora(42~43), can the temporary workaround be like this?
> 
> ```
> echo 'DRACUT_PATH="/usr/bin /bin /usr/sbin /sbin"' > /etc/kdump/dracut.conf.d/fix-path-order.conf
> ```

This is the wrong repository for the Fedora (42~43) discussion. A better repository would be the Fedora packaging repository. I would however want a more permanent fix, if I am the Fedora maintainer (which I am not).

As I mentioned earlier, this workaround does not work for the upstream dracut version (for any distribution).

bdrung:
Yes. The issue is that your system has the `/usr/sbin` -> `bin` symlink, but the initrd does not. We need to figure out how that could happen to fix it properly.

devkontrol:
@prabhakarpujeri - seems you're on Fedora. Perhaps you can try to help with this PR/issue as well. Thank you !

devkontrol:
https://github.com/dracut-ng/dracut/pull/2603 seems like a better direction, we should consider closing this PR.

challvy:
> #2603 seems like a better direction, we should consider closing this PR.

OK. Thanks!

PR #2461, opened 2026-05-31
fix(systemd-*): add new dlopen dependencies to modules lists

bluca:
systemd v261 moved more dependencies to dlopen, including libssl which is required by various components, like resolved

```
[    2.857336] systemd-resolved[311]: Shared library 'libssl.so.3' is not available: libssl.so.3: cannot open shared object file: No such file or directory
[    2.861594] systemd-resolved[311]: Could not create manager: Operation not supported
[    2.869073] systemd[1]: systemd-resolved.service: Main process exited, code=exited, status=1/FAILURE
```

### Checklist

### DIFF
--- a/modules.d/11systemd-creds/module-setup.sh
+++ b/modules.d/11systemd-creds/module-setup.sh
@@ -27,6 +27,11 @@ depends() {
 
 }
 
+# called by dracut
+config() {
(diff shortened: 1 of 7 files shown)


chewi (approved):
I trust you to know exactly where these are needed. 🙂

devkontrol (approved):

Conan-Kudo (approved):

PR #2462, opened 2026-05-31
fix(systemd-networkd): get DHCP options values from networkctl

bluca:
The internal lease is no longer saved to the private /run/ directory of networkd since v261, so it cannot be read from there. It was always clearly marked as private anyway.
Use the new 'dhcp-lease' networkctl verb to print it instead.

This fixes the various network boot tests in debci (60-nfs, 70-iscsi, 72-nbd).

### Checklist

### DIFF
--- a/modules.d/11systemd-networkd/module-setup.sh
+++ b/modules.d/11systemd-networkd/module-setup.sh
@@ -58,7 +58,7 @@ install() {
         "$systemdsystemunitdir"/systemd-networkd-wait-online.service \
         "$systemdsystemunitdir"/systemd-networkd-wait-online@.service \
         "$systemdsystemunitdir"/systemd-network-generator.service \
-        ip sed grep
+        ip sed grep networkctl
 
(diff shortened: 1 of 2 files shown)


devkontrol:
CC @BtbN

devkontrol (approved):

BtbN:
I think you also need to add a dependency to dbus for networkctl to work?
iirc all the various ctl tools communicate with their daemons via dbus exclusively.

bluca:
> I think you also need to add a dependency to dbus for networkctl to work? iirc all the various ctl tools communicate with their daemons via dbus exclusively.

No, this method uses a varlink interface, so that's not necessary

bluca:
We are getting closer to release systemd v261, and dracut won't work without these changes plus https://github.com/dracut-ng/dracut/pull/2461 - would be great to get them sorted, and a new release tagged, if possible, please. Thanks!

Conan-Kudo (approved):

PR #2465, opened 2026-06-05
ci: update to systemd v257 and Azure Linux v4

devkontrol:
Azure Linux v3.0 is quite dated and a door has opened to upgrade to systemd v257 and Azure Linux v4.

### Checklist

### DIFF
--- a/.github/workflows/container-extra.yml
+++ b/.github/workflows/container-extra.yml
@@ -40,7 +40,7 @@ jobs:
                     - {platform: 'linux/amd64', runner: 'ubuntu-24.04', tag: 'amd'}
                     - {platform: 'linux/arm64', runner: 'ubuntu-24.04-arm', tag: 'arm'}
                 config:
-                    - {dockerfile: 'Dockerfile-azurelinux', tag: 'azurelinux:3.0', registry: 'mcr.microsoft.com'}
+                    - {dockerfile: 'Dockerfile-azurelinux', tag: 'azurelinux:latest', registry: 'mcr.microsoft.com'}
                     - {dockerfile: 'Dockerfile-fedora', tag: 'centos:latest', registry: 'quay.io/centos'}
                     - {dockerfile: 'Dockerfile-debian', tag: 'debian:sid'}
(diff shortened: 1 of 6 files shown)


devkontrol:
CC @ddstreetmicrosoft @reubeno @christopherco

christopherco (approved):
Thanks @devkontrol for driving this! The Azure Linux 4.0 update to the CI and docker image build looks good, and the base container pointing to `mcr.microsoft.com/azurelinux-beta/base/core:4.0` is correct while we're in Beta / Preview.

I'm also seeing the CI failures on `azurelinux:latest`. From what I can tell, these seem like they might be expected since the workflow to create the new `azurelinux:latest` GHCR tag won't run naturally until this is merged.

Depending on how you'd want to approach this and keep CI green, we could make this a two-phased approach:
1. First PR to keep `azurelinux:3.0` in the test matrices for now, merge the Dockerfile & container-extra.yml changes
2. After the new container builds and is tagged, follow up PR to switch tests to `azurelinux:latest`

Happy to help out on this with additional PRs if desired.

devkontrol:
> I'm also seeing the CI failures on `azurelinux:latest`. From what I can tell, these seem like they might be expected since the workflow to create the new `azurelinux:latest` GHCR tag won't run naturally until this is merged.
> 
> Depending on how you'd want to approach this and keep CI green, we could make this a two-phased approach:

We have this situation quite often, and we usually opt for minimizing the number of PR, instead of keeping the CI green.

The project has CI for each PR and CI that runs daily - "only" the Daily Ci is impacted here and as you explained well, it is only transitional and should recover.

It usually takes up to 4-10 weeks to get reviews on PRs.

devkontrol:
My expectation is that this PR will also resolve the flakiness seen on the CI with the old version of systemd

```
[    6.006953] systemd[1]: Assertion 'close_nointr(fd) != -EBADF' failed at src/basic/fd-util.c:75, function safe_close(). Aborting.
[    6.011246] systemd[1]: Caught <ABRT>, from our own process.
[    6.012843] systemd[1]: Caught <ABRT>, core dump failed (child 704, code=killed, status=6/ABRT).
```

PR #2469, opened 2026-06-10
Fixes for various dhcp-related issues

pvalena:
Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2026-6893

and similar discovered issues.

If POSIX / dash compatibility is to be preserved, `printf %q` needs to be replaced -- but since it's used in `modules.d/35network-manager/nm-run.sh` already, I decided to keep it this way, until some project-wide decision is made. 

I created only one single PR for simplicity; I can split it into multiple MRs if desired.

I've removed `network-legacy`, since it has very little use at this point, but if we want to preserve it, I already have the patches for it.

Please let me know what you think.

### Checklist

### DIFF
--- a/modules.d/11systemd-networkd/networkd-run.sh
+++ b/modules.d/11systemd-networkd/networkd-run.sh
@@ -12,10 +12,12 @@ for ifpath in /sys/class/net/*; do
         leases_file="/run/systemd/netif/leases/$(cat "$ifpath"/ifindex)"
         dhcpopts_file="/tmp/dhclient.${ifname}.dhcpopts"
         if [ -r "$leases_file" ]; then
-            grep -E "^(NEXT_SERVER|ROOT_PATH)=" "$leases_file" \
-                | sed -e "s/NEXT_SERVER=/new_next_server='/" \
-                    -e "s/ROOT_PATH=/new_root_path='/" \
-                    -e "s/$/'/" > "$dhcpopts_file" || :
+            while IFS='=' read -r key val; do
+                case "$key" in
+                    NEXT_SERVER) printf 'new_next_server=%q\n' "$val" ;;
+                    ROOT_PATH) printf 'new_root_path=%q\n' "$val" ;;
+                esac
+            done < "$leases_file" > "$dhcpopts_file"
         else
             # Since systemd v261 the DHCP lease is no longer serialized to /run/, use the new networkctl command.
             lease=$(networkctl --no-pager --no-legend --full dhcp-lease "$ifname" 2> /dev/null) || lease=""
--- a/modules.d/35network-legacy/dhclient-script.sh
+++ b/modules.d/35network-legacy/dhclient-script.sh
@@ -48,22 +48,24 @@ setup_interface() {
     if [ -n "$gw" ]; then
         if [ "$mask" = "255.255.255.255" ]; then
             # point-to-point connection => set explicit route to gateway
-            echo ip route add "$gw" dev "$netif" > /tmp/net."$netif".gw
+            printf 'ip route add %q dev %q\n' "$gw" "$netif" > /tmp/net."$netif".gw
         fi
 
         echo "$gw" | {
             IFS=' ' read -r main_gw other_gw
-            echo ip route replace default via "$main_gw" dev "$netif" >> /tmp/net."$netif".gw
+            printf 'ip route replace default via %q dev %q\n' "$main_gw" "$netif" >> /tmp/net."$netif".gw
             if [ -n "$other_gw" ]; then
                 for g in $other_gw; do
-                    echo ip route add default via "$g" dev "$netif" >> /tmp/net."$netif".gw
+                    printf 'ip route add default via %q dev %q\n' "$g" "$netif" >> /tmp/net."$netif".gw
                 done
             fi
         }
     fi
 
     if getargbool 1 rd.peerdns; then
-        [ -n "${search}${domain}" ] && echo "search $search $domain" > /tmp/net."$netif".resolv.conf
+        if [ -n "${search}${domain}" ]; then
+            echo "search $search $domain" > /tmp/net."$netif".resolv.conf
+        fi
         if [ -n "$namesrv" ]; then
             for s in $namesrv; do
                 echo nameserver "$s"
@@ -72,7 +74,10 @@ setup_interface() {
     fi
     # Note: hostname can be fqdn OR short hostname, so chop off any
     # trailing domain name and explicitly add any domain if set.
-    [ -n "$hostname" ] && echo "echo ${hostname%."$domain"}${domain:+.$domain} > /proc/sys/kernel/hostname" > /tmp/net."$netif".hostname
+    if [ -n "$hostname" ]; then
+        safe_hostname=$(printf '%s' "${hostname%."$domain"}${domain:+.$domain}")
+        printf 'echo %q > /proc/sys/kernel/hostname\n' "$safe_hostname" > /tmp/net."$netif".hostname
+    fi
 }
 
 setup_interface6() {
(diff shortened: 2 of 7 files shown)


aafeijoo-suse:
> Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2026-6893

Can we have a proper fix for this, instead of just removing the `network-legacy` module? I don't know RH, but we cannot just remove it from old maintained versions. If you want to remove it from upstream, do it in a separate PR, but this is not a valid fix.


> fix(systemd-networkd): escape DHCP lease values in dhcpopts generation
> 
> networkd-run.sh converts DHCP lease values (ROOT_PATH, NEXT_SERVER)
> into shell variable assignments written to /tmp/dhclient.<ifname>.dhcpopts,
> which is later sourced by netroot.sh, net-lib.sh, and nfs-lib.sh. The
> previous sed-based pipeline wrapped values in single quotes without
> escaping embedded single quotes, allowing a rogue DHCP server to inject
> arbitrary shell commands via a crafted ROOT_PATH or NEXT_SERVER value.

You should have told Claude that systemd-networkd filters malicious DHCP values, so they never reach the values parsed from lease files. Anyway, this was already commented on May 18th opening a GSA.

pvalena:
> > Resolves: https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2026-6893
> 
> Can we have a proper fix for this, instead of just removing the `network-legacy` module? I don't know RH, but we cannot just remove it from old maintained versions. If you want to remove it from upstream, do it in a separate PR, but this is not a valid fix.

Sure, let me amend this MR (as mentioned in the description).

> You should have told Claude that systemd-networkd filters malicious DHCP values, so they never reach the values parsed from lease files. Anyway, this was already commented on May 18th opening a GSA.

Well, It wasn't used to implement / evaluate the actual fixes, so the blame lies on me. I did not know about it, but most importantly, since the fix is not introducing additional complexity, I went with it as a standartization / hardening instead of relying on some outside implementation.

devkontrol:
> If POSIX / dash compatibility is to be preserved, printf %q needs to be replaced -- but since it's used in modules.d/35network-manager/nm-run.sh already, I decided to keep it this way, until some project-wide decision is made.

We have a functioning dependency management on a per dracut module basis that this project adhere to for 10+ years. This PR would leave a lot of distributions, users and use-cases behind and breaks the CI.

`network-manager` dracut module depends on `bash`, so bashism is encouraged in the `network-manager` dracut module.

pvalena:
> > If POSIX / dash compatibility is to be preserved, printf %q needs to be replaced -- but since it's used in modules.d/35network-manager/nm-run.sh already, I decided to keep it this way, until some project-wide decision is made.
> 
> We have a functioning dependency management on a per dracut module basis that this project adhere to for 10+ years. This PR would leave a lot of distributions, users and use-cases behind and breaks the CI.

Does that mean we want to have `escape() { printf "'%s'\\n" "$(printf '%s' "$1" | sed -e "s/'/'\\\\''/g")"; }` (advised by shellcheck) or some equivalent in the dracut-lib and use that instead?

@aafeijoo-suse @jozzsi @Conan-Kudo WDYT?

aafeijoo-suse:
> Does that mean we want to have `escape() { printf "'%s'\\n" "$(printf '%s' "$1" | sed -e "s/'/'\\\\''/g")"; }` (advised by shellcheck) or some equivalent in the dracut-lib and use that instead?

At least for `network-legacy` makes sense. dracut-lib is POSIX, so I'd have this new `escape()` function using only a sed (personal opinion). But, the code you (or shellcheck) suggest is wrong, that only escapes single quotes, and you have to escape everything (search what `printf '%q'` does).

pvalena:
> > Does that mean we want to have `escape() { printf "'%s'\\n" "$(printf '%s' "$1" | sed -e "s/'/'\\\\''/g")"; }` (advised by shellcheck) or some equivalent in the dracut-lib and use that instead?
> 
> At least for `network-legacy` makes sense. dracut-lib is POSIX, so I'd have this new `escape()` function using only a sed (personal opinion). But, the code you (or shellcheck) suggest is wrong, that only escapes single quotes, and you have to escape everything (search what `printf '%q'` does).

Yes, that's why I opted for `printf %q` instead of implementing it :). 

Can you create [in a separate PR] it in dracut-lib.sh and I'll just use it? I just copied it from shellcheck page.

devkontrol:
Can this be closed now  @aafeijoo-suse @pvalena @bdrung ?

bdrung:
We have following commits in main now:
* 11577739221ff38c1fd29abbba51a6c797376ed6
* 584e95bea90cedd12d9190e9b09bb45e1aa26b08
* 207c339728eff81127469c2f6fe106447c781009
* 9c7459bde0e78dd2e1e14d35c5351ef627b1b1fc
* 4d2d812bed056ca2c671612d960911b075c99dd9
* b456aee2fc61c47db0d9a408634820969e7359df
* 589bef03b7acb26124b32011d3e39552490ebca3

bdrung:
Comparing those commits, we haven't updates for these files yet:
* modules.d/74iscsi/iscsiroot.sh
* modules.d/74iscsi/parse-iscsiroot.sh
* modules.d/77initqueue/initqueue.sh

You can use the `escape` function now. Using `printf "%q"` produces a different output than `escape` but both solutions should be safe.

bdrung:
iscsi has been addressed by https://github.com/dracut-ng/dracut/pull/2550

devkontrol:
@bdrung @pvalena @aafeijoo-suse  - do we need one more PR for `initqueue/initqueue.sh` ?

aafeijoo-suse:
> @bdrung @pvalena @aafeijoo-suse - do we need one more PR for `initqueue/initqueue.sh` ?

I don't think so. `/sbin/initqueue` writes a command line into a `.sh`, and that has to be executed. If we escape it, initqueue scripts would fail to be interpreted.

devkontrol:
> `/sbin/initqueue` writes a command line into a `.sh`, and that has to be executed. If we escape it, initqueue scripts would fail to be interpreted.

Fedora seem to have this change now - https://src.fedoraproject.org/rpms/dracut/blob/rawhide/f/0010-fix-base-escape-arguments-in-initqueue-hook-script-g.patch 

Let's close this PR than here, please reopen if needed.

aafeijoo-suse:
> Fedora seem to have this change now - https://src.fedoraproject.org/rpms/dracut/blob/rawhide/f/0010-fix-base-escape-arguments-in-initqueue-hook-script-g.patch

I've digging into this a bit, and asked Claude about this patch to check if I was wrong or not, an the answer is:

```
  1. Resume from hibernation stops working.

  upstream (bash):   echo 8:2 > /sys/power/resume
  fedora   (bash):   echo 8:2 \> /sys/power/resume

  Sourced, that prints 8:2 > /sys/power/resume to stdout and writes nothing to /sys/power/resume.

  2. root=liveiso: breaks.

  upstream: /sbin/dmsquash-live-root `/sbin/losetup -f --show /dev/sda1`
  fedora:   /sbin/dmsquash-live-root \`/sbin/losetup\ -f\ --show\ /dev/sda1\`

  The deferred losetup is the whole point of dmsquash-liveiso-genrules.sh:6.

  Fedora's patch is closing a genuine hole (parse-iscsiroot.sh:149 passes DHCP-derived $netroot unquoted today), so the motivation is sound;
  it just took collateral damage on hibernation and liveiso
```

Of course, not everything is true. The path Claude comments is without systemd, so Fedora is not affected.

But, this patch also detected a couple of problems in parse-iscsiroot.sh, which is why the Claude that wrote it decided to escape initqueue.sh. I've prepared an alternative patch for this: https://github.com/dracut-ng/dracut/pull/2581

PR #2470, opened 2026-06-12
fix(network-legacy): sanitize DHCP values in dhclient-script.sh

aafeijoo-suse:
A rogue DHCP server can inject shell commands through `hostname`, `domain`, `gateway`, and classless static route values. These values are written to `/tmp` files that `setup_net()` later sources as shell.

- Sanitize hostname to RFC 952 chars (`a-zA-Z0-9.-`)
- Strip control chars from `domain` (like `search` already does)
- Validate gateway IPs contain only digits and dots
- Single-quote interpolated values in generated shell commands
- Validate all `parse_option_121` octets are numeric upfront; the existing inline checks only validated `mask` and the first destination octet, leaving dest octets 2-4 and all four gateway octets unchecked — those flow into shell commands written to the `initqueue/setup_net_"$netif".sh` hook script via `echo`

Fixes: CVE-2026-6893
Co-developed-by: Claude Opus 4.6 <noreply@anthropic.com>

### Checklist

### DIFF
--- a/modules.d/35network-legacy/dhclient-script.sh
+++ b/modules.d/35network-legacy/dhclient-script.sh
@@ -14,11 +14,11 @@ setup_interface() {
     mask=$new_subnet_mask
     bcast=$new_broadcast_address
     gw=${new_routers%%,*}
-    domain=$new_domain_name
     # get rid of control chars
+    domain=$(printf -- "%s" "$new_domain_name" | tr -d '[:cntrl:]')
     search=$(printf -- "%s" "$new_domain_search" | tr -d '[:cntrl:]')
     namesrv=$new_domain_name_servers
-    hostname=$new_host_name
+    hostname=$(printf '%s' "$new_host_name" | tr -d -c 'a-zA-Z0-9.-')
     [ -n "$new_dhcp_lease_time" ] && lease_time=$new_dhcp_lease_time
     [ -n "$new_max_life" ] && lease_time=$new_max_life
     preferred_lft=$lease_time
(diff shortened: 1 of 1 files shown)


aafeijoo-suse:
Unrelated CI failure of TEST-72-NBD in openSUSE: https://github.com/dracut-ng/dracut/issues/2381#issuecomment-4691513677

devkontrol (dismissed):
Thank you @aafeijoo-suse

Conan-Kudo:
What should we do about test 72 failing?

devkontrol:
> What should we do about test 72 failing?

As @aafeijoo-suse mentioned, this failure is unrelated to this PR. 

It adds to the cognitive load for PR reviews to require this side-discussions on the CI unrelated to PR - this is why I think it is important to keep the CI green at all times so that side-discussion are not needed or part of a separate PR. 

If fix is not available, than temporary disable to the test run with another commit.

modules.d/35network-legacy/dhclient-script.sh
```diff
@@ -46,20 +46,30 @@ setup_interface() {
         ${preferred_lft:+preferred_lft ${preferred_lft}}
 
     if [ -n "$gw" ]; then
-        if [ "$mask" = "255.255.255.255" ]; then
-            # point-to-point connection => set explicit route to gateway
-            echo ip route add "$gw" dev "$netif" > /tmp/net."$netif".gw
-        fi
+        case "$gw" in
+            *[!0-9.]*)
```
bdrung: According to the code, `gw` can contain a space separated list of gateways. This code will spit out a warning.

aafeijoo-suse: Yes, sorry, you're right. PTAL.

devkontrol (dismissed):

aafeijoo-suse:
> > What should we do about test 72 failing?
> 
> If fix is not available, than temporary disable to the test run with another commit.

Removed the 2nd commit disabling test 72, not needed anymore since the issue is already fixed in Tumbleweed.

Now it fails test 40 in ubuntu devel, but I see the same failure in other PRs and it didn't fail before, so it's unrelated.

devkontrol (approved):

Conan-Kudo (approved):

PR #2472, opened 2026-06-19
test(run-qemu): use efi firmware when testing on loongarch64

shankerwangmiao:
When testing on loongarch64, we should use EFI firmware, since the UEFI runtime services and ACPI tables are required by the linux kernel for loongarch64.

### Checklist


bdrung (approved):
LGTM

Conan-Kudo (approved):

PR #2484, opened 2026-07-03
fix(dracut-systemd): add ordering constraints

shankerwangmiao:
When dracut-initqueue is not enabled, there would be no dracut-initqueue.service as a sequence point. So explicitly declare After = basic.target to order dracut-pre-mount.service after all the services ordered before dracut-initqueue.service.

Fixes: b1ae591945ac ("dracut-systemd: add back missing dependencies")

### Checklist


devkontrol (approved):
Thank you very much @shankerwangmiao 

CC @dtardon

Conan-Kudo (approved):

PR #2493, opened 2026-07-08
test(systemd): disable man page checks in verify test

### LINKED ISSUE #2492: TEST-40-SYSTEMD fails when systemd-analyze verify checks man page availability
**Describe the bug**
TEST-40-SYSTEMD appears to fail spuriously because systemd-analyze.sh calls systemd-analyze verify "$i" without disabling man-page checks.

In the failing run, systemd-analyze verify sysinit.target fails because it tries to check man systemd.special availability inside the initrd test environment:

`dracut-pre-pivot[420]: sysinit.target: Command 'man systemd.special(7)' failed with code 1
dracut-pre-pivot[379]: Warning: systemd-analyze verify sysinit.target failed`

The helper then powers off the guest early, so the test fails before reaching its success path.

This seems like a false negative: TEST-40-SYSTEMD is validating the initrd/systemd boot path, but man-page availability inside the initrd test environment is not directly related to whether that boot path works.

**Distribution used**
Ubuntu stonking development release.

**Dracut version**
I reproduced this with both upstream main and the upstream 111 tag.

The downstream Ubuntu package dracut 111-4 is also affected at least. Theoretically, other branches/tags should also have the same problem.

**Init system**
systemd.
The systemd version in the test environment was 259.5-0ubuntu3.

**To Reproduce**
In an upstream dracut checkout, run:
`make TESTS="40" clean check`

The test fails:
`TEST: root filesystem on a ext4 filesystem with systemd but without initqueue   [FAILED]
see /dracut/test/TEST-40-SYSTEMD/test.log`

The relevant log lines are:
`dracut-pre-pivot[420]: sysinit.target: Command 'man systemd.special(7)' failed with code 1
dracut-pre-pivot[379]: Warning: systemd-analyze verify sysinit.target failed`

Then change systemd-analyze.sh:
`sed -i 's/systemd-analyze verify/systemd-analyze --man=no verify/' test/TEST-40-SYSTEMD/systemd-analyze.sh`

and rerun:
`make TESTS="40" clean check`

The test then passes:
`TEST: root filesystem on a ext4 filesystem with systemd but without initqueue   [OK]`

**Expected behavior**
TEST-40-SYSTEMD should not fail only because man-page lookup is unavailable inside the initrd test environment.

The test should validate the systemd/initrd boot path and unit verification behavior, without treating missing or inaccessible man pages as a fatal boot-path failure.

**Additional context**
systemd-analyze verify supports --man=no, which disables man-page existence checks. Using:
`systemd-analyze --man=no verify "$i"`

in systemd-analyze.sh fixes the failure in the same environment and allows the test to reach [OK].

This also affects the downstream Ubuntu dracut 111-4 package autopkgtest, where 40-systemd fails with the same root cause. Since the same issue reproduces directly in upstream main and the 111 tag, this appears to be an upstream test false negative rather than an Ubuntu packaging-specific issue.
Add any other context you like about the problem here.

Bug-Ubuntu: https://launchpad.net/bugs/2160087

zhhuabj:
TEST-40-SYSTEMD verifies initrd systemd targets, but systemd-analyze verify can also check referenced man pages. In the initrd test environment this may fail on man systemd.special, causing a false negative before the test reaches its success path.

Use --man=no so the test validates unit semantics without depending on man page availability.

Fixes: https://github.com/dracut-ng/dracut/issues/2492

### Checklist

### DIFF
--- a/test/TEST-40-SYSTEMD/systemd-analyze.sh
+++ b/test/TEST-40-SYSTEMD/systemd-analyze.sh
@@ -8,8 +8,8 @@ for i in \
     initrd-switch-root.target \
     emergency.target \
     shutdown.target; do
-    if ! systemd-analyze verify "$i"; then
-        warn "systemd-analyze verify $i failed"
+    if ! systemd-analyze --man=no verify "$i"; then
+        warn "systemd-analyze --man=no verify $i failed"
         poweroff
     fi
 done


bdrung (approved):
Thanks for debugging and providing a fix! I verified that 40-systemd is passing with this fix.

bdrung:
You could add the Ubuntu bug reference to your commit:
```
Bug-Ubuntu: https://launchpad.net/bugs/2160087
```

zhhuabj:
> You could add the Ubuntu bug reference to your commit:
> 
> ```
> Bug-Ubuntu: https://launchpad.net/bugs/2160087
> ```

Hi @bdrung , done, thank you.

Conan-Kudo (approved):

PR #2494, opened 2026-07-08
fix(crypt): do not call return to exit a systemd generator

### LINKED ISSUE #2490: crypt-generator.sh uses return 0 instead of exit 0
Yo,

error when booting dracut main branch as of today with 6.18.38 on recent Arch:

> Jul 07 23:18:45 archlinux (gene[483]: '/usr/lib/systemd/system-generators/dracut-crypt-generator' failed with exit status 2.

Config:

```$ cat dracut.conf 
omit_dracutmodules+=" drm plymouth multipath "
add_dracutmodules+=" busybox ssh-client zfs "
kernel_cmdline="rd.shell rd.skipfsck zfs.force"
remove_items+=" etc/zfs/zpool.cache "
compress="zstd -19 -q -T0"
add_drivers+=" ast i915 "
hostonly=no
show_modules=yes
kernel_cmdline+=" rd.debug rd.break SYSTEMD_SULOGIN_FORCE=1 "
fileloglvl=6
stdloglvl=6
sysloglvl=6
```

Debug output was inconclusive. So I added
> exec >/run/1.log 2>&1

to head of /usr/lib/dracut/modules.d/70crypt/crypt-generator.sh and after this gruelling ;-) debug session am seeing:

```/usr/lib/systemd/system-generators/dracut-crypt-generator@30(): '[' -e /etc/crypttab ']'
/usr/lib/systemd/system-generators/dracut-crypt-generator@30(): return 0
/usr/lib/systemd/system-generators/dracut-crypt-generator: line 30: return: can only `return' from a function or sourced script
```

When I replace all return X statements with exit X the problem disappears. I have no LUKS et al on this machine, this exercise was to build a universal dracut image that can boot from ZFS with root on ZFS through ZFSBootMenu.

/bin/sh is a symlink to bash, bash version is 5.3.15(1).

aafeijoo-suse:
systemd generators implemented via shell scripts are not sourced by systemd, so calling `return` within them to exit is incorrect.

Fixes #2490

### Checklist

### DIFF
--- a/modules.d/70crypt/crypt-generator.sh
+++ b/modules.d/70crypt/crypt-generator.sh
@@ -22,13 +22,13 @@ generator_set_device_timeout() {
 
 if ! getargbool 1 rd.luks; then
     # crypto LUKS detection is disabled
-    return 0
+    exit 0
 fi
 
-[ -e /etc/crypttab ] || return 0
+[ -e /etc/crypttab ] || exit 0
 
 GENERATOR_DIR="$1"
-[ -n "$GENERATOR_DIR" ] || return 1
+[ -n "$GENERATOR_DIR" ] || exit 1
 [ -d "$GENERATOR_DIR" ] || mkdir -p "$GENERATOR_DIR"
 
 timeout=$(getarg rd.timeout)


Conan-Kudo (approved):

bdrung (approved):
Thanks.

Next time you could include the error message you saw (and that you described in the linked bug) in the git commit message as well.

PR #2497, opened 2026-07-10
test(run-qemu): use -accel kvm instead of -enable-kvm

shankerwangmiao:
-enable-kvm is believed to be deprecated and -accel kvm should be used instead.

### Checklist


bdrung:
Do you have a reference to this deprecation?

shankerwangmiao:
> Do you have a reference to this deprecation?

I noticed that `-enable-kvm` is not supported for loongarch while kvm is actually supported by loongarch and thus proposed to enable this option for loongarch. A qemu maintainer told me that `-enable-kvm` is seen as a legacy option and `-accel kvm` is recommended instead. 

The link to the relevant mail: https://lore.kernel.org/all/c7828107-d2c8-4cbd-9208-dcc637161597@oss.qualcomm.com/

bdrung:
Thanks. Can you include that information into the git commit message?

shankerwangmiao:
> Thanks. Can you include that information into the git commit message?

Btw, currently the tests for dracut on loongarch with tcg acceleration would fail with kernel soft lockup, as shown in the Debian CI system. I have isolated the reason, which lies in qemu and I am trying to propose a patch to have it fixed.

devkontrol (approved):

PR #2504, opened 2026-07-16
feat(resume): include in hostonly sloppy mode if supported by kernel

bdrung:
Systems might support hibernation. They need the resume module to get a working hibernation. Users might configure their systems to enable hibernation, but the correct order of steps are crucial for it. Currently this workflow does not work:

1. Make sure system is supported the power state:
```
$ cat /sys/power/state
freeze mem disk
```
2. Configured swap partition
3. Populated the resume kernel boot parameter `resume=/dev/<swap partition>`
4. update grub
5. regenerate initrd

Regenerating the initrd will not include the resume module, because it hits the "hibernation supported by the kernel, but not enabled" code path (`/proc/cmdline` doesn not contain `resume=` yet and `/sys/power/resume` contains `0:0`). A user would need to reboot once before regenerating the initrd.

To make the system more predictable, include the `resume` module in hostonly `sloppy` mode if hibernation is supported by the kernel.

Bug-Ubuntu: https://launchpad.net/bugs/2158694

### DIFF
--- a/modules.d/74resume/module-setup.sh
+++ b/modules.d/74resume/module-setup.sh
@@ -22,7 +22,10 @@ check() {
             return 0
         # resume= not set on kernel command line
         elif [[ -f /sys/power/resume ]]; then
-            if [[ "$(< /sys/power/resume)" == "0:0" ]]; then
+            if [[ $hostonly_mode != "strict" ]]; then
+                ddebug "Module resume: hibernation supported by the kernel"
+                return 0
+            elif [[ "$(< /sys/power/resume)" == "0:0" ]]; then
                 ddebug "Module resume: hibernation supported by the kernel, but not enabled"
                 return 255
             else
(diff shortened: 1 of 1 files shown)


devkontrol:
@pvalena , @dtardon, @YHNdnzj - can you please help review this PR ?

Related - https://github.com/systemd/systemd/issues/27247

devkontrol:
@pvalena - does the thumbs up emoji here means formal PR approval from you on behalf of Fedora ? 

I would be interested to know if this PR gets reverted downstream or not, given Fedora's history of heavy downstream activity. Thank you !

pvalena:
> @pvalena - does the thumbs up emoji here means formal PR approval from you on behalf of Fedora ?

Sorry, off course not. I just didn't have enough time to give it a thorough look. Unfortunately, I don't have time to test it, but otherwise it LGTM.

> I would be interested to know if this PR gets reverted downstream or not, given Fedora's history of heavy downstream activity. Thank you !

I don't think so, we don't have any patches in `resume/module-setup.sh` (https://github.com/redhat-plumbers/dracut-fedora/blame/main/modules.d/74resume/module-setup.sh), and I don't see it causing any issues. _(If there're any, I'll do so only temporarily until upstream solution is found- unless there's an obvious fix._

Thank you for your patience!

pvalena (approved):

devkontrol (approved):

Conan-Kudo (approved):

PR #2505, opened 2026-07-19
ci(debian:sid): install systemd-tpm package

devkontrol:
Debian:sid has split out systemd-pcrextend into systemd-tpm package.

See https://packages.debian.org/sid/systemd-tpm

This commit is required to resolve a regression for the kernel-install test - see https://github.com/dracut-ng/dracut/actions/workflows/daily-kernel-install-x64.yml

### Checklist

CC @bdrung


bdrung (approved):
This split has been done in systemd 261-2. You might want to add this to the git commit message.

Once systemd 261.1-2ubuntu2 migrates from stonking to stonking-proposed, this change needs to be applied to Ubuntu as well.

PR #2507, opened 2026-07-20
feat(network-manager): use escape to drop requiring bash

bdrung:
Use the new `escape` function to drop requiring bash in the network-manager module.

The output of `kf_parse` is written to `/tmp/dhclient."$ifname".dhcpopts` which is sourced by other dracut modules.

### Checklist

### DIFF
--- a/modules.d/35network-manager/module-setup.sh
+++ b/modules.d/35network-manager/module-setup.sh
@@ -10,7 +10,7 @@ check() {
 
 # called by dracut
 depends() {
-    echo bash net-lib kernel-network-modules initqueue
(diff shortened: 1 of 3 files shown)


devkontrol (approved):

devkontrol:
@bdrung - I should have raised this earlier; does this function work with busybox ash and busybox sed ?

bdrung:
Yes. Example:
```
$ value='example string with " and '"'"' in it with \n'
$ printf "variable='%s'\n" "$(escape "$value")"
variable='example string with " and '\'' in it with \n'
```

Issue #1962, opened 2025-12-20
revert editing 64-md-raid-assembly.rules


jozzsi:
Introduced in e3e5128cf20660c0789f9b4e2285dbc1f35f6799  in 2011.

Is this workaround still needed ? Is this workaround in place to be able to include `65-md-incremental-imsm.rules` with dracut ?

In general dracut should not edit distribution package files. 

Note that `65-md-incremental-imsm.rules` provided by dracut mdadm module.

Perhaps this is more related to dmraid which is being phased out ? 

CC @msoltyspl 

As an example mkinitcpio does not customize this file - https://github.com/archlinux/mkinitcpio/blob/master/install/mdadm_udev

jozzsi:
Thank you for the opportunity to be involved. At this time, I feel it is best for me to step away from contributing to the project. Closing this issue as not planning to work on it.

bdrung:
Still relevant.

bdrung:
Discussion in https://github.com/dracut-ng/dracut/pull/2520 shows that editing 64-md-raid-assembly.rules is still needed.

PR #2518, opened 2026-07-22
fix(mdraid): escape slash in sed regular expression

bdrung:
`sed` from busybox treats the `/` from `[[:alpha:]/]` inside `/pattern/d` as the closing delimiter for the match pattern, rather than a character inside the bracket class:

```
sed: bad regex '(RUN|IMPORT\{program\})\+?="[[:alpha:]': Missing ']'
```

Escape the slash in sed regular expression.

**Note**: Found by debugging https://github.com/dracut-ng/dracut/issues/2512

### Checklist

### DIFF
--- a/modules.d/70mdraid/module-setup.sh
+++ b/modules.d/70mdraid/module-setup.sh
@@ -83,7 +83,7 @@ install() {
     # assembled
     # shellcheck disable=SC2016
     if [ -f "${initdir}${udevdir}/rules.d/64-md-raid-assembly.rules" ]; then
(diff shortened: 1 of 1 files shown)


devkontrol:
Should we try to consider https://github.com/dracut-ng/dracut/issues/1962 instead ? 

Sadly I think this PR by itself as it is does not resolve the test failures for the busybox CI .

bdrung:
> Should we try to consider #1962 instead ?

Yes.

bdrung:
We can do the removal in a follow-up PR.

bdrung:
Since https://github.com/dracut-ng/dracut/pull/2520 raised quite some discussion, let's get the `sed` command fixed first. We still can remove our custom udev rules.

devkontrol (approved):

PR #2520, opened 2026-07-22
chore(mdraid): remove 65-md-incremental-imsm.rules and 59-persistent-storage-md.rules

### LINKED ISSUE #1962: revert editing 64-md-raid-assembly.rules

bdrung:
The task of `59-persistent-storage-md.rules` is to call `mdadm --export` which is already provided by `63-md-raid-arrays.rules` from mdadm.

Fixes: https://github.com/dracut-ng/dracut/issues/1962

### Checklist


devkontrol:
CC @puleglot @aafeijoo-suse @mtkaczyk  @mwilck

puleglot:
I think this will effectively make `rd.md.uuid=` cmdline option noop. `mdraid_start.sh` and a big part of `parse-md.sh` also become useless.

puleglot:
> I think this will effectively make `rd.md.uuid=` cmdline option noop.

Most (all?) of the `rd.md*` options actually will be broken.

mtkaczyk:
I will review this tomorrow.

devkontrol:
Let me share additional context, so that people helping out do not need to dig this out.

This code is under attention as it might not be working with busybox (perhaps it never did). I do not think it was fully understood that removing this code as part of this PR also regress `rd.md*` options.

There is however additional motivation here - there is expressed interest that dracut should allow an option for a "vanilla" udev/systemd initramfs building (without a bunch of custom shell execution), so it might be a worthwhile discussion how important to maintain all `rd.md*` options.

One possible path forward: We keep the custom udev rules, but we split the current `mdraid` dracut module into 2 dracut modules - one additional (optional one) for these dracut udev rules - that a "vanilla" udev/systemd initramfs might not need.

This new dracut module (`mdraid-dracut`) would install the custom udev rules and would also depend on the `mdraid` dracut for basic mdraid binaries.

bdrung:
> One possible path forward: We keep the custom udev rules, but we split the current `mdraid` dracut module into 2 dracut modules - one additional (optional one) for these dracut udev rules - that a "vanilla" udev/systemd initramfs might not need.
> 
> This new dracut module (`mdraid-dracut`) would install the custom udev rules and would also depend on the `mdraid` dracut for basic mdraid binaries.

How about talking to upstream mdadm to get the needed code there to support those `rd.md*` parameters?

bdrung:
For what do we need `59-persistent-storage-md.rules`? This seems to be the relevant udev code:

```
IMPORT{program}="/sbin/mdadm --detail --export $devnode"
IMPORT{builtin}="blkid"
OPTIONS+="link_priority=100"
OPTIONS+="watch"
OPTIONS+="db_persist"
```

The same code is in mdadm's `63-md-raid-arrays.rules`:

```
IMPORT{program}="BINDIR/mdadm --detail --no-devices --export $devnode"
ENV{DEVTYPE}=="disk", ENV{MD_NAME}=="?*", SYMLINK+="disk/by-id/md-name-$env{MD_NAME}", OPTIONS+="string_escape=replace"
ENV{DEVTYPE}=="disk", ENV{MD_UUID}=="?*", SYMLINK+="disk/by-id/md-uuid-$env{MD_UUID}"
ENV{DEVTYPE}=="disk", ENV{MD_DEVNAME}=="?*", SYMLINK+="md/$env{MD_DEVNAME}"
ENV{DEVTYPE}=="partition", ENV{MD_NAME}=="?*", SYMLINK+="disk/by-id/md-name-$env{MD_NAME}-part%n", OPTIONS+="string_escape=replace"
ENV{DEVTYPE}=="partition", ENV{MD_UUID}=="?*", SYMLINK+="disk/by-id/md-uuid-$env{MD_UUID}-part%n"
ENV{DEVTYPE}=="partition", ENV{MD_DEVNAME}=="*[^0-9]", SYMLINK+="md/$env{MD_DEVNAME}%n"
ENV{DEVTYPE}=="partition", ENV{MD_DEVNAME}=="*[0-9]", SYMLINK+="md/$env{MD_DEVNAME}p%n"

IMPORT{builtin}="blkid"
OPTIONS+="link_priority=100"
OPTIONS+="watch"
OPTIONS+="db_persist"
```

devkontrol:
> How about talking to upstream mdadm to get the needed code there to support those `rd.md*` parameters?

There is precedence for it already - https://github.com/md-raid-utilities/mdadm/pull/143

mtkaczyk:
>How about talking to upstream mdadm to get the needed code there to support those rd.md* parameters?

mdadm do not care about it. AFAIC this is just for dracut to ensure we are waiting for raid device to be up before we are moving forward with booting. In rare cases, especially with RAID1 we would boot from member drive directly..

So what you do here is looking for `dev/disk/by-uuid/` or something close to that.
 The question is if the check like that is worthy mdadm integration. It would open a room for more control on booting in mdadm  but for more than 10 years we didn't added nothing more advanced.

mtkaczyk:
@devkontrol you forgot about this one:
https://github.com/dracutdevs/dracut/blob/master/modules.d/90mdraid/parse-md.sh

it appears than during the installation/processing (this is installed on some hook?) `65-md-incremental*.rules` are edited to minimize harm and focus only on critical arrays (the one with rootfs I guess).

mtkaczyk:
Oh I found it, we use simple wait_for_dev to wait for rootfs to be up:
https://github.com/dracutdevs/dracut/blob/master/modules.d/90mdraid/parse-md.sh#L38

so the  rules removed here should be successfully replaced by mdadm rules. The question is we are fine with loosing timer funtionality. mdadm has separate one which is integrated with systemd but it is not used by every metadata format (AFAIC native handles this, exteral formats not supporting it). Till that moment dracut has separate timer which is obviously removed now:
```
RUN+="/sbin/initqueue --timeout --name 50-mdraid_start --onetime --unique /sbin/mdraid_start"
```

bdrung:
So summary from the discussiion here:
* 65-md-incremental-imsm.rules cannot be dropped. Bigger refactoring would be needed to do so.
* 59-persistent-storage-md.rules is not needed for mdadm >= 4.5. I created https://github.com/dracut-ng/dracut/pull/2524 for that.

So let's close this PR for now.

PR #2525, opened 2026-07-23
style: remove unneeded function keyword

bdrung:
`function` is a non-standard keyword that can be used to declare functions in Bash and Ksh, but this would not work with POSIX shell. Remove it to make the code more consistent (and potential easier to port to POSIX).

### Checklist


devkontrol:
> `function` is a non-standard keyword that can be used to declare functions in Bash and Ksh, but this would not work with POSIX shell. So remove it.

I must be missing something here. These files in this PR are all executed with bash explicitly.

bdrung:
Okay. Update the description to avoid such confusion. The change is more for style.

devkontrol (approved):
ok

PR #2531, opened 2026-07-24
ci: increase timeout for test 60-NFS from 10 to 15 min

bdrung:
Test 60-NFS is the biggest test and 10 minutes was not enough for it on `fedora:rawhide` (network) and `gentoo:latest` (network-legacy).

So increase the timeout for test 60-NFS from 10 to 15 minutes.

See also: 4f6f4c47a5eb ("ci: unify timeout to 10 min by default")

### Checklist


devkontrol (approved):
Thank you @bdrung

PR #2533, opened 2026-07-25
chore(Makefile): filter out chrony dracut module for non-systemd

devkontrol:
chrony dracut module in its current form depends on the systemd dracut module. Filter out chrony dracut module during build time for enviroments not supporting systemd to be consisent with othersystemd dependent dracut modules.

Follow-up to 4d64d0b .

### Checklist


devkontrol:
CC @jozzsi for Alpine

Conan-Kudo (approved):
I suspect eventually this module will get adapted for non-systemd, since it's very useful for those environments. But for now, this is fine.

PR #2535, opened 2026-07-28
fix(hwdb): always install this module

### LINKED ISSUE #2534: Should hwdb always be included with systemd?
In https://github.com/systemd/systemd/issues/40485, there were some comments stating that the initrd should contain hwdb:
- https://github.com/systemd/systemd/issues/40485#issuecomment-4109666284
- https://github.com/systemd/systemd/issues/40485#issuecomment-4109670638

mkinitcpio already added a patch including it: https://github.com/systemd/systemd/issues/40485#issuecomment-4145470359

Since 106, in hostonly mode dracut does not install the hwdb module unless it's explicitly added because there is no module that depends on it.

https://github.com/dracut-ng/dracut/blob/f837e50a3b02da7151c86e5c8633c704d716b719/modules.d/74hwdb/module-setup.sh#L5-L11

So the question is: is hwdb always required by systemd? or only by some of the systemd modules?

aafeijoo-suse:
This module is not being installed in hostonly mode, which causes udev rules that use the built-in hwdb program to fail.

This way, there won't be any surprises with udev rules not working, and it can be manually omitted if someone wants to save space.

Fixes #2534

### Checklist

### DIFF
--- a/modules.d/74hwdb/module-setup.sh
+++ b/modules.d/74hwdb/module-setup.sh
@@ -3,7 +3,9 @@
 # SPDX-License-Identifier: GPL-2.0-or-later
 
 check() {
-    if [[ $hostonly ]]; then
+    if [[ $hostonly_mode == "strict" ]]; then
+        dwarn "hwdb: module not automatically included in strict hostonly" \
+            "mode, udev rules that use the built-in hwdb program will fail."
         return 255
     fi
 


dtardon (approved):
LGTM

devkontrol:
My primary concern is the `kdump` and `hostonly_mode=strict` related use-cases. 

CC @prudo1  @coiby @lkundrak

@pvalena  - please share your learning if you are able to from https://github.com/redhat-plumbers/dracut-fedora/pull/22 - back from 2024-May

Conan-Kudo:
I'd like to see this fixed before landing it.

pvalena:
> My primary concern is the `kdump` and `hostonly_mode=strict` related use-cases.
> 
> CC @prudo1 @coiby @lkundrak
> 
> @pvalena - please share your learning if you are able to from [redhat-plumbers/dracut-fedora#22](https://github.com/redhat-plumbers/dracut-fedora/pull/22) - back from 2024-May

Actually, I just forwarded it upstream, afterwards: https://github.com/dracut-ng/dracut/pull/225 -- but didn't have time to finish it... Indeed I think it's it's good to have option to actually require the module based on acutal requirements, so it's not just shipped into every initrd that doesn't need it.

For RHEL-10 f.e. we ended up with: https://github.com/redhat-plumbers/dracut-rhel10/blob/c933e1bde17f6bd5e173541efa52f5a3ec8098b8/modules.d/95hwdb/module-setup.sh#L6

It seems it was dropped from upstream with similar changes: https://github.com/dracut-ng/dracut/commit/ac8cb5e5c16db52eadee5c3b32842780b09c5801#diff-b0bf6b78461071a68b0a2290df131937bb23ef2706cddb02d198d1e2c03937c5R6

modules.d/74hwdb/module-setup.sh
```diff
@@ -2,14 +2,6 @@
 # This file is part of dracut.
 # SPDX-License-Identifier: GPL-2.0-or-later
 
-check() {
-    if [[ $hostonly ]]; then
```
devkontrol, line 6: I think this should be

```
    if [[ $hostonly_mode == "strict" ]]; then
```

aafeijoo-suse: Would it be ok to add a warning message at least in this case?

devkontrol (approved):

modules.d/74hwdb/module-setup.sh
devkontrol: yes.

Conan-Kudo (approved):

PR #2537, opened 2026-07-29
feat(systemd-journald): remove libgcrypt from add_dlopen_features

aafeijoo-suse:
0406aa64a788c63a0bb2fd9d60a2d3d2d96320cb only removed the fallback if dlopen notes are not being parsed. Moreover, libgcrypt survives only in unit tests after https://github.com/systemd/systemd/pull/42695.

Follow-up for 0406aa64a788c63a0bb2fd9d60a2d3d2d96320cb

### Checklist

### DIFF
--- a/modules.d/11systemd-journald/module-setup.sh
+++ b/modules.d/11systemd-journald/module-setup.sh
@@ -28,7 +28,7 @@ depends() {
 
 # Config adjustments before installing anything.
 config() {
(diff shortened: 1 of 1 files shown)


devkontrol:
CC @pvalena @chewi

chewi:
If it's being dropped, there's no harm in keeping it here for older systemd versions. Then again, there was probably little need for it in the first place.

devkontrol (approved):

PR #2543, opened 2026-07-30
fix(rngd): add configuration file on ArchLinux

### LINKED ISSUE #2542: rngd.service fails on ArchLinux due to config file not being installed
ArchLinux in their `rngd.service` uses `EnvironmentFile=/etc/conf.d/rngd`, see [here](https://gitlab.archlinux.org/archlinux/packaging/packages/rng-tools/-/blob/main/rngd.service?ref_type=heads#L9).
The file is not declared as optional (`-` prefix), and thus if it's not present, the service will fail to start, see [the systemd docs](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#EnvironmentFile=).

However, dracut's `16rngd` module in its [module-setup.sh](https://github.com/dracut-ng/dracut/blob/main/modules.d/16rngd/module-setup.sh#L38-L40) does not consider including this file, as it only considers `/etc/sysconfig` that is used by Fedora and co.

Thus, if `rng-tools` is installed on Arch, Dracut will include it by default, and will therefore cause `rngd.service` to consistently fail in initrd. This is automatic if that package is just installed.

Affected: 111 or master

Steps to reproduce:
1. On Arch install `dracut` and `rng-tools`.
2. Use `dracut` to generate initramfs.
3. Startup will cause `rngd.service` failure.

Fix:
Works fine with `install_items+=" /etc/conf.d/rngd "` set as configuration. I would suggest adding a simple check if that file exists to `modules-setup.sh` and copy it over if so.

ChrisTX:
ArchLinux uses /etc/conf.d rather than /etc/sysconfig in its rngd.service to read environment variables from. So far this file was not included automatically, and only sysconfig was considered, leading to rngd.service failing on boot on ArchLinux if rng-tools is just installed. This commit adds the same handling for /etc/conf.d

Fixes: #2542

### Checklist

### DIFF
--- a/modules.d/16rngd/module-setup.sh
+++ b/modules.d/16rngd/module-setup.sh
@@ -37,6 +37,8 @@ install() {
 
     if [ -r /etc/sysconfig/rngd ]; then
         inst_simple "${moddir}/sysconfig" "/etc/sysconfig/rngd"
+    elif [ -r /etc/conf.d/rngd ]; then
+        inst_simple "${moddir}/conf.d" "/etc/conf.d/rngd"
     fi
 
     # make sure dependent libs are installed too


devkontrol:
Earlier discussion on this - https://github.com/dracut-ng/dracut/pull/2243

ChrisTX:
> Earlier discussion on this - https://github.com/dracut-ng/dracut/pull/2243

I can see that standardising the path upstream would be preferred in the medium term, but the status quo is a rather unfortunate automatic failure of a service if rng tools is just installed.

Wouldn't it make sense to merge this fix, as it doesn't have any downsides?

bdrung:
An alternative would be to enhance dracut to read the service file and look for `EnvironmentFile`.

Conan-Kudo:
I'm not really a fan of this, I'd rather have a more general solution to read and include EnvironmentFiles for any systemd units.

ChrisTX:
I can see how such a general solution would be overall preferable, but that's much different in terms of scope than a simple bugfix. As mentioned, this creates real issues right now for people who are using Dracut on Arch.

A general solution would require quite some design discussion, as .service files can be overridden. What if it read the EnvironmentFile=... line from the .service file, but a user had overridden it to use a different location? You couldn't read that from the .service alone.

As of such, and considering that the /etc/sysconfig solution exists right next to it as the same stop-gap until such a feature is designed, wouldn't it be an option to merge this as a bugfix until then?

Conan-Kudo:
My worry is that then this opens the doors for every distro's variants of this to be in _every_ module, and I don't really want to be in the business of figuring that out when someone does something stupid like have two distros' files because they cross-graded and so the wrong data is read.

This is already bad enough as it is. 😦

mtorromeo:
I don't know if this is acceptable but this could be used to read all potential environment files from the unit, and this would work even if the end user customized the units or with systemd drop-in files:

```
systemctl cat rngd | awk -F= '{ if ($1 == "EnvironmentFile") { sub(/^-/, "", $2); print $2 } }'
```

devkontrol:
> I don't know if this is acceptable but this could be used to read all potential environment files from the unit, and this would work even if the end user customized the units or with systemd drop-in files:
> 
> ```
> systemctl cat rngd | awk -F= '{ if ($1 == "EnvironmentFile") { sub(/^-/, "", $2); print $2 } }'
> ```

@bdrung has started similar work in another PR I think  - see https://github.com/dracut-ng/dracut/pull/1031 

We would have to understand the performance impact of such an extra computation before merging

PR #2544, opened 2026-07-31
fix(dracut): remove support for tmpfilesconfdir

devkontrol:
tmpfilesconfdir variable was used only by the systemd-tmpfiles dracut module, but after 73e9b64 that variable is no no longer in use.

The motivation for this PR is to explicitly disallow the use of tmpfilesconfdir going forward as it is not recommended for a dracut module to pick up files from /etc/tmpfiles.d.

Follow-up to 73e9b64 .

### Checklist


devkontrol:
From @bdrung 

>  I searched for all those four variables on https://codesearch.debian.net/ - and they are not used anywhere in Debian.

https://github.com/dracut-ng/dracut/pull/2245#issuecomment-5044546945

CC @aafeijoo-suse

aafeijoo-suse:
Fine by me.

I hadn't mentioned this before, because it's like opening a can of worms, but... here we go: #2546
Also, other some minor fixes that can be merged now without discussion: #2545

bdrung (approved):

PR #2548, opened 2026-07-31
Minor assorted fixes

aafeijoo-suse:
See individual commits.

### Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1312,7 +1312,7 @@ fw_dir=${fw_dir//:/ }
 # check for logfile and try to create one if it doesn't exist
 if [[ -n $logfile ]]; then
     if [[ ! -f $logfile ]]; then
-        if touch "$logfile"; then
+        if ! touch "$logfile"; then
             dwarn "touch $logfile failed."
         fi
     fi
--- a/dracut.sh
+++ b/dracut.sh
@@ -332,7 +332,7 @@ push_user_devs() {
     local _dev
     for _dev in "$@"; do
         [[ -z $_dev ]] && continue
-        [[ " ${user_devs[*]} " == *" $_dev "* ]] && return
+        [[ " ${user_devs[*]} " == *" $_dev "* ]] && continue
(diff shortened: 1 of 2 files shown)


dracut.sh
```diff
@@ -2794,7 +2794,7 @@ if [[ $no_kernel != yes ]]; then
         # shellcheck disable=SC2086
         hostonly='' instmods -c $force_drivers
         if [[ $kernel_only != yes ]]; then
-            rm -f "$initdir"/etc/cmdline.d/20-force_driver.conf
+            rm -f "$initdir"/etc/cmdline.d/20-force_drivers.conf
```
bdrung, line 2797: How about putting this file name into a variable to avoid having to specify it twice?

bdrung (commented):
Typo in commit: "is assigned an not used" -> should be "and"

dracut.sh
devkontrol: Is this line (removing the file) even important here ? It would be very unusual for the `rm` command to be NOT to be a no-op.

dracut.sh
```diff
@@ -2794,7 +2794,7 @@ if [[ $no_kernel != yes ]]; then
...
             for mod in $force_drivers; do
                 echo "rd.driver.pre=$mod" >> "$initdir"/etc/cmdline.d/20-force_drivers.conf
             done
```
bdrung, line 2800: In case you want to avoid the `rm` command, you probably would need to do something like:
```
    { for mod in $force_drivers; do echo "rd.driver.pre=$mod"; done; } > "$initdir"/etc/cmdline.d/20-force_drivers.conf
```

bdrung: Since we are close to a release, I went ahead and implemented that: https://github.com/dracut-ng/dracut/pull/2551

devkontrol:
Superseded by https://github.com/dracut-ng/dracut/pull/2551 , closing. Please reopen if this PR is still needed.

PR #2554, opened 2026-08-03
ci(alpine): remove duplicated packages

devkontrol:
After a recent upstream change in the dracut-test package, we can remove most of the remaining downstream Alpine packages to improve readability and reduce maintenance.

See https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/103160/diffs .

### Checklist


bdrung:
Having the directly used projects listed there is useful IMO.

PR #2560, opened 2026-08-03
fix(net-lib): validate iSCSI port parameters

aafeijoo-suse:
Similar to 05c02db994ca322371a40695a3cc223faa1c7258, the iSCSI port is a number.

### Checklist

### DIFF
--- a/modules.d/45net-lib/net-lib.sh
+++ b/modules.d/45net-lib/net-lib.sh
@@ -367,19 +367,14 @@ normalized_iscsi_name() {
     printf "%s" "$1" | tr '[:upper:]' '[:lower:]' | tr -cd 'a-z0-9.:-'
 }
 
-_validate_iscsi_lun() {
-    local raw_lun="$1"
-
-    case "$raw_lun" in
-        *[!0-9]*)
-            warn "Given iSCSI LUN '$raw_lun' contains non-digits. Stripping non-digit characters."
-            printf "%s" "$raw_lun" | tr -cd '[:digit:]'
-            ;;
-        *)
-            # Purely decimal digits
-            echo "$raw_lun"
-            ;;
-    esac
+strip_non_digits() {
+    local str="$1"
+    local msg="$2"
+    if [ -n "$str" ] && ! isdigit "$str"; then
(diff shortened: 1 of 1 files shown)


modules.d/80base/dracut-lib.sh
```diff
@@ -261,6 +261,16 @@ isdigit() {
     return 0
 }
 
+strip_non_digits() {
+    local _str="$1"
+    local _msg="$2"
```
bdrung: since we declare these variables as local, the `_` is not needed.

bdrung (dismissed):

modules.d/80base/dracut-lib.sh
aafeijoo-suse: I know it's not necessary, it's an old habit that helps to identify a local variable just by seeing its name. Anyway, I can remove the `_` if you want.

Conan-Kudo: I would prefer they be removed, as they are increasingly unconventional in the codebase.

bdrung (dismissed):

devkontrol:
What is the motivation to move the `strip_non_digits` function to base out of net ?

aafeijoo-suse:
> What is the motivation to move the `strip_non_digits` function to base out of net ?

It's a generic function, not related to or tied to networking.

devkontrol:
> > What is the motivation to move the `strip_non_digits` function to base out of net ?
> 
> It's a generic function, not related to or tied to networking.

I do not see an urgency to move `strip_non_digits` to base as part of this PR. We can always move `strip_non_digits` to base when there is a second user of this function.

aafeijoo-suse:
> I do not see an urgency to move `strip_non_digits` to base as part of this PR. We can always move `strip_non_digits` to base when there is a second user of this function.

It's not urgency, it's what seems natural to me. Since it was already approved by @bdrung in the current form, I would like confirmation as to whether I need to change it or not.

bdrung:
I agree with @devkontrol to keep `strip_non_digits` in the net libs until it is needed somewhere else.

bdrung (approved):

devkontrol (approved):

PR #2565, opened 2026-08-06
Follow-ups for additional hook locations

aafeijoo-suse:
See individual commits.

### Checklist

### DIFF
--- a/modules.d/77dracut-systemd/dracut-mount.sh
+++ b/modules.d/77dracut-systemd/dracut-mount.sh
@@ -23,9 +23,14 @@ while :; do
         . "$f"
         if ismounted "$NEWROOT"; then
             usable_root "$NEWROOT" && break
-            warn "$NEWROOT has no proper rootfs layout, ignoring and removing offending mount hook"
             umount "$NEWROOT"
-            rm -f -- "$f"
+            # hooks can be in read-only locations
+            if [ -w "$f" ]; then
+                warn "$NEWROOT has no proper rootfs layout, ignoring and removing offending mount hook"
+                rm -f -- "$f"
+            else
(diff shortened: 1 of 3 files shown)


devkontrol:
> hooks can be in read-only locations
> `rm` will print a permission denied message.

Dracut should indeed try to do better here. 

It would be even better to have the same flow for read-only and read-write hook locations.

Would it be possible never delete hooks ?

CC @mwilck

aafeijoo-suse:
> Would it be possible never delete hooks ?

Background on why we thought that removing initqueue/finished hooks is a good idea: https://github.com/dracutdevs/dracut/pull/2335

devkontrol:
> Background on why we thought that removing initqueue/finished hooks is a good idea: [dracutdevs/dracut#2335](https://github.com/dracutdevs/dracut/pull/2335)

Thanks @aafeijoo-suse for the pointer, it helps.

> delete finished initqueue scripts when they succeed. 

Could we perhaps have a marker when scripts succeed somewhere under /run (which is always rw) instead of deleting the hook ? 

CC @vittyvk

vittyvk:
> > Background on why we thought that removing initqueue/finished hooks is a good idea: [dracutdevs/dracut#2335](https://github.com/dracutdevs/dracut/pull/2335)
> 
> Thanks @aafeijoo-suse for the pointer, it helps.
> 
> > delete finished initqueue scripts when they succeed.
> 
> Could we perhaps have a marker when scripts succeed somewhere under /run (which is always rw) instead of deleting the hook ?
> 

Yes, I think it is possible: hooks in /etc, /usr, and /var override each other when they have the same name. It gives us an opportunity to just create a /var/lib/dracut/hooks/succeeded/[queue]/[hookname] or /run/dracut/... as an indicator.

PR #2574, opened 2026-08-08
feat(rngd): only install rngd dracut module if explicitly included

devkontrol:
rngd dracut module is not needed anymore since kernel 5.6 (released 2020) because /dev/random does not block anymore.

This PR does not remove support for the rngd dracut module - it simply now not installed by default.

According to dracut compatibility table, currently our CI containers are all kernel version 6.12 or higher - https://dracut-ng.github.io/dracut/developer/compatibility.html

See https://github.com/torvalds/linux/commit/30c08efec8884fb106b8e57094baa51bb4c44e32

### Checklist

### DIFF
--- a/modules.d/16rngd/module-setup.sh
+++ b/modules.d/16rngd/module-setup.sh
@@ -23,7 +23,7 @@ check() {
     require_binaries rngd || return 1
     [[ -e "${systemdsystemunitdir}/rngd.service" ]] || return 1
 
-    return 0
+    return 255
 }
 
 depends() {


devkontrol:
@ChrisTX - can you please confirm, that this resolves the issue that you reported ?

This PR is not intended to close https://github.com/dracut-ng/dracut/issues/2542 , but for most users it will take away the warning/error that is being reported without any adverse implications.

Conan-Kudo (approved):

bdrung:
> because /dev/random does not block anymore

https://github.com/torvalds/linux/commit/30c08efec8884fb106b8e57094baa51bb4c44e32 says:

> /dev/random [...] reads will block until the CRNG is ready.

If I read that correctly, `/dev/random` will still block. Can you check?

PR #2621, opened 2026-08-19
Log missing files from dracut-install and inst_simple

airtower-luna:
Files missing without any indication why causes trouble when debugging initramfs builds. I added these while working on #2588, @bdrung suggesting splitting the commits out so they can be merged already.

### Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1150,9 +1150,14 @@ inst_script() {
 inst_simple() {
     local dstdir="${dstdir:-"$initdir"}"
     local _ret _hostonly_install
-    if [[ $1 == "-H" ]] && [[ $hostonly ]]; then
-        _hostonly_install="-H"
-        shift
+    if [[ $1 == "-H" ]]; then
+        if [[ $hostonly ]]; then
+            _hostonly_install="-H"
+            shift
+        else
+            ddebug "skipping install of '$2', not in hostonly mode"
+            return 0
+        fi
     fi
     [[ -e ${dstdir}/"${2:-$1}" ]] && return 0 # already there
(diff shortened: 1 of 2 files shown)


devkontrol (approved):

devkontrol (dismissed):

dracut-functions.sh
```diff
@@ -1155,10 +1155,12 @@ inst_simple() {
...
     fi
     [[ -e ${dstdir}/"${2:-$1}" ]] && return 0 # already there
-    if [[ $1 == /* ]]; then
-        [[ -e ${dracutsysrootdir-}/${1#"${dracutsysrootdir-}"} ]] || return 1 # no source
-    else
-        [[ -e $1 ]] || return 1 # no source
+    if [[ $1 == /* ]] && [[ ! -e ${dracutsysrootdir-}/${1#"${dracutsysrootdir-}"} ]]; then
+        dwarn "no source: '$1'!"
+        return 1
+    elif [[ ! -e $1 ]]; then
```
bdrung, line 1168: Isn't this elif part hit in this case: `[[ $1 == /* ]] && [[ -e ${dracutsysrootdir-}/${1#"${dracutsysrootdir-}"} ]]`?

airtower-luna: You're right, I was trying to simplify the first two lines into one `if`, but it's going to break if `$1` is absolute, exists within the sysroot, but does not contain the sysroot path. Going to fix that.

airtower-luna: Should be fixed now. I went back to the inner condition in the `if` block above.

devkontrol (approved):

bdrung (approved):

PR #2622, opened 2026-08-20
fix(dracut-install): bounds-check RUNPATH walk and ELF string derefs

prabhakarpujeri:
**Stacked on #2615** (depends on its `PARSE_ELF_START` extent checks and `elf_sect_name` helper). Will be rebased after #2615 merges.

## Problem

`FIND_LIBRARY_RUNPATH_FOR_BITS` (in `src/install/dracut-install.c`) walks the ELF `.dynamic` section's `Elf_Dyn` array and pulls out the DT_RUNPATH / DT_RPATH strings to search for libraries. Unlike the sibling macros `RESOLVE_DEPS_NEEDED_FOR_BITS` and `RESOLVE_DEPS_DLOPEN_FOR_BITS`, this walk validates only `e_shoff`/`e_shstrndx` — the per-iteration pointers into the map are never bounds-checked. Additionally, in the interpreter/`DT_SONAME`/`DT_NEEDED` code paths the strings extracted with `(char *)map + shdr[i].sh_offset + d->d_un.d_val` are handed to `strlen`/`strdup`/`hashmap_get`/`expand_runpath`/`regexec` without verifying they are NUL-terminated within the mapped file (a malformed ELF reading past the end of the map can hang on a never-terminated `strcmp`/`strdup`).

Also found one off-by-boundary check in the interpreter path: `interpreter > (char *)map + src_len` accepts the pointer at exactly one-past-end (should be `>=`).

## Fix

- New shared helper `elf_map_string(map, src_len, offset)`: returns NULL if the offset is outside the map or the string there isn't NUL-terminated within the map. Used for the runpath string, interpreter string, DT_SONAME and DT_NEEDED strings.
- The RUNPATH walk now mirrors its siblings' per-iteration bounds checks on `dyn` and `d`.

## Verification

Crafted 3 malformed ELF64 inputs exercising each new check; ran the real `dracut-install -l` binary built before/after, and `valgrind` on the difference-maker:

| input | before | after |
|---|---|---|
| `H7A2`: DT_RPATH whose strtab pointer lands 4 bytes before map end, runpath not NUL-terminated | valgrind `Invalid read of size 1` → SIGSEGV | clean exit 0 (section rejected) |
| `H7B2`: PT_INTERP at exactly `src_len` one-past-end | no crash (never dereferenced) | clean exit 0 (no change needed⁠—it's never actually reached the new helper with that offset as a valid string; check now correctly excludes) |
| `H7C2`: PT_INTERP in-map but non-terminated | clean | clean exit 0 |

(The last two are safe in both versions only because no code path dereferences them — they're defensive/paranoia checks. The RUNPATH one is the reachable crash.)

**Regression control:** `-l` dependency resolution of `/bin/bash`, `/bin/ls`, and `/usr/bin/ssh` — byte-identical 33-entry file list before vs after. **H6 test set re-verified** — the 6 malformed ELFs from #2615's verification table still all exit cleanly.

### DIFF
--- a/src/install/dracut-install.c
+++ b/src/install/dracut-install.c
@@ -865,6 +865,17 @@ static const char *elf_sect_name(const char *shstrtab, size_t shstrtab_len, uint
         return shstrtab + name_off;
 }
 
+/* Return a pointer to the NUL-terminated string at the given offset within
+   the mapped file, or NULL if the offset is out of bounds or the string
+   would not be NUL-terminated within the map. */
+static const char *elf_map_string(const char *map, size_t src_len, size_t offset)
+{
+        if (offset >= src_len || !memchr(map + offset, '\0', src_len - offset))
+                return NULL;
+
+        return map + offset;
+}
+
 /* Get a pointer to the ELF header map's section header string table, where B is
    64 or 32 bit. Sanity checks the ELF structure to avoid crashes. */
 #define PARSE_ELF_START(B, map) \
@@ -906,7 +917,12 @@ static const char *elf_sect_name(const char *shstrtab, size_t shstrtab_len, uint
                         continue; \
 \
                 Elf##B##_Dyn *dyn = (Elf##B##_Dyn *)((char *)map + ELF_BYTESWAP(B, shdr[i].sh_offset)); \
+                if ((char *)dyn < (char *)map || (char *)dyn > (char *)map + src_len) \
+                        break; \
+\
                 for (Elf##B##_Dyn *d = dyn; ELF_BYTESWAP(32, d->d_tag) != DT_NULL; d++) { \
+                        if ((char *)d < (char *)map || (char *)d + sizeof(Elf##B##_Dyn) > (char *)map + src_len) \
+                                break; \
                         if (ELF_BYTESWAP(B, d->d_tag) == DT_RUNPATH) \
                                 seen_runpath = true; /* RUNPATH has precedence over RPATH. */ \
                         else if (seen_runpath || ELF_BYTESWAP(B, d->d_tag) != DT_RPATH) \
@@ -915,7 +931,10 @@ static const char *elf_sect_name(const char *shstrtab, size_t shstrtab_len, uint
                         if (ELF_BYTESWAP(32, shdr[i].sh_link) >= ELF_BYTESWAP(16, ehdr->e_shnum)) \
                                 break; \
 \
-                        char *runpath = (char *)map + ELF_BYTESWAP(B, shdr[ELF_BYTESWAP(32, shdr[i].sh_link)].sh_offset) + ELF_BYTESWAP(B, d->d_un.d_val); \
+                        const char *runpath = elf_map_string((const char *)map, src_len, ELF_BYTESWAP(B, shdr[ELF_BYTESWAP(32, shdr[i].sh_link)].sh_offset) + ELF_BYTESWAP(B, d->d_un.d_val)); \
+                        if (!runpath) \
+                                break; \
+\
                         _cleanup_free_ char *expanded = expand_runpath(runpath, src, match64); \
                         if (!expanded) \
                                 continue; \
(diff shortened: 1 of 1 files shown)


src/install/dracut-install.c
```diff
@@ -875,17 +912,29 @@ static char *expand_runpath(const char *input, const char *src, const Elf64_Ehdr
         bool seen_runpath = false; \
 \
         for (size_t i = 0; i < ELF_BYTESWAP(16, ehdr->e_shnum); i++) { \
-                if (strcmp(&shstrtab[ELF_BYTESWAP(32, shdr[i].sh_name)], ".dynamic") != 0) \
+                const char *sect_name = elf_sect_name(shstrtab, shstrtab_len, ELF_BYTESWAP(32, shdr[i].sh_name)); \
+                if (!sect_name || strcmp(sect_name, ".dynamic") != 0) \
                         continue; \
 \
                 Elf##B##_Dyn *dyn = (Elf##B##_Dyn *)((char *)map + ELF_BYTESWAP(B, shdr[i].sh_offset)); \
+                if ((char *)dyn < (char *)map || (char *)dyn > (char *)map + src_len) \
```
chewi: I think `void *` rather than `char *` here and below would be equivalent but less confusing?

prabhakarpujeri: Addressed in a0e87b34 — switched all the raw-map pointer casts (`map` arithmetic, `dyn`/`d`/`phdr`/`shdr` bounds comparisons, and the `elf_map_string()` call sites) from `char *` to `void *`. Built clean with the project's `-std=gnu99 -Wall` flags; behavior is unchanged.

chewi: Apologies, I had forgotten that I'd used `char *` myself, so you were just following suit. I thought it at least made sense where we were assigning to a `char *`, but on reflection, the `map` is obviously not a string, so what you have now is better.

chewi: Aah, I see CI fails now. Maybe that's why I went with `char *` in the end. Best stick with what you had originally, I guess. Sorry!

prabhakarpujeri: Reverted — dropped a0e87b34 and force-pushed back to 26faa415, so the branch is back to the original `char *` casts.

For the record, the CI failure root cause: clang's `-Werror=compare-distinct-pointer-types` fires on the `.note.dlopen` lines (`note_offset`/`note_end` are `const char *`, so `note_offset < (void *)map` compares distinct pointer types). So `char *` is indeed the right choice here — thanks for the suggestion and the quick flag anyway.

chewi (approved):

devkontrol (dismissed):

src/install/dracut-install.c
```diff
@@ -875,17 +912,29 @@ static char *expand_runpath(const char *input, const char *src, const Elf64_Ehdr
...
+                        break; \
+\
                 for (Elf##B##_Dyn *d = dyn; ELF_BYTESWAP(32, d->d_tag) != DT_NULL; d++) { \
+                        if ((char *)d < (char *)map || (char *)d + sizeof(Elf##B##_Dyn) > (char *)map + src_len) \
+                                break; \
                         if (ELF_BYTESWAP(B, d->d_tag) == DT_RUNPATH) \
```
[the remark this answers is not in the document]
prabhakarpujeri: The loop condition is already safe here: the `(char *)d + sizeof(Elf##B##_Dyn) <= (char *)map + src_len` check runs first and `&&` short-circuits, so `d->d_tag` is only dereferenced after the current entry has been proven to lie fully inside the mapping. The same guard order is used at the other two DT loops. No change needed for this one.

src/install/dracut-install.c
```diff
@@ -875,17 +912,29 @@ static char *expand_runpath(const char *input, const char *src, const Elf64_Ehdr
...
                         else if (seen_runpath || ELF_BYTESWAP(B, d->d_tag) != DT_RPATH) \
                                 continue; \
 \
-                        char *runpath = (char *)map + ELF_BYTESWAP(B, shdr[ELF_BYTESWAP(32, shdr[i].sh_link)].sh_offset) + ELF_BYTESWAP(B, d->d_un.d_val); \
+                        if (ELF_BYTESWAP(32, shdr[i].sh_link) >= ELF_BYTESWAP(16, ehdr->e_shnum)) \
+                                break; \
+\
+                        const char *runpath = elf_map_string((const char *)map, src_len, ELF_BYTESWAP(B, shdr[ELF_BYTESWAP(32, shdr[i].sh_link)].sh_offset) + ELF_BYTESWAP(B, d->d_un.d_val)); \
+                        if (!runpath) \
+                                break; \
```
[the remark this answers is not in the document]
prabhakarpujeri: Addressed in 3c101aab: the string-table offset and `d->d_un.d_val` are now each clamped to the mapped range before the addition, so the sum passed to `elf_map_string()` can no longer wrap. Applied at all three call sites the comment flagged (RUNPATH, SONAME, and NEEDED).

PR #2623, opened 2026-08-20
perf(bluetooth): do not try to install 60-persistent-input.rules

aafeijoo-suse:
It is always installed unconditionally via the udev-rules module, so it is redundant in this module.

https://github.com/dracut-ng/dracut/blob/ed81bbd9199df18c1452fc58b5554ed6213cae48/modules.d/74udev-rules/module-setup.sh#L39

### Checklist

### DIFF
--- a/modules.d/70bluetooth/module-setup.sh
+++ b/modules.d/70bluetooth/module-setup.sh
@@ -23,8 +23,7 @@ check() {
 
 # Module dependency requirements.
 depends() {
-    # This module has external dependencies on the systemd and dbus modules.
-    echo systemd dbus
+    echo dbus systemd-udevd
     # Return 0 to include the dependent modules in the initramfs.
     return 0
 }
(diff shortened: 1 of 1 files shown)


bdrung (changes requested):
If you remove this file, this module would need to depend on the udev-rules package.

modules.d/70bluetooth/module-setup.sh
```diff
@@ -24,7 +24,7 @@ check() {
 # Module dependency requirements.
 depends() {
     # This module has external dependencies on the systemd and dbus modules.
```
bdrung: This comment is outdated then. I do not see that this comment provides any useful information. IMO it could go.

bdrung (commented):
Then depending on systemd-udevd make this dependency clearer.

bdrung (approved):

devkontrol (approved):

PR #2624, opened 2026-08-20
perf(btrfs): remove dead code

aafeijoo-suse:
64-btrfs.rules is installed by udev [1] / eudev [2] since Sep 2012, and tried to be used by dracut instead of its custom btrfs implementation since Feb 2013 [3]. This code has been dead ever since: `inst_rules()` always returns 0, both today and even back then [4]. That means 80-btrfs.rules, btrfs_device_ready.sh and btrfs_finished.sh are not being installed since 2013, so it's safe to remove them from our codebase.

Also, `btrfs-zero-log` was removed in btrfs-progs-4.16.1 (2018-04-24) [1], its functionality is condensed in `btrfs`, via `btrfs rescue zero-log`.

[1] https://github.com/systemd/systemd/commit/0bb91b50100e4633a0e68135854e606653055748
[2] https://github.com/eudev-project/eudev/commit/0bb91b50100e4633a0e68135854e606653055748
[3] f12f694355c8105438be2b7ee9d3e84dd4568f6a
[4] http://github.com/dracut-ng/dracut/blob/f12f694355c8105438be2b7ee9d3e84dd4568f6a/dracut-functions.sh#L937-L963

### Checklist


bdrung (approved):
Thanks. The change looks good.

One suggestion: The title "perf(btrfs): remove dead code" could be made more detailed. E.g. "perf(btrfs): remove dead code for missing 64-btrfs.rules case"

devkontrol (approved):

PR #2630, opened 2026-08-21
Assorted fixes for DRACUT_EXTRA_ARGS

aafeijoo-suse:
See individual commits.

### Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -609,7 +609,7 @@ eval set -- "$TEMP"
 if [[ -n ${DRACUT_EXTRA_ARGS-} ]]; then
     reset_lastpipe="$(shopt -p lastpipe)"
     shopt -s lastpipe
-    if xargs printf '%s\0' <<< "$DRACUT_EXTRA_ARGS" 2> /dev/null | mapfile -d '' _extra_args; then
+    if xargs -r printf '%s\0' <<< "$DRACUT_EXTRA_ARGS" 2> /dev/null | mapfile -d '' _extra_args; then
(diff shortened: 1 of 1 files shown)


devkontrol (dismissed):

dracut.sh
```diff
@@ -607,10 +607,13 @@ PARMS_TO_STORE=""
 eval set -- "$TEMP"
 
 if [[ -n ${DRACUT_EXTRA_ARGS-} ]]; then
-    mapfile -d '' _extra_args < <(xargs printf '%s\0' <<< "$DRACUT_EXTRA_ARGS" 2> /dev/null \
-        || printf "%s\n" "dracut[W]: Ignoring malformed DRACUT_EXTRA_ARGS: $DRACUT_EXTRA_ARGS" >&2)
-    set -- "${_extra_args[@]}" "$@"
-    unset _extra_args
+    if xargs true <<< "$DRACUT_EXTRA_ARGS" 2> /dev/null; then
+        mapfile -d '' _extra_args < <(xargs printf '%s\0' <<< "$DRACUT_EXTRA_ARGS")
```
bdrung: Why do you call xargs twice? Couldn't you just skip the `set` call in case of the xargs call fails?

aafeijoo-suse: Enable/disable `lastpipe` only for this? IMO it just adds complexity for the little value it provides, but if that is what you want.

bdrung: Okay. The previous version was more readable that this `lastpipe` one.

Would it be possible to use a function and `return` in the error case to avoid `lastpipe`?

PR #2649, opened 2026-08-25
fix(busybox): allow overriding list of applets

airtower-luna:
If the sysroot is cross-compiled, running `busybox --list` is usually not possible. Introduce the environment variable `DRACUT_MODULE_BUSYBOX_LINKS`, which overrides autodetection to work around the problem.

Split out of #2588 as suggest by @devkontrol.

### Checklist

### DIFF
--- a/man/dracut.8.adoc
+++ b/man/dracut.8.adoc
@@ -791,6 +791,20 @@ _DRACUT_INSTALL_LOG_LEVEL_::
 Default:
     _DRACUT_LOG_LEVEL_
 
+_DRACUT_MODULE_BUSYBOX_LINKS_::
+    overrides the list of busybox applets to link in the initramfs. Used for
+    **--sysroot** if the sysroot is cross-compiled and running `busybox
+    --list-full` is not possible. The list must contain full paths of the links
+    to create (leading slash optional).
++
+If you compile busybox, using the **busybox.links** file produced by the
+busybox build is recommended, e.g.:
++
+`export DRACUT_MODULE_BUSYBOX_LINKS="$(cat busybox.links)"`
++
+Using the output of `busybox --list-full` (e.g. running using Qemu userspace
+emulation) is also possible.
+
 FILES
 -----
 _/var/log/dracut.log_::
--- a/modules.d/10busybox/module-setup.sh
+++ b/modules.d/10busybox/module-setup.sh
@@ -15,14 +15,25 @@ check() {
 
 # called by dracut
 install() {
-    local _path _p _busybox _busybox_path
+    local _path _p _busybox _busybox_applets _busybox_path
     local _dstdir="${dstdir:-"$initdir"}"
     local _progs=()
     _busybox=$(find_binary busybox)
     _busybox_path="/usr/bin/busybox"
 
+    # get list of available applets
+    if [[ ${DRACUT_MODULE_BUSYBOX_LINKS-} ]]; then
+        _busybox_applets="$DRACUT_MODULE_BUSYBOX_LINKS"
+    else
+        _busybox_applets="$($_busybox --list-full)"
+    fi
+    if ! [[ ${_busybox_applets} ]]; then
+        derror "Could not get list of busybox applets!"
+        dinfo "If cross-building, try setting DRACUT_MODULE_BUSYBOX_LINKS."
+    fi
+
     # install busybox symlinks manually
-    for _path in $($_busybox --list-full); do
+    for _path in ${_busybox_applets}; do
         # if busybox is built without CONFIG_FEATURE_INSTALLER=y, the list
         # has only plain names
         if [[ ${_path##*/} == "${_path}" ]]; then
--- a/test/TEST-15-BUSYBOX/test.sh
+++ b/test/TEST-15-BUSYBOX/test.sh
@@ -8,6 +8,12 @@ test_check() {
     require_binaries_for_test busybox
 }
 
+test_setup() {
+    busybox --list-full > "${TESTDIR}/busybox.links"
+    # add an extra link to see if the list is used
+    echo "usr/bin/silly-serval" >> "${TESTDIR}/busybox.links"
+}
+
 check_applets_from_busybox() {
     local _applet _listing ret=0
     local initrd="$1"
@@ -42,7 +48,7 @@ test_run() {
 
     # Check applets the base module would otherwise install from the host.
     # Each must resolve to busybox in the resulting initrd.
-    check_applets_from_busybox "$initrd" cp ip ls mv mkdir sleep tr || ret=1
+    check_applets_from_busybox "$initrd" ash cp ip ls mv mkdir sleep tr || ret=1
 
     # switch_root must NOT be a busybox symlink as the base module reinstalls
     # the host util-linux version on top of any symlink the busybox module left
@@ -50,6 +56,15 @@ test_run() {
         echo "FAIL: switch_root is a busybox symlink, host version was not preserved" >&2
         ret=1
     fi
+
+    # repeat the same test, but override the list of applets
+    rm "$initrd"
+    DRACUT_MODULE_BUSYBOX_LINKS="$(cat "${TESTDIR}/busybox.links")" \
+        test_dracut \
+        --no-hostonly --no-kernel --drivers "" \
+        --modules "base busybox" \
+        "$initrd"
+    check_applets_from_busybox "$initrd" ash cp ip ls mv mkdir silly-serval sleep tr || ret=1
 
     return "$ret"
 }


chewi (approved):
> Would it be an option to look at the host/chroot/sysroot filesystem and discover the list of applets from there (at to match it) ?

There's a fair chance the distro you're building on is not the same distro you're building for. This is especially true for Buildroot. Even then, we've generally avoid taking anything from the host when cross building. I think that between `DRACUT_MODULE_BUSYBOX_LINKS`, QEMU likely just working if BusyBox is static, and the proposed BusyBox introspection feature, we're sufficiently covered.

Tests aside, I think this looks good.

modules.d/81busybox/module-setup.sh
bdrung: You would still skip applets that are not available on the host. Here is the fix for that: https://github.com/dracut-ng/dracut/pull/2653

airtower-luna: You're right, anything that doesn't exist on the path would be skipped. I'll rebase after that PR is merged.

airtower-luna: And rebased. :slightly_smiling_face:

modules.d/81busybox/module-setup.sh
```diff
@@ -13,15 +13,24 @@ check() {
...
     local _dstdir="${dstdir:-"$initdir"}"
     local _progs=()
     _busybox=$(find_binary busybox)
     _busybox_path="/usr/bin/busybox"
 
     # do not depend on CONFIG_FEATURE_INSTALLER
     # install busybox symlinks manually
-    for _path in $($_busybox --list-all); do
+    if [[ $DRACUT_MODULE_BUSYBOX_LINKS ]]; then
+        _busybox_applets="$(cat "$DRACUT_MODULE_BUSYBOX_LINKS")"
```
bdrung: Why does the env variable point to a file? Shouldn't the env variable just contain the paths (space separated)?

airtower-luna: That's what my initial draft did, until you asked if we can use something from the build and I discovered the `busybox.links` file. :wink: 

Would be easy enough to change though.

bdrung: Ah, okay. The `busybox.links` file is not part of the Debian/Ubuntu busybox packages. So it is not generally be usable. When asked for "something from the build" I had something like https://github.com/uapi-group/specifications/issues/224 in mind.

airtower-luna: It is generally usable in cross-compile contexts, people who use distro packages will have to figure out how to make the list in any case as long as there's no metadata. :wink: 

Anyway, it's no big difference to me, do you want me to change it?

bdrung: Yes. Please.

airtower-luna: Done.

devkontrol (dismissed):

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
```diff
@@ -0,0 +1,51 @@
...
+set -eu
+
+# shellcheck disable=SC2034
+TEST_DESCRIPTION="override list of busybox applets using DRACUT_MODULE_BUSYBOX_LINKS"
+
+test_check() {
+    if ! command -v busybox &> /dev/null; then
+        echo "Test needs busybox on host... Skipping"
+        return 1
+    fi
```
bdrung: `require_binaries_for_test` can be used here now.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env bash
+
```
bdrung: For consistency, please remove this empty line.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
```diff
@@ -0,0 +1,51 @@
...
+    if ! command -v busybox &> /dev/null; then
+        echo "Test needs busybox on host... Skipping"
+        return 1
+    fi
+}
+
+test_setup() {
+    busybox --list-full > "${TESTDIR}/busybox.links"
+    # add an extra link to see if the list is used
+    echo "usr/bin/${TESTDIR##*/}" >> "${TESTDIR}/busybox.links"
```
bdrung: For reproducibility, could you just pick a non-existing name instead of `${TESTDIR##*/}`?

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
```diff
@@ -0,0 +1,51 @@
...
+}
+
+test_run() {
+    set -x
```
bdrung: I am not a big fan of using `set -x`. If you want to keep `set -x` it should probably be hidden behind `V=2`.

.github/workflows/daily-basic.yml
```diff
@@ -53,6 +53,7 @@ jobs:
                     - "80"
                     - "82"
                     - "83"
+                    - "84"
```
bdrung: The number 84 is in the wrong range. see test/README.md.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
```diff
@@ -0,0 +1,51 @@
...
+        "$initrd"
+
+    # Check for the test link and basename. Match layout-agnostically: the
+    # applet may live under bin/ or usr/bin/, and the symlink target may be
+    # relative (e.g. bin/basename -> ../usr/bin/busybox on alpine) or just
+    # "busybox"
+    local _applet ret=0
+    local _listing
+    _listing=$(lsinitrd "$initrd")
+    for _applet in basename ${TESTDIR##*/}; do
```
bdrung: Could this test check one or two more applets besides just basename?

bdrung (commented):
I think the commit "fix(busybox): allow overriding list of applets" is more a feature than a fix.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
airtower-luna: That's actually exactly why I'm using `${TESTDIR##*/}` here: It's extremely unlikely there'll actually be a busybox applet by that name. Would you prefer some silly static name?

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
airtower-luna: I tried, but got CI failures. It's difficult to find something that's both enabled in all the distros (e.g. opensuse doesn't have `ash`, which all the others seem to have) and also not shadowed by some other binary. Do you have a suggestion?

.github/workflows/daily-basic.yml
airtower-luna: Would "10-19: core modules" be the right one?

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
bdrung: Yes, I prefer a silly static name that is easy to grep for and that makes in clear that this is used for testing.

.github/workflows/daily-basic.yml
bdrung: Yes. That fits the best. Sadly we currently do not differentiate between unit, integration, and system tests.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
airtower-luna: From the logs `base64` and `top` seem to be available everywhere, I hope I didn't miss anything there.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
airtower-luna: Done.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
airtower-luna: Done.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
airtower-luna: Done.

.github/workflows/daily-basic.yml
airtower-luna: Done (using 15).

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
airtower-luna: Done.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
bdrung: Opensuse has `ash` (at least according to the busybox output):
```
$ podman run -it --rm ghcr.io/dracut-ng/opensuse:latest busybox --list | tr -d '\r' | grep '^a\?sh$' 
ash
sh
```

Small hacky script:
```
for x in $(podman images | grep dracut-ng | sed 's/ \+/:/g' | cut -d: -f 1,2); do y=$(basename "$x"); podman run -it --rm "$x" busybox --list | sort > "$y"; done
files=$(ls -1 | wc -l); cat * | sort | uniq -c | grep -e "^ *${files} " | cut -c 9-
```
I have 14 dracut docker containers on my machine. So maybe I miss some.

The list is long:
```
arch
ash
awk
base64
basename
blkid
brctl
bunzip2
bzcat
bzip2
cat
chgrp
chmod
chown
chroot
chvt
clear
cmp
cp
cut
date
dd
depmod
df
diff
dirname
dmesg
dnsdomainname
du
echo
egrep
env
expand
expr
factor
false
fatattr
fdisk
fgrep
find
findfs
free
getopt
grep
groups
gunzip
gzip
head
hexdump
hostname
id
ifconfig
insmod
install
ip
kill
killall
less
link
linux32
linux64
ln
loadkmap
losetup
ls
lsmod
lzcat
md5sum
mkdir
mkfifo
mknod
mkswap
mktemp
modinfo
modprobe
more
mount
mv
nc
netstat
nl
nproc
nslookup
paste
pidof
ping
ping6
printf
ps
pwd
readlink
realpath
renice
reset
rm
rmdir
rmmod
sed
seq
setpriv
setsid
sh
sha1sum
sha256sum
sha3sum
sha512sum
shred
shuf
sleep
sort
stat
strings
stty
switch_root
sync
sysctl
tac
tail
tar
tee
test
time
top
touch
tr
true
truncate
tty
udhcpc
umount
uname
unexpand
uniq
unlink
unshare
unxz
unzip
uptime
vi
wc
wget
which
whoami
xargs
xxd
xzcat
yes
zcat
```

bdrung (dismissed):

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
airtower-luna: I see the same output from `busybox --list`, but no link in the initramfs, which is odd.

airtower-luna: Thanks for insisting, that turned up a bug in the final fallback (in case `--list-full` doesn't give paths, and there's no installed link/binary):

https://github.com/dracut-ng/dracut/blob/10fec41461dce3d81ad726b3573e5be6dcaffb63/modules.d/81busybox/module-setup.sh#L37-L46

Line 44 had `_path="usr/bin/${_p}"`, which is empty if we get to that point. Fixed now, and I've added `ash` to the test.

bdrung: So it would be useful to also check more applets that we would expect in the initrd like sh, find, grep, cut, tr, etc. In other words check for the applets in the list above but exclude applets that are not used in the initrd.

man/dracut.8.adoc
```diff
@@ -765,6 +765,20 @@ _SYSTEMCTL_::
 _NM_VERSION_::
     overrides the NetworkManager version. Used for **--sysroot**.
 
+_DRACUT_MODULE_BUSYBOX_LINKS_::
```
bdrung: Could we sort it alphabetical? Since it is not fully sorted alphabetical, I would place it after DRACUT_INSTALL_LOG_LEVEL

modules.d/81busybox/module-setup.sh
```diff
@@ -13,15 +13,38 @@ check() {
...
-    local _path _busybox _busybox_path
+    local _path _p _busybox _busybox_applets _busybox_path
     local _dstdir="${dstdir:-"$initdir"}"
     local _progs=()
     _busybox=$(find_binary busybox)
     _busybox_path="/usr/bin/busybox"
 
-    # do not depend on CONFIG_FEATURE_INSTALLER
+    # get list of available applets
+    if [[ $DRACUT_MODULE_BUSYBOX_LINKS ]]; then
```
bdrung: Just in case:

```suggestion
    if [[ ${DRACUT_MODULE_BUSYBOX_LINKS-} ]]; then
```

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
bdrung: At least check one applet per /usr/bin, /usr/sbin, /sbin, /bin.

airtower-luna: That's tricky, because those are also the most likely to be shadowed by something else, rather than links to busybox. That would be different after #2451 or a similar change lands.

Of course a general check if required tools are present could also be useful, but that'd be less about the busybox module. :wink:

airtower-luna: > At least check one applet per /usr/bin, /usr/sbin, /sbin, /bin.

I cannot check that in the initramfs, because merged-usr/merged-bin are going to change the paths. Best I can do is test one from each category in the `busybox --list-full` output (on Debian). Going to try with: `arp ash basename nologin top`, I hope that doesn't lead to a failure in CI because one is missing/shadowed on some distro.

If shadowing is an issue, would you approve skipping an applet if a non-link exists?

man/dracut.8.adoc
airtower-luna: Done.

modules.d/81busybox/module-setup.sh
airtower-luna: Sure.

test/TEST-84-BUSYBOX-APPLETS-LIST/test.sh
airtower-luna: I've checked, there isn't a single `usr/sbin/` applet that is available and not shadowed in all images. `nbd-client` and `arping` are very close (missing on void), so I guess the best I can do is add a condition for that in `test_check`. `chroot` is shadowed on a bunch of distros (fedora, centos, azurelinux).

airtower-luna: Found the next best option I think, skip checking `arping` if it's not built into busybox.

airtower-luna: With #2689 merged we could probably drop the check and use `chroot`.

devkontrol: > With #2689 merged we could probably drop the check and use `chroot`.

Ok, will review #2689 first !

bdrung: After looking at the list myself, I think we can drop checking `usr/sbin` and only test binaries in /usr/bin, /sbin, /bin.

airtower-luna: Looking at #2689 I'm just going to use the list from there, plus `ash` (because it caught a bug) and the test link. :slightly_smiling_face:

devkontrol (approved):

bdrung (approved):

PR #2650, opened 2026-08-25
fix(tests): don't fail a passing test when server teardown races

prabhakarpujeri:
## Problem

The integration-test runner arms `trap print_test_result ERR` around the whole `setup`+`run`+`cleanup` subshell, so a failure in the **teardown** phase turns an already-passing test red. The teardown steps are not robust against the daemonized server:

- `qemu -daemonize -pidfile …/server.pid` removes the pidfile itself when the daemon exits, which races the test's `kill`/`rm` cleanup.
- `print_test_result()`'s success path used a bare `rm` of the test log, and `client_test_end()` / `stop_webserver()` used bare `rm` / an unguarded `kill` — each can fail on a file/process that's already gone.

Observed in CI (e.g. TEST-71 on ubuntu:devel) and reproducible locally: all functional boots pass (`All tests passed [OK]`), then a cleanup `rm`/`kill` fails with `No such file or directory` / `No such process`, the ERR trap fires, and the job reports `[FAILED]`.

## Fix

- `test/test-functions`: in the `--all` flow, disable the ERR trap and `set -e` for the best-effort cleanup tail (`trap - ERR; set +e; …; exit 0`) so a teardown race cannot flip a passing test; make the teardown `rm`s `rm -f`; make `stop_webserver`'s `kill` tolerant.
- Add a shared `stop_server()` helper that tolerates the pidfile already being removed (or holding a dead pid) and use it in TEST-60-NFS, TEST-70-ISCSI, TEST-71-ISCSI-MULTI and TEST-72-NBD instead of the duplicated inline blocks.
- This also fixes the leftover-server check in TEST-60-NFS and TEST-72-NBD, which tested `server.pid` relative to the CWD (where it never exists) instead of `$TESTDIR`, so a daemonized server from an interrupted run was never killed before the next one started.

## Verification

- Unit-tested `stop_server()` under `set -eu` for: missing pidfile, empty pidfile, live pid, dead pid, and repeated calls — all return 0.
- Exercised the real `--all` dispatch with the actual `test-functions`: (a) run passes + cleanup races/fails → now `[OK]`/exit 0 (previously `[FAILED]`); (b) run fails → still exactly one `[FAILED]`/exit 1.
- Ran TEST-60-NFS, TEST-70-ISCSI and TEST-71-ISCSI-MULTI in the `ghcr.io/dracut-ng/ubuntu:devel` container with KVM: all three `[OK]`.
- `make syncheck` clean; `bash -n` clean on all touched files.

### DIFF
--- a/test/test-functions
+++ b/test/test-functions
@@ -303,6 +303,22 @@ wait_for_server_startup() {
     fi
 }
 
+# Stop the daemonized server QEMU whose pidfile is in $TESTDIR.
+# QEMU removes the pidfile itself when the daemon exits, so the file may
+# already be gone (or hold a dead pid); neither case must fail the test.
+# Only remove the pidfile when the process is already gone, otherwise QEMU
+# deletes it when it exits.
+stop_server() {
+    local server_pid
+    server_pid=$(cat "$TESTDIR"/server.pid 2> /dev/null) || server_pid=
+    [[ -n $server_pid ]] || return 0
+    if ! kill "$server_pid"; then
+        # process already gone, drop the now-stale pidfile
+        rm -f -- "$TESTDIR"/server.pid
+    fi
+    return 0
+}
+
 command -v test_check &> /dev/null || test_check() {
     :
 }
--- a/test/test-functions
+++ b/test/test-functions
@@ -205,7 +205,7 @@ client_test_end() {
     [[ -e "$TESTDIR/client_test_name.txt" ]] || return 0
     test_name=$(cat "$TESTDIR/client_test_name.txt")
     echo "CLIENT TEST END: $test_name [$status]"
-    rm "$TESTDIR/client_test_name.txt"
+    rm -f -- "$TESTDIR/client_test_name.txt"
 }
 
 require_binaries_for_test() {
@@ -269,8 +269,8 @@ stop_webserver() {
     if [[ -s "$TESTDIR/webserver.pid" ]]; then
         pid=$(cat "$TESTDIR/webserver.pid")
         echo "Stopping web server (pid $pid)..." >&2
-        kill "$pid"
-        rm "$TESTDIR/webserver.pid"
+        kill "$pid" || true
+        rm -f -- "$TESTDIR/webserver.pid"
     fi
 }
 
@@ -428,7 +428,7 @@ print_test_result() {
     local ret=${1-1}
     if [ "$ret" -eq 0 ]; then
         client_test_end
-        rm -- test${TEST_RUN_ID:+-$TEST_RUN_ID}.log
+        rm -f -- test${TEST_RUN_ID:+-$TEST_RUN_ID}.log
         echo -e "TEST: $TEST_DESCRIPTION " "$COLOR_SUCCESS" "[OK]" "$COLOR_NORMAL"
     else
         client_test_end "FAILED"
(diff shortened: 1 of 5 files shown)


test/test-functions
```diff
@@ -303,6 +303,17 @@ wait_for_server_startup() {
     fi
 }
 
+# Stop the daemonized server QEMU and remove its pidfile.
+# QEMU removes the pidfile itself when the daemon exits, so it may already
+# be gone (or hold a dead pid); neither case may fail the test.
+stop_server() {
```
bdrung, line 311: This function is added to a commit that does not use it.

test/test-functions
```diff
@@ -269,8 +269,8 @@ stop_webserver() {
     if [[ -s "$TESTDIR/webserver.pid" ]]; then
         pid=$(cat "$TESTDIR/webserver.pid")
         echo "Stopping web server (pid $pid)..." >&2
-        kill "$pid"
-        rm "$TESTDIR/webserver.pid"
+        kill "$pid" 2>/dev/null || true
```
bdrung: Please do not redirect the output to /dev/null.

test/test-functions
```diff
@@ -303,6 +303,17 @@ wait_for_server_startup() {
     fi
 }
 
+# Stop the daemonized server QEMU and remove its pidfile.
+# QEMU removes the pidfile itself when the daemon exits, so it may already
+# be gone (or hold a dead pid); neither case may fail the test.
+stop_server() {
+    local server_pid
+    [[ -s $TESTDIR/server.pid ]] || return 0
+    server_pid=$(cat "$TESTDIR"/server.pid 2>/dev/null) || server_pid=
```
bdrung: This can fail with TOCTOU

bdrung (changes requested):
Can you squash the style fix changes into the relevant commits?

test/test-functions
```diff
@@ -489,15 +489,24 @@ while (($# > 0)); do
                 (
                     test_setup
                     test_run
+                    # teardown is best effort: the daemonized QEMU removes its
```
devkontrol, line 508: Do we need to do the same for the `--clean` case as well ?

test/test-functions
```diff
@@ -303,6 +303,16 @@ wait_for_server_startup() {
...
 }
 
+# Stop the daemonized server QEMU and remove its pidfile.
+# QEMU removes the pidfile itself when the daemon exits, so it may already
+# be gone (or hold a dead pid); neither case may fail the test.
+stop_server() {
+    local server_pid
+    server_pid=$(cat "$TESTDIR"/server.pid 2> /dev/null) || server_pid=
+    [[ $server_pid ]] && kill "$server_pid" || true
+    rm -f -- "$TESTDIR"/server.pid
```
bdrung: Shouldn't we only remove the file in case the process is gone?

test/test-functions
prabhakarpujeri: Fixed by the history rewrite: stop_server() is now introduced in the same commit that adds its four call sites (currently 8c130507).

test/test-functions
prabhakarpujeri: Dropped the redirection: stop_webserver() now uses `kill "$pid" || true` with the output left visible.

test/test-functions
prabhakarpujeri: Fixed: the `[[ -s ]]` test before reading the pidfile is gone. stop_server() now reads the pidfile in one step and falls back to an empty pid, so there is no check-then-act window.

test/test-functions
prabhakarpujeri: No change needed for --clean: it runs test_cleanup/test_common_cleanup directly and propagates the exit status, which is the behavior you want from an explicit clean. The hazard this PR addresses only existed under --all, where teardown ran with `trap print_test_result ERR` armed and could flip a passing test to FAILED. The pidfile race itself is handled inside stop_server(), so the --clean path benefits from the same fix.

test/test-functions
prabhakarpujeri: Yes, and the current version does exactly that: the pidfile is only removed when `kill` reports the process is already gone (stale pid). After a successful kill, QEMU removes its own pidfile on exit, so we leave it alone.

PR #2653, opened 2026-08-25
fix(busybox): use full path provided by busybox --list-all

bdrung:
`busybox --list` only lists the names of the included applets but does not show their paths. The busybox module looks for the path on the host system. It will skip the applet in case the binary is not found on the host system.

Rely on `busybox --list-all` to provide the path for the applets. That way all applets can be installed. Support getting relative paths (like `bin/busybox`) in addition to absolute paths (like `/bin/busybox`).

### Checklist

### DIFF
--- a/modules.d/81busybox/module-setup.sh
+++ b/modules.d/81busybox/module-setup.sh
@@ -13,26 +13,27 @@ check() {
 
 # called by dracut
 install() {
-    local _i _path _busybox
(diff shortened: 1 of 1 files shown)


airtower-luna (approved):
Looks correct to me, thanks.

devkontrol (approved):

PR #2663, opened 2026-08-26
support grep applet from busybox

bdrung:
Use short parameter form instead of long one.

### Checklist


chewi (approved):
Looks good. Be aware that `-a` doesn't do anything in BusyBox as it's effectively always applied.

devkontrol (approved):

PR #2681, opened 2026-08-27
feat(rt): make dracut-initramfs-restore.sh POSIX compatible

airtower-luna:
Most changes replace bash's `[[` with `[` and quote parameters where needed. Also avoid `&>` redirect shorthand.

The one more complex change is in `find_initrd_for_kernel_version()`, where "create array from glob, use element 0 if existing path" is replaced with "iterate over glob, use element if existing path" to do the same thing: unless the glob doesn't match anything, the first element will exist and be used.

This makes `dracut-initramfs-restore.sh` work on systems without bash (e.g. busybox shell).

### Checklist

### DIFF
--- a/dracut-initramfs-restore.sh
+++ b/dracut-initramfs-restore.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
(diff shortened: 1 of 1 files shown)


dracut-initramfs-restore.sh
```diff
@@ -13,17 +13,17 @@ trap 'echo "Received SIGTERM signal, ignoring!" >&2' TERM
 
 KERNEL_VERSION="$(uname -r)"
 
-[[ $dracutbasedir ]] || dracutbasedir=/usr/lib/dracut
+[ "$dracutbasedir" ] || dracutbasedir=/usr/lib/dracut
 
 find_initrd_for_kernel_version() {
     local kernel_version="$1"
-    local base_path files initrd machine_id
+    local base_path _f initrd machine_id
```
bdrung: For consistency: all other variable do not start with an underscore.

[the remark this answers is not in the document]
bdrung: dracut expects the shell to support `local`.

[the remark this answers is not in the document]
devkontrol: @bdrung - could we capture this understanding in the documentation somewhere ?

bdrung: Good point. We have a mixture of using underscore in functions or not. We should be consistent in a single function (but also recommend one preferred way).

airtower-luna: I've changed it to just `f`.

bdrung (approved):
Thanks.

devkontrol (approved):

PR #2682, opened 2026-08-27
fix(base): support readlink applet from busybox

### LINKED ISSUE #2731: Daily Tests systemd fails on ubuntu:devel on arm: systemd-sysext-initrd.service failed
```
made it to the test rootfs!
● systemd-sysext-initrd.service not-found failed failed systemd-sysext-initrd.service
```

See https://github.com/dracut-ng/dracut/actions/workflows/daily-systemd.yml

CC @bdrung

bdrung:
busybox's `readlink` applet does not support the option `-e` option.

These `readlink` calls can be replaced by `realpath`, but `realpath` is not included in the initrd. So support using `readlink` (when busybox is used) and fall back to keep using `readlink` otherwise. Since `realpath` will return successfully on non-existing path, check the path exists if needed.

Fixes: https://github.com/dracut-ng/dracut/issues/2731
Part of: https://github.com/dracut-ng/dracut/issues/2437

### Checklist

### DIFF
--- a/modules.d/80base/dracut-lib.sh
+++ b/modules.d/80base/dracut-lib.sh
@@ -512,9 +512,18 @@ udevproperty() {
     done
 }
 
+_resolve_path() {
+    # busybox readlink does not support the option -e. So rely on the realpath applet.
+    if command -v realpath > /dev/null; then
+        realpath "$@"
+    else
+        readlink -e -q "$@"
+    fi
+}
+
 find_mount() {
     local dev wanted_dev
-    wanted_dev="$(readlink -e -q "$1")"
+    wanted_dev="$(_resolve_path "$1")"
     while read -r dev _ || [ -n "$dev" ]; do
         [ "$dev" = "$wanted_dev" ] && echo "$dev" && return 0
     done < /proc/mounts
@@ -645,7 +654,7 @@ copytree() {
     local src="$1" dest="$2"
     [ -d "$src" ] || return 1
     mkdir -p "$dest" || return 1
-    dest=$(readlink -e -q "$dest") || return 1
+    dest=$(_resolve_path "$dest") || return 1
     (
         cd "$src" || exit 1
         cp -af . -t "$dest"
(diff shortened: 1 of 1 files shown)


modules.d/80base/dracut-lib.sh
```diff
@@ -512,9 +512,17 @@ udevproperty() {
     done
 }
 
+_resolve_path() {
+    if command -v pidof > /dev/null; then
```
devkontrol: pidof ?

bdrung: Ups. Copy-paste bug. Fixed.

modules.d/80base/dracut-lib.sh
```diff
@@ -645,7 +653,7 @@ copytree() {
     local src="$1" dest="$2"
     [ -d "$src" ] || return 1
     mkdir -p "$dest" || return 1
-    dest=$(readlink -e -q "$dest") || return 1
+    dest=$(resolve_path "$dest") || return 1
```
aafeijoo-suse: Typo, missing `_`: `resolve_path` -> `_resolve_path`

bdrung: Thanks. Fixed.

devkontrol (approved):

modules.d/80base/dracut-lib.sh
```diff
@@ -512,9 +512,17 @@ udevproperty() {
     done
 }
 
+_resolve_path() {
+    if command -v realpath > /dev/null; then
```
devkontrol, line 517: I see no dracut module that would copy `realpath` to the generated initrd, so most of the cases, this condition is not mean (just dead code). What am I missing here ?

devkontrol (changes requested):

modules.d/80base/dracut-lib.sh
bdrung: Busybox will install the realpath symlink and that is the case where it is needed. Should I add a comment explaining this?

devkontrol: Yes please. Also perhaps add that this is only the case if busybox was built with realpath enabled . The error handling of this code is now a bit degarded as there is no warning if realpath is actually not enabled applet from busybox.

bdrung: Added a comment. Suggestions how to improve the comment are welcome.

devkontrol (approved):

Conan-Kudo (approved):

PR #2695, opened 2026-08-31
fix(busybox): do not install blkid if udev rules need -o

### LINKED ISSUE #2512: busybox blkid is not compatible with dracut for lvm
Test 20 and Test 26 are failing with the busybox CI container.

They use to pass, this could be a busybox regression as the regression is not triggered by test case changes.

See https://github.com/dracut-ng/dracut/actions/workflows/daily-busybox-x64.yml

Last time these tests passed is on Apr-12 (and the days leading to that day) - https://github.com/dracut-ng/dracut/actions/runs/24319501762 

CC @bdrung @nadzyah

bdrung:
The busybox `blkid` applet does not support the option `-o`. The `13-dm-disk.rules` udev rule might either use the builtin blkid or call `/sbin/blkid -o udev -p $tempnode`. The latter is not supported by busybox.

So check the udev rule to call `blkid` with `-o` and do not install this busybox applet in this case.

Fixes: https://github.com/dracut-ng/dracut/issues/2512

### Checklist

### DIFF
--- a/modules.d/10busybox/module-setup.sh
+++ b/modules.d/10busybox/module-setup.sh
@@ -34,6 +34,12 @@ install() {
 
     # install busybox symlinks manually
     for _path in ${_busybox_applets}; do
+        if [[ ${_path##*/} == blkid ]] && grep -qr "blkid -o" "${udevdir}/rules.d" 2> /dev/null; then
+            # The busybox blkid applet does not support the option -o.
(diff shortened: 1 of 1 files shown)


devkontrol (approved):

Conan-Kudo (approved):
This kind of stuff is sketchy, but yeah, I don't see a better solution.

PR #2696, opened 2026-08-31
fix(busybox): do not install mount applet

bdrung:
The busybox `mount` applet does not resolve "auto" to the actual filesystem and therefore passes `auto` directly to mount(2). If the target partition's filesystem driver (e.g., `ext4`, `xfs`) is compiled as a module and not yet loaded into `/proc/filesystems`, the kernel skips `request_module()` and returns `EINVAL` (Invalid argument). For example `test/test.sh debian:sid 10` fails with, because `ext4` is not loaded:

```
systemd[1]: Mounting sysroot.mount - /sysroot...
mount: mounting /dev/disk/by-label/dracut on /sysroot failed: Invalid argument
```

Instead of adding a quirky workaround to probe the device and load the relevant filesystem driver, do not install the `mount` applet.

### Checklist

### DIFF
--- a/modules.d/10busybox/module-setup.sh
+++ b/modules.d/10busybox/module-setup.sh
@@ -39,6 +39,11 @@ install() {
             # See https://github.com/vda-linux/busybox_mirror/issues/33
             continue
         fi
+        if [[ ${_path##*/} == mount ]]; then
+            # The busybox mount applet does not resolve "auto" to the actual filesystem.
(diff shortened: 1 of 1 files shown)


devkontrol (approved):
Ok.

1./ I would do the same with blkid, not just with mount.
2./ Make sure it is really only mount the issue and not, e.g. kmod

PR #2697, opened 2026-09-01
test(BUSYBOX): also test booting

bdrung:
Extend the busybox test case to test that booting the initrd works.

This PR needs https://github.com/dracut-ng/dracut/pull/2696 to succeed.

### Checklist

### DIFF
--- a/test/TEST-10-BASIC/test.sh
+++ b/test/TEST-10-BASIC/test.sh
@@ -23,7 +23,8 @@ test_setup() {
     build_client_rootfs "$TESTDIR/rootfs"
     build_ext4_image "$TESTDIR/rootfs" "$TESTDIR"/root.img dracut
 
-    test_dracut
+    # Busybox is tested in a separate test case
+    test_dracut --omit busybox
 }
 
 # shellcheck disable=SC1090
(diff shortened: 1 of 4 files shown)


devkontrol (approved):
Thanks @bdrung

test/TEST-10-BASIC/test.sh
```diff
@@ -23,7 +23,8 @@ test_setup() {
     build_client_rootfs "$TESTDIR/rootfs"
     build_ext4_image "$TESTDIR/rootfs" "$TESTDIR"/root.img dracut
 
-    test_dracut
```
devkontrol, line 26: @bdrung I think this hunk should be reverted as it will prevent CI continuers that only has `busybox` installed to pass this test - e.g. https://github.com/dracut-ng/dracut/actions/workflows/daily-busybox-x64.yml

Regardless of the regression, the spirit of this test is to test default config, I do not think the test case should force a particular config.

bdrung: `test/test.sh alpine:busybox-edge 10` still works, because it does not need the busybox module. This container uses the busybox applets on the host.

PR #2699, opened 2026-09-01
Install dracut-cpio to bindir

airtower-luna:
As discussed in #2588, installing tools used to build an initramfs to `bindir` and expecting them on `PATH` makes it easier to replace them in a cross-build situtation, specifically: if dracut is installed in a cross-sysroot (so running scripts from the sysroot works, but running executables does not). If tools are installed to `PATH` dracut can transparently use tools installed on the host, and users can adjust `PATH` if they want a specific version, or have dracut installed in a non-default location (e.g. a development build instead of installed distro package).

This PR moves `dracut-cpio`, and adds a manpage for it (based on its `--help` output).

### Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,6 +59,10 @@ man8pages = man/dracut.8 \
             modules.d/77dracut-systemd/dracut-pre-udev.service.8 \
             modules.d/77initqueue/dracut-initqueue.service.8
 
+ifeq ($(enable_dracut_cpio),yes)
+man8pages += man/dracut-cpio.8
+endif
+
 manpages = $(man1pages) $(man5pages) $(man7pages) $(man8pages)
 
 .PHONY: install clean distclean archive test all check AUTHORS CONTRIBUTORS doc
--- a/Makefile
+++ b/Makefile
@@ -257,7 +257,7 @@ endif
 		install -m 0755 dracut-util $(DESTDIR)$(pkglibdir)/dracut-util; \
 	fi
 ifeq ($(enable_dracut_cpio),yes)
-	install -m 0755 dracut-cpio $(DESTDIR)$(pkglibdir)/dracut-cpio
+	install -m 0755 dracut-cpio $(DESTDIR)$(bindir)/dracut-cpio
 endif
 	mkdir -p $(DESTDIR)${prefix}/lib/kernel/install.d
 	install -m 0755 install.d/50-dracut.install $(DESTDIR)${prefix}/lib/kernel/install.d/50-dracut.install
--- a/dracut.sh
+++ b/dracut.sh
@@ -2088,7 +2088,7 @@ if command -v 3cpio > /dev/null; then
 fi
 
 if [[ $enhanced_cpio == "yes" ]]; then
-    enhanced_cpio="$dracutbasedir/dracut-cpio"
+    enhanced_cpio="$(command -v dracut-cpio)"
     if [[ $threecpio_help_output == *--data-align* ]]; then
         # align based on statfs optimal transfer size
         cpio_align=$(stat -f -c "%s" -- "$initdir")
--- a/dracut.sh
+++ b/dracut.sh
@@ -175,9 +175,9 @@ Creates initial ramdisk images for preloading modules
   --tmpdir [DIR]        Temporary directory to be used instead of default
                          ${TMPDIR:-/var/tmp}.
   -r, --sysroot [DIR]   Specify sysroot directory to collect files from.
-  -l, --local           Local mode. Use modules from the current working
-                         directory instead of the system-wide installed in
-                         /usr/lib/dracut/modules.d.
+  -l, --local           Local mode. Use modules and executables from the
+                         current working directory instead of the system-wide
+                         install in /usr/lib/dracut/modules.d.
                          Useful when running dracut from a git checkout.
   -H, --hostonly        Host-only mode: Install only what is needed for
                          booting the local host instead of a generic host.
@@ -860,8 +860,12 @@ while :; do
                     ;;
                 -l | --local)
                     allowlocal="yes"
-                    [[ -f "$(readlink -f "${0%/*}")/dracut.sh" ]] \
-                        && dracutbasedir="$(readlink -f "${0%/*}")"
+                    if [[ -f "$(readlink -f "${0%/*}")/dracut.sh" ]]; then
+                        dracutbasedir="$(readlink -f "${0%/*}")"
+                        # dracutbasedir is repository, add to PATH to use built
+                        # executables
+                        export PATH="${dracutbasedir}:${PATH}"
+                    fi
                     ;;
                 -H | --hostonly | --host-only)
                     hostonly_l="yes"
--- a/man/dracut-cpio.8.adoc
+++ b/man/dracut-cpio.8.adoc
@@ -1,0 +1,57 @@
+DRACUT-CPIO(8)
+==============
+:doctype: manpage
+:man source:   dracut
+:man manual:   dracut
+
+NAME
+----
+dracut-cpio - builds cpio archives, optimized for copy-on-write filesystems
+
+SYNOPSIS
+--------
+**dracut-cpio** [_OPTIONS_...] _OUTPUT_
+
+DESCRIPTION
+-----------
+dracut-cpio builds cpio archives, optimized for copy-on-write filesystems by
+using the man:copy_file_range[2] syscall via Rust's io::copy(). See the
+man:dracut[8] **--enhanced-cpio** option. _OUTPUT_ is the path of the cpio
+archive to create. The list of files to include is read from stdin.
+
+**Note:** This is a dracut internal tool, do not expect a stable interface
+across versions.
+
+OPTIONS
+-------
+**--data-align** _ALIGNMENT_::
+    Attempt to pad archive to achieve _ALIGNMENT_ (in bytes) for file data.
+
+**-0**, **--null**::
+    Expect null delimiters in stdin filename list instead of newline.
+
+**--mtime** _EPOCH_::
+    Use _EPOCH_ for archived mtime instead of filesystem reported values.
+
+**--owner** _UID:GID_::
+    Use _UID_ and _GID_ instead of filesystem reported owner values.
+
+**--truncate-existing**::
+    Truncate and overwrite any existing _OUTPUT_ file, instead of appending.
+
+**-h**, **--help**::
+    Print help message.
+
+EXAMPLE
+-------
+
+`cd fs-tree && find . -print0 | dracut-cpio -0 ../archive.cpio`
+
+AVAILABILITY
+------------
+The dracut-cpio command is part of the dracut package and is available from
+link:$$https://github.com/dracut-ng/dracut$$[https://github.com/dracut-ng/dracut]
+
+SEE ALSO
+--------
+man:dracut[8]
--- a/man/dracut.8.adoc
+++ b/man/dracut.8.adoc
@@ -387,8 +387,8 @@ Default:
     _/var/log/dracut.log_
 
 **-l, --local**::
-    Activates the local mode. dracut will use modules from the current working
-    directory instead of the system-wide installed modules in
+    Activates local mode. Dracut will use modules and executables from the
+    current working directory instead of the system-wide installed modules in
     _/usr/lib/dracut/modules.d_.
     This is useful when running dracut from a git checkout.
 
--- a/test/TEST-82-DRACUT-CPIO/test.sh
+++ b/test/TEST-82-DRACUT-CPIO/test.sh
@@ -8,10 +8,7 @@ TEST_DESCRIPTION="kernel cpio extraction tests for dracut-cpio"
 # see dracut-cpio source for unit tests
 
 test_check() {
-    if ! [[ -x "$PKGLIBDIR/dracut-cpio" ]]; then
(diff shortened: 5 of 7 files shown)


test/TEST-82-DRACUT-CPIO/test.sh
bdrung, line 12: Could we tweak `PATH` and then just use `require_binaries_for_test`?

dracut.sh
```diff
@@ -2071,7 +2071,11 @@ if command -v 3cpio > /dev/null; then
 fi
 
 if [[ $enhanced_cpio == "yes" ]]; then
-    enhanced_cpio="$dracutbasedir/dracut-cpio"
+    if [[ -x "${BASH_SOURCE[0]%/*}/dracut-cpio" ]]; then
+        enhanced_cpio="${BASH_SOURCE[0]%/*}/src/dracut-cpio/target/release/dracut-cpio"
+    else
+        enhanced_cpio="$(command -v dracut-cpio)"
+    fi
```
bdrung: We move the code to `PATH` to make the code simpler, don't we?

airtower-luna: Not really, the point discussed in #2588 was to avoid needing a separate override (`DRACUT_BINDIR`) for the location of executables. Other than that I'm trying not to break anything else, so I'm keeping the option to use build artifacts over `PATH`. :wink:

test/TEST-82-DRACUT-CPIO/test.sh
airtower-luna: Technically, I see no problem. In general the test functions seem to avoid modifying global environment though. I guess checking in `test/test-functions` if `$basedir` looks like a repository and extending `PATH` if yes might work.

airtower-luna: Oh, but there's another problem: Extending `PATH` and `require_binaries_for_test dracut-cpio` would also pass if the build didn't actually build `dracut-cpio`, but it's already installed, say, from a distro package (not going to be the case *now*, but once a release has been made).

airtower-luna: So I think it's probably best to keep the strict preference for build artifacts in their location (also in dracut.sh, lsinitrd.sh), if only to make sure test are actually using them.

airtower-luna: Okay, switched to updating `PATH` and `require_binaries_for_test`.

man/dracut-cpio.8.adoc
```diff
@@ -0,0 +1,55 @@
+DRACUT-CPIO(8)
```
bdrung, line 1: Maybe the section should be 1?

Makefile
```diff
@@ -50,6 +50,7 @@ man7pages = man/dracut.cmdline.7 \
 
 man8pages = man/dracut.8 \
             man/dracut-catimages.8 \
+            man/dracut-cpio.8 \
```
bdrung: This man page should not be installed in case dracut-cpio is not built/installed.

man/dracut-cpio.8.adoc
airtower-luna: I think one could argue either, it's neither strictly an admin-only tool, nor a "normal" user tool either because it's mostly meant for dracut internal use. But sure, moved.

Makefile
airtower-luna: Right, made it conditional.

man/dracut-cpio.1.adoc
```diff
@@ -0,0 +1,55 @@
...
+using the man:copy_file_range[2] syscall via Rust's io::copy(). See the
+man:dracut[8] **--enhanced-cpio** option. _OUTPUT_ is the path of the cpio
+archive to create. The list of files to include is read from stdin.
+
+OPTIONS
+-------
+**--data-align** _ALIGNMENT_::
+    Attempt to pad archive to achieve ALIGNMENT (in bytes) for file data.
+
+**-0, --null**::
```
bdrung:
```suggestion
**-0**, **--null**::
```

man/dracut-cpio.8.adoc
bdrung: @Conan-Kudo @devkontrol What do you think?

I have no strong opinion about this. Should we somehow mention that these tools are meant to be dracut internals?

devkontrol: no opinion on this at this time, sorry.

airtower-luna: I guess the key question is if you want/need to warn people not to assume a stable interface. :thinking:

man/dracut-cpio.1.adoc
airtower-luna: Changed.

man/dracut-cpio.8.adoc
airtower-luna: How about I add something like the following and we call it done? I mostly want to get cross-builds working, if you want to guarantee stability later the note would be easy to remove.

> **Note**: This is a dracut internal tool, do not expect a stable interface across versions.

bdrung: Yes. That note is useful. So add this note and keep it in section 8?

airtower-luna: Sure, will do that.

airtower-luna: And done. :slightly_smiling_face:

devkontrol (dismissed):

man/dracut-cpio.8.adoc
```diff
@@ -0,0 +1,57 @@
...
+**--truncate-existing**::
+    Truncate and overwrite any existing _OUTPUT_ file, instead of appending.
+
+**-h**, **--help**::
+    Print help message.
+
+EXAMPLE
+-------
+
+`find fs-tree/ | dracut-cpio archive.cpio`
```
bdrung: You probably want to cd into the fs-tree directory:
```suggestion
`cd fs-tree/; find . | dracut-cpio ../archive.cpio`
```

airtower-luna: Depends on what a user wants, but yeah, that is closer to building from a sysroot dir. Changed.

devkontrol (dismissed):

dracut.sh
```diff
@@ -175,9 +175,9 @@ Creates initial ramdisk images for preloading modules
   --tmpdir [DIR]        Temporary directory to be used instead of default
                          ${TMPDIR:-/var/tmp}.
   -r, --sysroot [DIR]   Specify sysroot directory to collect files from.
-  -l, --local           Local mode. Use modules from the current working
-                         directory instead of the system-wide installed in
-                         /usr/lib/dracut/modules.d.
+  -l, --local           Local mode. Use modules and executables from the
+                         current working directory instead of the system-wide
+                         install in /usr/lib/dracut/modules.d.
```
bdrung, line 180: man/dracut.8.adoc should get the same change.

test/test-functions
```diff
@@ -12,11 +12,21 @@ export TESTDIR
 export BOOT_ROOT="$TESTDIR"
 export QEMU_LOGFILE="$TESTDIR/qemu.log"
 
-if [[ -z ${basedir-} ]]; then basedir="$(realpath ../..)"; fi
+if [[ -z ${basedir-} ]]; then
+    basedir="$(realpath ../..)"
+else
+    basedir="$(realpath "${basedir}")"
+fi
```
bdrung: Why do you call realpath on a user-specfied path?

man/dracut-cpio.8.adoc
```diff
@@ -0,0 +1,57 @@
...
+man:dracut[8] **--enhanced-cpio** option. _OUTPUT_ is the path of the cpio
+archive to create. The list of files to include is read from stdin.
+
+**Note:** This is a dracut internal tool, do not expect a stable interface
+across versions.
+
+OPTIONS
+-------
+**--data-align** _ALIGNMENT_::
+    Attempt to pad archive to achieve ALIGNMENT (in bytes) for file data.
```
bdrung:
```suggestion
    Attempt to pad archive to achieve _ALIGNMENT_ (in bytes) for file data.
```

man/dracut-cpio.8.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+**-0**, **--null**::
+    Expect null delimiters in stdin filename list instead of newline.
+
+**--mtime** _EPOCH_::
+    Use EPOCH for archived mtime instead of filesystem reported values.
```
bdrung:
```suggestion
    Use _EPOCH_ for archived mtime instead of filesystem reported values.
```

man/dracut-cpio.8.adoc
```diff
@@ -0,0 +1,57 @@
...
+**--truncate-existing**::
+    Truncate and overwrite any existing _OUTPUT_ file, instead of appending.
+
+**-h**, **--help**::
+    Print help message.
+
+EXAMPLE
+-------
+
+`cd fs-tree/; find . | dracut-cpio ../archive.cpio`
```
bdrung:
```suggestion
`cd fs-tree/ && find . -print0 | dracut-cpio -0 ../archive.cpio`
```

Since that example came from the dracut-cpio code, I created https://github.com/dracut-ng/dracut/pull/2736 to change it there as well.

man/dracut-cpio.8.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+AVAILABILITY
+------------
+The dracut-cpio command is part of the dracut package and is available from
+link:$$https://github.com/dracut-ng/dracut$$[https://github.com/dracut-ng/dracut]
```
bdrung, line 53: This link is weirdly formatted. But since we have the same text in other man pages, addressing/fixing this can be done in a future PR.

bdrung (commented):
So. I finally found time to fully review this PR. I found only minor things.

I wouldn't classify the commit "refactor: install dracut-cpio to bindir" as "refactor" because this change can be noticed by the user. So I probably would use "feat" here.

test/test-functions
airtower-luna: I add it to `PATH` a few lines below, and want to be sure it's absolute for that, without forcing the user to provide an already absolute path (which might be inconvenient), and where a relative path works an absolute one should, too.

bdrung: Then I would suggest to call `realpath` when setting `PATH` to make to code easier to read:
```
export PATH="$(realpath "${basedir}"):${PATH}"
```

or:

```
basedir="$(realpath "${basedir}")"
export PATH="${basedir}:${PATH}"
```

man/dracut-cpio.8.adoc
airtower-luna: It looks weird in `man`, too, but because it's that way in the other manpages I assumed it was intended that way for some reason. And yes, I'd very much prefer handling that separately.

man/dracut-cpio.8.adoc
airtower-luna: Seems reasonable, I've adopted that variant in the manpage.

man/dracut-cpio.8.adoc
airtower-luna: Changed.

man/dracut-cpio.8.adoc
airtower-luna: Changed.

dracut.sh
airtower-luna: Done.

test/test-functions
airtower-luna: I originally used the first variant, and then shellcheck complained, so I'm going to use the second one. 😉

man/dracut.8.adoc
```diff
@@ -387,10 +387,10 @@ Default:
     _/var/log/dracut.log_
 
 **-l, --local**::
-    Activates the local mode. dracut will use modules from the current working
-    directory instead of the system-wide installed modules in
-    _/usr/lib/dracut/modules.d_.
-    This is useful when running dracut from a git checkout.
+    Activates local mode. dracut will use modules and executables from the
```
bdrung:
```suggestion
    Activates local mode. Dracut will use modules and executables from the
```

man/dracut.8.adoc
```diff
@@ -387,10 +387,10 @@ Default:
     _/var/log/dracut.log_
 
 **-l, --local**::
-    Activates the local mode. dracut will use modules from the current working
-    directory instead of the system-wide installed modules in
-    _/usr/lib/dracut/modules.d_.
-    This is useful when running dracut from a git checkout.
```
bdrung, line 393: I would keep this sentence on its own line. I created https://github.com/dracut-ng/dracut/issues/2737 to discuss if we want to use sembr.

bdrung (changes requested):

test/test-functions
bdrung: The third one is:
```
PATH="$(realpath "${basedir}"):${PATH}"
export PATH
```

That has less side effects.

man/dracut.8.adoc
airtower-luna: Changed.

man/dracut.8.adoc
airtower-luna: Changed.

test/test-functions
airtower-luna: Makes sense, changed.

bdrung (approved):

devkontrol (approved):

Issue #2688, opened 2026-08-28
Test 60 NFS times out on opensuse:latest


airtower-luna:
The test gets stuck and (in CI) killed after the 15 min timeout. Log shows watchdog failures that are absent from runs with other images:
https://github.com/dracut-ng/dracut/actions/runs/33144973057/job/98763884103#step:4:1162

devkontrol:
CC @dracut-ng/opensuse-maint

devkontrol:
Is https://github.com/dracut-ng/dracut/issues/2522 related ?

airtower-luna:
I'm not sure, but the log indicates kernel version 7.2.0-1, maybe the regression described in the Fedora bug made it into the release? And there is an error message from nfsd during startup: <https://github.com/dracut-ng/dracut/actions/runs/33144973057/job/98763884103#step:4:1200>

bdrung:
Maybe related: https://bugs.launchpad.net/ubuntu/+source/dracut/+bug/2166082

I suspect that the new systemd version causes the regression. Can someone bisect the change in the container that causes it?

mwilck:
So does not seem to be openSUSE-specific. Rather, multiple "cutting edge" rolling distros are affected?

aafeijoo-suse:
If I'm not mistaken, it's a problem that already existed and is now showing up with kernel 7.2.
I submitted a proposed patch: https://lore.kernel.org/linux-nfs/20260902133554.28863-1-antonio.feijoo@suse.com/T/#u

bdrung:
I can also see this on Ubuntu stonking with linux 7.0.0-14.14.

aafeijoo-suse:
> I can also see this on Ubuntu stonking with linux 7.0.0-14.14.

https://autopkgtest.ubuntu.com/results/autopkgtest-stonking/stonking/amd64/d/dracut/20260901_034959_62fd0@/log.gz points to 7.2.0-5, can you confirm that what you see in 7.0.0-14 is the same issue?

bdrung:
You right! Now it makes sense why I could see the failure sometimes, but sometimes not. stonking has 7.0.0-14, but stonking-proposed has 7.2.0-5. The Dracut CI does not have the proposed pocket enabled and therefore still uses linux 7.0.

PR #2714, opened 2026-09-02
test(NFS): fix for #2688

aafeijoo-suse:
### Checklist
- [X]  I have tested it locally

### DIFF
--- a/test/TEST-60-NFS/test.sh
+++ b/test/TEST-60-NFS/test.sh
@@ -225,7 +225,7 @@ make_server_rootfs() {
         --add-confdir test-root \
         -a "bash $USE_NETWORK nfs" \
         --add-drivers "nfsd sunrpc lockd" \
-        -I "exportfs pidof rpc.nfsd rpc.mountd dnsmasq tcpdump nfsdcld" \
+        -I "exportfs pidof rpc.nfsd rpc.mountd dnsmasq tcpdump nfsdcld ${SERVER_DEBUG:+rpcdebug}" \
         --install-optional "/etc/netconfig /etc/nsswitch.conf /etc/rpc /etc/protocols /etc/services /usr/etc/nsswitch.conf /usr/etc/rpc /usr/etc/protocols /usr/etc/services rpc.idmapd /etc/idmapd.conf" \
         -f "$TESTDIR"/initramfs.root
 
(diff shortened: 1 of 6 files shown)


test/TEST-60-NFS/test.sh
```diff
@@ -225,7 +225,7 @@ make_server_rootfs() {
         --add-confdir test-root \
         -a "bash $USE_NETWORK nfs" \
         --add-drivers "nfsd sunrpc lockd" \
-        -I "exportfs pidof rpc.nfsd rpc.mountd dnsmasq tcpdump" \
+        -I "exportfs pidof rpc.nfsd rpc.mountd dnsmasq tcpdump nfsdcld" \
```
bdrung: nfsdcld should be added to the required binaries.

modules.d/74nfs/module-setup.sh
```diff
@@ -52,6 +52,9 @@ cmdline() {
     local nfs_address
     local lookup
 
+    # NFSv4's nfs_layout_flexfiles can pull in an NFSv3 data server at any time
+    printf "rd.driver.pre=nfsv3 "
```
bdrung: This solution would not work on systems that do not use `--hostonly-cmdline`.

test/TEST-60-NFS/test.sh
aafeijoo-suse: Also `rpc.nfsd`, `rpc.mountd`... but all these binaries are installed with `nfs-utils` and none of them are on the list of required binaries.

modules.d/74nfs/module-setup.sh
aafeijoo-suse: `--hostonly-cmdline` is for storing kernel command line options needed in the initrd, i.e., write some conf files in `/etc/cmdline.d/*.conf`. See `70btrfs` or `70multipath`: with `--hostonly-cmdline` some required kernel modules are loaded.

test/TEST-60-NFS/test.sh
bdrung: It doesn't hurt to check for binaries that the test calls directly.

bdrung (approved):
Thanks.

Conan-Kudo (approved):

PR #2728, opened 2026-09-05
fix(busybox): busybox blkid, losetup are not compatible with dracut

devkontrol:
Remove busybox version of blkid and losetup.

Fixes 864029e.

### Checklist

### DIFF
--- a/test/container/Dockerfile-alpine
+++ b/test/container/Dockerfile-alpine
@@ -45,11 +45,12 @@ RUN \
 
 # busybox container
 # replace GNU coreutils, util-linux, kmod binaries with busybox binaries
-# workaround for busybox blkid compatibility, see https://github.com/dracut-ng/dracut/issues/2512
+# blkid compatibility, see https://github.com/vda-linux/busybox_mirror/issues/33
+# losetup compatibility, see https://github.com/vda-linux/busybox_mirror/issues/37
 RUN \
 if [ "$OPTION" == "busybox" ] ; then \
   mkdir /busybox && cd /busybox && /bin/busybox --install -s /busybox ;\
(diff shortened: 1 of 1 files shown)


test/container/Dockerfile-alpine
```diff
@@ -45,14 +45,15 @@ RUN \
 
 # busybox container
 # replace GNU coreutils, util-linux, kmod binaries with busybox binaries
-# workaround for busybox blkid compatibility, see https://github.com/dracut-ng/dracut/issues/2512
+# blkid compatibility, see https://github.com/vda-linux/busybox_mirror/issues/33
+# losetup compatibility, see https://github.com/vda-linux/busybox_mirror/issues/37
 RUN \
 if [ "$OPTION" == "busybox" ] ; then \
   mkdir /busybox && cd /busybox && /bin/busybox --install -s /busybox ;\
+  rm -f /busybox/blkid /busybox/losetup ;\
```
bdrung, line 53: Instead of removing those files, would it make sense to mimic the logic from the busybox dracut module?

devkontrol: > mimic the logic from the busybox

Since the logic can not be reused as is, I assume you meant here to copy filtering out what files gets copied as part of the copying process - instead of filtering out before the copy starts as this PR does it.

I can see how that would make sense, but that would make the code much harder to read (compared to this one liner), and would lead to more or less the same container rootfs. 

That said, if you you make code change proposal, I think we can consider it.

bdrung: Challenge accepted. I think the outcome is readable: https://github.com/dracut-ng/dracut/pull/2730

PR #2732, opened 2026-09-07
Two follow-ups for --prefer

aafeijoo-suse:
Propagate `prefer_dracutmodules` to modules: some internal (e.g. `bluetooth`) and external (e.g. [openSUSE `kdump`](https://github.com/openSUSE/kdump/blob/6e399fc783871b2087b1c0b59a67051b602c02cf/dracut/module-setup.sh#L8)) dracut modules perform certain checks based on the other `*dracutmodules` variables.

### Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -2575,7 +2575,7 @@ done
 
 export initdir dracutbasedir \
     dracutmodules force_add_dracutmodules add_dracutmodules omit_dracutmodules \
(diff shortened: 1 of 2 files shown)


devkontrol (approved):

bdrung (approved):
I thought that we do not need to expose `prefer_dracutmodules`, but we do.

PR #2736, opened 2026-09-07
feat(cpio): use a safer, more realistic example

bdrung:
The dracut-cpio example would create a cpio with all files in a directory named `fs-tree`. Normally users would change the current directory to `fs-tree` and then call `dracut-cpio` from there. Also use null delimiters in the example to make it safer.

### Checklist

### DIFF
--- a/src/dracut-cpio/src/main.rs
+++ b/src/dracut-cpio/src/main.rs
@@ -618,7 +618,7 @@ fn archive_loop<R: BufRead, W: Seek + Write>(
 
 fn params_usage(params: &[Argument]) {
     argument::print_help("dracut-cpio", "OUTPUT", params);
-    println!("\nExample: find fs-tree/ | dracut-cpio archive.cpio\n");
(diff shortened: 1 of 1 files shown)


src/dracut-cpio/src/main.rs
```diff
@@ -618,7 +618,7 @@ fn archive_loop<R: BufRead, W: Seek + Write>(
 
 fn params_usage(params: &[Argument]) {
     argument::print_help("dracut-cpio", "OUTPUT", params);
-    println!("\nExample: find fs-tree/ | dracut-cpio archive.cpio\n");
```
Conan-Kudo, line 621: I think it's valid to document both ways.

Conan-Kudo (changes requested):

src/dracut-cpio/src/main.rs
bdrung: IMO one example is enough for the help output. We can have more examples in the man page.

PR #2738, opened 2026-09-07
fix(functions): do not try to install rules from $dracutbasedir/rules.d

airtower-luna:
The rules.d directory was removed from Dracut source with commit 751c4d43ebf62d2c4f23ce18af9530e820048ed2 (in 2009), existing modules install included rule files from `$moddir`.

I found this while working on #2588: In a cross-build using host dracut installing from `$dracutbasedir` doesn't make sense, and in trying to figure out how to fix it I noticed it doesn't may sense any more in general (see rules from modules above).

### Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1302,6 +1302,7 @@ inst_rules() {
     for _rule in "$@"; do
         unset _found
         if [ "${_rule#/}" = "$_rule" ]; then
+            # _rule contains a relative path
             for r in ${hostonly:+"${dracutsysrootdir-}"/etc/udev/rules.d} "${dracutsysrootdir-}${udevdir}/rules.d"; do
                 [[ -e $r/$_rule ]] || continue
                 _found="$r/$_rule"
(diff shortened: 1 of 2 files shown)


dracut-functions.sh
```diff
@@ -1311,16 +1311,14 @@ inst_rules() {
                 inst_simple "$_found"
             done
         fi
```
bdrung: Could an `elif` be used make it easier to understand?

airtower-luna: Should work, if `[ "${_rule#/}" = "$_rule" ]` matches, `[[ $_rule == /* ]]` won't.

airtower-luna: Works in my test (`dm` module installs `$moddir/11-dm.rules` as expected), changed.

dracut-functions.sh
```diff
@@ -1310,17 +1310,14 @@ inst_rules() {
                 _inst_rule_initqueue "$_found"
                 inst_simple "$_found"
             done
-        fi
-        for r in '' "${dracutsysrootdir-}$dracutbasedir/rules.d/"; do
-            # skip rules without an absolute path
-            [[ "${r}$_rule" != /* ]] && continue
-            [[ -f ${r}$_rule ]] || continue
-            _found="${r}$_rule"
+        elif [[ $_rule == /* ]] && [[ -f $_rule ]]; then
```
bdrung: Isn't `[[ $_rule == /* ]]` obsolete since it's the opposite of `[ "${_rule#/}" = "$_rule" ]`?

airtower-luna: Should be, yes. I've added a comment to the `if` to hopefully make clear how we know `_rule` is absolute in the `elif`.

bdrung (approved):

devkontrol (approved):

PR #2745, opened 2026-09-08
fix(functions): honor DRACUT_INSTALL_PATH in find_binary

airtower-luna:
find_binary is used to find binaries to install into the initramfs, so if `DRACUT_INSTALL_PATH` is set it needs to be used over `PATH`.

Split out of #2588.

### Checklist


bdrung (dismissed):

devkontrol (approved):

bdrung:
>  * `modules.d/73crypt-gpg/module-setup.sh`: runs `gpg` to get version
>  * `modules.d/45url-lib/module-setup.sh`: runs `curl` to get build options (specifically, if NSS is enabled)
>  * `modules.d/70multipath/module-setup.sh`: runs `multipath` to get path of config dir
>  * `modules.d/70mdraid/module-setup.sh`: runs `mdadm` to examine devices in `host_fs_types` in `cmdline` function. It's not obvious to me if that's implicitly skipped if not in hostonly mode (if it is, that should be fine).
> 
> That's definitely out of scope for this PR though, should I create issues for them?

Yes, please.

airtower-luna:
I've created the issues: #2765, #2766, #2767, #2768

airtower-luna:
Is there anything left for me to do here? Those issues aren't caused by the cross-sysroot fixes, it just becomes possible to hit them instead of failing earlier, so I don't think they should delay merging.

PR #2740, opened 2026-09-08
Install dracut-extractinitrd to bindir

airtower-luna:
As discussed in #2588, installing tools used to build an initramfs to `bindir` and expecting them on `PATH` makes it easier to replace them in a cross-build situtation, specifically: if dracut is installed in a cross-sysroot (so running scripts from the sysroot works, but running executables does not). If tools are installed to `PATH` dracut can transparently use tools installed on the host, and users can adjust `PATH` if they want a specific version, or have dracut installed in a non-default location (e.g. a development build instead of installed distro package).

This PR renames `extractinitrd` to `dracut-extractinitrd` (to avoid conflicts with other tools), moves it to `bindir`, and adds a manpage for it (based on its `--help` output).

### Checklist
- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #2747, opened 2026-09-08
feat: set dracutbasedir based on install location

airtower-luna:
`dracutbasedir` is where scripts, modules, and `dracut-util` are loaded from. The default should match the running dracut script, wherever it is installed, assuming the usual `${prefix}/bin/dracut` and `${prefix}/lib/dracut/` locations.

The `--local` option works as before because it overrides `dracutbasedir` before defaults can apply.

Split out of #2588. Together with #2740 and #2745 this should allow running dracut as installed in a cross-sysroot, provided compatible host tools (`dracut-cpio`, `dracut-extractinitrd`, `dracut-install`) are in `PATH`.

### Checklist

### DIFF
--- a/dracut.sh
+++ b/dracut.sh
@@ -1182,7 +1182,6 @@ drivers_dir="${drivers_dir%"${drivers_dir##*[!/]}"}"
 [[ $use_fstab_l ]] && use_fstab=$use_fstab_l
 [[ $mdadmconf_l ]] && mdadmconf=$mdadmconf_l
 [[ $lvmconf_l ]] && lvmconf=$lvmconf_l
-[[ $dracutbasedir ]] || dracutbasedir="${dracutsysrootdir-}"/usr/lib/dracut
 [[ $fw_dir ]] || {
     [[ -e /sys/module/firmware_class/parameters/path ]] \
         && fw_path_para=$(< /sys/module/firmware_class/parameters/path)
--- a/dracut.sh
+++ b/dracut.sh
@@ -997,7 +997,11 @@ export DRACUT_LOG_LEVEL=warning
     debug=yes
 }
 
-[[ ${dracutbasedir-} ]] || dracutbasedir="${dracutsysrootdir-}"/usr/lib/dracut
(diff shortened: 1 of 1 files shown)


dracut.sh
```diff
@@ -997,7 +997,7 @@ export DRACUT_LOG_LEVEL=warning
     debug=yes
 }
 
-[[ ${dracutbasedir-} ]] || dracutbasedir="${dracutsysrootdir-}"/usr/lib/dracut
+[[ ${dracutbasedir-} ]] || dracutbasedir="$(realpath -e "${BASH_SOURCE[0]%/*}/../lib/dracut")"
```
bdrung: Busybox does not provide the `-e` option. We should check the existence of the path afterwards and print a proper error message plus exit code in case of failure.

bdrung (changes requested):
I found a second `[[ $dracutbasedir ]] || dracutbasedir="${dracutsysrootdir-}"/usr/lib/dracut`. We should check why we have it twice and probably remove the second one.

dracut.sh
airtower-luna: Fixed.

dracut.sh
```diff
@@ -997,7 +997,11 @@ export DRACUT_LOG_LEVEL=warning
     debug=yes
 }
 
-[[ ${dracutbasedir-} ]] || dracutbasedir="${dracutsysrootdir-}"/usr/lib/dracut
+[[ ${dracutbasedir-} ]] || dracutbasedir="$(realpath "${BASH_SOURCE[0]%/*}/../lib/dracut")"
+if [[ ! -d $dracutbasedir ]]; then
+    echo "Error: dracut base directory '$dracutbasedir' does not exist." >&2
```
bdrung: To be consistent with the other error code:
```suggestion
    printf "dracut[F]: base directory '%s' does not exist.\n" "$dracutbasedir" >&2
```

bdrung (approved):

devkontrol (approved):

PR #2748, opened 2026-09-08
fix(base): allow loading dracut-lib.sh with set -eu

airtower-luna:
I was testing a custom module, and noticed the following in an initrd script would fail:

```
set -e
set -u
command -v getarg > /dev/null || . /lib/dracut-lib.sh
```

`NEWROOT` and `RD_DEBUG` were used without ensuring they were set (fixed by providing empty defaults), and the exit status of `[ "$RD_DEBUG" = "yes" ] && set -x` wasn't ignored if debug was disabled.

### Checklist

### DIFF
--- a/modules.d/80base/dracut-lib.sh
+++ b/modules.d/80base/dracut-lib.sh
@@ -4,7 +4,7 @@ command -v wait_for_dev > /dev/null || . /lib/dracut-dev-lib.sh
 
 export DRACUT_SYSTEMD
 export NEWROOT
-if [ -n "$NEWROOT" ]; then
+if [ -n "${NEWROOT-}" ]; then
     [ -d "$NEWROOT" ] || mkdir -p -m 0755 "$NEWROOT"
 fi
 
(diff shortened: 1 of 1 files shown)


modules.d/80base/dracut-lib.sh
```diff
@@ -24,7 +24,7 @@ debug_off() {
 }
 
 debug_on() {
-    [ "$RD_DEBUG" = "yes" ] && set -x
+    [ "$RD_DEBUG" = "yes" ] && set -x || :
```
bdrung: I suggest to inverse the logic:
```suggestion
    [ "$RD_DEBUG" != "yes" ] || set -x
```
This is more readable IMO.

airtower-luna: It is, but would change behavior, empty string is != yes here:
https://github.com/dracut-ng/dracut/blob/c8ab9b5388072747f3f73848af0dcb651d42abb1/modules.d/80base/dracut-lib.sh#L873-L884

I'm not sure what's the point of setting `RD_DEBUG=''` there though. `getarg` does an unconditional `debug_off` at the start, so unless I'm missing something it only changes the behavior of `debug_on` at return.

airtower-luna: Sorry, got confused last night, that change works. I've applied it.

bdrung (approved):

devkontrol (approved):

PR #2752, opened 2026-09-10
Allow installing from dracut dir when using sysroot

airtower-luna:
In a cross-build situation where a build host is preparing an initramfs for a target system (which may have a different architecture) there is no reason to install Dracut into the sysroot, and for embedded systems where rootfs size should be kept low it is explicitly undesirable to do so. Installing `dracut-util` (for use in the initramfs) and possibly `dracut-extractinitrd` (if `dracut-initramfs-restore.sh` is used) is sufficient.

Scripts, udev rules, and systemd units (if applicable) need to be installed from the modules directories that are part of Dracut as installed on the host. The new `--dracutdir`/`-B` parameter for `dracut-install` allows installing from the specified directory if `--sysroot` is used, the `inst{,_script,_simple}` helper functions use the new parameter if `dracutsysrootdir` is set. Binaries must still come from the sysroot.

Split out of #2588.

### Checklist

### DIFF
--- a/dracut-functions.sh
+++ b/dracut-functions.sh
@@ -1114,11 +1114,11 @@ inst() {
     fi
     [[ -e ${dstdir}/"${2:-$1}" ]] && return 0 # already there
     [[ ${DRACUT_RESOLVE_LAZY-} ]] || _resolve_deps=1
-    if $DRACUT_INSTALL ${dracutsysrootdir:+-r "$dracutsysrootdir"} ${dstdir:+-D "$dstdir"} ${loginstall:+-L "$loginstall"} ${_resolve_deps:+-l} ${DRACUT_FIPS_MODE:+-f} ${_hostonly_install:+-H} "$@"; then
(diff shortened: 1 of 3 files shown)


chewi (approved):

bdrung:
Do we need to pass `DRACUTDIR` to `dracut-install`? Shouldn't we know which files come from `DRACUTDIR` and specify the path for them explicitly?

airtower-luna:
> Do we need to pass `DRACUTDIR` to `dracut-install`? Shouldn't we know which files come from `DRACUTDIR` and specify the path for them explicitly?

The problem is not that we don't know where the files come from (the modules already install them by absolute path), but that `dracut-install` does not allow installing files from outside the sysrootdir if that is set (which makes sense in principle to avoid mistakes). If we use host dracut (installed outside the sysroot) to build the initrd, we need to allow installing files from its dracutbasedir in addition to the sysroot, so modules can install their scripts, udev rules, etc.

The reason I use `DRACUTDIR` instead of `DRACUTBASEDIR` is that there's a separate patch to allow installing cross-built `dracut-util` from outside the sysrootdir in #2588. Point is, it's a location outside the sysroot that provides dracut components for the initrd.

PR #2753, opened 2026-09-10
Don't use require_binaries for host tools

airtower-luna:
Working on #2745 I found two modules that use `require_binary` to check for tools they use while building the initrd, rather than installing them into it:

* `modules.d/74squash-erofs/module-setup.sh`: `mkfs.erofs`
* `modules.d/74squash-squashfs/module-setup.sh`: `mksquashfs`

Using `require_binary` for both host and target tools cannot work with a cross sysroot, because binaries valid for one purpose will not work for the other. Add a new helper function `require_binaries_host` to check for host tools, and fix the modules.

### Checklist


devkontrol (dismissed):
I find this welcome change as it makes the code more readable as well. 

In theory there might be some out of tree dracut modules impacted as well, but in practice i could not find any and even if there is one, that would just mean that those out of tree dracut modules would fail in sysroot the same way as before this PR, but at least this PR now allows a possible way forward.

devkontrol:
CC @chewi

chewi:
Something isn't making sense to me. `require_binaries` doesn't actually install the binaries. Even as things stand, I cannot see anything that actually installs fsck.erofs or unsquashfs even though they are used by dracut-initramfs-restore.sh.

airtower-luna:
`dracut-initramfs-restore.sh` runs on shutdown (if at all) so the tools don't need to be installed in the initrd, but the matching one (for squashfs or erofs) must be available in the sysroot or it won't work.

chewi (approved):
Ah yes, of course. All good then!

dracut-functions.sh
```diff
@@ -1485,6 +1485,22 @@ require_any_binary() {
...
+# helper function for check() in module-setup.sh
+# to check for required binaries used while building initrd
+# issues a standardized warning message
+require_binaries_host() {
+    local _module_name="${moddir##*/}"
+    local _ret=0
+
+    for cmd in "$@"; do
+        if ! command -v "$cmd" &> /dev/null; then
+            ddebug "Module '${_module_name#[0-9][0-9]}' will not be installed, because host command '$cmd' could not be found!"
```
bdrung: Isn't "will not be installed" misleading since those are used for the host?

The doc says "Issues a standardized warning message" but only ddebug is used here. Shoudn't a warning level be produced?

man/dracut.modules.7.adoc
```diff
@@ -395,15 +395,20 @@ dracut releases.
...
+All of the binaries listed in the arguments of this function must be available in `DRACUT_INSTALL_PATH` if set, otherwise in `PATH`. If using **-sysroot**, the search path is within the sysroot directory.
 Issues a standardized warning message.
 
 ==== require_any_binary
 
-Helper function for check() in module-setup.sh to check for required installed binaries.
-At least one of the binaries listed in the arguments of this function should exists in
-the host.
+Helper function for check() in module-setup.sh to check for required binaries to be installed into initrd.
+At least one of the binaries listed in the arguments of this function must be available in `DRACUT_INSTALL_PATH` if set, otherwise in `PATH`. If using **-sysroot**, the search path is within the sysroot directory.
```
bdrung: `--sysroot` misses one dash and the long line should be wrapped:

```suggestion
At least one of the binaries listed in the arguments of this function must be available in `DRACUT_INSTALL_PATH` if set, otherwise in `PATH`.
If using **--sysroot**, the search path is within the sysroot directory.
```

man/dracut.modules.7.adoc
```diff
@@ -395,15 +395,20 @@ dracut releases.
 
 ==== require_binaries
 
-Helper function for check() in module-setup.sh to check for required installed binaries.
-All of the binaries listed in the arguments of this function should exists in the host.
+Helper function for check() in module-setup.sh to check for required binaries to be installed into initrd.
+All of the binaries listed in the arguments of this function must be available in `DRACUT_INSTALL_PATH` if set, otherwise in `PATH`. If using **-sysroot**, the search path is within the sysroot directory.
```
bdrung:
```suggestion
All of the binaries listed in the arguments of this function must be available in `DRACUT_INSTALL_PATH` if set, otherwise in `PATH`.
If using **--sysroot**, the search path is within the sysroot directory.
```

bdrung (commented):
For the latest two commit can you add a commit description with the reasoning why they need to use `require_binaries_host`?

bdrung:
Side note: It's refreshing to review your PR compared to some LLM generated PRs.

dracut-functions.sh
airtower-luna: > Isn't "will not be installed" misleading since those are used for the host?

The function is supposed to be used in the `check()` function of `module-setup.sh`, and if it fails to find the binaries the module cannot be installed (into the initrd), so I don't think it's wrong. I guess I could change it to something more generic like "cannot be used"?

> The doc says "Issues a standardized warning message" but only ddebug is used here. Shoudn't a warning level be produced?

Maybe? The existing `require_binaries` and `require_any_binary` are inconsistent there, using `ddebug` and `dinfo` with the same statement in documentation. I simply adapted the `require_binaries` function for `require_binaries_host`, if you say this should be a warning I should probably change them all.

bdrung: Ah, the "will not be installed" referred to the Dracut module, not the specified binaries. Now I see the "Module". So discard my comment.

Yes, we should fix that inconsistency (probably in a separate PR). `dinfo` is probably the right level (but that should be tested with real world data).

man/dracut.modules.7.adoc
airtower-luna: Fixed, thank you!

man/dracut.modules.7.adoc
airtower-luna: Fixed.

dracut-functions.sh
airtower-luna: I've changed the message in `require_binaries_host` to warning and "cannot be used". For `require_binaries` and `require_any_binary` I've created #2764 (which can be reviewed and merged independently).

airtower-luna:
> Side note: It's refreshing to review your PR compared to some LLM generated PRs.

Thank you, much appreciated! :slightly_smiling_face:

dracut-functions.sh
airtower-luna: Okay, changed to info, I didn't see your reply when writing my latest one (and changed in the other PR, too). What kind of real world data would be needed for testing?

bdrung: Calling `dracut -v` on some machines to see how many times this warning/info/debug is printed.

airtower-luna: On my desktop system I get about 75 instances of the `require_binaries` message (changed to `dinfo`), and none of the others. Less than 20 when building my cross-sysroot (where a lot of modules are disabled already). That does not seem overwhelming, and in any case why a module isn't used is the kind of thing I'd expect to see with verbose output.

PR #2754, opened 2026-09-11
fix(extractinitrd): create leading directories when extracting a selection

aafeijoo-suse:
cpio without the `-d/--make-directories` option does not create leading directories where needed (see cpio(1)). Extracting a selection of files fails if the archive entry for a parent directory is not part of that. E.g.:

```
$ sudo dracutbasedir=. ./lsinitrd.sh --unpack /boot/initrd etc/cmdline.d/*
cpio: etc/cmdline.d/00-btrfs.conf: Cannot open: No such file or directory
cpio: etc/cmdline.d/90crypt.conf: Cannot open: No such file or directory
cpio: etc/cmdline.d/95resume.conf: Cannot open: No such file or directory
cpio: etc/cmdline.d/95root-dev.conf: Cannot open: No such file or directory
dracut-extractinitrd: cpio failed
```

### Checklist

### DIFF
--- a/src/extractinitrd/extractinitrd.c
+++ b/src/extractinitrd/extractinitrd.c
@@ -719,8 +719,12 @@ int main(int argc, char **argv)
                 cpio_optv[cpio_optc++] = "-t";
         if (verbose)
                 cpio_optv[cpio_optc++] = "-v";
-        if (!do_list && !to_stdout && cpio_supports_no_absolute_filenames())
-                cpio_optv[cpio_optc++] = "--no-absolute-filenames";
+        if (!do_list && !to_stdout) {
+                cpio_optv[cpio_optc++] = "-d";
+
+                if (cpio_supports_no_absolute_filenames())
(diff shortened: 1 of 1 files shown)


src/extractinitrd/extractinitrd.c
```diff
@@ -721,6 +721,12 @@ int main(int argc, char **argv)
                 cpio_optv[cpio_optc++] = "-v";
         if (!do_list && !to_stdout && cpio_supports_no_absolute_filenames())
                 cpio_optv[cpio_optc++] = "--no-absolute-filenames";
```
bdrung: The new code could be integrated here to not duplicate the `!do_list && !to_stdout` check.

aafeijoo-suse: Ok. Should I keep the comment or can I drop it? I think is a bit redundant.

bdrung: You can drop the comment IMO. It does not add much information.

PR #2755, opened 2026-09-12
refactor(def_fs,dmsquash-live-root,iso-scan/filename): series to address portability & potential fs corruption

### LINKED ISSUE #2562: iso-scan may destroy hibernation image or cause some file corruption
As shown below, the iso-scan feature will try to mount every block device recognized by the system in readonly mode. This is [not safe](https://bugs.launchpad.net/ubuntu/+source/casper/+bug/230703) for most/all journalled filesystems that have been [hibernated to](https://bugs.launchpad.net/ubuntu/+source/casper/+bug/2051898). This is a problem even if the filesystem does not contain the iso.

https://github.com/dracut-ng/dracut/blob/3a737fc50c8b5e366d134a1a07bf91cbc022a705/modules.d/70dmsquash-live/iso-scan.sh#L18-L22

### Potential solutions
I see several potential solutions:
1. Set the block device to readonly: `blockdev --setro $dev`.
    - The simplest, but perhaps won't allow some block devices to mount.
2. Use `losetup -r` to create a readonly loopback device to be used to mount the filesystem.
    -  Creating a new readonly device seems more assured that no writes will make it to the block device.
3. Use `dm-snapshot` with snapshot file in tmpfs.
    - This is the most complicated (though not super difficult), but if the filesystem needs to be read-write this will allow it to be mounted with no modifications. I've seen xfs filesystems refuse to mount because the journal needs to be replayed.

These would be setup and tore down for every block device. Do we want to do a "bare" mount of the filesystem  once we find the iso? Or, for instance, keep the `dm-snapshot` device and use the found iso on it? If loading whole iso to ram, this problem goes away.

FGrose:
This series
- Fixes: https://github.com/dracut-ng/dracut/issues/2562,
- addresses portability with blkid within its scope,
- simplifies a convoluted dmsquash-live-root,
- avoids potential file corruption on mounting journaled filesystems,
- improves error checking on mounts,
- enables iso-scan on ntfs (with dmsquash-live-ntfs loaded),
- includes a number of micro optimizations
- improves debugging by providing a more informative prompt &
     reducing debug transcript spew,
 - avoids redundant pre-pivot hooks on non-network boots,
 - enables disc designation by ID_SERIAL_SHORT.

### Checklist
- [x] I have tested it locally on Fedora Workstation 44
- I have NOT provided new test(s) for it

### DIFF
--- a/man/dracut.cmdline.7.adoc
+++ b/man/dracut.cmdline.7.adoc
@@ -112,14 +112,33 @@ resume=UUID=3f5ad593-4546-4a94-a374-bcfb68aa11f7
 
 iso-scan/filename
 ~~~~~~~~~~~~~~~~~
+Requires the dracut 'dmsquash-live' module. +
+Using the GRUB command `loopback <devicename> </path/to/imagefile>`, the GRUB
+boot loader can make a disk device identified by `<devicename>` from an
+image file on the boot or other specified device. This allows GRUB to load
+the boot kernel and initramfs from the device made from the image file.
 
-Mount all mountable devices and search for ISO pointed by the argument. When
-the ISO is found set it up as a loop device. Device containing this ISO
-image will stay mounted at /run/initramfs/isoscandev.
+**iso-scan/filename=**__<filepath>__::
+    _dracut_ must independently identify the disk device holding the image file
+    using _dmsquash-live/iso-scan.sh_ code and the *_iso-scan/filename_*
+    parameter.
++
+All mountable devices will be mounted readonly in `/dev/disk/by-uuid/*` order.
+The first matching *_<filepath>_* is loop mounted. +
++
+NOTE: _ext[43]_, _f2fs_, & _xfs_ filesystems will be mounted with
+`-o ro,noload` or `ro,norecovery` and kernel messages will be checked for a
+dirty or hibernated filesytem, in which case the image will be checked with
+`checkisomd5` for integrity.)
++
+The parent disk device is always unmounted after the .iso image file has been
+loop setup. +
+A link to this disk is saved at `/run/initramfs/isoscandev`. +
+A link to the image loop device is saved at `/run/initramfs/isoloop`.
++
 Using iso-scan/filename with a Fedora/Red Hat/CentOS Live iso should just work
 by copying the original kernel cmdline parameters.
-Requires the dracut 'dmsquash-live' module.
-
++
 [listing]
 .Example
 --
--- a/man/dracut.cmdline.7.adoc
+++ b/man/dracut.cmdline.7.adoc
@@ -118,37 +118,93 @@ boot loader can make a disk device identified by `<devicename>` from an
 image file on the boot or other specified device. This allows GRUB to load
 the boot kernel and initramfs from the device made from the image file.
 
-**iso-scan/filename=**__<filepath>__::
+**iso-scan/filename=**[__<devspec>__:]__<filepath>__::
     _dracut_ must independently identify the disk device holding the image file
     using _dmsquash-live/iso-scan.sh_ code and the *_iso-scan/filename_*
     parameter.
+    * *_<devspec>_* may be specified as
+    *{`block:|UUID=|PARTUUID=|LABEL=|PARTLABEL=|serial=<devserial>/serial/<partspec>`}*
+    * *_<filepath>_* specifies the image file for loop mounting by _dracut_ to
+    access the root filesystem within. +
 +
-All mountable devices will be mounted readonly in `/dev/disk/by-uuid/*` order.
-The first matching *_<filepath>_* is loop mounted. +
+NOTE: _Without_ the optional *_<devspec>_* argument, all mountable devices will
+be mounted readonly in `/dev/disk/by-uuid/*` order. +
+(_ext[43]_, _f2fs_, & _xfs_ filesystems will be mounted with `-o ro,noload` or
+`ro,norecovery`.) +
+The first matching *_<filepath>_* that is successfully loop setup and mounted
+is used.
 +
-NOTE: _ext[43]_, _f2fs_, & _xfs_ filesystems will be mounted with
-`-o ro,noload` or `ro,norecovery` and kernel messages will be checked for a
-dirty or hibernated filesytem, in which case the image will be checked with
-`checkisomd5` for integrity.)
+Kernel messages from mounting will be checked for a dirty or hibernated
+filesytem, in which case the image will be checked with `checkisomd5` for
+integrity.
 +
-The parent disk device is always unmounted after the .iso image file has been
-loop setup. +
+The parent disk device is always unmounted after the .iso image's loop device
+has been mounted. +
 A link to this disk is saved at `/run/initramfs/isoscandev`. +
-A link to the image loop device is saved at `/run/initramfs/isoloop`.
+A link to the image loop device primary partition is saved at
+`/run/initramfs/isoloop`.
 +
 Using iso-scan/filename with a Fedora/Red Hat/CentOS Live iso should just work
 by copying the original kernel cmdline parameters.
 +
+The GRUB boot loader will need its appropriate modules for the partition and
+filesystem bearing the .iso file, such as __part_gpt__ or __part_msdos__ and
+_ext2_, _btrfs_, _f2fs_, or _xfs_ and _iso9660_ for the iso filesystem.
++
+In the first example below, the .iso image file is stored in the `/isos/`
+directory of the BOOT PARTITION OF THE HOST OPERATING SYSTEM so that GRUB
+can address the isofile for loop setup and reach the linux kernel and initrds.
++
+NOTE: The filepaths to the linux kernel and initrd below will differ depending
+on the .iso's OS distribution and vintage. +
+For example, Fedora 37-41 used paths `/images/pxeboot/vmlinuz` &
+`/images/pxeboot/initrd.img`.
++
 [listing]
-.Example
+.GRUB menuentry example employing the legacy all device scan.
 --
-menuentry 'Live' --class gnu-linux --class gnu --class os {
-    set isolabel=Live-x86_64
-    set isofile="/boot/iso/Live-x86_64.iso"
-    loopback loop $isofile
-    linux (loop)/isolinux/vmlinuz0 boot=isolinux iso-scan/filename=$isofile root=live:LABEL=$isolabel ro rd.live.image quiet
-    initrd (loop)/isolinux/initrd0.img
+menuentry 'Live .iso' --class gnu-linux --class gnu --class os {
+  insmod part_gpt
+  insmod ext2
+  insmod iso9660
+  insmod loopback
+  set isopath=isos/Live.x86_64.iso
+  loopback loop "$isopath"
+  linux (loop)/boot/x86_64/loader/linux iso-scan/filename="$isopath" root=live:iso-scan ro quiet splash
+  initrd (loop)/boot/x86_64/loader/initrd
 }
+--
+In the following example, the filesystem uuid of a storage partition
+is set to the variable `dev_uuid`, enabling direct addressing of an isofile on
+that filesystem.  On btrfs filesystems, the default root subvolume may need to
+be included as an additional directory level for GRUB as indicated by
+<btrfs_root_subvol/> in the example below.
++
+NOTE: This requires an initramfs built with an updated version of dracut.
++
+[listing]
+.grub.cfg script example for direct isofile access. Also accommodates multiple initrd images, if employed.
+--
+insmod part_gpt
+insmod ext2
+insmod iso9660
+insmod probe
+insmod loopback
+menuentry "Start 'Live .iso' --class gnu-linux --class gnu --class os {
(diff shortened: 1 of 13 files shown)


man/dracut.cmdline.7.adoc
```diff
@@ -112,24 +112,99 @@ resume=UUID=3f5ad593-4546-4a94-a374-bcfb68aa11f7
...
 [listing]
-.Example
+.GRUB menuentry example employing the legacy all device scan.
+--
+menuentry 'Live .iso' --class gnu-linux --class gnu --class os {
+  insmod part_gpt
+  insmod ext2
+  insmod iso9660
+  insmod loopback
+  set isopath=isos/Live.x86_64.iso
```
crass, line 171:
```suggestion
  set isopath=isos/Live.x86_64.iso
  search --no-floppy --set root --file "$isopath"
```

This shows root getting set and I think is less confusing and not showing root getting set and implicitly using $root.

modules.d/70dmsquash-live/dmsquash-live-root.sh
```diff
@@ -13,8 +13,82 @@ if getargbool 0 rd.live.debug; then
...
+    case $DEVTYPE in
+        partition)
+            parent=$(readlink -f "$syspath/..")
+            diskDevice="/dev/${parent##*/}"
+            ;;
+        *) diskDevice="/dev/$dev" ;;
+    esac
+}
+
+set +x
```
crass, line 40: What is the purpose of this `set +x` and others? Seems like they might prevent someone from doing a shell trace for debugging.

modules.d/70dmsquash-live/dmsquash-live-root.sh
```diff
@@ -13,8 +13,82 @@ if getargbool 0 rd.live.debug; then
...
+            mntcmd=/sbin/mount-ntfs-3g
+        else
+            die "mount-ntfs-3g is needed for ntfs filesystem mounting."
+        fi
+        ;;
+    *)
+        if [ -f "$livedev" ]; then
+            FSIMG=$livedev
+        else
+            mntcmd="mount -m -n -t $FS"
```
crass: should `FS` be `livedev_fstype`?

modules.d/70dmsquash-live/iso-scan.sh
```diff
@@ -1,38 +1,106 @@
...
+    mntcmd="mount -m -n -t $fsType"
+    # Prevent writing to filesystems with journals that may have been hibernated.
+    case $fsType in
+        btrfs) d=d ;;
+        ext[43]) opt=,noload ;;
+        xfs | f2fs) opt=,norecovery ;;
+        ntfs)
+            if [ -x /sbin/mount-ntfs-3g ]; then
+                mkdir -m 0755 -p /run/initramfs/live
+                mntcmd=/sbin/mount-ntfs-3g
```
crass, line 33: IIRC, this might modify the ntfs volume, perhaps replaying the log or something else that would make the windows hibernation image unusable.

modules.d/70dmsquash-live/iso-scan.sh
```diff
@@ -1,38 +1,106 @@
...
+                ;;
+        esac
     done
+    debug_on
+    fsType=$(blkid "$_isopath")
+    fsType="${fsType#* TYPE=\"}"
+    fsType="${fsType%%\"*}"
+    losetup -fPr "$_isopath" || return 1
+    isoloop=$(losetup -j "$_isopath")
+    isoloop="${isoloop%%:*}p1"
```
crass, line 70: Why add the partition here when not having it should work just as well? I think there are some bootable isos that do not have an embedded partition table and would not work here.

modules.d/70dmsquash-live/iso-scan.sh
```diff
@@ -1,38 +1,106 @@
...
-        else
-            umount "/run/initramfs/isoscan"
-        fi
+isospec=$1
+
+[ "$isospec" ] || die "An path to the .iso was not provided."
+
+ismounted /run/initramfs/live && exit 0
+
+isopath="${isospec##*:}"
```
crass, line 16: If we stick with this implementation, it should probably documented that the path to the iso cannot have colon characters anywhere in the path. I would prefer not to have that restriction though.

man/dracut.cmdline.7.adoc
```diff
@@ -112,24 +112,99 @@ resume=UUID=3f5ad593-4546-4a94-a374-bcfb68aa11f7
...
+boot loader can make a disk device identified by `<devicename>` from an
+image file on the boot or other specified device. This allows GRUB to load
+the boot kernel and initramfs from the device made from the image file.
+
+**iso-scan/filename=**[__<devspec>__:]__<filepath>__::
+    _dracut_ must independently identify the disk device holding the image file
+    using _dmsquash-live/iso-scan.sh_ code and the *_iso-scan/filename_*
+    parameter.
+    * *_<devspec>_* may be specified as
+    *{`block:|UUID=|PARTUUID=|LABEL=|PARTLABEL=|serial=<devserial>/serial/<partspec>`}*
```
crass, line 126: This specification of `_<devspec>_` seems incomplete. Presumably `block:` through `PARTLABEL=` have arguments, but I don't see them here. I guess many people would guess correctly what is meant here. But, still not completely specified.

I would think to be consistent `SERIAL=` would be preferable to `serial=`. Also perhaps it should be called `SERIALID` to distinguish from a serial port, which was where my mind went on first seeing this. I also don't like the `/serial/` delimiter, feels too hacky. And what is the `<partspec>`?

Why is `block:` prefix even necessary? I would think the device path would be sufficient. If the devspec starts with `/dev/` then its a block device path and won't be confused for the other types of specifiers.

man/dracut.cmdline.7.adoc
```diff
@@ -112,24 +112,99 @@ resume=UUID=3f5ad593-4546-4a94-a374-bcfb68aa11f7
...
-Mount all mountable devices and search for ISO pointed by the argument. When
-the ISO is found set it up as a loop device. Device containing this ISO
-image will stay mounted at /run/initramfs/isoscandev.
+Requires the dracut 'dmsquash-live' module. +
+Using the GRUB command `loopback <devicename> </path/to/imagefile>`, the GRUB
+boot loader can make a disk device identified by `<devicename>` from an
+image file on the boot or other specified device. This allows GRUB to load
+the boot kernel and initramfs from the device made from the image file.
+
+**iso-scan/filename=**[__<devspec>__:]__<filepath>__::
```
crass, line 121: I think it makes more sense to have a separate boot parameter called `iso-scan/device` whose argument can take the form similar to the `root` boot parameter. This would remove any ambiguity when the devspec have colons and the filepath has colons.

modules.d/70dmsquash-live/iso-scan.sh
```diff
@@ -1,38 +1,106 @@
...
+
+loopmountiso() {
+    local - opt _isopath fsType d mntcmd line
+    fsType=$(blkid "$dev")
```
crass, line 20: I would think it would be better to pass `$dev` at the first argument to `loopmountiso` (and the second argument could be `$isopath`). Is there a reason not to do this?

modules.d/80base/dracut-dev-lib.sh
```diff
@@ -126,3 +126,31 @@ cancel_wait_for_dev() {
         /sbin/initqueue --onetime --unique --name daemon-reload systemctl daemon-reload
     fi
 }
+
+# Trigger a disk or partition having property spec
+#  $1 - {LABEL=|UUID=|PARTLABEL=|PARTUUID=|serial=<SERIAL_SHORT>/serial/[pt_spec]}
+#  for action $2 - [add|remove|change|move|online|offline|bind|unbind] default: add
+label_uuid_udevadm_trigger() {
+    local "devspec=${1#block:}" "act=${2:-add}" prop=''
```
crass, line 134: Why is `block:` treated specially? Is `block:LABEL=...` a valid device specifier?

crass (changes requested):
This wasn't a comprehensive review, just somethings I noticed. I would still prefer, in addition most of what's here, using a readonly loopback device. I don't trust the file system drivers to not write to the file system when mounting as readonly, even with `noload` or `norecovery`. At a minimum `blockdev --setro` could be used, but I don't trust that fully either.

modules.d/70dmsquash-live/iso-scan.sh
```diff
@@ -1,38 +1,143 @@
...
+    esac
+    discloop=$(losetup -f)
+    losetup -r "$discloop" "$dev" || {
+        ret 1
+        return 1
+    }
+    dmesg -c > /dev/null
+    $mntcmd -o "ro$opt" "$discloop" /run/initramfs/isoscan 2> /dev/kmsg || {
+        ret 2
+        return1
```
crass, line 73:
```suggestion
        return 1
```

modules.d/70dmsquash-live/iso-scan.sh
```diff
@@ -1,38 +1,143 @@
...
+                mntcmd=/sbin/mount-ntfs-3g
+            else
+                warn "mount-ntfs-3g is needed for ntfs filesystem mounting."
+                ret 1
+                return 1
+            fi
+            ;;
+    esac
+    discloop=$(losetup -f)
+    losetup -r "$discloop" "$dev" || {
```
crass, line 66: This is racy in a way that probably really does not matter (not likely anything else is setting up a loop device in this moment). I think the following would be better.

```suggestion
    discloop=$(losetup --show -r -f "$discloop" "$dev") || {
```

modules.d/70dmsquash-live/iso-scan.sh
```diff
@@ -1,38 +1,143 @@
...
+        isopath="$devspec"
+        unset -v devspec
+        return 0
+    }
+    isopath="$1"
+}
+IFS='
+' read_isospec
+
+ret() {
```
crass, line 31: A better name might be `cleanup_loopmountiso`.

man/dracut.cmdline.7.adoc
```diff
@@ -110,26 +110,106 @@ resume=UUID=3f5ad593-4546-4a94-a374-bcfb68aa11f7
...
-.Example
+.GRUB menuentry example employing the legacy all device scan.
+--
+insmod part_gpt
+insmod ext2
+insmod iso9660
+insmod loopback
+menuentry 'Live .iso' --class gnu-linux --class gnu --class os {
+  set isopath=isos/Live.x86_64.iso
+  loopback loop "$isopath"
```
crass, line 177: There is still an implied root here. But maybe that's fine.

man/dracut.cmdline.7.adoc
```diff
@@ -112,24 +112,99 @@ resume=UUID=3f5ad593-4546-4a94-a374-bcfb68aa11f7
...
++
+[listing]
+.grub.cfg script example for direct isofile access. Also accommodates multiple initrd images, if employed.
+--
+insmod part_gpt
+insmod ext2
+insmod iso9660
+insmod probe
+insmod loopback
+menuentry "Start 'Live .iso' --class gnu-linux --class gnu --class os {
```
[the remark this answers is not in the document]
crass: Yes, I believe this is still an issue.

man/dracut.cmdline.7.adoc
```diff
@@ -110,26 +110,106 @@ resume=UUID=3f5ad593-4546-4a94-a374-bcfb68aa11f7
...
+  set isopath=isos/Live.x86_64.iso
+  loopback loop ($dev)/<btrfs_root_subvol/>"$isopath"
+  initrds=''
+  for f in (loop)/boot/x86_64/loader/initrd*; do
+    initrds="$initrds $f"
+  done
+  linux (loop)/boot/x86_64/loader/linux root=live:iso-scan ro quiet splash \
+    iso-scan/devspec=UUID=${dev_uuid} iso-scan/filename="$isopath"
+  initrd $initrds
+  loopback -d loop
```
crass, line 210: This should be removed as its unneeded. When GRUB boots this menu entry, it doesn't matter what loopback devices are enabled as GRUB gets blown away and GRUB doesn't need to do cleanup of loop devices.

modules.d/80base/dracut-dev-lib.sh
crass, line 1: This file was not reviewed.

modules.d/80base/dracut-lib.sh
crass, line 1: Also, not reviewed.

crass (changes requested):
I still see my comments from the last review and can't resolve them for some reason. Can you try resolving the ones you feel you addressed?

modules.d/80base/dracut-dev-lib.sh
FGrose: Apparently, see [dracut-lib.sh](https://github.com/dracut-ng/dracut/blob/main/modules.d/80base/dracut-lib.sh#L578C1-L598C2)

modules.d/70dmsquash-live/iso-scan.sh
FGrose: losetup option --show is not available in dash/busybox.  get_loop() from /sys/block/loop*/loop/backing_file used instead.

PR #2759, opened 2026-09-14
chore(nvmf): drop supporting nvmi-cli < 2.5

bdrung:
The code to check the `nvme` command to support `--nbft` does not work with busybox:

```sh
$ /usr/sbin/nvme connect-all --help 2>&1 | busybox sed -n '/[[:space:]]--nbft[[:space:]]/q1;$q0'
sed: unsupported command ;
```

Remove support for nvmi-cli < 2.5 because all containers have a recent enough version (except for opensuse where `nvme` is not installed):

```sh
$ containers=(alpine:busybox-edge alpine:edge arch:latest azurelinux:latest centos:latest debian:latest debian:sid fedora:latest fedora:rawhide gentoo:amd64-openrc gentoo:latest opensuse:latest ubuntu:devel ubuntu:rolling void:latest)
$ for container in "${containers[@]}"; do podman run -it --rm "ghcr.io/dracut-ng/$container" /usr/sbin/nvme connect-all --help | grep -q -- --nbft || echo "$container"; done
Error: runc: runc create failed: unable to start container process: error during container init: exec: "/usr/sbin/nvme": stat /usr/sbin/nvme: no such file or directory: OCI runtime attempted to invoke a command that was not found
opensuse:latest
```

### Checklist


devkontrol:
CC @mwilck @aafeijoo-suse @tbzatek

tbzatek:
Isn't it easier to fix the sed pattern instead of bumping version requirement? The nvmf module is not used just for NBFT.

mwilck:
nvme-cli < 2.5 shouldn't be dropped just yet.

bdrung:
If there a reason for keeping supporting nvme-cli, we can keep it. Looking at Debian: stable (trixie) has nvmi-cli 2.13-2.

mwilck:
TBH, I don't remember why I  didn't simply  use "grep" instead of that weird sed construct. I guess I thought that we aren't supposed to use "grep" in module code, but we pretty obviously do elsewhere.

PR #2763, opened 2026-09-16
fix(i18n): install locale.conf in hostonly mode with i18n_install_all

TomaszBojanowski:
This pull request restores installing /etc/locale.conf in hostonly mode
when i18n_install_all="yes" (Fedora's default). It was dropped from the
systemd module in c1975f65, but the i18n module only installs it when
i18n_install_all is not set. See the commit message for details.

This affects e.g. plymouth, which needs the system language in the
initrd to show translated messages on the offline update screen
(related: https://gitlab.freedesktop.org/plymouth/plymouth/-/merge_requests/498).

### Checklist


TomaszBojanowski:
The only failing check is TEST-46 on ubuntu:rolling, which looks
unrelated: the new code only runs in hostonly mode, and the test suite
builds images without -H. The same test passes on all other
distributions.

modules.d/20i18n/module-setup.sh
```diff
@@ -133,6 +133,11 @@ install() {
         if dracut_module_included "systemd"; then
             # https://github.com/dracutdevs/dracut/issues/796
             [[ -f ${dracutsysrootdir-}${VCONFIG_CONF} ]] && inst_simple ${VCONFIG_CONF}
+            # install locale.conf here, as install_local_i18n() is
+            # skipped when i18n_install_all is set
+            if [[ $hostonly ]] && [[ -f ${dracutsysrootdir-}${I18N_CONF} ]]; then
+                inst_simple ${I18N_CONF}
```
aafeijoo-suse: Shouldn't it be marked as a hostonly file, like it was before c1975f654864a12a7bbe45935b04f69b4da50ca6?
```sh
inst_simple -H ${I18N_CONF}
```

aafeijoo-suse (commented):
I guess that now this code in `install_local_i18n()` is dead:

https://github.com/dracut-ng/dracut/blob/8453799245898461e7e0b192a03cbe41b0d63cec/modules.d/20i18n/module-setup.sh#L224-L225

TomaszBojanowski:
Thanks! Both done: locale.conf is now installed with inst_simple -H,
and the redundant installation in install_local_i18n() is removed.

Tested on Fedora 44 (dracut 108): hostonly images contain locale.conf,
and with the -H handling from main, generic images don't.

devkontrol (approved):

PR #2769, opened 2026-09-18
feat(dracut): use exit code of 3cpio

bdrung:
In case creating the initrd with 3cpio fails, use the same exit code as 3cpio to pass that information to the users.

### Checklist


Conan-Kudo:
This doesn't make sense to do when we don't do it for other cpio backends.

PR #2774, opened 2026-09-22
Docs: Fix wrong usage of initrd

ypnos:
Just after stating that dracut creates initramfs, in the next paragraph, its output is referred to as an initrd image.


devkontrol:
https://github.com/dracut-ng/dracut/issues/1588

ypnos:
Thanks, I see. It might be worthwile to merge this, because it is right at the front page of the documentation, which is otherwise consistent.

